Issue #478, opened 2014-06-11
expose ide mode as a flag


ghosttie:
I would like to use IDE mode (i.e. error checking but no JS output) but it isn't exposed as a flag.

I think this could be a popular request - see [this discussion](https://groups.google.com/forum/#!topic/closure-compiler-discuss/9AdNq4wmqEE).

danbeam:
what's the resolution to this thread?  is there a CHECKS_ONLY or IDE mode equivalent from flags?

ghosttie:
--checks-only was added in https://github.com/google/closure-compiler/commit/601ce79200d657ebe959181a29c12ee66fc0d43c

PR #592, opened 2014-08-23
Add --checks-only

pauldraper:
This is at least a partial resolution for https://github.com/google/closure-compiler/issues/478.

Specifying the `--no-output` (EDIT: this has been changed to `--checks-only`) command line parameter skips output and runs only checks.
It works in with each of the compilation levels.

---

Alternate approaches I considered:

* **Add a new compilation level** However, unlike GCC's `-O`, Closure's compilation level determines a lot more than just the optimizations. For example, `checkTypes` is enabled by default `ADVANCED_OPTIMIZATIONS`, but not `SIMPLE_OPTIMIZATIONS`, and never for `WHITESPACE_ONLY`.

* **Use `skipNormalOutputs` in `CommandLineConfig`**. This  sounds similar, except it also skips checks. (I believe it is meant for outputting only dependency info.)

---

Performance benefits:

* 6000-line file with `WHITESPACE_ONLY` -- from 2.6 to 1.7 seconds
* 6000-line file with `SIMPLE_OPTIMIZATIONS` -- from 6.1 seconds to 3.8 seconds
* Large project with `ADVANCED_OPTIMIZATIONS` -- from 135 seconds to 70 seconds

### DIFF
--- a/src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
+++ b/src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
@@ -925,7 +925,7 @@ abstract class AbstractCommandLineRunner<A extends Compiler,
       outputBundle();
       outputModuleGraphJson();
       return 0;
-    } else if (result.success) {
+    } else if (!options.checksOnly && result.success) {
       outputModuleGraphJson();
       if (modules == null) {
         outputSingleBinary();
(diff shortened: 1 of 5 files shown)


concavelenz:
@MatrixFrog Do you have any outstanding concerns?  I think we should merge this.

MatrixFrog:
There is the general problem that we have a few different "only" flags and it can get confusing. But this is a good one to have so I'm not too worried about it.

pauldraper:
I rebased to make sure tests still pass.

PR #706, opened 2014-11-20
Ensure synchronized access to Flags.guardLevels in CommandLineRunner

crdev:
Otherwise, running a few CommandLineRunners in parallel threads
results in intermittent racy crashes.


MatrixFrog:
Thanks, sorry for the delay.

PR #731, opened 2014-12-05
Add event handlers for HTMLMediaElement

adelespinasse:
Most of the event handler attributes of HTMLMediaElement, such as onabort and oncanplay, weren't declared in the default externs, so they were getting renamed by the compiler. The one such attribute that was declared, oncanplaythrough, was erroneously declared as a non-nullable type (these attributes can be null).

One test fails with this change (ES6ModuleLoaderFileSystemTest.testFileSystem), but that seems to be failing at HEAD also.

### DIFF
--- a/externs/html5.js
+++ b/externs/html5.js
@@ -1338,11 +1338,79 @@ HTMLMediaElement.prototype.load = function() {};
  */
 HTMLMediaElement.prototype.canPlayType = function(type) {};
 
-/**
- * Callback when the media is buffered and ready to play through.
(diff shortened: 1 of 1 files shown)


adelespinasse:
Can this get reviewed and pulled? Or is there some reason why it's not a good idea?

MatrixFrog:
Ah, sorry. I think I was waiting for Travis and then forgot? Anyway I'll take a look.

adelespinasse:
Thanks!

PR #737, opened 2014-12-08
Add compiler pass to remove console statements

tntim96:
There doesn't seem to be any non-intrusive way to remove 'console.log' statements from JavaScript source (for example see http://www.elijahmanor.com/grunt-away-those-pesky-console-log-statements/). Attached is a patch to try to achieve this via a compiler pass.

Produces a warning if it detects that removal of the 'console.log' statement has any side-effects.

Any chance of working this into the compiler, and perhaps adding a specific command-line switch? Any feedback is welcome.


dimvar:
Closing this. We prefer to avoid passes that are too specific.

PR #750, opened 2014-12-19
Fix typo errors which caused rollback

ChadKillingsworth:
aea6e011c13bb183d1bbc74b1f30d240e976ccb7 mentions a typo caused the rollback. I manual review and local tests of the externs brought me to these corrections. However, if this isn't the correct change, can someone chime in with more details?

### DIFF
--- a/contrib/externs/angular_ui_router.js
+++ b/contrib/externs/angular_ui_router.js
@@ -36,7 +36,6 @@ var ui = {};
 ui.router = {};
 
 
-// TODO: Provide stronger types for properties on $state.
 /**
  * @typedef {{
  *   params: Object,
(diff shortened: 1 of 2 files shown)


contrib/externs/angular_ui_router.js
```diff
@@ -55,6 +54,93 @@ ui.router.$state;
...
+/**
+ * @typedef {{
+ *   location: (boolean|string|undefined),
+ *   inherit: (boolean|undefined),
+ *   relative: (Object|undefined),
+ *   notify: (boolean|undefined),
+ *   reload: (boolean|undefined)
+ * }}
+ */
+ui.router.$state.GoOptions_;
```
mprobst: I think JSCompiler used to have a bug that it would (silently!) not understand type definitions in externs that were used before they were defined. If you're not 100% sure this works as is, please move up front before any use sites.

ChadKillingsworth:
Done

ChadKillingsworth:
Can this be merged?

mprobst:
Looks ok, though we really should have tests verifying at least the syntactical correctness of these externs...

PR #761, opened 2014-12-23
Allow literal as method names in ObjectLiteral.

tuchida:
```js
// test.js
var a = { 0 () {} };
```

```
$ java -jar compiler.jar --language_in ECMASCRIPT6 --language_out ECMASCRIPT3 test.js
test.js:1: ERROR - Parse error. '[' expected
var a = { 0 () {} };
          ^
```

NumericLiteral and StringLiteral is valid methods name.
https://people.mozilla.org/~jorendorff/es6-draft.html#sec-method-definitions

I was confirmed that the code is valid on Firefox37 and Chrome41.


MatrixFrog:
Even though this is technically valid, I think it will be nice to keep the parser simpler by not allowing it in the Closure compiler, since it's a pretty strange construct that is unlikely to come up in real code.

tuchida:
I see. Thanks!

PR #767, opened 2014-12-31
Improve jquery 1.9 externs

pauldraper:
* Remove jQuery.event
* Remove jQuery.deferred
* Add non-nullable returns,
* Add jQuery.Event properties
* Deprecate .pipe()

### DIFF
--- a/contrib/externs/jquery-1.9.js
+++ b/contrib/externs/jquery-1.9.js
@@ -79,7 +79,7 @@ var jQueryAjaxSettings;
  * @param {(jQuerySelector|Element|Object|Array.<Element>|jQuery|string|
  *     function())=} arg1
  * @param {(Element|jQuery|Document|
- *     Object.<string, (string|function(!jQuery.event=))>)=} arg2
+ *     Object.<string, (string|function(!jQuery.Event=))>)=} arg2
  * @return {!jQuery}
  */
 function jQuery(arg1, arg2) {}
@@ -90,7 +90,7 @@ function jQuery(arg1, arg2) {}
  * @param {(jQuerySelector|Element|Object|Array.<Element>|jQuery|string|
  *     function())=} arg1
  * @param {(Element|jQuery|Document|
- *     Object.<string, (string|function(!jQuery.event=))>)=} arg2
+ *     Object.<string, (string|function(!jQuery.Event=))>)=} arg2
  * @return {!jQuery}
  */
 function $(arg1, arg2) {}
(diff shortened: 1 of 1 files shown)


ChadKillingsworth:
Spent some time looking at this more. As history, the externs were created by scraping the API documentation site (xml feed) and the types were not up to the normal standards of a Closure project. Since the public api has slowed in changes, manual changes have replaced this process and the types are slowly migrating to match the actual source rather than the documentation.

For the `jQuery.Event` changes - I agree. This just needs changed and is outdated. Any pain in the migration period is justified since `jQuery.event` is a non-public namespace for helper methods in the source.

As for `jQuery.Deferred`, it isn't a true constructor. The `jQuery.Deferred` function returns an anonymous object conforming to an interface (which I had called `jQuery.deferred`). See https://github.com/jquery/jquery/blob/master/src/deferred.js

To further complicate things, `jQuery.Deferred` can be used as a mixin function to add the `jQuery.deferred` interface methods to an object passed to the `jQuery.Deferred` function.

I'm open to suggestions on improving this, but it seems wrong to continue to mark `jQuery.Deferred` as a constructor. I'm also at a loss as to how to properly annotate this in a way that is both usable and correct.

pauldraper:
Yeah, it's too bad the API isn't cleaner.

`jQuery.Deferred` might not be a constructor in that it doesn't use prototypes, but the objects that it produces do adhere to the type definition currently in the externs.

Marking is a constructor permits the use of `new jQuery.Deferred()`, which is allowed.

And it isn't used as a mix-in function. It takes one parameter, which is a function that is called when it is constructed. (Not sure why...seems weird.)

I believe the type definitions are such that the compiler will type-check correctly.

(There is one exception, albeit uncommon: `goog.inherits(Foo, jQuery.Deferred)`. `@final` could fix this except that we have `$.jQuery` extending it.)

ChadKillingsworth:
It's worse than just inheriting. For instance:

```JavaScript
jQuery.Deferred() instanceof jQuery.Deferred !== true
(new jQuery.Deferred()) instanceof jQuery.Deferred !== true
```
Also, the jQuery documentation references a `deferred` object (lowercase): http://api.jquery.com/category/deferred-object/

I think the `jQuery.Deferred` function needs annotated to return either a typedef of a record object or an interface with the correct signature.

Separately, I'm going to issue a pull request to the jQuery API site to correct the reference to jQuery.Deferred being a constructor. That might take some doing though because they do not view types the same way as the compiler does.

ChadKillingsworth:
Just looked at Callbacks - it's the same as Deferred. Should return either a record or interface type (lowercase). http://api.jquery.com/category/callbacks-object/

ChadKillingsworth:
@pauldraper My jQuery documentation PR was merged. `jQuery.Deferred` is no longer described as a constructor. Do you want to finish this up now?

ChadKillingsworth:
Closing this for now.

Issue #412, opened 2014-05-08
Update Contributors file to reference GitHub usernames


ChadKillingsworth:
While reviewing a pull request, I realized that how the contributors file is arranged isn’t much help on GitHub. Are there thoughts for this going forward?

How do we recognize Google employees (or are their commits now strictly limited to the internal version)?

ChadKillingsworth:
We deleted the contributors file - it's now the github commit history.

PR #768, opened 2014-12-31
Decouple jsdoc documentation parsing from IDE mode.

kayahr:
Currently the parsing of JSDoc documentation is coupled to the IDE mode
option so normally JSDoc documentation strings are not parsed.  This
prevents the implementation of custom compiler passes like API Documentation
generation.  This patch introduces a new compiler option
(parseJsDocDocumentation) which can be used to enable the documentation
parsing even when not in IDE mode.  The functionality is backward
compatible, when IDE mode is enabled then documentation parsing is always
enabled)

### DIFF
--- a/src/com/google/javascript/jscomp/Compiler.java
+++ b/src/com/google/javascript/jscomp/Compiler.java
@@ -2175,6 +2175,7 @@ public class Compiler extends AbstractCompiler {
   protected Config createConfig(Config.LanguageMode mode) {
     return ParserRunner.createConfig(
         isIdeMode(),
+        options.isParseJsDocDocumentation(),
         mode,
         acceptConstKeyword(),
         options.extraAnnotationNames);
--- a/src/com/google/javascript/jscomp/CompilerOptions.java
+++ b/src/com/google/javascript/jscomp/CompilerOptions.java
@@ -121,6 +121,8 @@ public class CompilerOptions implements Serializable, Cloneable {
    */
   public boolean ideMode;
 
+  private boolean parseJsDocDocumentation = false;
+
   boolean saveDataStructures = false;
 
   /**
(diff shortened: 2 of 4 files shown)


MatrixFrog:
While we're generally hesitant to add more flags (because it adds complexity and provides new and interesting ways that things can go wrong) this makes sense to me. I wonder if it would make sense to get rid of "IDE mode" boolean on the `CompilerOptions` object, and just have the `--ide_mode` flag be shorthand for both `parseJsDocumentation` and the other things that are controlled by IDE mode (I don't recall what all of those are right now).

kayahr:
Sorry, just noticed you added a "needs-cla" label. I already signed a CLA, please check your [CONTRIBUTORS](https://github.com/google/closure-compiler/blob/35fbb325b06328ce7e2754d3dfb100509f188c01/CONTRIBUTORS#L55) file.

MatrixFrog:
I don't see an entry for `kayahr`. Did you switch usernames at some point?

kayahr:
Nope. I already signed the CLA before you switched to GitHub so you may have no reference to my GitHub account. The CONTRIBUTORS file lists E-Mail addresses and not usernames. My email address is k@ailis.de as you can see in the details of my Github account.

ChadKillingsworth:
Someday I'll get around to cleaning that up. See #412

MatrixFrog:
Looking closer, you're not adding a new flag to the command-line, just a new option in the CompilerOptions, which is good as it keep the command-line interface from becoming any more complex than it already is. Cool :)

concavelenz:
@MatrixFrog ide_mode prevents the compiler from failing early on errors (so it can cause instability due to unexpected AST contents) and skips the optimization phase

I'm ok with change, did anyone have any outstanding concerns or should we ask the author to update pull request so we can merge it?

MatrixFrog:
Sure, I think we can merge this.

MatrixFrog:
Had to do the merge from the command line so I uploaded it as a separate PR in order to get Travis to run again. Committed as 7e9f3b2855fa40e678c2b70254f7d886802bb409

PR #770, opened 2015-01-03
Fix index.js resolution for non-relative require invocations.

sayrer:
Resolve absolute references against the module root, so directories containing index.js files will be handled as they are for relative references.

### DIFF
--- a/src/com/google/javascript/jscomp/ES6ModuleLoader.java
+++ b/src/com/google/javascript/jscomp/ES6ModuleLoader.java
@@ -107,7 +107,7 @@ abstract class ES6ModuleLoader {
 
     private NaiveModuleLoader(AbstractCompiler compiler, String moduleRoot) {
       this.moduleRoot = moduleRoot;
-      this.moduleRootURI = new File(moduleRoot).toURI();
+      this.moduleRootURI = createUri(moduleRoot);
 
       // Precompute the module name of each source file.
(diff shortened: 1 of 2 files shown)


sayrer:
(I have already signed the CLA, and am listed in contributors)

sayrer:
cc @nicks

MatrixFrog:
This seems fine but I'm not as familiar with this as Nick so I'll wait to see if he has any comments on it.

src/com/google/javascript/jscomp/ES6ModuleLoader.java
```diff
@@ -117,11 +117,10 @@ private NaiveModuleLoader(AbstractCompiler compiler, String moduleRoot) {
 
     @Override
     String locate(String name, CompilerInput referrer) {
-      if (isRelativeIdentifier(name)) {
-        return convertSourceUriToModuleAddress(
-            createUri(referrer).resolve(createUri(name)));
-      }
-      return createUri(name).normalize().toString();
+      URI base = isRelativeIdentifier(name) ? createUri(referrer)
+          : createUri(moduleRoot);
```
nicks: should this use moduleRootURI?

sayrer:
moduleRootURI previously had a file: scheme, which made the resolution absolute in this case. Its other use works fine with no file: scheme, so I changed it.

nicks:
LGTM!

PR #771, opened 2015-01-04
Opera vendor-prefixed extern for MatchesSelector

sayrer:
(CLA signed)

### DIFF
--- a/externs/w3c_selectors.js
+++ b/externs/w3c_selectors.js
@@ -51,6 +51,14 @@ Element.prototype.querySelector = function(selectors) {};
 Element.prototype.querySelectorAll = function(selectors) {};
 
 /**
+ * https://dom.spec.whatwg.org/#dom-element-matches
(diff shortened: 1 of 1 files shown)


externs/w3c_selectors.js
```diff
@@ -51,6 +51,14 @@ Element.prototype.querySelector = function(selectors) {};
 Element.prototype.querySelectorAll = function(selectors) {};
 
 /**
+ * https://dom.spec.whatwg.org/#dom-element-matches
+ * https://developer.mozilla.org/en-US/docs/Web/API/Element.matches
+ * @param {string} selectors
+ * @return {boolean}
+ */
+Element.prototype.matches = function(selectors) {};
```
MatrixFrog: So it sounds like the standard name will eventually be `matches`, not `matchesSelector`, right? Perhaps we should mark `matchesSelector` as deprecated, if not remove it entirely. (But we can do that later, doesn't need to be in this PR)

Issue #763, opened 2014-12-24
Add externs for Encoding API


cemerick:
i.e. https://encoding.spec.whatwg.org, currently supported in Firefox and Chrome, polyfilled via https://github.com/inexorabletash/text-encoding. Usage of this API is munged as part of advanced compilation, looks like no externs are defined for it yet.

I am not familiar with the finer points of writing externs files, so my submitting a PR wouldn't be productive. However, I'm using the below with success locally, so other users might find it useful until an official externs file is shipped.

```js
/**
This Source Code Form is subject to the terms of the Mozilla Public License,
v. 2.0. If a copy of the MPL was not distributed with this file, You can
obtain one at http://mozilla.org/MPL/2.0/.
**/

/**
 * @fileoverview Externs for Encoding API
 *
 * @see https://encoding.spec.whatwg.org
 * @externs
 */

/**
 * @constructor
 * @param {string=} encoding
 * @param {Object=} options
 */
function TextDecoder(encoding, options) {}

TextDecoder.prototype = {
    /**
     * @param {Uint8Array=} bytes
     * @param {Object=} options
     * @return {string}
     */
    decode: function decode(input, options) {}
};

/**
 * @constructor
 * @param {string=} encoding
 * @param {Object=} options
 */
function TextEncoder(encoding, options) {}

TextEncoder.prototype = {
    /**
     * @param {string=}
     * @param {Object=} options
     * @return {Uint8Array}
     */
    encode: function encode(opt_string, options) {}
};
```

PR #775, opened 2015-01-13
add externs file for W3C Encoding API, fixes gh-763

cemerick:
I'm using this successfully in a couple of projects. CLA was filed this morning.

### DIFF
--- a/externs/w3c_encoding.js
+++ b/externs/w3c_encoding.js
@@ -1,0 +1,57 @@
+/*
+ * Copyright 2015 The Closure Compiler Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * @fileoverview Definitions for W3C's Encoding specification
+ *     https://encoding.spec.whatwg.org
+ * @externs
+ */
+
+/**
+ * @constructor
+ * @param {string=} encoding
+ * @param {Object=} options
+ */
+function TextDecoder(encoding, options) {}
+
+/** @type {string} **/ TextDecoder.prototype.encoding;
+/** @type {boolean} **/ TextDecoder.prototype.fatal;
+/** @type {boolean} **/ TextDecoder.prototype.ignoreBOM;
+
+/**
+ * @param {!Uint8Array} bytes
+ * @param {Object=} options
+ * @return {string}
+ */
+TextDecoder.prototype.decode = function decode(input, options) {};
+
+/**
+ * @constructor
+ * @param {string=} encoding
+ * @param {Object=} options
+ */
+function TextEncoder(encoding, options) {}
+
+/** @type {string} **/ TextEncoder.prototype.encoding;
+
+/**
+ * @param {!string=}
+ * @param {Object=} options
+ * @return {!Uint8Array}
+ */
+TextEncoder.prototype.encode = function(string, options) {};
+
+
--- a/src/com/google/javascript/jscomp/CommandLineRunner.java
+++ b/src/com/google/javascript/jscomp/CommandLineRunner.java
@@ -1227,6 +1227,7 @@ public class CommandLineRunner extends
     "v8.js",
     "webstorage.js",
     "w3c_anim_timing.js",
+    "w3c_encoding.js",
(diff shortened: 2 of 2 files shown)


externs/w3c_encoding.js
```diff
@@ -0,0 +1,57 @@
...
+ */
+function TextDecoder(encoding, options) {}
+
+/** @type {string=} **/ TextDecoder.prototype.encoding;
+/** @type {boolean} **/ TextDecoder.prototype.fatal;
+/** @type {boolean} **/ TextDecoder.prototype.ignoreBOM;
+
+/**
+ * @param {Uint8Array=} bytes
+ * @param {Object=} options
```
MatrixFrog: As written these params are both nullable and optional. Would it make sense to make them just optional by adding `!`s?

cemerick:
@MatrixFrog Yeah, you're right. This is my first go at writing externs for others to consume, so sorry for the sloppiness. I've updated the PR with a squashed commit that includes more precise type expressions.

externs/w3c_encoding.js
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * @fileoverview Definitions for W3C's Encoding specification
+ *     https://encoding.spec.whatwg.org
+ * @externs
+ */
+
+/**
+ * @constructor
+ * @param {!string=} encoding
+ * @param {Object=} options
```
MatrixFrog: should this be `@param {!Object=}`?

externs/w3c_encoding.js
```diff
@@ -0,0 +1,57 @@
...
+ * @param {!Uint8Array} bytes
+ * @param {Object=} options
+ * @return {string}
+ */
+TextDecoder.prototype.decode = function decode(input, options) {};
+
+/**
+ * @constructor
+ * @param {!string=} encoding
+ * @param {Object=} options
```
MatrixFrog: here too

externs/w3c_encoding.js
```diff
@@ -0,0 +1,57 @@
...
+ */
+function TextEncoder(encoding, options) {}
+
+/** @type {!string} **/ TextEncoder.prototype.encoding;
+
+/**
+ * @param {!string=}
+ * @param {Object=} options
+ * @return {Uint8Array}
```
MatrixFrog: `@return {!Uint8Array}`?

externs/w3c_encoding.js
cemerick: Both constructor options params are nullable AFAICT; there's no mention of that constraint in the spec, and two implementations of it (FF, Chrome) allow nulls, e.g. `new TextEncoder("UTF8", null).encode("foo")`.

externs/w3c_encoding.js
cemerick: Sure, I can change that; was mostly just following what looks like convention in existing externs, where return types aren't qualified as non-nullable, even when the methods/fns in question always return something non-null.

MatrixFrog: Yeah, we've historically been a little lazy/sloppy about nullness and we're trying to gradually fix that.

cemerick:
Updated the PR with more stringent return types.

externs/w3c_encoding.js
MatrixFrog: You probably wouldn't bother doing that, you'd just write `new TextEncoder("UTF8").encode("foo")`. But I guess it's more correct this way so that's fine.

Small nit: Leave off the '!' on string/boolean since it's implied there. Then we should be good to go!

cemerick: Right, I've never used the options option.

Will update the PR again shortly.

externs/w3c_encoding.js
```diff
@@ -0,0 +1,57 @@
...
+function TextDecoder(encoding, options) {}
+
+/** @type {string} **/ TextDecoder.prototype.encoding;
+/** @type {boolean} **/ TextDecoder.prototype.fatal;
+/** @type {boolean} **/ TextDecoder.prototype.ignoreBOM;
+
+/**
+ * @param {!Uint8Array} bytes
+ * @param {Object=} options
+ * @return {!string}
```
MatrixFrog: Sorry one last nitpick: leave off the `!` for string and boolean since it's auto implied in that case.

cemerick:
Let's hope that does the trick. I didn't realize that such things were implicit, which I take as my defense w.r.t. return type nullability in general. :-)

MatrixFrog:
Sorry, I'm way behind on pull requests. Thanks!

externs/w3c_encoding.js
```diff
@@ -0,0 +1,57 @@
...
+function TextEncoder(encoding, options) {}
+
+/** @type {string} **/ TextEncoder.prototype.encoding;
+
+/**
+ * @param {!string=}
+ * @param {Object=} options
+ * @return {!Uint8Array}
+ */
+TextEncoder.prototype.encode = function(string, options) {};
```
MatrixFrog: Actually, if I'm reading this right, encode() just takes one string argument, without the `options` argument. https://encoding.spec.whatwg.org/#interface-textencoder

PR #777, opened 2015-01-15
Remove Eclipse specific .classpath and .project.

mprobst:
Closure Compiler's Maven pom.xml should be the authoritative source of
project configuration. Users can run
`mvn eclipse:eclipse -DdownloadSources=true` to create a correctly
configured Eclipse project.


ChadKillingsworth:
A lot of users utilize Eclipse - and this will make it much harder for them to build the compiler from source.

At a minimum - removal of this would need to have corresponding updates to the readme file on building with Eclispe.

concavelenz:
I don't see that I we need to remove this.  I never use maven and always use eclipse.  Is there a motivation for this?  Is it out of date?

ChadKillingsworth:
I use eclipse as well - and none of the projects I work on use maven.

mprobst:
I brought it up on the mailing list. We currently manually maintain a checked in .classpath and .project, whereas you can easily generate both from the `pom.xml`. Normally you wouldn't check in sources that can be generated from an authoritative source (the `pom.xml`).

An added benefit is that maven will automatically download source attachments for all required libraries, which is nice for debugging.

I don't feel particularly strong about this. If we agree not to do this, we should fix .classpath though ;-)

dimvar:
I'm fine with doing this, but also don't have a strong preference.

As Chad said, if this PR is accepted, you should also update the README file about building with Eclipse.

ChadKillingsworth:
Why can't we just generate and commit the eclipse project files whenever pom.xml is updated? Could we add a travis check to make sure this happens?

I'm all for fixing the classpath settings for eclipse.

mprobst:
Chad: a generated .classpath/.project has references to the jars (both implementation and source) in your local Maven repository. It shouldn't be checked in, it's an artifact of the build process.

In any case, I wasn't aware people were not really using maven for the build (it's a bit misleading to have the pom around). I'll drop this pull request and send a CL to just fix .classpath for the time being.

mprobst:
FYI I sent #778 for the .classpath fix.

supersteves:
See https://github.com/google/closure-compiler/pull/1940

PR #778, opened 2015-01-16
Add truth.jar to .classpath.

mprobst:
Include .properties files in the Eclipse compile operation so that
unit tests can be run from Eclipse, even if there was a clean build
since the last command line/ant build.

### DIFF
--- a/.classpath
+++ b/.classpath
@@ -1,24 +1,17 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <classpath>
-        <classpathentry including="**/*.java" kind="src" output="build/test-classes" path="test"/>
-        <classpathentry including="**/*.java" kind="src" path="src"/>
-        <classpathentry including="**/*.java" kind="src" path="gen"/>
-        <classpathentry kind="lib" path="lib/ant.jar"/>
-        <classpathentry kind="lib" path="lib/ant-launcher.jar"/>
-        <classpathentry kind="lib" path="lib/args4j.jar"/>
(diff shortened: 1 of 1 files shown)


.classpath
```diff
@@ -1,24 +1,17 @@
...
-        <!--
-        You need to tell Eclipse to build the compiler jar first, and then
-        normal source warnings will work because it will pick up the artifact.
-
-        If anyone has any better ideas, I'd be glad to hear them.
-        -->
-
-        <classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER"/>
-        <classpathentry kind="output" path="build/classes"/>
+	<classpathentry including="**/*.java|**/*.properties" kind="src" path="src"/>
```
MatrixFrog: Github is showing almost every line as changed, I guess because you switched from tabs to spaces. But the addition of |**/*.properties here is the only significant change, right?

MatrixFrog: er, spaces to tabs.

mprobst: Yes, the new file is what Eclipse generates when you use its UI to configure things. Happy to change this back, but I think it's better to just go with what Eclipse does here. .classpath is IMHO not really a human written, humand read source file, so I think it's fair to with simply what works best with the tooling.

mprobst:
Also FYI, I'm not entirely sure what behaviour comment was referencing, but with the .properties change everything seems to work fine without building compiler.jar first.

mprobst:
@MatrixFrog PTAL, does this make sense to you?

PR #783, opened 2015-01-17
Include a space before single quote and bang when pretty-printing

alexeagle:
This fix includes a refactoring. I would need to pass yet another boolean (setPreferSingleQuotes) on the options, which in the original design means yet another overload of the parsePrint method. It's a smell that all the callsites of this method are in other utility classes (not in test* methods) and they pass false, true arguments which don't mean anything unless you find that positional argument to parsePrint to see what that boolean controls.
So instead I've made a minimal interface so that each callsite can explicitly set the options it cares about. It's more lines overall, but more readable IMHO. This also makes it easy to let individual tests alter the options for one particular assertion, as I needed to do.

I'd be happy to put the refactoring in a separate PR if you want.

### DIFF
--- a/src/com/google/javascript/jscomp/CodePrinter.java
+++ b/src/com/google/javascript/jscomp/CodePrinter.java
@@ -340,6 +340,7 @@ public final class CodePrinter {
         append(op);
         append(" ");
       } else {
+        append(" ");
         append(op);
       }
     }
--- a/src/com/google/javascript/jscomp/CodePrinter.java
+++ b/src/com/google/javascript/jscomp/CodePrinter.java
@@ -333,15 +333,12 @@ public final class CodePrinter {
 
     @Override
     void appendOp(String op, boolean binOp) {
+      if (getLastChar() != ' ' && op.charAt(0) != ',') {
+        append(" ");
+      }
+      append(op);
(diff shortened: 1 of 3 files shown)


mprobst:
Could you change the header line of the commit message to actually say what is being fixed, instead of just referencing the issue #?

src/com/google/javascript/jscomp/CodePrinter.java
```diff
@@ -340,6 +340,7 @@ void appendOp(String op, boolean binOp) {
         append(op);
         append(" ");
       } else {
+        append(" ");
```
mprobst: Would this be easier to read if you reformulated the conditions? Essentially you're now always prepending a whitespace, except for the lastchar/',' case (where op == ',' should always be a binOp, right?). So you could just check for the binOp case for the appended whitespace.

alexeagle: Sure, done.

mprobst:
LGTM. @dimvar do you want to merge this?

concavelenz:
I've been pretty unhappy with the code printer class and interface for a while but it hasn't bubbled up as a priority.

PR #790, opened 2015-01-22
Include a space before single quote and bang when pretty-printing.

alexeagle:
Fixes issue 782.

### DIFF
--- a/src/com/google/javascript/jscomp/CodeGenerator.java
+++ b/src/com/google/javascript/jscomp/CodeGenerator.java
@@ -188,6 +188,7 @@ class CodeGenerator {
       case Token.THROW:
         Preconditions.checkState(childCount == 1);
         add("throw");
+        cc.maybeInsertSpace();
         add(first);
 
         // Must have a ';' after a throw statement, otherwise safari can't
@@ -198,6 +199,7 @@ class CodeGenerator {
       case Token.RETURN:
         add("return");
         if (childCount == 1) {
+          cc.maybeInsertSpace();
           add(first);
         } else {
           Preconditions.checkState(childCount == 0);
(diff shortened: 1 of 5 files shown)


alexeagle:
Thanks, I looked through the unary operators but wasn't thinking about how they can appear prefix or suffix, and that makes a difference. Added the missing test cases and will find a fix.

alexeagle:
Ping!

src/com/google/javascript/jscomp/CodeGenerator.java
```diff
@@ -198,6 +198,7 @@ void add(Node n, Context context) {
       case Token.RETURN:
         add("return");
         if (childCount == 1) {
+          cc.maybeInsertSpace();
```
blickly: Let's also do the same for yield and throw, while we're here.

alexeagle: Done.

test/com/google/javascript/jscomp/CodePrinterTest.java
```diff
@@ -978,6 +967,50 @@ public void testPrettyPrinter_defaultValue() throws Exception {
     assertPrettyPrint("(a=1)=>123;", "(a = 1) => 123;\n");
   }
 
+  // For https://github.com/google/closure-compiler/issues/782
+  public void testPrettyPrinter_spaceBeforeSingleQuote() throws Exception {
+    assertPrettyPrint("var f = function() { return 'hello';};",
```
dimvar: Was the missing space before } intentional in this test?

dimvar:
OK, looks good. I just had a minor comment. Can you resolve the conflicts and squash the commits? And then I'll merge it.

test/com/google/javascript/jscomp/CodePrinterTest.java
alexeagle: No, it's inconsequential. Added a space.

alexeagle:
All set.

Issue #798, opened 2015-01-28
ES6 Compilation - Internal Compiler Error


si-robertson:
I have been working on adding a new `--allow_es6_to_es6` flag (see #780), which went well, but the compiler is tripping over itself when it reaches a class constructor.

The console output is as follows:

````
java.lang.RuntimeException: INTERNAL COMPILER ERROR.
Please report this problem.

null
  Node(EXPR_RESULT): test-es6-to-es6.js:5:4
    console.log('Foo');
  Parent(BLOCK): test-es6-to-es6.js:4:16
  constructor() {

        at com.google.javascript.jscomp.OptimizeArgumentsArray.tryReplaceArguments(OptimizeArguments Array.java:182)
        at com.google.javascript.jscomp.OptimizeArgumentsArray.exitScope(OptimizeArgumentsArray.java:127)
        at com.google.javascript.jscomp.NodeTraversal.popScope(NodeTraversal.java:689)
````

alexeagle:
This is the first of many issues with updating the optimization passes to work with es6 inputs. It's probably a good idea to add your flag to resolve #780 but it will only be useful to allow us to discover everything that is broken.

MatrixFrog:
Right, NPEs like this are the reason I added the restriction in the first place :) Sorry, I probably should have been clearer about that.

si-robertson:
Should I create a pull request for the new flag or wait until things have settled down a bit at your end?

MatrixFrog:
The flag will be good to have, so we can find all the places where we don't
support ES6. The documentation for the flag should probably say "Note that
the compiler will probably crash if you use this flag" :)

si-robertson:
Done.

alexeagle:
Tyler: have a look at
https://github.com/alexeagle/closure-compiler/compare/allow_es6_to_es6?expand=1

for a first stab at disabling these broken passes when Si's flag is turned on.

si-robertson:
I'm a bit stuck creating a new pull request at the moment because something somewhere keeps causing entire Java scripts to be flagged as modified by Git. I'm pretty sure it's a line-ending problem but I have no idea what's causing it. Git's autocrlf option is set to true. There's nothing in Eclipse that automatically modifies line-endings.

I will have to get this sorted out tomorrow.

si-robertson:
Done.

concavelenz:
es6-to-es6 should only work in whitespace only mode right now.  If it doesn't that should be readily fixable.

Issue #780, opened 2015-01-16
Make "ES6 is only supported for transpilation" a command line option


agentx-cgn:
Currently the compiler can't output ES6 JavaScript. Supported is transpiling to e.g ES5, though. However, there are already use cases for at least minimal whitespace compression for ES6 without performance impact as SpiderMonkey, IO.js and recent browsers already support ES6. History: https://github.com/google/closure-compiler/issues/221#issuecomment-70284785

si-robertson:
I should be able to sort this one out this week, as discussed [here](https://groups.google.com/forum/#!topic/closure-compiler-discuss/fWsa3YzrRXA), so consider it to be in-progress.

agentx-cgn:
si-robertson, I read in the group you hit a road block. Could put here a minimal description to document progress?

si-robertson:
Yep, see issue #798 :)

alexeagle:
FYI we also need a flag to allow ES5->ES6_TYPED, so I'm planning to rename your flag to
--allow_es6_out

si-robertson:
@alexeagle What's the purpose of the ES5 to ES6_TYPED conversion?

alexeagle:
Moving the type declarations from JSDoc to typescript-style inline types,
so we can start on inter-op between closure code and typescript code.

PR #800, opened 2015-01-28
Added --allow_es6_to_es6 flag.

si-robertson:
Allows ES6 to ES6 compilation but is currently experimental and may cause the compiler to crash.

References: #780 #798

### DIFF
--- a/src/com/google/javascript/jscomp/CompilerOptions.java
+++ b/src/com/google/javascript/jscomp/CompilerOptions.java
@@ -69,6 +69,11 @@ public class CompilerOptions implements Serializable, Cloneable {
   private LanguageMode languageOut;
 
   /**
+   * If true, allows experimental ES6 to ES6 compilation.
+   */
+  boolean allowEs6ToEs6;
+
+  /**
    * If true, transpile ES6 to ES3 only. All others passes will be skipped.
    */
   boolean transpileOnly;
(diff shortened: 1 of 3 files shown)


src/com/google/javascript/jscomp/CompilerOptions.java
```diff
@@ -69,6 +69,11 @@
   private LanguageMode languageOut;
 
   /**
+   * If true, allows experimental ES6 to ES6 compilation.
+   */
+  boolean allowEs6ToEs6;
```
MatrixFrog: Let's make this private with a getter and setter. I know it feels overkill but it may actually make things easier if we need to refactor.

src/com/google/javascript/jscomp/CompilerOptionsPreprocessor.java
```diff
@@ -43,14 +43,14 @@ static void preprocess(CompilerOptions options) {
 
     if (options.getLanguageIn() == options.getLanguageOut()) {
       // No conversion.
-    } else if (!options.getLanguageIn().isEs6OrHigher()) {
+    } else if (!options.getLanguageIn().isEs6OrHigher() && !options.allowEs6ToEs6) {
```
MatrixFrog: I don't think we need this. This is to prevent things like --language_in=ES5 --language_out=ES3

si-robertson:
No problem, I will make those changes now and create another pull request. Closing this one now.

PR #804, opened 2015-01-29
Added --allow_es6_to_es6 flag.

si-robertson:
Allows experimental ES6 to ES6 compilation but may cause the compiler to crash at the moment.

Reference: #780 #798


MatrixFrog:
At some point we want to do something more nuanced -- only report an error if the rest of the options specified indicate that you're going to run a pass that might crash (or worse, produce incorrect results). @alexeagle is working on this. But for now this is fine, I think.

PR #805, opened 2015-01-29
Add @nocollapse annotation

ChadKillingsworth:
This isn't ready to merge yet, but since this is a significant change.  Would someone be willing to test this internal to Google? 

This adds the `@nocollapse` annotation which will prevent a property from being completely collapsed. The annotation does NOT block sub-properties from being collapsed.

Example:

```JavaScript
var a = {};
/** @nocollapse */
a.b = {}; // this property will NOT be collapsed
a.b.c = {}; //this property will still be collapsed
```

There are multiple variations using the `@nocollapse` annotation. The unit tests show comprehensive examples.

### DIFF
--- a/src/com/google/javascript/jscomp/CollapseProperties.java
+++ b/src/com/google/javascript/jscomp/CollapseProperties.java
@@ -92,6 +92,10 @@ class CollapseProperties implements CompilerPass {
       "Variable {0} aliases a constructor, "
       + "so it cannot be assigned multiple times");
 
+  static final DiagnosticType INVALID_NOCOLLAPSE = DiagnosticType.warning(
+      "JSC_INVALID_NOCOLLAPSE",
+      "@nocollapse is not permitted here");
+
   private AbstractCompiler compiler;
 
   /** Global namespace tree */
@@ -506,7 +510,7 @@ class CollapseProperties implements CompilerPass {
    */
   private void flattenReferencesToCollapsibleDescendantNames(
       Name n, String alias) {
-    if (n.props == null) {
+    if (n.props == null || n.isCollapsingExplicitlyDenied()) {
       return;
     }
 
@@ -696,6 +700,14 @@ class CollapseProperties implements CompilerPass {
 
     // Handle this name first so that nested object literals get unrolled.
     if (n.canCollapse()) {
+      // Enum properties are always collapsed. Warn when @nocollapse
+      // is used on an enum property
+      if (n.docInfo != null && n.docInfo.isNoCollapse() &&
+          n.isDescendantOfEnum()) {
+        compiler.report(
+            JSError.make(n.docInfo.getAssociatedNode(), INVALID_NOCOLLAPSE));
+      }
+
       updateObjLitOrFunctionDeclaration(n, alias, canCollapseChildNames);
     }
 
@@ -997,7 +1009,7 @@ class CollapseProperties implements CompilerPass {
    */
   private void updateFunctionDeclarationAtFunctionNode(
       Name n, boolean canCollapseChildNames) {
-    if (!canCollapseChildNames) {
+    if (!canCollapseChildNames || !n.canCollapse()) {
       return;
     }
 
@@ -1118,7 +1130,7 @@ class CollapseProperties implements CompilerPass {
    * @param parent The node to which new global variables should be added
    *     as children
    * @param addAfter The child of after which new
-   *     variables should be added (may be null)
+   *     variables should be added
    * @return The number of variables added
    */
   private int addStubsForUndeclaredProperties(
--- a/src/com/google/javascript/jscomp/GlobalNamespace.java
+++ b/src/com/google/javascript/jscomp/GlobalNamespace.java
@@ -907,9 +907,10 @@ class GlobalNamespace
 
     Type type;
     private boolean declaredType = false;
-    private boolean hasDeclaredTypeDescendant = false;
+    private boolean isDeclared = false;
     int globalSets = 0;
     int localSets = 0;
+    int localSetsWithNoCollapse = 0;
     int aliasingGets = 0;
     int totalGets = 0;
     int callGets = 0;
@@ -964,6 +965,7 @@ class GlobalNamespace
 
     void addRef(Ref ref) {
       addRefInternal(ref);
+      JSDocInfo info;
       switch (ref.type) {
         case SET_FROM_GLOBAL:
           if (declaration == null) {
@@ -974,6 +976,11 @@ class GlobalNamespace
           break;
         case SET_FROM_LOCAL:
           localSets++;
+          info = ref.getNode() == null ? null :
+              NodeUtil.getBestJSDocInfo(ref.getNode());
+          if (info != null && info.isNoCollapse()) {
+            localSetsWithNoCollapse++;
+          }
           break;
         case PROTOTYPE_GET:
         case DIRECT_GET:
@@ -1009,12 +1016,18 @@ class GlobalNamespace
           }
         }
 
+        JSDocInfo info;
         switch (ref.type) {
           case SET_FROM_GLOBAL:
             globalSets--;
             break;
           case SET_FROM_LOCAL:
             localSets--;
+            info = ref.getNode() == null ? null :
+                NodeUtil.getBestJSDocInfo(ref.getNode());
+            if (info != null && info.isNoCollapse()) {
+              localSetsWithNoCollapse--;
+            }
             break;
           case PROTOTYPE_GET:
           case DIRECT_GET:
@@ -1074,10 +1087,28 @@ class GlobalNamespace
       return false;
     }
 
+    boolean isCollapsingExplicitlyDenied() {
+      // Enum keys are always collapsed. @nocollapse annotations are ignored
+      if (isDescendantOfEnum()) {
+        return false;
+      }
+
+      if (docInfo == null) {
+        Ref ref = getDeclaration();
+        if (ref != null) {
+          docInfo = getDocInfoForDeclaration(ref);
+        }
+      }
+
+      return docInfo != null && docInfo.isNoCollapse();
+    }
+
     boolean canCollapse() {
-      return !inExterns && !isGetOrSetDefinition() && (declaredType ||
+      return !inExterns && !isGetOrSetDefinition() &&
+          !isCollapsingExplicitlyDenied() &&
+          (declaredType ||
           (parent == null || parent.canCollapseUnannotatedChildNames()) &&
-          (globalSets > 0 || localSets > 0) &&
+          (globalSets > 0 || localSets > 0) && localSetsWithNoCollapse == 0 &&
           deleteProps == 0);
     }
 
@@ -1096,6 +1127,10 @@ class GlobalNamespace
       // it's probably not worth the effort.
       Preconditions.checkNotNull(declaration);
       if (declaration.getTwin() != null) {
+        return false;
+      }
+
+      if (isCollapsingExplicitlyDenied()) {
         return false;
       }
 
@@ -1119,18 +1154,20 @@ class GlobalNamespace
 
     /** Whether this is an object literal that needs to keep its keys. */
     boolean shouldKeepKeys() {
-      return type == Type.OBJECTLIT && aliasingGets > 0;
+      return type == Type.OBJECTLIT &&
+          (aliasingGets > 0 || isCollapsingExplicitlyDenied());
     }
 
     boolean needsToBeStubbed() {
-      return globalSets == 0 && localSets > 0;
+      return globalSets == 0 && localSets > 0 && localSetsWithNoCollapse == 0 &&
+          !isCollapsingExplicitlyDenied();
     }
 
     void setDeclaredType() {
       declaredType = true;
       for (Name ancestor = parent; ancestor != null;
            ancestor = ancestor.parent) {
-        ancestor.hasDeclaredTypeDescendant = true;
+        ancestor.isDeclared = true;
       }
     }
 
@@ -1155,7 +1192,7 @@ class GlobalNamespace
      * considered namespaces.
      */
     boolean isNamespaceObjectLit() {
-      return hasDeclaredTypeDescendant && type == Type.OBJECTLIT;
+      return isDeclared && type == Type.OBJECTLIT;
     }
 
     /**
@@ -1164,6 +1201,24 @@ class GlobalNamespace
      */
     boolean isSimpleName() {
       return parent == null;
+    }
+
+    /**
+     * Determines whether a node is a property of an enum.
+     * This is recursive because static properties can be added to enums after
+     * declaration.
+     */
+    boolean isDescendantOfEnum() {
+      if (parent == null) {
+        return false;
+      }
+
+      if (parent.type == Type.OBJECTLIT && parent.docInfo != null &&
+          parent.docInfo.hasEnumParameterType()) {
+        return true;
+      }
+
+      return parent.isDescendantOfEnum();
     }
 
     @Override public String toString() {
@@ -1190,6 +1245,8 @@ class GlobalNamespace
           case Token.VAR:
             return ref.node == refParent.getFirstChild() ?
                 refParent.getJSDocInfo() : ref.node.getJSDocInfo();
+          case Token.OBJECTLIT:
+            return ref.node.getJSDocInfo();
         }
       }
 
--- a/src/com/google/javascript/rhino/Messages.properties
+++ b/src/com/google/javascript/rhino/Messages.properties
@@ -235,6 +235,9 @@ msg.jsdoc.externs =\
 msg.jsdoc.nocompile =\
     extra @nocompile tag
 
+msg.jsdoc.nocollapse =\
+    extra @nocollapse tag
+
 msg.jsdoc.seemissing =\
     @see tag missing description
 
--- a/test/com/google/javascript/jscomp/CollapsePropertiesTest.java
+++ b/test/com/google/javascript/jscomp/CollapsePropertiesTest.java
@@ -16,8 +16,6 @@
 
 package com.google.javascript.jscomp;
 
-import static com.google.javascript.jscomp.CollapseProperties.UNSAFE_THIS;
-
 import com.google.javascript.rhino.Node;
 
 /**
@@ -61,6 +59,10 @@ public class CollapsePropertiesTest extends CompilerTestCase {
   public void testMultiLevelCollapse() {
     test("var a = {}; a.b = {}; a.b.c = {}; var d = a.b.c;",
          "var a$b$c = {}; var d = a$b$c;");
+
+    test("var a = {}; a.b = {}; "
+        + "/** @nocollapse */ a.b.c = {}; var d = a.b.c;",
+        "var a$b = {}; a$b.c = {}; var d = a$b.c;");
   }
 
   public void testDecrement() {
@@ -76,6 +78,11 @@ public class CollapsePropertiesTest extends CompilerTestCase {
   public void testObjLitDeclaration() {
     test("var a = {b: {}, c: {}}; var d = a.b; var e = a.c",
          "var a$b = {}; var a$c = {}; var d = a$b; var e = a$c");
+
+    test("var a = {b: {}, /** @nocollapse */ c: {}}; "
+        + "var d = a.b; var e = a.c",
+        "var a$b = {}; var a = {c: {}}; "
+        + "var d = a$b; var e = a.c");
   }
 
   public void testObjLitDeclarationWithGet1() {
@@ -120,16 +127,32 @@ public class CollapsePropertiesTest extends CompilerTestCase {
   public void testObjLitAssignmentDepth1() {
     test("var a = {b: {}, c: {}}; var d = a.b; var e = a.c",
          "var a$b = {}; var a$c = {}; var d = a$b; var e = a$c");
+
+    test("var a = {b: {}, /** @nocollapse */ c: {}}; var d = a.b; var e = a.c",
+        "var a$b = {}; var a = {c: {}}; var d = a$b; var e = a.c");
   }
 
   public void testObjLitAssignmentDepth2() {
     test("var a = {}; a.b = {c: {}, d: {}}; var e = a.b.c; var f = a.b.d",
          "var a$b$c = {}; var a$b$d = {}; var e = a$b$c; var f = a$b$d");
+
+    test("var a = {}; a.b = {c: {}, /** @nocollapse */ d: {}}; "
+        + "var e = a.b.c; var f = a.b.d",
+        "var a$b$c = {}; var a$b = {d: {}}; var e = a$b$c; var f = a$b.d");
   }
 
   public void testObjLitAssignmentDepth3() {
     test("var a = {}; a.b = {}; a.b.c = {d: 1, e: 2}; var f = a.b.c.d",
          "var a$b$c$d = 1; var a$b$c$e = 2; var f = a$b$c$d");
+
+    test("var a = {}; a.b = {}; a.b.c = {d: 1, /** @nocollapse */ e: 2}; "
+        + "var f = a.b.c.d; var g = a.b.c.e",
+        "var a$b$c$d = 1; var a$b$c = {e: 2}; var f = a$b$c$d; "
+        + "var g = a$b$c.e");
+
+    testSame("var a = {}; /** @nocollapse*/ a.b = {}; "
+        + "a.b.c = {d: 1, e: 2}; "
+        + "var f = a.b.c.d; var g = a.b.c.e");
   }
 
   public void testObjLitAssignmentDepth4() {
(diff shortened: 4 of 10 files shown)


ChadKillingsworth:
I mainly want to make sure this doesn't break existing code.

concavelenz:
What is the state of things here?

MatrixFrog:
I'll try it out internally and see what breaks.

MatrixFrog:
It seems that the script I'm using for converting PRs to internal CLs is unhappy about trailing whitespace. Can you set your editor to automatically remove trailing whitespace on save?

ChadKillingsworth:
Done. I recently switched editors and thought it was already set that way.

Also, while I'd like the unit tests to be finished at some point, I see no reason we can't go ahead and get this merged in and documented for people to use (as long as it doesn't break internal code). We can handle any cases where the `@nocollapse` annotation isn't properly honored via issues and patches.

test/com/google/javascript/jscomp/CollapsePropertiesTest.java
```diff
@@ -61,6 +61,10 @@ public void testCollapse() {
   public void testMultiLevelCollapse() {
     test("var a = {}; a.b = {}; a.b.c = {}; var d = a.b.c;",
          "var a$b$c = {}; var d = a$b$c;");
+
+    test("var a = {}; a.b = {}; " +
+        "/** @nocollapse */ a.b.c = {}; var d = a.b.c;",
```
MatrixFrog: The internal linter is complaining here that the + should be on the beginning of the next line, instead of the end of the previous line.

ChadKillingsworth: There's going to be a lot of those. I'll get them all fixed then.

MatrixFrog:
Okay, got it patched in and the presubmit is running. I'm not sure if the whitespace was actually throwing it off or that was just a red herring, but it's nice to get rid of it anyway.

test/com/google/javascript/jscomp/CollapsePropertiesTest.java
ChadKillingsworth: Is this a new style requirement? There are lots of these cases in the file that I didn't change.

MatrixFrog: New-ish, yes. Don't worry about fixing it in lines that you're not touching, but let's use the new style in new code.

MatrixFrog:
Seeing some failures with the internal presubmit that may actually be caused by this. I'll look into them.

ChadKillingsworth:
That doesn't surprise me. If you can give me a repo case that will help.

MatrixFrog:
Got a minimal repro case:

    var FOO = {
      'bar': {
        'baz,qux': {
          'beep': 'xxxxx',
        },
      }
    };
    
    alert(FOO);

If you compile this with the current compiler you get just `'use strict';alert({bar:{"baz,qux":{beep:"xxxxx"}}});` (i.e. no interesting changes) but with this PR I get `'use strict';alert({bar:{}});` -- most of the object literal has been eliminated. It seems to be caused by the fact that the key includes a comma, so (I guess) it can't collapse that part of the object.

ChadKillingsworth:
It looks like we handle quoted keys on object literals differently than assignments. https://github.com/google/closure-compiler/blob/master/test/com/google/javascript/jscomp/CollapsePropertiesTest.java#L437

I think quoted properties should be blocked from collapsing (object literal keys included).

@concavelenz thoughts?

dimvar:
> I think quoted properties should be blocked from collapsing (object literal keys included).

If this is independent from the nocollapse annotation, I would rather have only the nocollapse change in this PR, and keep the rest of the behavior the same. It simplifies things.

dimvar:
Hey Chad, I'll do this review. I will try to have comments by the end of this week. If not, definitely before our next meeting.

ChadKillingsworth:
>  I would rather have only the nocollapse change in this PR, and keep the rest of the behavior the same.

This is taking a bit to figure out. The unit tests don't define this well. I'm still unraveling how this is handled.

ChadKillingsworth:
This should be ready to test again.

test/com/google/javascript/jscomp/CollapsePropertiesTest.java
```diff
@@ -1617,92 +1917,91 @@ public void testLocalAliasOfEnumWithInstanceofCheck() {
         " var Fruit = null;\n" +
         " if (f == Enums$Fruit$APPLE) alert('apple');\n" +
         " if (f == Enums$Fruit$BANANA) alert('banana');\n" +
-        "}",
-        null);
```
MatrixFrog: It looks like this `null` was already removed a while ago, so the patch is not applying cleanly. I think if you merge or rebase that will fix it?

ChadKillingsworth: The merge conflict is caused by 76f458e596389ebba6927ea53f8073528b33e085. I can reverse that commit for now (it was just linter/formatting changes).

MatrixFrog:
Since you already have it as a separate commit, maybe let's merge it as its own PR (which will also make this one quicker to review)?

ChadKillingsworth:
Sure thing.

ChadKillingsworth:
That looks better.

src/com/google/javascript/rhino/Messages.properties
```diff
@@ -235,6 +235,9 @@ msg.jsdoc.externs =\
 msg.jsdoc.nocompile =\
     extra @nocompile tag
 
+msg.jsdoc.nocollapse =\
```
dimvar: Add test in JsDocInfoParserTest.java that has two <code>@nocollapse</code> in a jsdoc comment.

src/com/google/javascript/jscomp/GlobalNamespace.java
```diff
@@ -1075,9 +1088,22 @@ boolean isSimpleStubDeclaration() {
...
+            isAnnotatedNoCollapse = info.isNoCollapse();
+          }
+        }
+      }
+
+      return !inExterns && !isGetOrSetDefinition() && !isAnnotatedNoCollapse &&
+          (declaredType ||
           (parent == null || parent.canCollapseUnannotatedChildNames()) &&
-          (globalSets > 0 || localSets > 0) &&
+          (globalSets > 0 || localSets > 0) && localSetsWithNoCollapse <= 0 &&
```
dimvar: when can localSetsWithNoCollapse be negative?

src/com/google/javascript/jscomp/GlobalNamespace.java
```diff
@@ -1075,9 +1088,22 @@ boolean isSimpleStubDeclaration() {
     }
 
     boolean canCollapse() {
-      return !inExterns && !isGetOrSetDefinition() && (declaredType ||
+      boolean isAnnotatedNoCollapse = docInfo == null ?
+          false : docInfo.isNoCollapse();
+      if (!isAnnotatedNoCollapse) {
+        Ref ref = this.getDeclaration();
+        if (ref != null && ref.node != null) {
+          JSDocInfo info = NodeUtil.getBestJSDocInfo(ref.node);
```
dimvar: I would think that docInfo is always the same as info (the jsdoc at the assignment). Why is this not the case?

src/com/google/javascript/jscomp/GlobalNamespace.java
```diff
@@ -910,6 +910,7 @@ Name getOrCreateName(String name) {
     private boolean hasDeclaredTypeDescendant = false;
```
dimvar: change the name of the declaredType field to isDeclared.

src/com/google/javascript/jscomp/GlobalNamespace.java
dimvar: The expression <code> localSetsWithNoCollapse <= 0 </code> can only be reached if isAnnotatedNoCollapse is false, so in this case localSetsWithNoCollapse will always be 0, right?

src/com/google/javascript/jscomp/GlobalNamespace.java
```diff
@@ -1123,7 +1149,7 @@ boolean shouldKeepKeys() {
     }
 
     boolean needsToBeStubbed() {
-      return globalSets == 0 && localSets > 0;
+      return globalSets == 0 && localSets > 0 && localSetsWithNoCollapse <= 0;
```
dimvar: I think <code>@nocollapse</code> should only have an effect on the declaration site of a name, not on any random assignment to the name.

test/com/google/javascript/jscomp/CollapsePropertiesTest.java
```diff
@@ -76,6 +80,11 @@ public void testIncrement() {
   public void testObjLitDeclaration() {
     test("var a = {b: {}, c: {}}; var d = a.b; var e = a.c",
          "var a$b = {}; var a$c = {}; var d = a$b; var e = a$c");
+
+    test("var a = {b: {}, /** @nocollapse */ c: {}}; "
```
dimvar: what's the use case for not collapsing just one property of an object literal?

test/com/google/javascript/jscomp/CollapsePropertiesTest.java
```diff
@@ -202,11 +274,22 @@ public void testGlobalObjectNameInBooleanExpressionDepth3() {
...
+        "function a() {} a.c = 1; if (a) x(a.c);");
   }
 
   public void testGlobalFunctionNameInBooleanExpressionDepth2() {
     test("var a = {b: function(){}}; a.b.c = 1; if (a.b) x(a.b.c);",
          "var a$b = function(){}; var a$b$c = 1; if (a$b) x(a$b$c);");
+
+    test("var a = {/** @nocollapse */ b: function(){}}; a.b.c = 1; "
+        + "if (a.b) x(a.b.c);",
+        "var a = {b: function(){}}; var a$b$c = 1; if (a.b) x(a$b$c);");
```
dimvar: If we can't collapse a.b, the we should have a.b$c (or a.b.c) in this example, not a$b$c ?

test/com/google/javascript/jscomp/CollapsePropertiesTest.java
```diff
@@ -233,11 +320,36 @@ public void testAliasCreatedForObjectDepth2_2() {
          "var a$b = {c: 0}; for (var p in a$b) { e(a$b[p]); }");
   }
 
+  public void testEnumDepth1() {
+    test("/** @enum */ var a = {b: 0, c: 1};",
+         "var a$b = 0; var a$c = 1;");
+
+    test("/** @enum */ var a = { /** @nocollapse */ b: 0, c: 1};",
```
dimvar: Enums have two kinds of properties (it may or may not be possible to distinguish between the two kinds in this pass, I'm not sure). The first kind is the properties declared in the enum's object literal. These are the enumerated values, and we should not allow nocollapse there (if possible). The second kind is the properties added after the enum declaration. This is when the enum is treated as a namespace, and we can allow nocollapse for these properties.

Actually, John mentioned to me recently that maybe we should stop supporting enums as namespaces. So, if you can't distinguish between the two kinds of enum properties here, you can even disallow nocollapse on enum properties altogether.

test/com/google/javascript/jscomp/CollapsePropertiesTest.java
```diff
@@ -331,6 +460,11 @@ public void testAliasCreatedForCtorDepth1_1() {
     test("/** @constructor */ var a = function(){}; a.b = 1; "
          + "var c = a; c.b = 2; a.b != c.b;",
          "var a = function(){}; var a$b = 1; var c = a; c.b = 2; a$b != c.b;");
+
+    // Sometimes we want to prevent static members of a constructor from
+    // being collapsed.
+    testSame("/** @constructor */ var a = function(){};"
+        + "/** @nocollapse */ a.b = 1; var c = a; c.b = 2; a.b == c.b;");
```
dimvar: When we use nocollapse, would we want to prevent collapsing on all sub-properties of that property as well?

dimvar:
I did a round of comments, I'll do another one when these are addressed. Also, can you squash all the commits into one? It helps when we pull the change in and test it.

test/com/google/javascript/jscomp/CollapsePropertiesTest.java
ChadKillingsworth: If you are exporting a property and need it to remain mutable, then `@nocollapse` is needed. Restricting definitions to object literals annotated with `const` seems unnecessarily strict and would not be expected by users. 

```JavaScript
var foo = {
  /** @nocollapse */
  bar1: 10,
  bar2: 20
};
window['foo'] = window['foo'] || {};
window['foo']['bar1'] = foo.bar1;
```

test/com/google/javascript/jscomp/CollapsePropertiesTest.java
ChadKillingsworth: It may be possible to specifically deny `@nocollapse` annotations on enums. I'll look into that - I don't see any use cases that would require `@nocollapse` on an enum.

test/com/google/javascript/jscomp/CollapsePropertiesTest.java
ChadKillingsworth: I think the answer should be no we don't. From a user perspective, I'd like `@nocollapse` to be analogous to `@export` which only applies to the immediate property. If sub-properties need protected from collapsing, then they would also need `@nocollapse`.

I'd like to prevent the issue that occurs when an `@nocollapse` annotation on a top-level property blocks optimizations on entire chunks of code. I specifically wrote it to apply to only the associated property.

src/com/google/javascript/rhino/Messages.properties
ChadKillingsworth: Done

src/com/google/javascript/jscomp/GlobalNamespace.java
ChadKillingsworth: It shouldn't ever be negative - that would be a bug. I'll change that.

localSetsWithNoCollapse was added for this test case: https://github.com/google/closure-compiler/pull/805/files#diff-f6ebc91c417a43a65475b9a83521bc54R772

The test case fails without that check.

src/com/google/javascript/jscomp/GlobalNamespace.java
ChadKillingsworth: For an object literal, the docInfo points at the root object. Since a `@nocollapse` annotation could be used on a property key, we need to check that node as well. See test https://github.com/google/closure-compiler/pull/805/files#diff-f6ebc91c417a43a65475b9a83521bc54R84

I've added a comment to clarify.

src/com/google/javascript/jscomp/GlobalNamespace.java
ChadKillingsworth: Done. Though that one isn't my code.

src/com/google/javascript/jscomp/GlobalNamespace.java
ChadKillingsworth: In the case where the object is global, but the property is assigned in a local scope, this is needed.

See https://github.com/google/closure-compiler/pull/805/files#diff-f6ebc91c417a43a65475b9a83521bc54R855

test/com/google/javascript/jscomp/CollapsePropertiesTest.java
ChadKillingsworth: I specifically implemented this such that `@nocollapse` should not impact sub-properties. In this case, I believe `a.b.c`  should be fully collapsed and not remain a property since it doesn't have its own `@nocollapse` annotation.

ChadKillingsworth:
Apparently CLAs need github usernames on file now.

Enum handling is still outstanding, I've responded to the rest of the comments.

I'll squash the commits when I get that one handled.

ChadKillingsworth:
`@nocollapse` annotations on enum keys are now ignored. However, there isn't convenient place to issue warnings that these will be ignored. There are comments in the unit tests however.

I've squashed the commits as requested. I still have a local branch pointer to the original commits for my own reference.

test/com/google/javascript/jscomp/CollapsePropertiesTest.java
dimvar: If the point of nocollapse is to allow mutable properties to be exported, it's not correct to collapse the subproperties. Eg, one can mutate the uncollapsed a.b in a way that impacts a.b.c, so if we have already collapsed it to a$b$c, we produce wrong code.

Or are you expecting mutating the uncollapsed property in a restricted way, that doesn't impact subproperties? If so, do we have a well-defined rule about what we allow and what we don't?

ChadKillingsworth: I've had cases where the property was valid to be renamed, but it was not exported. Specifically this came up using modules. Here's an example:

```JavaScript
// Code in module 1
var foo = {};
(function() {
  var a = 0; //some calculated value
  foo.bar = a;
})();
function showFooBar() {
  console.log(foo.bar);
}

// Code in module 2
foo.bar = 42;
showFooBar();
foo.bar = 0;
```

However, this is still more or less the same use case as exporting. I can come up with contrived examples where collapsing sub-properties is used, but not a single real world example.

The rationale behind implementing `@nocollapse` such that sub-properties were still collapsed was a fear of blocking further optimizations. However, perhaps that fear is unfounded.

ChadKillingsworth: I should mention that it significantly simplifies the code if we do not collapse sub properties.

dimvar: Perfect. It seems unsafe to me to collapse subproperties if we allow the non-collapsed property to be mutable. If it also simplifies the code to not collapse them, let's do that.

ChadKillingsworth:
Ready for review/testing again:
 * `@nocollapse` now blocks sub-properties from being collapsed.
 * Since enums are always collapsed, a warning is now issued when `@nocollapse` is encountered on an enum key.
 * Code in GlobalNamespace.java was refactored and simplified.

dimvar:
Great, thanks. I'll take a look tomorrow again. Do you mind squashing the new commits as well?

ChadKillingsworth:
Sure. I'm looking into why their are unit tests failing at the moment in the ProcessDefines pass. I don't know why that would be related.

dimvar:
Sounds good. ProcessDefines uses GlobalNamespace, so it's probably some of the changes there that cause the failures.

ChadKillingsworth:
Fixed - took care of a couple TODOs at the same time.

src/com/google/javascript/jscomp/CollapseProperties.java
```diff
@@ -696,6 +700,16 @@ private void collapseDeclarationOfNameAndDescendants(Name n, String alias) {
 
     // Handle this name first so that nested object literals get unrolled.
     if (n.canCollapse()) {
+
```
dimvar: remove blank line

src/com/google/javascript/jscomp/CollapseProperties.java
```diff
@@ -696,6 +700,16 @@ private void collapseDeclarationOfNameAndDescendants(Name n, String alias) {
...
+      // Enum properties are always collapsed. Warn when @nocollapse
+      // is used on an enum property
+      if (n.docInfo != null && n.docInfo.isNoCollapse() &&
+          n.parent != null && n.parent.type == Name.Type.OBJECTLIT &&
+          n.parent.docInfo != null && n.parent.docInfo.hasEnumParameterType()) {
```
dimvar: here, I would create a helper method in the Name class, called isDescendantOfEnum, and it would check not only the parent, but all ancestors of the name.

src/com/google/javascript/jscomp/CollapseProperties.java
```diff
@@ -918,6 +932,7 @@ private void updateObjLitOrFunctionDeclarationAtAssignNode(
     }
   }
 
+
```
dimvar: remove blank line

src/com/google/javascript/jscomp/GlobalNamespace.java
```diff
@@ -1074,10 +1087,29 @@ boolean isSimpleStubDeclaration() {
       return false;
     }
 
+    boolean isCollapsingExplicitlyDenied() {
+      // Enum keys are always collapsed. @nocollapse annotations are ignored
+      if (parent != null && parent.type == Type.OBJECTLIT &&
```
dimvar: use the helper method for enum ancestors here

src/com/google/javascript/jscomp/GlobalNamespace.java
dimvar: Wrong field changed. Change the declaredType one instead. (I find it confusing that it's called declaredType when its type is boolean.)

test/com/google/javascript/jscomp/CollapsePropertiesTest.java
dimvar: Sounds good, let's keep it like this for now.

FYI, in the new type inference we only handle obj lits declared with const as namespaces. (It's both easier to implement, and it's clearer for the programmer to explicitly declare a namespace when they use const.)

test/com/google/javascript/jscomp/CollapsePropertiesTest.java
```diff
@@ -141,49 +168,94 @@ public void testObjLitAssignmentDepth4() {
...
+    test("var a = a || {}; /** @nocollapse */ a.c = 1;",
+        "var a = a || {}; a.c = 1;");
+ }
 
   public void testGlobalObjectDeclaredToPreserveItsPreviousValue3() {
     test("var a = a || {get b() {}}; a.c = 1;",
          "var a = a || {get b() {}}; var a$c = 1;");
+
+    test("var a = a || {get b() {}}; /** @nocollapse */ a.c = 1;",
+        "var a = a || {get b() {}}; a.c = 1;");
```
dimvar: I don't see any difference between the before/after here, except that the nocollapse annotation is removed. But in other tests you are using testSame. Why is the annotation removed in some tests and not in others? Or can you just use testSame here?

test/com/google/javascript/jscomp/CollapsePropertiesTest.java
```diff
@@ -260,6 +368,17 @@ public void testAliasCreatedForEnumDepth2_1() {
          + "var d = a.b; d.c = 1; a.b.c != d.c;",
          "var a$b$c = 0; var a$b = {c: a$b$c};"
          + "var d = a$b; d.c = 1; a$b$c != d.c;");
+
+    test("var a = {}; /** @nocollapse @enum */ a.b = {c: 0};"
+        + "var d = a.b; d.c = 1; a.b.c == d.c;",
+        "var a = {}; a.b = {c: 0};"
+        + "var d = a.b; d.c = 1; a.b.c == d.c;");
```
dimvar: testSame here?

src/com/google/javascript/jscomp/CollapseProperties.java
ChadKillingsworth: All ancestors? Can enums be nested somehow?

test/com/google/javascript/jscomp/CollapsePropertiesTest.java
```diff
@@ -415,48 +568,100 @@ public void testAliasCreatedForClassProperty() {
...
+        "var a$b = function() {}; a$b['d'] = 1;");
   }
 
   public void testFunctionWithQuotedPropertyThatGetsRead() {
     test("var a = {}; a.b = function() {}; a.b['d'] = 1; f(a.b['d']);",
          "var a$b = function() {}; a$b['d'] = 1; f(a$b['d']);");
+
+    test("var a = {}; /** @nocollapse */  a.b = function() {}; a.b['d'] = 1;"
+        + "f(a.b['d']);",
+        "var a = {}; a.b = function() {}; a.b['d'] = 1; f(a.b['d']);");
```
dimvar: testSame?

test/com/google/javascript/jscomp/CollapsePropertiesTest.java
```diff
@@ -470,6 +675,15 @@ public void testObjLitRedefinedInGlobalScope() {
   public void testObjLitRedefinedInLocalScope() {
     test("var a = {}; a.b = {c: 0}; function d() { a.b = {c: 1}; } e(a.b.c);",
          "var a$b = {c: 0}; function d() { a$b = {c: 1}; } e(a$b.c);");
+
+    test("var a = {};/** @nocollapse */ a.b = {c: 0}; "
+        + "function d() { a.b = {c: 1}; } e(a.b.c);",
+        "var a = {}; a.b = {c: 0}; function d() { a.b = {c: 1}; } e(a.b.c);");
```
dimvar: testSame?

In general, all tests without a $ in the output code should probably be testSame.

dimvar:
Looks great. I only had minor comments. I'll need to test this internally and then we can merge.

src/com/google/javascript/jscomp/CollapseProperties.java
dimvar: No, but the enumerated type can be an object, eg,

    /** @enum { { a: { b: number}} } */
    var e = {
      KEY1: { a: { /** @nocollapse */ b: 123}},
      KEY2: { a: { b: 456}}
    };

A more realistic usecase (and then one we'd like to prevent) is when people use enums as namespaces. They declare a bunch of "static" properties on the enum name after the definition. These properties can have any type. Sth like:

    /** @enum */
    var e = { A: 1, B: 2 };
    /** @type { c: { d: number } } */
    e.name1 = { c: { /** @nocollapse */ d: 123 } };

ChadKillingsworth: Ugh that's horrible. I can see why it should not be allowed.

ChadKillingsworth:
I've addressed the review issues, resquashed the commits and added the weird enum cases as an additional unit test.

dimvar:
Sounds good. I'm waiting for the results of the tests and then I'll let you know.

PR #823, opened 2015-02-11
Add externs for chrome.serial

abrander:
All methods and types from https://developer.chrome.com/apps/serial added to chrome_extensions.js.

I have resisted making types for the pseudo-types used in ```chrome.serial.send()```, ```chrome.serial.getControlSignals()```, ```chrome.serial.setControlSignals()``` and events as requested in the header of chrome_extensions.js.


MatrixFrog:
CLA confirmed.

MatrixFrog:
Some teams inside Google have written externs for chrome.serial already, so we'll probably have to submit this internally and delete those at the same time. I'll kick that off and hopefully it will be in `master` sometime next week. Thanks!

Issue #840, opened 2015-02-22
nullable type in externs file in HTMLSelectElement


myphysicslab:
With new type inference, I get errors like this one when using an `HTMLSelectElement`:

```
src/lab/controls/ParameterChoice.js:241: WARNING - Attempt to use nullable type HTMLOptionsCollection|null.
  this.selectMenu_.options.length = 0;
  ^
```

where `selectMenu_` is defined like this:

```javascript
  /**
  * @type {!HTMLSelectElement}
  * @private
  */
  this.selectMenu_ =
      /** @type {!HTMLSelectElement} */(document.createElement('select'));
  this.selectMenu_.type = 'select-one';
```

I don't see anything in the www.w3.org documentation stating that options can be null.  Should the extern file `w3c_dom2.js` be changed so that options is a non-nullable type?

```javascript
/**
 * @type {HTMLOptionsCollection}
 * @see http://www.w3.org/TR/2000/CR-DOM-Level-2-20000510/html.html#ID-30606413
 */
HTMLSelectElement.prototype.options;
```

Or do I need to test for null whenever using options?

ChadKillingsworth:
```JavaScript
var a = document.createElement('select');
a.options.length === 0; //not null
```

I think it can be safely marked as non-null. Would you like to create a pull request for this?

myphysicslab:
The pull request https://github.com/myphysicslab/closure-compiler/pull/1 has fixed this.

PR #853, opened 2015-02-25
make HTMLOptionsCollection non-nullable

myphysicslab:
[Issue #840] Make HTMLOptionsCollection non-nullable in externs for HTMLSelectElement.options.


MatrixFrog:
lgtm. I think we can merge this once Travis is done.

PR #867, opened 2015-03-03
Fix angular.bind jsdoc to return the actual function type

gberaudo:
Angular.bind was not returning the actual type of the passed function but
'Function'. This issue was addressed by using the jsdoc from goog.bind.

### DIFF
--- a/contrib/externs/angular-1.3.js
+++ b/contrib/externs/angular-1.3.js
@@ -58,10 +58,14 @@ var JQLiteSelector;
 var angular = {};
 
 /**
- * @param {Object} self
- * @param {Function} fn
- * @param {...*} args
- * @return {Function}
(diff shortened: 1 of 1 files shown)


gberaudo:
The few lines of code come from the goog project: do I need to sign a CLA for this?
See https://github.com/google/closure-library/blob/6f9347f4f22aec2db8d084056dec39d63b277c33/closure/goog/base.js#L1744

Is something else required?

MatrixFrog:
Yeah, we should still get a CLA for this.

It's probably worth mentioning that the compiler has special logic for Function.prototype.bind and goog.bind so if you use angular.bind your code will probably not be typechecked quite as thoroughly as it would with one of the other binds.

gberaudo:
Hi Tyler, the CLA has been submitted.

PR #868, opened 2015-03-03
Enable the use_types_for_optmization flag by default

### LINKED ISSUE #812: Turning on the typed-based optimizations by default
This seems to come up a lot. Unless you are well versed in the compiler project, the expectation is that type information would be used for renaming.

* http://stackoverflow.com/q/28318862/1211524
* http://stackoverflow.com/q/28196051/1211524
* http://stackoverflow.com/q/12671041/1211524

Is it time to turn on the type based optimizations by default?

ChadKillingsworth:
Enables type optimizations by default. Fixes #812.

Based off of the discussion at https://groups.google.com/forum/#!topic/closure-compiler-discuss/JPzrMiD_6dk

Question: Should we have a temporary warning or banner for the compiler when this would take effect?

I plan to update the FAQ and any other documentation pieces about this. There are also a few stackoverflow posts I'll edit. Someone should probably change the web service options to match as well (after release).

### DIFF
--- a/src/com/google/javascript/jscomp/CommandLineRunner.java
+++ b/src/com/google/javascript/jscomp/CommandLineRunner.java
@@ -359,12 +359,12 @@ public class CommandLineRunner extends
     private String compilationLevel = "SIMPLE";
     private CompilationLevel compilationLevelParsed = null;
 
-    @Option(name = "--use_types_for_optimization",
+    @Option(name = "--disable_type_optimizations",
         hidden = true,
-        usage = "Experimental: perform additional optimizations " +
-        "based on available information. Inaccurate type annotations " +
+        usage = "Disable the optimizations " +
+        "based on available type information. Inaccurate type annotations " +
         "may result in incorrect results.")
-    private boolean useTypesForOptimization = false;
+    private boolean disableTypeOptimizations = false;
 
     @Option(name = "--warning_level",
         aliases = {"-W"},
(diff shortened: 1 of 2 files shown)


MatrixFrog:
Can we change the default value of the flag without changing the name? I think it would be clearer.

ChadKillingsworth:
I thought of that, but there were several issues:

* Most people use the binary flags only for the true state. For instance, I've never seen someone do this:

        --debug false

* I was also trying to make it clearer that this was a significant change and thought that changing the default value of the flag would hide the change too much.

ChadKillingsworth:
Oh and the very name of the flag infers that it is opt-in rather than opt-out.

dimvar:
It is a common convention for a flag's name to describe what it does and then we have on and off states. As such, I agree with Tyler that the current name is more appropriate than disable_type_optimizations.

To avoid hiding the change, we can announce it at the appropriate lists/forums.

ChadKillingsworth:
I can make that change, but it will be the only boolean flag that is on by default.

ChadKillingsworth:
Nevermind - I don't think boolean flags can be disabled like that with args4j. http://args4j.kohsuke.org/args4j/apidocs/org/kohsuke/args4j/Option.html

None of the following worked:

    --use_types_for_optimization false
    --use_types_for_optimization=false
    --use_types_for_optimization=0

MatrixFrog:
Well that's annoying :|

Okay, I guess changing the name makes sense then.

ChadKillingsworth:
Reverted to the original option name and it is now on by default.

dimvar:
lgtm once the travis build is done

MatrixFrog:
lgtm

Issue #433, opened 2014-05-22
Reinstate --variable_map_output_file


cugh:
The --variable_map_output_file command line option has been dropped in v20140508.

This was a very useful option. I'm getting bug reports with obfuscated stack trace all the time and I used the variable map to quickly figure out what went wrong. The variable map was very easy to use because the original and obfuscated name are on the same line.

The source map file is not useful at all, as it is nearly impossible to open it in an editor and manually figure out quickly what is the original name of an obfuscated variable.

Could you please reinstate the --variable_map_output_file?

PR #869, opened 2015-03-04
Obfuscation exclusions

supersteves:
Allowing exclusion of properties from obfuscation using "old:old"
mapping. Wasn't working when the same property existed elsewhere as a
quoted string.

We use input renaming maps to preserve names of a large number of 'schema' objects which are exchanged elsewhere.  We supply e.g. "manifest:manifest" to ensure that particular object member is never renamed.

However in rare cases it ignores this, and you see "manifest:xyz" in the output. This fixes it.

### DIFF
--- a/src/com/google/javascript/jscomp/RenameProperties.java
+++ b/src/com/google/javascript/jscomp/RenameProperties.java
@@ -289,7 +289,9 @@ class RenameProperties implements CompilerPass {
       String prevName = prevUsedPropertyMap.lookupNewName(prop.oldName);
       if (!generatePseudoNames && prevName != null) {
         // We can reuse prevName if it's not reserved.
-        if (reservedNames.contains(prevName)) {
+      	// HOWEVER, if prevName is an "exclude obfuscation" mapping (matches oldName),
+      	// retain it, to ensure the exclusion remains:
+        if (reservedNames.contains(prevName) && !prevName.equals(prop.oldName)) {
           continue;
         }
 


dimvar:
Don't use renaming maps to prevent renaming. The compiler does not guarantee that it will respect the renaming map. See the last paragraph in this post.
http://closuretools.blogspot.com/2011/01/property-by-any-other-name-part-3.html

ChadKillingsworth:
For reference, I believe this was the last official discussion: #433 

In this case though I believe that the compiler was doing the correct thing. It shouldn't rename a property to something that conflicts with an extern property.

@dimvar is correct though - you are using renaming maps for an unintended purpose. If you can generate renaming maps, you should be able to generate externs to solve the same problem.

Using your above example, the following will disable property renaming for specific property names as long as the type based optimizations are **not** enabled:

```JavaScript
/** @externs */

/** @const */
var myUnusedNamespace = {};
myUnusedNamespace.manifest;
myUnusedNamespace.old;
```

concavelenz:
This should work regardless of type based optimizations:

```JavaScript
/** @externs */

/** @type {Object} */
var props;

props.manifest;
props.old;
```

supersteves:
I'm using the approach suggested above from @concavelenz. But now a property "start" has appeared in my schema, so my externs look like this:
``` javascript
/** @type {Object} */
var xxx;

xxx.start;
```
This leads to a fairly obscure warning:
```
extern:11: WARNING - property start already defined on interface EventTarget; use @override to override it
MessagePort.prototype.start = function() {};
^
```
Presumably writing @ type {Object} for xxx means I am actually augmenting the type definition for "Object". Hence Closure Compiler now thinks all objects have this property? In this case, the conflicts are with these default externs:
``` javascript
/**
 * @interface
 */
function EventTarget() {}

/**
 * @constructor
 * @implements {EventTarget}
 */
function MessagePort() {}
MessagePort.prototype.start = function() {};
```

https://closure-compiler-debugger.appspot.com/#input0%26input1%26conformanceConfig%26externs%3D%252F**%250A%2520*%2520%2540interface%250A%2520*%252F%250Afunction%2520EventTarget()%2520%257B%257D%250A%250A%252F**%250A%2520*%2520%2540constructor%250A%2520*%2520%2540implements%2520%257BEventTarget%257D%250A%2520*%252F%250Afunction%2520MessagePort()%2520%257B%257D%250AMessagePort.prototype.start%2520%253D%2520function()%2520%257B%257D%253B%250A%250A%250A%252F**%2520%2540type%2520%257BObject%257D%2520*%252F%250Avar%2520xxx%253B%250A%250Axxx.start%253B%26refasterjs-template%26CHECK_REPORT_MISSING_OVERRIDE%3D1%26CHECK_TYPES%3D1

Do you have any suggestions to resolve this warning?

PR #884, opened 2015-03-19
Deprecate the @expose annotation

ChadKillingsworth:
With `@nocollapse` now supported, `@expose` can be deprecated. The compiler will only warn on the first occurrence of `@expose`.

I currently have this as part of the `nonStandardJsDocs` diagnostic group. I'm not completely convinced that's the correct place for it.

### DIFF
--- a/src/com/google/javascript/jscomp/parsing/IRFactory.java
+++ b/src/com/google/javascript/jscomp/parsing/IRFactory.java
@@ -196,6 +196,9 @@ class IRFactory {
 
   static final String UNDEFINED_LABEL = "undefined label \"%s\"";
 
+  static final String ANNOTATION_DEPRECATED =
+      "The %s annotation is deprecated.%s";
+
   private final String sourceString;
   private final List<Integer> newlines;
   private final StaticSourceFile sourceFile;
@@ -470,6 +473,24 @@ class IRFactory {
     validateTypeAnnotations(n);
     validateFunctionJsDoc(n);
     validateMsgJsDoc(n);
+    validateDeprecatedJsDoc(n);
+  }
+
+  /**
+   * Checks that deprecated annotations such as @expose are not present
+   */
+  private void validateDeprecatedJsDoc(Node n) {
+    JSDocInfo info = n.getJSDocInfo();
+    if (info == null) {
+      return;
+    }
+    if (info.isExpose()) {
+      errorReporter.warning(
+          String.format(ANNOTATION_DEPRECATED, "@expose",
+              " Use @nocollapse or @export instead."),
+          sourceName,
+          n.getLineno(), n.getCharno());
+    }
   }
 
   /**
--- a/src/com/google/javascript/jscomp/DiagnosticGroups.java
+++ b/src/com/google/javascript/jscomp/DiagnosticGroups.java
@@ -82,19 +82,18 @@ public class DiagnosticGroups {
   // If a group is suppressible on a per-file basis, it should be added
   // to parser/ParserConfig.properties
   static final String DIAGNOSTIC_GROUP_NAMES =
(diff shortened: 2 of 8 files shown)


dimvar:
Let's give it until next week to make sure the nocollapse change sticks before we merge this one. I can do the review.

src/com/google/javascript/jscomp/Normalize.java
```diff
@@ -73,6 +73,12 @@
         "The use of scope variable {0} is not allowed within a catch block " +
         "with a catch exception of the same name.");
```
dimvar: 1) Warnings should happen in checks; normalize is run too late. Move the warning to IRFactory.java.
2) Warn for every use of <code>@expose</code>, not just the first one.

ChadKillingsworth:
Refactored and squashed commits. Warning is now a parse warning in IRFactory for every occurrence.

src/com/google/javascript/jscomp/parsing/IRFactory.java
```diff
@@ -196,6 +196,10 @@
 
   static final String UNDEFINED_LABEL = "undefined label \"%s\"";
 
+  static final String EXPOSE_DEPRECATED_WARNING =
```
dimvar: Modify this file
https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/RhinoErrorReporter.java
to match the text of the warning, and create a DiagnosticGroup, so that the warning can be suppressed. Add your new diagnostic here.
https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/DiagnosticGroups.java#L127

src/com/google/javascript/jscomp/parsing/IRFactory.java
```diff
@@ -470,6 +474,22 @@ private void validateJsDoc(Node n) {
     validateTypeAnnotations(n);
     validateFunctionJsDoc(n);
     validateMsgJsDoc(n);
+    validateDeprecatedJsDoc(n);
+  }
+
+  /**
+   * Checks that deprecated annotations such as @expose are not present
+   */
+  private void validateDeprecatedJsDoc(Node n) {
```
dimvar: add a unit test for this in
https://github.com/google/closure-compiler/blob/master/test/com/google/javascript/jscomp/parsing/NewParserTest.java

src/com/google/javascript/jscomp/parsing/IRFactory.java
ChadKillingsworth: Done. It's not suppressible with an `@suppress` annotation. I'm assuming that's because it's a parser warning.

src/com/google/javascript/jscomp/parsing/IRFactory.java
ChadKillingsworth: Done.

src/com/google/javascript/jscomp/parsing/IRFactory.java
dimvar: It should be suppressible. What's the example program that you tried? Can you try suppressing other warnings in IRFactory and see what happens?

ChadKillingsworth: I've tried both in the `@fileoverview` block and at a function. Neither worked. I got the same behavior with the "nonStandardJsDocs" group.

Test case:
```JavaScript
/**
 * @fileoverview
 * @suppress {nonStandardJsDocs}
 */
function f1() {
  /**
   * @foo
   * @expose
   */
  window.x = true;
}
```

Both work when using the `--jscomp_off` flag.

dimvar: OK, sounds good. LGTM then. Can you squash the commits?

ChadKillingsworth:
Squashed

Issue #381, opened 2014-04-29
Feature request: dead code removal without property renaming


blickly:
This issue was imported from Closure Compiler's previous home at http://closure-compiler.googlecode.com

The original discussion is archived at:
http://blickly.github.io/closure-compiler-issues/#183

donmccurdy:
I see that #886 has been closed, but this issue is still open. Would a PR that exposes only the OFF and (default) ALL_UNQUOTED policies be accepted? 

Setting CompilationLevel does not work in my case; SIMPLE still modifies too much, and WHITESPACE_ONLY results in code that does not run, like this:

```js
goog.provide('goog.disposable.IDisposable');
goog.disposable.IDisposable = ...
// → TypeError: Cannot set property 'IDisposable' of undefined
```

I have a project that needs to do some inspection of function parameter names, e.g. `fn.toString()`, and any renaming whatsoever will break this. This is similar, for example, to the AngularJS injection pattern. I'm fine with the additional filesize overhead; in the contexts where this function inspection is needed, it's acceptable.

Proposed syntax:

```
--property_renaming ALL_UNQUOTED | OFF
```

Thanks

ChadKillingsworth:
Property renaming only affects `ADVANCED` mode.

If `WHITESPACE_ONLY` mode isn't working, then you have either encountered a bug or more probably are not using the correct set of flags. I recommend posting in the [Google Group](https://groups.google.com/forum/#!forum/closure-compiler-discuss) or on [Stackoverflow](http://stackoverflow.com/questions/tagged/google-closure-compiler).

PR #886, opened 2015-03-21
Added force_property_renaming_policy flag.

iyegoroff:
Exposed property renaming policy as a command line flag '--force_property_renaming_policy' to resolve issue #381. 

Suggested usage:
java -jar compiler.jar --compilation_level ADVANCED --force_property_renaming_policy OFF

### DIFF
--- a/src/com/google/javascript/jscomp/CommandLineRunner.java
+++ b/src/com/google/javascript/jscomp/CommandLineRunner.java
@@ -584,6 +584,16 @@ public class CommandLineRunner extends
         hidden = true,
         usage = "A list of JS Conformance configurations in text protocol buffer format.")
     private List<String> conformanceConfigs = new ArrayList<>();
+    
+    @Option(name = "--force_property_renaming_policy",
+            hidden = true,
+            usage = "Forces specified property renaming policy. Options: " +
+                "OFF, " +
+                "HEURISTIC, " +
+                "AGGRESSIVE_HEURISTIC, " +
+                "ALL_UNQUOTED")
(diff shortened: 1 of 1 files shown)


iyegoroff:
@googlebot OK, CLA signed.

ChadKillingsworth:
The heuristic based property renaming policies are not used. They should not be available. The `ALL_UNQUOTED` policy is the actively developed renaming policy. You might want to read about the renaming policy in the blog posts referenced here: https://github.com/google/closure-compiler/wiki/FAQ#some-of-my-properties-are-getting-renamed-but-some-arent-why

Also, what is your use case for doing this?

iyegoroff:
>Also, what is your use case for doing this?

I was converting some three.js examples to Clojurescript and performance of unoptimized code was quite bad. So I need an option to remove unused parts of Clojurescript runtime and avoid renaming properties of three.js objects.

dimvar:
Heuristic property renaming has been deprecated for years. The only reason we haven't deleted it yet is that we haven't found the time to move legacy clients (at google) off of it.

dimvar:
Try simple optimizations, they don't include property renaming.

ChadKillingsworth:
You might also try disabling property renaming here: https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/DefaultPassConfig.java#L665

And then try leaving the type based optimizations on. I've experimented with this configuration before. It should only rename properties for which it knows the type. In theory, external properties in three.js will be left alone.

PR #916, opened 2015-04-22
Add Navigator.prototype.msMaxTouchPoints property in IE extern

fredj:
Note: as of Internet Explorer 11, this property is deprecated and
Navigator.prototype.maxTouchPoints should be used instead (defined in
contrib/externs/w3c_pointer_events.js).


ChadKillingsworth:
:+1:

PR #921, opened 2015-04-25
Add hashCode and equals method. Removing the need for a wrapper class wh...

lennel:
...en populating SourceFiles in any hash structure, avoiding a "cache miss" situation.


dimvar:
Closing due to inactivity

PR #928, opened 2015-05-01
Fix externs for FontFace constructor

bramstein:
According to the W3C CSS font loading specification (http://dev.w3.org/csswg/css-font-loading/#font-face-constructor) the FontFace constructor takes three types: `string`, `ArrayBuffer`, and `ArrayBufferView`. The current version of the w3c_css externs file only lists `string`. This pull request adds `ArrayBuffer` and `ArrayBufferView` so it follows the specification.


bramstein:
I signed it!

blickly:
Thanks for the fix!

Issue #636, opened 2014-09-24
@expose annotation causes JSC_UNSAFE_NAMESPACE warning


meyertee:
When compiling the following code in advanced mode it produces a `JSC_UNSAFE_NAMESPACE` warning. The problem appears when exposing a field that has a name that's equal to a component in a namespace, even though they are unrelated. Setting the `--use_types_for_optimization` makes no difference.

    // ==ClosureCompiler==
    // @compilation_level ADVANCED_OPTIMIZATIONS
    // @output_file_name default.js
    // @formatting pretty_print
    // ==/ClosureCompiler==
    
    goog.provide("com.whatever.A");
    goog.provide("com.problem.B");

    /**
     * @constructor
     */
    com.whatever.A = function () {
    	/**
    	 * @expose
    	 * @type {string}
    	 */
    	this.problem = "A here!";
    };

    /**
     * @constructor
     */
    com.problem.B = function () {
    	/**
    	 * @type {string}
    	 */
    	this.problem = "B here!";
    };

    var a = new com.whatever.A();
    var b = new com.problem.B();

    console.log(a.problem, b.problem);

You can paste the code into http://closure-compiler.appspot.com/home to see the warnings.

moderndegree:
+1 --export_local_property_definitions would be very helpful

moderndegree:
Any word on this? @meyertee Could you share the changes you made to CommandLineRunner?

moderndegree:
@meyertee I think I got it figured out. I just submitted a pull request that should solve the problem. It appears to be working now.

dpashkevich:
Thanks for merging in the `--export_local_property_definitions` flag into the published jar! I'm working on an Angular app and would really benefit from it. For now I'm using the deprecated `@expose` in the same way. I assume it will come out with the next release?

I have another quesion. The [guide on using Angular with Closure Compiler](http://google-styleguide.googlecode.com/svn/trunk/angularjs-google-style.html#compilerflags) also mentions:

> If you are using @export for properties, you will need to add the flags:
> "--generate_exports",
> "--remove_unused_prototype_props_in_externs=false",
> "--export_local_property_definitions",

It looks like the `remove_unused_prototype_props_in_externs` isn't present in the published compiler jar either, do I actually need it? Sorry if this is a lame quesion, I don't have a lot of experience with the closure compiler.

moderndegree:
[...]

PR #930, opened 2015-05-04
[WIP] Improved @export support

ChadKillingsworth:
The only method for exporting code using `@export` is through shadowing. Ideally, we would also have a method to export that simply doesn't rename.

Past discussions related to this:

 * https://code.google.com/p/closure-compiler/issues/detail?id=50
 * https://github.com/google/closure-compiler/issues/636 - See the end of the discussion

### DIFF
--- a/src/com/google/javascript/jscomp/CommandLineRunner.java
+++ b/src/com/google/javascript/jscomp/CommandLineRunner.java
@@ -388,11 +388,12 @@ public class CommandLineRunner extends
         usage = "Enable debugging options")
     private boolean debug = false;
 
-    @Option(name = "--generate_exports",
+    @Option(name = "--exports_mode",
         hidden = true,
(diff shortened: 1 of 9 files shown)


ChadKillingsworth:
Design document for `@export` improvements: https://gist.github.com/ChadKillingsworth/b86a4cffaa71571b5d01

I've currently only documented the current behavior, but I believe I have captured the majority of use cases. I still need to flush out the proposed implementation.

ChadKillingsworth:
I have updated the design doc with my recommended implementation.

dimvar:
Thanks Chad. I will take a look later this week.

ChadKillingsworth:
Based on discussion, I'm closing this PR. Instead I'll work on support for an `@exernal` annotation.

smudgeface:
Is there information on @external and how this annotation will address the concerns behind this PR?

PR #936, opened 2015-05-07
Add ServiceWorker externs to html5.js

nicks:


externs/html5.js
```diff
@@ -3242,3 +3242,104 @@ function HTMLTemplateElement() {}
...
  */
 HTMLTemplateElement.prototype.content;
+
+
+/**
+ * @constructor
+ * @extends {Worker}
+ * @see http://www.w3.org/TR/service-workers/#service-worker-obj
+ */
+function ServiceWorker() {}
```
ChadKillingsworth: Should this be an interface? Would you ever call "new ServiceWorker()"?

nicks: i don't think it can be, because it extends Worker?

externs/html5.js
```diff
@@ -3242,3 +3242,104 @@ function HTMLTemplateElement() {}
...
+ServiceWorker.prototype.onstatechange;
+
+
+
+/**
+ * @constructor
+ * @implements {EventTarget}
+ * @see http://www.w3.org/TR/service-workers/#service-worker-registration-obj
+ */
+function ServiceWorkerRegistration() {}
```
ChadKillingsworth: Is there a reason to omit the `installing`, `waiting` and `active` properties?

externs/html5.js
```diff
@@ -3242,3 +3242,104 @@ function HTMLTemplateElement() {}
...
+ServiceWorkerContainer.prototype.addEventListener;
+
+/** @override */
+ServiceWorkerContainer.prototype.removeEventListener;
+
+/** @override */
+ServiceWorkerContainer.prototype.dispatchEvent;
+
+/** @type {!ServiceWorker} */
+ServiceWorkerContainer.prototype.controller;
```
ChadKillingsworth: I think this is supposed to be nullable.

externs/html5.js
```diff
@@ -3242,3 +3242,104 @@ function HTMLTemplateElement() {}
...
+
+/**
+ * @param {string} scriptURL
+ * @param {{scope: string}=} options
+ * @return {!Promise<ServiceWorkerRegistration>}
+ */
+ServiceWorkerContainer.prototype.register = function (scriptURL, options) {};
+
+/**
+ * @param {string} clientURL
```
ChadKillingsworth: This is an optional param

externs/html5.js
ChadKillingsworth: You are correct. I thought that might be allowed in externs, but no:

> JSC_CONFLICTING_EXTENDED_TYPE: ServiceWorker cannot extend this type; interfaces can only extend interfaces at line 11 character 0
function ServiceWorker() {}

MatrixFrog:
We already have some ServiceWorker externs in our internal repo. Didn't realize that we forgot to put them on the open-source-able list.

nicks:
ah, ok :+1:

PR #955, opened 2015-05-18
Apply pull request #768 by @kayahr.

MatrixFrog:
Decouple jsdoc documentation parsing from IDE mode.

Currently the parsing of JSDoc documentation is coupled to the IDE mode
option so normally JSDoc documentation strings are not parsed.  This
prevents the implementation of custom compiler passes like API Documentation
generation.  This patch introduces a new compiler option
(parseJsDocDocumentation) which can be used to enable the documentation
parsing even when not in IDE mode.  The functionality is backward
compatible, when IDE mode is enabled then documentation parsing is always
enabled)


MatrixFrog:
Just making a new pull request since the other one can't be auto-merged.

PR #959, opened 2015-05-21
Add mockDate function to jasmine clock extern

calebegg:


calebegg:
I am covered by the Google Corporate CLA (I am a Google employee).

MatrixFrog:
In that case, it's preferable to send changes internally so we don't have to re-LGTM them when MOE pulls them in.

But if you do send changes internally, I've found that if the commit has your @google.com address the googlebot is happy.

MatrixFrog:
Er, I meant, if you do send changes _externally_ of course.

calebegg:
Sent. Thanks for the heads up.

PR #987, opened 2015-06-04
[WIP] Jquery externs cleanup

ChadKillingsworth:
Due to the removal of `jQuery.event`, this has a high likelihood of breaking existing projects due to now incorrect type annotations. This change is however necessary. See discussion in #767

I've marked this as a Work in Progress because I would like to find users with large jQuery code bases to test this against first.

I'm unsure what impact this will have internal to Google.

### DIFF
--- a/contrib/externs/jquery-1.9.js
+++ b/contrib/externs/jquery-1.9.js
@@ -83,7 +83,7 @@ var jQueryAjaxSettings;
  * @param {(jQuerySelector|Element|Object|Array<Element>|jQuery|string|
  *     function())=} arg1
  * @param {(Element|jQuery|Document|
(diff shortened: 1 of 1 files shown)


dimvar:
Can you split this in two PRs? The changes that are safe (eg, the change of Foo.<Bar> to Foo<Bar>) can land easily. And the 2nd one will only have the serious changes, with a smaller diff.

ChadKillingsworth:
Sure thing.

ChadKillingsworth:
I'm actually going to split this up even further into two more PRs. Closing this one.

ChadKillingsworth:
#990 has fixes to the `jQuery.Event` object.

PR #990, opened 2015-06-05
Cleanup types related to the jQuery Event objects.

### LINKED ISSUE #876: jQuery externs for event handlers are too loose
For example, `$.on` is annotated as follows:

```
/**
 * @param {(string|Object.<string,*>)} arg1
 * @param {*=} selector
 * @param {*=} data
 * @param {function(!jQuery.event=)=} handler
 * @return {!jQuery}
 */
jQuery.prototype.on = function(arg1, selector, data, handler) {};
```

The event handler is defined as taking an argument of type `!jQuery.event=`, or `jQuery.event|undefined`.  But I can't think of any circumstance in which jQuery _wouldn't_ pass an event object to an event handler...  so why is the `=` there?

I'm trying to upgrade Closure Compiler with an existing annotated codebase, and this is causing a slew of problems in code like:

```
/**
 * @param {jQuery.event} e
 **/
function quux(e) { ... }

$(...).on('.foo', '.bar', {}, function(e) {
    quux(e);
});
```

The compiler now rejects this, because according to the externs, the argument to the callback might be `undefined`.  But that should never happen.  Am I missing something?

ChadKillingsworth:
jQuery.event is a mostly internal namespace - not a type
jQuery.Event is the correct type. Added missing properties.

Any type annotation for `jQuery.event` should be replaced with `jQuery.Event`. I don't forsee any other conflicts with the type name changes.

Event listener callbacks no longer have the `jQuery.Event` argument marked as optional. Fixes #876

### DIFF
--- a/contrib/externs/jquery-1.9.js
+++ b/contrib/externs/jquery-1.9.js
@@ -83,7 +83,7 @@ var jQueryAjaxSettings;
  * @param {(jQuerySelector|Element|Object|Array<Element>|jQuery|string|
  *     function())=} arg1
  * @param {(Element|jQuery|Document|
- *     Object<string, (string|function(!jQuery.event=))>)=} arg2
+ *     Object<string, (string|function(!jQuery.Event))>)=} arg2
  * @return {!jQuery}
  */
 function jQuery(arg1, arg2) {}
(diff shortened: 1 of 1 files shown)


dimvar:
Merging with the caveat that we may have to roll back if it breaks someone at Google.

ChadKillingsworth:
I'll block any other jQuery changes for a week or so to see how this goes.

PR #998, opened 2015-06-10
Added pass-through of angularPass option to ant task

wolfgangmeyers-donuts:
Although the "--angular-pass" option is available when invoking the
compiler from the command line, it was not yet available in the
ant plugin. This change adds support for angular pass by accepting
it as an attribute and passing in the compiler options.


ChadKillingsworth:
:+1:

Issue #984, opened 2015-06-03
Warn about placing @nocollapse on a prototype property


dimvar:
Folks at Google are starting to use nocollapse in their projects, and they are putting it in all sorts of places. The following should be a warning, that the annotation does nothing, since collapse-properties doesn't look at prototype properties anyway.

    /** @constructor */
    function Foo() {}
    
    /** @nocollapse */
    Foo.prototype.dprop = 234;
    
    var y = Foo.prototype.dprop;

ChadKillingsworth:
This isn't as straight forward as I had hoped. Since the CollapseProperties pass uses the GlobalNamespace class, prototype properties are ignored.

I can try to shoe horn this into GlobalNamespace with a flag to indicate whether warnings should be emitted or not (since GlobalNamespace is used in multiple passes), or I can try to find another pass to issue the warning from.

Thoughts?

blickly:
I'd prefer not to put more warnings into the optimizations phase. Especially not something like this that can just be checked syntactically regardless of whether the properties are actually being collapsed.

Maybe CheckJSDoc.java ?

dimvar:
I was gonna suggest parsing/IRFactory.java, but I very much like the idea of CheckJSDoc.java.

It's good to have a pass that detects misplaced jsdocs. It is quite common that the passes who need a jsdoc annotation only look at the nodes that might have it, so they don't warn about the annotation placed on irrelevant nodes. It can be the first check in DefaultPassConfig#getChecks, always on (without a compiler option).

ChadKillingsworth:
That's about where I had landed as well. I'll start work on the CheckJSDoc.java pass then.

ChadKillingsworth:
Closed by #1006

PR #1006, opened 2015-06-16
Add CheckJSDoc pass to verify annotations are used appropriately

ChadKillingsworth:
Moves most JSDoc warnings out of IRFactory and makes them supressible.

It seems like the diagnostic groups related to JSDocs should be cleaned up:

* deprecatedAnnotations
* misplacedTypeAnnotation

Perhaps combine to `invalidJsDoc`?

### DIFF
--- a/src/com/google/javascript/jscomp/CheckJSDoc.java
+++ b/src/com/google/javascript/jscomp/CheckJSDoc.java
@@ -1,0 +1,253 @@
+/*
+ * Copyright 2015 The Closure Compiler Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.javascript.jscomp;
+
+import com.google.javascript.jscomp.NodeTraversal.AbstractPostOrderCallback;
+import com.google.javascript.rhino.JSDocInfo;
+import com.google.javascript.rhino.Node;
+import com.google.javascript.rhino.Token;
+
+/**
+ * Checks for misplaced, misused or deprecated JSDoc annotations.
+ *
+ * @author chadkillingsworth@gmail.com (Chad Killingsworth)
+ */
+final class CheckJSDoc implements CompilerPass {
+
+  public static final DiagnosticType MISPLACED_ANNOTATION =
+      DiagnosticType.warning("JSC_MISPLACED_ANNOTATION",
+          "Misplaced {0} annotation.{1}");
+
+  public static final DiagnosticType ANNOTATION_DEPRECATED =
+      DiagnosticType.warning("JSC_ANNOTATION_DEPRECATED",
+          "The {0} annotation is deprecated.{1}");
+
+  private final AbstractCompiler compiler;
+
+  CheckJSDoc(AbstractCompiler compiler) {
+    this.compiler = compiler;
+  }
+
+  @Override
+  public void process(Node externs, Node root) {
+    NodeTraversal.traverse(compiler, externs, new ProcessNodes());
+    NodeTraversal.traverse(compiler, root, new ProcessNodes());
+  }
+
+  final class ProcessNodes extends AbstractPostOrderCallback {
+
+    @Override
+    public void visit(NodeTraversal t, Node n, Node parent) {
+      JSDocInfo info = n.getJSDocInfo();
+      validateTypeAnnotations(t, n, info);
+      validateFunctionJsDoc(t, n, info);
+      validateMsgJsDoc(t, n, info);
+      validateDeprecatedJsDoc(t, n, info);
+      validateNoCollapse(t, n, info);
+    }
+
+    /**
+     * Checks that deprecated annotations such as @expose are not present
+     */
+    private void validateDeprecatedJsDoc(NodeTraversal t, Node n,
+        JSDocInfo info) {
+      if (info != null && info.isExpose()) {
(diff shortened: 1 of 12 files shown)


ChadKillingsworth:
Fixes #984

src/com/google/javascript/jscomp/CheckJSDoc.java
```diff
@@ -0,0 +1,274 @@
...
+      }
+    }
+
+    /**
+     * Checks that JSDoc intended for a function is actually attached to a
+     * function.
+     */
+    private void validateNoCollapse(NodeTraversal t, Node n) {
+      JSDocInfo info;
+      if (this.isExternsPass_) {
```
dimvar: You don't need this field. You can just use n.isFromExterns()

https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/rhino/Node.java#L1225

src/com/google/javascript/jscomp/CheckJSDoc.java
```diff
@@ -0,0 +1,274 @@
...
+        t.getCompiler().report(
+            t.makeError(n, ANNOTATION_DEPRECATED, "@expose",
+                " Use @nocollapse or @export instead."));
+      }
+    }
+
+    /**
+     * Checks that JSDoc intended for a function is actually attached to a
+     * function.
+     */
```
dimvar: wrong jsdoc here

src/com/google/javascript/jscomp/CheckJSDoc.java
```diff
@@ -0,0 +1,274 @@
...
+        // @nocollapse should be attached to a property node or an object
+        // literal key.
+        switch (n.getType()) {
+          case Token.GETELEM:
+          case Token.GETPROP:
+            if (NodeUtil.isPrototypeProperty(n)) {
+              t.getCompiler().report(t.makeError(n, MISPLACED_ANNOTATION,
+                  "@nocollapse", "This JSDoc has no effect on prototype"
+                      + "properties."));
+              break;
```
dimvar: this break can be removed

src/com/google/javascript/jscomp/CheckJSDoc.java
```diff
@@ -0,0 +1,274 @@
...
+    }
+
+    /**
+     * Checks that JSDoc intended for a function is actually attached to a
+     * function.
+     */
+    private void validateNoCollapse(NodeTraversal t, Node n) {
+      JSDocInfo info;
+      if (this.isExternsPass_) {
+        info = n.getJSDocInfo();
```
dimvar: You can start the method like this:

JSDocInfo info = n.getJSDocInfo();
if (info == null || !info.isNoCollapse()) {
  return;
}

and then remove the corresponding checks from various places in the method

src/com/google/javascript/jscomp/CheckJSDoc.java
```diff
@@ -0,0 +1,274 @@
...
+              "message", " @desc, @hidden, and @meaning annotations should only"
+                  + "be on message nodes."));
+        }
+      }
+    }
+
+    /**
+     * Check that JSDoc with a {@code @type} annotation is in a valid place.
+     */
+    @SuppressWarnings("incomplete-switch")
```
dimvar: it seems that you can just add a default: case for the switch and remove this suppression

test/com/google/javascript/jscomp/parsing/NewParserTest.java
```diff
@@ -538,10 +532,9 @@ public void testJSDocAttachment5() {
    * propagate to following code due to {@link JSDocInfo} aggregation.
    */
   public void testJSDocAttachment6() throws Exception {
```
dimvar: For every test you've added in the new file, delete it from here if it was only testing bad jsdocs

src/com/google/javascript/jscomp/parsing/IRFactory.java
```diff
@@ -143,15 +143,6 @@
       "Non-JSDoc comment has annotations. " +
```
dimvar: https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/RhinoErrorReporter.java#L116

In RhinoErrorReporter, we turn some parser warnings into diagnostic-types by regex matching the warning text :). Now that you've moved some of these to checkjsdoc, you can remove the relevant parts of the code from rhinoerrorreporter.

dimvar:
Looks good. I left some comments.

src/com/google/javascript/jscomp/CheckJSDoc.java
ChadKillingsworth: I'm not sure what you mean.

test/com/google/javascript/jscomp/parsing/NewParserTest.java
ChadKillingsworth: That test specifically looked to be testing more than simple misplacement of the JSDoc. It seemed to be testing whether misplaced annotations affected later code which is why I left it.

ChadKillingsworth:
Fixed most of these, but had some followup questions.

Also, what do you think about combining the diagnostic groups to simply `invalidJsDoc`?

src/com/google/javascript/jscomp/CheckJSDoc.java
dimvar: It's copied over from validateFunctionJsDoc, but you forgot to edit it to talk about nocollapse

test/com/google/javascript/jscomp/parsing/NewParserTest.java
dimvar: OK, that's good. I just made a general comment for the tests being rewritten here. If some of them can be deleted, just go ahead and remove them.

dimvar:
> what do you think about combining the diagnostic groups to simply invalidJsDoc?

No, let's leave them separate. I can see people wanting to ignore the deprecation warnings, but keep the misplaced jsdoc warnings.

The misplaced jsdoc warnings are our version of parse errors, for the syntax we'd like to have but doesn't exist in JS.

ChadKillingsworth:
Done - commits squashed. It took three tries to get all the merge conflicts handled as other conflicting commits were being pushed while I was trying to get Travis to finish.

dimvar:
Thanks Chad! This will be a useful pass.

PR #1009, opened 2015-06-17
Fixed record type truncation in generated externs

LorenzoDV:
When automatically generating an externs file for exported symbols, the compiler will truncate longer record types used as type arguments.

Given the following input code:
```javascript
/**
 * @export
 * @param {!Array<{a:string, b:string, c:string, d:string, e:string, f:string}>} arrayOfObjects
 * @return {Object}
 */
function test(arrayOfObjects) {
	// ...
	return null;
}
```
this is the generated externs file:
```javascript
/**
 * @param {Array<{a: string, b: string, c: string, d: string, ...}>} arrayOfObjects
 * @return {(Object|null)}
 */
var test = function(arrayOfObjects) {
};
```

Note how properties after the first four are dropped from the annotation.

### DIFF
--- a/src/com/google/javascript/rhino/jstype/TemplatizedType.java
+++ b/src/com/google/javascript/rhino/jstype/TemplatizedType.java
@@ -39,9 +39,11 @@
 
 package com.google.javascript.rhino.jstype;
 
+import com.google.common.base.Function;
 import com.google.common.base.Joiner;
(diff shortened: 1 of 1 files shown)


MatrixFrog:
Oh nice. Thanks! lgtm. Just waiting for Travis to finish.

PR #1016, opened 2015-06-19
Emit a warning message when the compiler is waiting for input

### LINKED ISSUE #1008: CommandLineRunner should display a message stating it is waiting for input
I wanted to use cc for optimizing javascript code written using commonjs modules. But when invoking cc, the compiler just loops and outputs nothing.

I then created a very simple test, it' only one file, with no requires and only one export:

```
$ cat cc-test.js
function genTs(date) {
    var d = date || new Date();
    var pad = function (n) {
        return (n < 10 ? '0' + n : n.toString());
    };
    return d.getUTCFullYear() + '-' + pad(d.getUTCMonth() + 1) + '-' + pad(d.getUTCDate()) + 'T' +
        pad(d.getUTCHours()) + ':' + pad(d.getUTCMinutes()) + ':' + pad(d.getUTCSeconds()) + 'Z';
}
exports.genTs = genTs;
```

But the compiler still runs forever, without producing any output (and without consuming much CPU time):

```
$ java -jar ./compiler.jar --compilation_level=WHITESPACE_ONLY --logging_level=FINEST --process_common_js_modules --common_js_entry_module cc-test.js
```

I guess I'm using the latest stable build:

```
$ java -jar compiler.jar --version
Closure Compiler (http://github.com/google/closure-compiler)
Version: v20150609
Built on: 2015/06/09 16:35
```

ChadKillingsworth:
Fixes #1008

### DIFF
--- a/src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
+++ b/src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
@@ -103,6 +103,9 @@ abstract class AbstractCommandLineRunner<A extends Compiler,
       "JSC_NO_TREE_GENERATED_ERROR",
       "Code contains errors. No tree was generated.");
 
+  static final String WAITING_FOR_INPUT_WARNING =
+      "The compiler is waiting for input via stdin.";
+
   private final CommandLineConfig config;
 
   private final PrintStream defaultJsOutput;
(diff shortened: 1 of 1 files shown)


dimvar:
This is too complex. Just emit the text that the compiler is waiting for input immediately when <code>usingStdin = true;</code>

ChadKillingsworth:
Done.

Issue #979, opened 2015-05-29
sourceMappingURL # vs. @


bardware:
[Google's own document](https://docs.google.com/document/d/1U1RGAehQwRypUTovF1KRlpiOFze0b-_2gc6fAH0KY0k/edit#) reads: 

```
The prefix for this annotation was initially “//@” however this conflicts with
Internet Explorer’s Conditional Compilation and was changed to “//#”.
It is reasonable for tools to also accept “//@” but “//#” is preferred.
```
Will you accordingly change the ant task `\closure-compiler\src\com\google\javascript\jscomp\ant\CompileTask.java` lin 378 to `source.append("//# sourceMappingURL=" + sourceMapOutputFile.getName());`

concavelenz:
Please submit a pull request.  The ant task is maintained by the community.

bardware:
I didn't want to create a pull request for just one character but I had another issue so I now created a pull request with another modification. #1017

dimvar:
PRs for a single-character change are fine.

dimvar:
Fixed by https://github.com/google/closure-compiler/commit/d16b3c3aa888e11a0842b1b940a364973b037b0b

PR #1017, opened 2015-06-19
expose compiler option PreferSingleQuotes in ant task

bardware:
I added code to expose the compiler option `PreferSingleQuotes` in the ant task.
I changed the character referencing the sourcemap from `@` to `#` according to google's documentation. I opened an issue earlier: #979


bardware:
I signed it!

PR #1041, opened 2015-07-14
Support a wildcard value for  --jscomp_XXX flags.

ChadKillingsworth:
Shortcut methods for adding all supported warning guards to a specific level.

`java -jar compiler.jar -O=ADVANCED -W=VERBOSE --jscomp_off=* mysrc.js`

### DIFF
--- a/src/com/google/javascript/jscomp/DiagnosticGroups.java
+++ b/src/com/google/javascript/jscomp/DiagnosticGroups.java
@@ -26,8 +26,11 @@ import com.google.javascript.jscomp.lint.CheckNullableReturn;
 import com.google.javascript.jscomp.lint.CheckPrototypeProperties;
 import com.google.javascript.jscomp.newtypes.JSTypeCreatorFromJSDoc;
 
+import java.util.Arrays;
 import java.util.HashMap;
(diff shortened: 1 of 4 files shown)


src/com/google/javascript/jscomp/DiagnosticGroups.java
```diff
@@ -37,6 +40,9 @@
   static final DiagnosticType UNUSED =
       DiagnosticType.warning("JSC_UNUSED", "{0}");
 
+  public static final Set<String> wildcardExcludedGroups =
+      new HashSet<String>(Arrays.asList(new String[] { "reportUnknownTypes" }));
```
MatrixFrog: :+1: 
There might be other things that belong in this list but none come to mind immediately.

MatrixFrog: I think you can use `ImmutableSet.of("reportUnknownTypes")`. We don't need this to be mutable, right?

ChadKillingsworth: Since this is the only project I work on that uses Guava - I just never think of those methods. Fixed.

PR #1042, opened 2015-07-15
CommonJS processing: Consider "module" scope

MNeise:
When rewriting a CommonJS module, only process `module.exports` when
`module` is not defined in current scope.

For example, if a function has a parameter `module`, references to `module.exports` should not be replaced with the new module name (e.g. `module$foobar`)

```
var foo = function (module) {
    module.exports = "foo";
};
module.exports = "bar";
```

https://groups.google.com/forum/#!topic/closure-compiler-discuss/AklSy5PNQ_I

### DIFF
--- a/src/com/google/javascript/jscomp/ProcessCommonJSModules.java
+++ b/src/com/google/javascript/jscomp/ProcessCommonJSModules.java
@@ -44,6 +44,7 @@ public final class ProcessCommonJSModules implements CompilerPass {
   private static final String MODULE_NAME_PREFIX = "module$";
 
   private static final String EXPORTS = "exports";
+  private static final String MODULE = "module";
 
   private final Compiler compiler;
   private final ES6ModuleLoader loader;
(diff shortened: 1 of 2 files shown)


concavelenz:
Martin, does this conflict with these your proposed changes?

src/com/google/javascript/jscomp/ProcessCommonJSModules.java
```diff
@@ -168,7 +169,10 @@ public void visit(NodeTraversal t, Node n, Node parent) {
 
       if (n.isGetProp() &&
           "module.exports".equals(n.getQualifiedName())) {
-        moduleExportRefs.add(n);
+        Var v = t.getScope().getVar(MODULE);
+        if (v == null) {
```
mprobst: Could you add a comment explaining the significance? E.g. `// Only rewrite the special top-level 'module' variable, but not local variables or parameters.`

mprobst:
It might merge conflict due to the change from `test(...)` to `testModules(...)`, but content wise it's fine.

src/com/google/javascript/jscomp/ProcessCommonJSModules.java
MNeise: Sure, no problem ;)

mprobst:
LGTM.

PR #1043, opened 2015-07-16
Rewrite ES6ModuleLoader.

mprobst:
* support multiple module roots, akin to inclusion paths
* resolve imports relative to the current file, then relative to the
  module roots
* do not depend on the file system, the list of inputs is sufficient
* do not depend on the compiler - the loader is only created after all
  inputs are known, there's no need to lazy initialise.
* remove code duplicated across CommonJS and ES6 handling
* remove loader exception & associated code
  it was never thrown, and the existing code purely worked by accident.


mprobst:
Not quite done yet, there were some more surprises in the tests...

mprobst:
Should be good now.

dimvar:
For large changes like this, create a CL internally. It's much easier for
us to review it and test it and be confident that it works.

MatrixFrog:
True, but I don't think anyone uses CommonJS internally. So since this PR is here, we can try asking some CommonJS users to patch it in and make sure it doesn't break them.

concavelenz:
Is this still pending?

dimvar:
No, it was submitted internally and pushed externally.

PR #1047, opened 2015-07-22
Initial support for the --env flag

### LINKED ISSUE #966: Add a simple command-line option for the environment the code is going to run in
I think we've discussed this but I didn't see a bug filed. This would probably look something like --env=WEB or --env=NODEJS or maybe --env=MODERN_WEB (meaning, latest version or two of the major browsers) and you would automatically get the correct set of externs, instead of having to specify each externs file with the `--externs` flag.

This would also help a lot with making the compiler easier to use.

ChadKillingsworth:
I expect this PR to have a rather large impact. Definitely should block this until after the next release.

* Builtin externs are now arranged by environment specific folders
* Adds a NODEJS environment
* Replaces the --use_only_custom_externs flag with the --env=CUSTOM argument
* esX.js externs are now included for every environment - including CUSTOM
* Makes it easy to add future environments

@nicks Would you like to take a look at the NODEJS part? I think the externs may need tweaking.

Fixes #966


dimvar:
I'll take a look and let's discuss it at the meeting on Monday.

ChadKillingsworth:
Per our conversation today, this is now ready for review. Language level externs are now also loaded based on the highest language level specified by the `language_in` and `language_out` flags.

ChadKillingsworth:
Restricting the `esX.js` externs included based on language mode is proving to be problematic. We currently don't have forward declares for the default set of externs which might use those types.

ChadKillingsworth:
I've removed the conditional inclusion of the esX.js externs. With the way our current externs are handled there is no way to easily split out the externs by language level. Nor do I see a good way to forwardDeclare many of the types.

build.xml
```diff
@@ -173,15 +173,16 @@
     <copy todir="${classes.dir}/com/google/javascript/jscomp/js">
       <fileset dir="${src.dir}/com/google/javascript/jscomp/js" />
     </copy>
+
+    <!-- Zip the externs and move them to the classes folder -->
+    <zip destfile="${classes.dir}/externs.zip" basedir="${externs.dir}" includes="**/*.js" />
```
dimvar: is this supposed to be

    includes="../*.js"

?

src/com/google/javascript/jscomp/CompilerOptions.java
```diff
@@ -2551,4 +2591,20 @@ public void addAlias(String alias, String definition) {
...
+
+  /**
+   * An environment specifies the built-in externs that are loaded for a given
+   * compilation.
+   */
+  public static enum Environment {
+    /**
+     * Hand crafted externs that have traditionally been the default externs.
+     */
+    LEGACY,
```
dimvar: Is this the same set of externs someone would want in the browser enviroment with es6? If so, we can pick some name related to that.

src/com/google/javascript/jscomp/CompilerOptions.java
```diff
@@ -2411,6 +2423,34 @@ public static LanguageMode fromString(String value) {
       }
       return null;
     }
+
+    public int toInteger() {
```
dimvar: This should probably be private

src/com/google/javascript/jscomp/CompilerOptions.java
```diff
@@ -2411,6 +2423,34 @@ public static LanguageMode fromString(String value) {
...
+        case ECMASCRIPT5_STRICT:
+          return 5;
+
+        case ECMASCRIPT3:
+        default:
+          return 3;
+      }
+    }
+
+    public static int maxLevel(LanguageMode... modes) {
```
dimvar: I don't see any calls to this method

src/com/google/javascript/jscomp/CompilerOptions.java
```diff
@@ -2369,20 +2388,13 @@ public boolean isStrict() {
...
       Preconditions.checkState(this != NO_TRANSPILE);
-      switch (this) {
-        case ECMASCRIPT6:
-        case ECMASCRIPT6_STRICT:
-        case ECMASCRIPT6_TYPED:
-          return true;
-        default:
-          return false;
-      }
+      return this.toInteger() >= 6;
```
dimvar: return this.toInteger() == 6;

Since toInteger never returns > 6, I find the >= a bit confusing (even though it's more future-proof).

src/com/google/javascript/jscomp/CommandLineRunner.java
```diff
@@ -1191,15 +1193,16 @@ protected Compiler createCompiler() {
...
+      throws FlagUsageException, IOException {
+    List<SourceFile> externs = super.createExterns(options);
+    if (isInTestMode()) {
       return externs;
     } else {
-      List<SourceFile> defaultExterns = getDefaultExterns();
-      defaultExterns.addAll(externs);
-      return defaultExterns;
+      List<SourceFile> builtinExterns =
+          getBuiltinExterns(options);
```
dimvar: move up

src/com/google/javascript/jscomp/CommandLineRunner.java
```diff
@@ -1320,30 +1286,53 @@ private static ConformanceConfig loadConformanceConfig(String configFile)
...
-        externsMap.keySet().equals(new HashSet<>(DEFAULT_EXTERNS_NAMES)),
-        "Externs zip must match our hard-coded list of externs.");
-
-    // Order matters, so the resources must be added to the result list
-    // in the expected order.
     List<SourceFile> externs = new ArrayList<>();
-    for (String key : DEFAULT_EXTERNS_NAMES) {
-      externs.add(externsMap.get(key));
+
+    for (String key : BUILTIN_LANG_EXTERNS) {
```
dimvar: Add comment

    // The externs for code JS objects are loaded in all environments.

src/com/google/javascript/jscomp/CommandLineRunner.java
```diff
@@ -1320,30 +1286,53 @@ private static ConformanceConfig loadConformanceConfig(String configFile)
...
+      Preconditions.checkState(
+          externsMap.containsKey(key),
+          "Externs zip must contain " + key + ".");
+      externs.add(externsMap.remove(key));
+    }
+
+    // Order matters, so extern resources which have dependencies must be added
+    // to the result list in the expected order.
+    for (String key : BUILTIN_EXTERN_DEP_ORDER) {
+      if (key.indexOf(envPrefix) < 0) {
```
dimvar: Here you only want to add them if the environment is legacy, right? So, you can check

    if (key.indexOf("legacy") < 0) {

dimvar:
The PR looks fine to me. Let me summarize what it does and tell me if my understanding is correct.

- Move non-js externs to subfolder
- Always load js externs
- Only load externs in the subfolder when the env is legacy

Why did you move the location of externs.zip from build/ to build/classes?

Can we find a better name for legacy, eg, web? Or are these externs not what we want by default on the web environment?

I didn't get why the language-level-specific inclusion failed. Can you explain it more?

build.xml
ChadKillingsworth: No. Doing that didn't include the subfolders and files. I explicitly had to change it so that the externs folder structure was included/preserved.

src/com/google/javascript/jscomp/CompilerOptions.java
ChadKillingsworth: Yes. At this point we only have one set of externs. And they definitely depend on es6 types.

src/com/google/javascript/jscomp/CompilerOptions.java
ChadKillingsworth: I can remove all of this for now. Originally the intent was that this was a much more flexible way throughout the compiler to test for specific code levels rather than continuing to add functions like `es5orHigher`.

src/com/google/javascript/jscomp/CompilerOptions.java
ChadKillingsworth: I removed the call - I can remove this now: https://github.com/ChadKillingsworth/closure-compiler/commit/2126d1eaf30e7007acfba9d84c14f1ab5731c70a#diff-17c34c561974efe2470d59d8ba5f4efaL1308

src/com/google/javascript/jscomp/CompilerOptions.java
ChadKillingsworth: I definitely did it to be future proof.

src/com/google/javascript/jscomp/CommandLineRunner.java
ChadKillingsworth: The idea was that all extern environments would be added to the same list, but we would only load the ones from the specific folder. Example:

```JavaScript
private static final List<String> BUILTIN_EXTERN_DEP_ORDER = ImmutableList.of(
   //-- Legacy externs --
   "legacy/intl.js",
   "legacy/w3c_event.js",
   "legacy/w3c_event3.js",

   //-- NodeJS exters --
   "nodejs/stream.js",
   "nodejs/path.js"
);
```

ChadKillingsworth:
> * Move non-js externs to subfolder
> * Always load js externs

Yes

 > * Only load externs in the subfolder when the env is legacy

So for this first round, we only have one environment. This should be "Only load externs from the folder whose name matches the `env` provided name." This was done specifically to make it easier to add more environments.

> Can we find a better name for legacy, eg, web? Or are these externs not what we want by default on the web environment?

Yes. Or "Browser"? I was simply trying to think ahead to a time when we move to auto-generated externs from IDL files and determining how to differentiate the two.

> I didn't get why the language-level-specific inclusion failed. Can you explain it more?

Our externs require parts from ALL of the esX language levels. For instance, [fetchapi.js](https://github.com/google/closure-compiler/blob/master/externs/fetchapi.js#L90) uses types from the `es6.js` extern. So If the `es6.js` extern is not loaded, there are type errors from what will be the default extern set.

This problem is not near as pronounced with the `es5.js` externs as they don't really introduce new types. My best option was to find a way to determine which language level was required for each extern file and conditionally load it. That seemed like a harder problem.

src/com/google/javascript/jscomp/CommandLineRunner.java
dimvar: oh got it, thanks.

src/com/google/javascript/jscomp/CompilerOptions.java
dimvar: sounds good, let's remove it then

dimvar:
OK, everything looks good. Change the name from legacy to browser, and other than that, LGTM.

(You'll have to squash the commits and fix the merge conflicts before I can merge the PR.)

ChadKillingsworth:
Oh I missed one:

> Why did you move the location of externs.zip from build/ to build/classes?

By moving the externs.zip to the classes folder, it was accessible in unit tests - and it simplified the build file. Seemed like a win all around.

src/com/google/javascript/jscomp/CommandLineRunner.java
ChadKillingsworth: Looks like I missed this one. What did you mean here?

dimvar: Oh just to do

    List<SourceFile> builtinExterns = getBuiltinExterns(options);

src/com/google/javascript/jscomp/ant/CompileTask.java
```diff
@@ -173,6 +173,25 @@ public void setWarning(String value) {
   }
 
   /**
+   * Set the environment which determines the builtin extern set.
+   * @param value The name of the environment.
+   *     (LEGACY, NODEJS, CUSTOM).
```
dimvar: change to BROWSER

src/com/google/javascript/jscomp/ant/CompileTask.java
```diff
@@ -173,6 +173,25 @@ public void setWarning(String value) {
...
+   */
+  public void setEnvironment(String value) {
+    switch (value) {
+      case "LEGACY":
```
dimvar: change to BROWSER

src/com/google/javascript/refactoring/RefasterJs.java
```diff
@@ -67,9 +68,10 @@
       usage = "Location of the JS file to use as the RefasterJS template.")
   private String refasterJsTemplate = null;
 
-  @Option(name = "--include_default_externs",
-      usage = "Whether to include the standard JavaScript externs. Defaults to true.")
-  private boolean includeDefaultExterns = true;
+  @Option(name = "--env",
+      usage = "Which set of externs to include. Defaults to LEGACY.")
```
dimvar: change to BROWSER

ChadKillingsworth:
Ok done - should be gtg as soon as Travis finishes.

PR #1056, opened 2015-08-02
Fixed bugs in the html5 extern relating to Web Sockets.  The readySta…

mmacfadden:
Fixed a few issues in the WebSockets externs.  The changes all align the extern with the spec here:

http://www.w3.org/TR/websockets/#websocket

1. The OPEN, CONNECTING, and CLOSED constants were moved from the prototype to the class.
2. The missing WebSocket.CLOSING constant was added and set to 2.
3. The missing WebSocket.CLOSED constant was changed from 2 (incorrect) to 3 (correct).
4. The WebSocket.prototype.URL property was changed to The WebSocket.prototype.url.
5. The WebSocket.prototype.close method was updated to add the two optional parameters (code, and reason).

This PR addresses the discussion here:

https://groups.google.com/forum/#!topic/closure-compiler-discuss/FYK7rx04U4M

As well as a few other issues, I was seeing.

### DIFF
--- a/externs/html5.js
+++ b/externs/html5.js
@@ -2144,30 +2144,6 @@ TimeRanges.prototype.end = function(index) { return 0; };
 function WebSocket(url, opt_protocol) {}
 
 /**
- * The connection has not yet been established.
- * @type {number}
- */
-WebSocket.CONNECTING = 0;
-
-/**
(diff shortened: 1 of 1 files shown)


mmacfadden:
Need to resubmit this with the correct email.

PR #1071, opened 2015-08-12
CommonJS: Remove redundant namespace declaration.

mneise:
Namespace should already be declared and initialized by `goog.provide`. So instead of emitting this:

```
goog.provide("module$libs$foo");
var module$libs$foo = {};
module$libs$foo.sayHelloInEnglish = function() {
  return "Hello";
};
```
this should be emitted:
```
goog.provide("module$libs$foo");
module$libs$foo.sayHelloInEnglish = function() {
  return "Hello";
};
```

This is also consistent with the way "normal" Google Closure modules are written.

Adjusted the tests to match changes.

See the mailing list discussion: https://groups.google.com/forum/#!topic/closure-compiler-discuss/t8u_YhL2Onk

### DIFF
--- a/src/com/google/javascript/jscomp/ProcessCommonJSModules.java
+++ b/src/com/google/javascript/jscomp/ProcessCommonJSModules.java
@@ -244,16 +244,14 @@ public final class ProcessCommonJSModules implements CompilerPass {
     private void processExports(Node script, String moduleName) {
       if (hasOneTopLevelModuleExportAssign()) {
         // One top-level assign: transform to
-        // /** @const */ var moduleName = rhs
+        // moduleName = rhs
         Node ref = moduleExportRefs.get(0);
         Node newName = IR.name(moduleName);
         newName.putProp(Node.ORIGINALNAME_PROP, ref.getQualifiedName());
 
-        Node newVar = IR.var(newName)
-            .copyInformationFromForTree(ref.getParent());
         Node rhsValue = ref.getNext().detachFromParent();
-        newVar.getFirstChild().addChildToFront(rhsValue);
-        newVar.setJSDocInfo(NodeUtil.createConstantJsDoc());
+        Node newExprResult = IR.exprResult(IR.assign(newName, rhsValue)
+          .copyInformationFromForTree(ref.getParent()));
 
         // If the rValue is an object literal, check each property to see if
         // it's an alias, and if it is, copy the annotation over.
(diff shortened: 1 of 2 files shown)


src/com/google/javascript/jscomp/ProcessCommonJSModules.java
```diff
@@ -244,16 +244,15 @@ private void visitScript(NodeTraversal t, Node script) {
...
         newName.putProp(Node.ORIGINALNAME_PROP, ref.getQualifiedName());
 
-        Node newVar = IR.var(newName)
-            .copyInformationFromForTree(ref.getParent());
         Node rhsValue = ref.getNext().detachFromParent();
-        newVar.getFirstChild().addChildToFront(rhsValue);
-        newVar.setJSDocInfo(NodeUtil.createConstantJsDoc());
+        Node newExprResult = IR.exprResult(IR.assign(newName, rhsValue)
+          .copyInformationFromForTree(ref.getParent()));
+        newExprResult.setJSDocInfo(NodeUtil.createConstantJsDoc());
```
dimvar: Please remove this line. <code>@const</code> for variables only makes sense with VAR, not at other assignments.

test/com/google/javascript/jscomp/ProcessCommonJSModulesTest.java
```diff
@@ -100,7 +97,7 @@ public void testExports() {
         "goog.provide('module$test');"
             + "goog.require('module$other');"
             + "var name$$module$test = module$other;"
-            + "/** @const */ var module$test = function () {};");
+            + "/** @const */ module$test = function () {};");
```
dimvar: Per previous comment, there should be no <code>@const</code> generated here.

dimvar:
Looks fine. Only a minor comment.

mneise:
Ah, good to know. Made the changes as suggested. Thank you for reviewing :smiley:

src/com/google/javascript/jscomp/ProcessCommonJSModules.java
```diff
@@ -244,16 +244,14 @@ private void visitScript(NodeTraversal t, Node script) {
     private void processExports(Node script, String moduleName) {
       if (hasOneTopLevelModuleExportAssign()) {
         // One top-level assign: transform to
-        // /** @const */ var moduleName = rhs
+        // /** @const */ moduleName = rhs
```
dimvar: can you remove const from the comment as well? Sorry, missed that the first time.

dimvar:
LGTM. One more minor thing to fix and then i'll merge it.

mneise:
Good find! Fixed the comment.

PR #1074, opened 2015-08-13
Fix Regression: source location of node of type Token.NAME should actually refer to variable name.

aslushnikov:
Currently, source location of nodes of type Token.NAME refer to the full variable declaration.
This patch fixes this to refer to the variable name only.

This is a regression caused by 544de7369955ba21cbab4217d7f0799d81c957a0.

Please, take a look.

### DIFF
--- a/src/com/google/javascript/jscomp/parsing/IRFactory.java
+++ b/src/com/google/javascript/jscomp/parsing/IRFactory.java
@@ -1811,7 +1811,7 @@ class IRFactory {
       if (decl.initializer != null) {
         Node initializer = transform(decl.initializer);
         node.addChildToBack(initializer);
-        maybeSetLength(node, decl.location.start, decl.location.end);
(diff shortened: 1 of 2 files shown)


MatrixFrog:
That was intentional, to fix a bug with the refactoring tools IIRC. In general, the size of a node should include its children I think, though I know that's a bit weird in this case.

aslushnikov:
How to figure out the variable name then?

MatrixFrog:
You mean how to figure out the exact position of it? I guess you can look at the original source and go from the beginning of the node to the first whitespace or `=` character. But this is probably a discussion for the discuss list.

aslushnikov:
Forcing clients to re-parse text returned by utility that parses text is weird. Can we have nested nodes that provide variable name separately?

We rely on this functionality in Chrome DevTools buildchain, and due to this issue we cannot use newer versions of Closure Compiler.

MatrixFrog:
Please start a thread on the mailing list.

PR #1079, opened 2015-08-13
Normalize the separator for file paths in source files.

### LINKED ISSUE #1070: @package visibility level does not work in Windows
I am never getting "Access to package-private property XXX of YYY not allowed here" when compile on Windows. While on Linux I get them.
It seems that problem is in method getPackageName of DefaultCodingConvention class.
It uses platform-independent forward slashes.

Example:
```javascript
test/ns1/class1.js:
goog.provide('NS1.Class1');
/** @constructor */
NS1.Class1 = function(a)
{
    /** @package */
    this.a_ = a;
}

test/ns2/class2.js:
goog.require('NS1.Class1');
goog.provide('NS2.Class2');
/** 
 * @constructor
 * @param {NS1.Class1} c1
 */
NS2.Class2 = function(c1)
{
    /** @type {NS1.Class1} */
    this.c1_ = c1;
}
NS2.Class2.prototype.meth = function()
{
    alert(this.c1_.a_);
}

test/main.js:
goog.require('NS1.Class1');
goog.require('NS2.Class2');
var c1 = new NS1.Class1('aaa');
var c2 = new NS2.Class2(c1);
c2.meth();
```

build.bat:
java -jar ..\compiler.jar --js ns1/class1.js --js ns2/class2.js --js main.js --compilation_level ADVANCED -W VERBOSE --jscomp_error visibility --jscomp_error accessControls 

Result: 
I get `@package` - access error on Linux
I get no `@package` - access error on Windows

ChadKillingsworth:
Allows most passes to assume a separator of "/".

Fixes #1070

### DIFF
--- a/src/com/google/javascript/jscomp/Compiler.java
+++ b/src/com/google/javascript/jscomp/Compiler.java
@@ -18,6 +18,7 @@ package com.google.javascript.jscomp;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Function;
+import com.google.common.base.Joiner;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Supplier;
 import com.google.common.base.Throwables;
@@ -48,6 +49,7 @@ import com.google.javascript.rhino.Token;
 import com.google.javascript.rhino.TypeIRegistry;
 import com.google.javascript.rhino.jstype.JSTypeRegistry;
 
+import java.io.File;
 import java.io.IOException;
 import java.io.PrintStream;
 import java.nio.file.FileSystems;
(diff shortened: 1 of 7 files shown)


ChadKillingsworth:
Ideally we would be able to test for these variants regardless of the build environment. I found that to be difficult. Instead I simply made sure the tests all passed in both my unix and windows build environments.

MatrixFrog:
Maybe we can configure Travis to run on Windows?

ChadKillingsworth:
Doesn't look like windows is supported: http://docs.travis-ci.com/user/multi-os/

MatrixFrog:
There must be some standard "join" function such that `join("a", "b") == "a/b"` on Unixish systems and `join("a", "b") == "a\\b"` on Windows?

dimvar:
I didn't find a standard function, but you can add a static method in Compiler that takes two strings and does the concatenation, and then you won't have to import File in all the other places.

ChadKillingsworth:
Done. Commits squashed as well.

PR #1081, opened 2015-08-14
ES6: Remove redundant namespace declaration

mneise:
When converting ECMAScript 6 modules, it is not necessary to add an extra namespace declaration. The namespace will be declared through `goog.provide`. When processing the `goog.provide`, the compiler will then add a JSDoc `@const` tag to the namespace. (https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/ProcessClosurePrimitives.java#L1423).

For example, instead of converting the following ES6 module:

```javascript
// calculator.js
export function square(x) {
        return x * x;
}
```

to:

```javascript
goog.provide("module$calculator");
function square$$module$calculator(x) {
  return x * x;
}
var module$calculator = {};
module$calculator.square = square$$module$calculator;
```

it will now be converted to:

```javascript
goog.provide("module$calculator");
function square$$module$calculator(x) {
  return x * x;
}
module$calculator.square = square$$module$calculator;
```

which will later be converted to:

```javascript
var /** @const */ module$calculator = {};
function square$$module$calculator(x) {
  return x * x;
}
module$calculator.square = square$$module$calculator;
```

### DIFF
--- a/src/com/google/javascript/jscomp/ProcessEs6Modules.java
+++ b/src/com/google/javascript/jscomp/ProcessEs6Modules.java
@@ -307,18 +307,6 @@ public final class ProcessEs6Modules extends AbstractPostOrderCallback {
     URI normalizedAddress = loader.normalizeInputAddress(t.getInput());
     String moduleName = ES6ModuleLoader.toModuleName(normalizedAddress);
 
-    if (!exportMap.isEmpty()) {
-      // Creates an export object for this module.
-      // var moduleName = {};
-      Node objectlit = IR.objectlit();
-      Node varNode = IR.var(IR.name(moduleName), objectlit)
-          .useSourceInfoIfMissingFromForTree(script);
-      JSDocInfoBuilder info = new JSDocInfoBuilder(true);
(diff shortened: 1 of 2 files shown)


Dominator008:
We want people to be able to use ES6 modules without having to process Closure primitives.

mneise:
Ah, yes, that is exactly what we are trying to do. We are using the `ProcessEs6Modules` class directly from Clojure to convert ES6 modules. Unfortunately, the extra `var` declaration is causing problems when trying to load the converted module in a ClojureScript Node REPL, since the Node REPL is running scripts in a separate context.

How would the changes prevent people from using converted ES6 modules?

dimvar:
Michael, the proposed translation seems completely fine to me. What do you mean by not using process-closure-primitives? The pass emits goog.provide/require anyway, so we can't avoid that. Do you mean sth about when running uncompiled?

Dominator008:
Ok after discussing with @dimvar I think this PR looks good to me :)

MatrixFrog:
lgtm

PR #1085, opened 2015-08-18
Add externs for the Pointer Lock API

fredj:
The W3C spec: https://w3c.github.io/pointerlock/

### DIFF
--- a/contrib/externs/w3c_pointerlock.js
+++ b/contrib/externs/w3c_pointerlock.js
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2015 The Closure Compiler Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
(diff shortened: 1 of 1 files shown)


contrib/externs/w3c_pointerlock.js
```diff
@@ -0,0 +1,51 @@
...
+
+
+/**
+ * @see https://w3c.github.io/pointerlock/#widl-Element-requestPointerLock-void
+ */
+Element.prototype.requestPointerLock = function() {};
+
+/**
+ * @see https://w3c.github.io/pointerlock/#widl-Document-pointerLockElement
+ * @type {Element}
```
MatrixFrog: I assume this can be null? Let's make it {?Element} so it's explicit.

contrib/externs/w3c_pointerlock.js
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * @fileoverview Definitions for W3C's Pointer Lock API.
+ * @see https://w3c.github.io/pointerlock/
+ *
+ * @externs
+ */
+
+
+/**
+ * @see https://w3c.github.io/pointerlock/#widl-Element-requestPointerLock-void
```
MatrixFrog: Add `@return {void}` here so that the typechecker knows this is a function.

contrib/externs/w3c_pointerlock.js
```diff
@@ -0,0 +1,51 @@
...
+Element.prototype.requestPointerLock = function() {};
+
+/**
+ * @see https://w3c.github.io/pointerlock/#widl-Document-pointerLockElement
+ * @type {Element}
+ */
+Document.prototype.pointerLockElement;
+
+/**
+ * @see https://w3c.github.io/pointerlock/#widl-Document-exitPointerLock-void
```
MatrixFrog: here too

fredj:
Thanks for the review @MatrixFrog, I've addressed all your comments.

PR #1087, opened 2015-08-18
Better control over what gets renamed

dchambers:
The heuristic used to determine whether member properties can be renamed is very liberal (it accepts any names containing symbols or non lower-case characters) which reduces the effectiveness of the `CodeConventions.isPrivate()` method for determining whether properties can be renamed. By providing a `PropertyRenamingPolicy.PRIVATE` flag, it becomes possible to let the `CodeConventions` instance better drive which properties get renamed or not.

I've also taken the opportunity to update the documentation so it better represents what actually happens in code, and to make the code easier to follow.

### DIFF
--- a/src/com/google/javascript/jscomp/DefaultPassConfig.java
+++ b/src/com/google/javascript/jscomp/DefaultPassConfig.java
@@ -2377,22 +2377,12 @@ public final class DefaultPassConfig extends PassConfig {
     char[] reservedChars = options.anonymousFunctionNaming.getReservedCharacters();
     switch (options.propertyRenaming) {
       case HEURISTIC:
+      case AGGRESSIVE_HEURISTIC:
+      case PRIVATE:
         RenamePrototypes rproto =
             new RenamePrototypes(compiler, options.propertyRenaming, reservedChars, prevPropertyMap);
         rproto.process(externs, root);
         return rproto.getPropertyMap();
-
-      case AGGRESSIVE_HEURISTIC:
-        RenamePrototypes rproto2 =
-            new RenamePrototypes(compiler, options.propertyRenaming, reservedChars, prevPropertyMap);
-        rproto2.process(externs, root);
-        return rproto2.getPropertyMap();
-
-      case PRIVATE:
-          RenamePrototypes rproto3 =
-              new RenamePrototypes(compiler, options.propertyRenaming, reservedChars, prevPropertyMap);
-          rproto3.process(externs, root);
-          return rproto3.getPropertyMap();
 
       case ALL_UNQUOTED:
         RenameProperties rprop =
--- a/src/com/google/javascript/jscomp/DefaultPassConfig.java
+++ b/src/com/google/javascript/jscomp/DefaultPassConfig.java
@@ -2378,15 +2378,21 @@ public final class DefaultPassConfig extends PassConfig {
     switch (options.propertyRenaming) {
       case HEURISTIC:
         RenamePrototypes rproto =
-            new RenamePrototypes(compiler, false, reservedChars, prevPropertyMap);
+            new RenamePrototypes(compiler, options.propertyRenaming, reservedChars, prevPropertyMap);
         rproto.process(externs, root);
         return rproto.getPropertyMap();
 
       case AGGRESSIVE_HEURISTIC:
         RenamePrototypes rproto2 =
-            new RenamePrototypes(compiler, true, reservedChars, prevPropertyMap);
+            new RenamePrototypes(compiler, options.propertyRenaming, reservedChars, prevPropertyMap);
         rproto2.process(externs, root);
         return rproto2.getPropertyMap();
+
+      case PRIVATE:
+          RenamePrototypes rproto3 =
+              new RenamePrototypes(compiler, options.propertyRenaming, reservedChars, prevPropertyMap);
+          rproto3.process(externs, root);
+          return rproto3.getPropertyMap();
 
       case ALL_UNQUOTED:
         RenameProperties rprop =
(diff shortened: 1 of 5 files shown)


ChadKillingsworth:
The heuristic renaming policies are deprecated and slated for removal at some point. We actually don't support them any longer.

There is actually a series of blog posts that describe the history of the renaming policies: http://closuretools.blogspot.com/2011/01/property-by-any-other-name-part-1.html

dchambers:
Hi @ChadKillingsworth, thanks for the quick response. I wasn't actually trying to use heuristic renaming, I just wanted to be able to provide a custom coding-convention that defined private members as being anything starting with `_` (e.g. `_privateMethod`).

Since `CodingConvention.isPrivate()` is only used by `RenamePrototypes` (heuristic renaming), but not by `RenameProperties` (all-unquoted renaming), that's the route I went down. Unfortunately, I quickly found that the heuristic used to determine if things should be renamed breaks my coding convention, since it thinks `publicMethod` should be renamed since it has an upper-case letter!

We've been using Google Closure for years, but we've realized that we're never going to be able to make use of `ADVANCED_OPTIMIZATIONS` since some of the code that forms the app won't always be part of the main bundle. On the other hand, we've consistently adopted the norm of indicating private methods and variables using `_`.

What if I submitted a pull-request that causes re-naming of private members to happen even under `SIMPLE_OPTIMIZATIONS`? This would have no effect for most users since only `GoogleCodingConvention` currently defines what private members look like (anything ending with an `_`).

dimvar:
RenamePrototypes is part of heuristic property renaming and will be deleted soon. Either the next release or the one after that. It would have been deleted already, but we're blocked by an internal team trying to migrate off of it. Therefore, we won't be merging PRs related to this pass.

Regarding custom coding conventions. Any project can have a different coding convention that suits their needs. It is not feasible for us to host all conventions in the compiler code base, so each project can write their coding convention and host it in their own code base.

Also, this option (that enables disambiguation of private properties) may be of interest to you.
https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/CompilerOptions.java#L583

dchambers:
Hi @dimvar,

> Regarding custom coding conventions. Any project can have a different coding convention that suits their needs. It is not feasible for us to host all conventions in the compiler code base, so each project can write their coding convention and host it in their own code base.

I'm already hosting the coding conventions in our own code base, but they don't work because _heuristic-renaming_ is too permissive (it renames names anything with upper-case characters despite what `CodingConvention.isPrivate()` says), and since _all-unquoted_ doesn't take `CodingConvention.isPrivate()` into account.

I'd be happy to create a separate pull-request that adds a new compiler option called `renamePublicProperties` which, when set to `false`, would prevent _all-unquoted_ from renaming private properties as defined by `CodingConvention.isPrivate()`. Would this be more compatible with the future direction of the project?

> Also, this option (that enables disambiguation of private properties) may be of interest to you.
https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/CompilerOptions.java#L583

AFAICT, `disambiguatePrivateProperties` doesn't prevent renaming, it just controls whether the new names are unique, or whether they're also used by other objects too.

dimvar:
Let's forget heuristic renaming for the rest of this discussion, since it's going to be deleted.

What is it that you are trying to do? Are you trying to prevent renaming of private properties? If so, why?

ChadKillingsworth:
@dimvar I think it's the other way around. I think he wants renaming of private properties, in the global scope, with SIMPLE_OPTIMIZATIONS.

dchambers:
> @dimvar I think it's the other way around. I think he wants renaming of private properties, in the global scope, with SIMPLE_OPTIMIZATIONS.

Yes @ChadKillingsworth, it's exactly that.

dimvar:
Got it. Thanks for the clarification.

So, the reason we can't do that, is that simple-optimizations doesn't do type checking, so the compiler can't enforce that a private property is actually used only privately. So, the optimization would break code that misuses private properties. The general idea for simple optimizations is that it's safe to do for the vast majority of code (there are still some code patterns that break, but they are rare.)

I would encourage you to make your code compatible with advanced optimizations, and then you'll get property renaming and many other benefits.

ChadKillingsworth:
I don't think we do any property renaming in simple optimizations, correct? And without type based optimizations, renaming properties is just a map. Any property named `.foo` becomes `.z` - regardless of what object it is attached to.

As @dimvar said, the compiler has no way to know if a reference to `.foo` is actually a private property or simply a property that shares the same name as a private property.

nicks:
it might be valuable to have a 'rename underscored properties to random stuff' pass, similar to what Python does for `self.__privateProperty`, and separate from ADVANCED_OPTIMIZATIONS? 

it would certainly make closure-compiler more valuable to people that want a nice incremental step towards advanced mode

totally agree that trying to implement this by modifying the old HEURISTIC renaming pass will expose you to lots of nasty surprises

ChadKillingsworth:
@nicks That solution would work. As long as it was a pass that wasn't enabled by default (required a custom build), there shouldn't be any major issues.

In fact, you could probably modify the existing RenameProperties pass to do that pretty easily.

dchambers:
> it might be valuable to have a 'rename underscored properties to random stuff' pass, similar to what Python does for self.__privateProperty, and separate from ADVANCED_OPTIMIZATIONS?

If the determination of which properties are private could be controlled by the coding conventions that would be ideal for us since ourselves and our clients have loads of code that uses `m_` to denote private member variables, and we'd ideally like those renamed too.

> it would certainly make closure-compiler more valuable to people that want a nice incremental step towards advanced mode

:+1: 

> @nicks That solution would work. As long as it was a pass that wasn't enabled by default (required a custom build), there shouldn't be any major issues.

This would work perfectly for us.

> In fact, you could probably modify the existing RenameProperties pass to do that pretty easily.

Right, that's what I was proposing earlier, though for reasons I can l no longer remember I thought to require `renamePublicProperties` to be set to `false` to tone down _all-unquoted_, rather than more sensibly requring `renamePrivateProperties` to be set to `true`, to enable the use of a toned-down `RenameProperties` pass for even `SIMPLE_OPTIMIZATIONS` users.

concavelenz:
We have a "disambiguate private properties" pass that renames properties based on the naming convention.  It doesn't work well with legacy code as too much code breaks the convention and it doesn't back off.

PR #1094, opened 2015-08-21
New pass WhitespaceWrapGoogModules in specialist "always pre-check" phase (fixes #611)

### LINKED ISSUE #611: Whitespace only mode breaks goog.module
goog.module needs to be wrapped or rewritten and whitespace only mode does neither.

supersteves:
I predict this will be rejected since it is arguable that the ```WHITESPACE_ONLY``` mode should mean just that.  However my argument is that goog.modules are unusable without this (#611), therefore it is forgiveable.

For a large codebase, you need to be able to develop rapidly without waiting for the ~ 40 second compile time we're currently seeing. Whitespace only mode drops this to ~2 seconds.

Also I'm adding a new pass phase "always pre-check" (which happens always, even in whitespace only mode) immediately prior to the "checks" phase (which only happens in simple or advanced mode). Even more debatable! I first tried using a custom pass in the options (see reverted commits), but I don't have the Compiler instance yet (needed for ```compiler.reportCodeChange()```) when I build the options in my external Java compiler front-end.  If you don't like this, I could find a way to do it using custom passes, but that's not going to help other users as much.

Currently WhitespaceWrapGoogModules is opt-in, but should be made default in WHITESPACE_ONLY once approved. 

I expect this to fail Travis, since the parent commit is already failing (@blickly). If you reject because of Travis failures, how should I proceed?


supersteves:
I signed it!  However this link isn't showing it:
https://cla.developers.google.com/clas
Signed as a corporation, name: Visokio

supersteves:
NB. the 2 CI failures are pre-existing.

ChadKillingsworth:
Thanks for pointing out the pre-existing CI failures. They will be fixed very shortly. However, you'll want to rebase your changes off of the latest master once #1096 is merged.

blickly:
Sorry about the existing travis failures. Can you rebase now that travis is green?

concavelenz:
Do you want whitespace only mode or simply a means of concat-ing files using goog.module?

supersteves:
We need a rapid refresh Dev mode. (Simple and Advanced take about 30 seconds).

Before switching to goog.module this was possible by serving up all source files individually along with a deps file. However this was also a performance issue on slower devices and made Chrome Dev tools sluggish.

This fork gives us 2 second refresh with goog module support and in a single http resource. And now Chrome Canary supports ES6 classes, let, const, fat arrows, we have a huge step forward in our dev platform. 

Whether it should be called whitespace mode or something else, I don't know. Also happy to rework it into a different approach should you prefer, rather than have a long term reliance on a fork.

I'm on holiday for a couple of weeks so will rebase when I'm back.

concavelenz:
I'm content with a change like this but I also want to expose the bundling logic (which would preserve comments, etc).

concavelenz:
@blickly Ben if you want to review this for consistency with the ES6->ES5 transpilation support in "whitespace" mode, I'll review it the bundling logic.

src/com/google/javascript/jscomp/Compiler.java
```diff
@@ -668,6 +668,8 @@ private void compileInternal() {
     if (!precheck()) {
       return;
     }
+    
+    alwaysPreChecks(); // even in whitespace only mode
 
     if (!options.skipNonTranspilationPasses || options.lowerFromEs6()) {
       check();
```
blickly: We actually already have a case where whitespace only mode runs the checks (when we are transpiling from ES6 down to ES3/5), so it's reasonable to just run check unconditionally, and add this logic to the beginning of getChecks in DefaultPassConfig.

supersteves: Yes, I saw that. However I am minimising passes and AST changes to be as
fast as possible, and to make debugging as close to original code as
possible without sourcemaps (I don't thinks these are supported unless
doing advanced compilation, or at least this was the case a while back).
And now that Chrome Canary supports all the ES6 killer features we use, we
don't use transpilation, hence none of the checks run.

Ideally there wouldn't be the lack of clarity as to which checks run.
Perhaps keep WS only mode as pure no-checks, and have another mode which
does the minimum needed for latest Chrome Canary build (I.e. concat only,
keeping comments/ws)?

Incidentally how practical is it to putting sourcemaps support in to all
comp levels (unless it already works), and to chaining sourcemaps I.e.
providing input sourcemaps for previous external transformations? I'm
looking at adding a JSX transpilation pass (initially using Babel in
Nashorn and it would emit sourcemaps - adapting your parser to recognise
JSX is a bit beyond my abilities).

supersteves:
I did author these commits.

concavelenz:
I think you need to rebase to the top of tree.

supersteves:
Yes. Travis is currently failing again. So I tried to rebase to the last successful Travis build. But the pull request presumably runs it against a merge with the head. I'll wait for a successful Travis build on master before going through the rebase process again.

blickly:
OK, travis on master should be back to passing now. Sorry for the hiccups.

supersteves:
Closing, replacing with:
https://github.com/google/closure-compiler/pull/1123

PR #1095, opened 2015-08-21
Fix maven tests

ChadKillingsworth:
Travis is currently failing because the externs.zip folder is not copied into the target/classes folder to be available for tests. I made that change for the ant build, but I forgot about Maven.

This may not be the most efficient way to do this. I barely use maven so I'm more or less hacking at the configuration to get it working.

### DIFF
--- a/externs/pom.xml
+++ b/externs/pom.xml
@@ -48,7 +48,8 @@
             <phase>generate-sources</phase>
             <configuration>
               <tasks>
-                <zip destfile="target/generated-resources/externs.zip" basedir="." includes="*.js"/>
+                <zip destfile="../target/classes/externs.zip" basedir="." includes="**/*.js"/>
+                <zip destfile="target/classes/externs.zip" basedir="." includes="**/*.js"/>
               </tasks>
             </configuration>
             <goals>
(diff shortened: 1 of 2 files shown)


dimvar:
I also never use maven. What did you do to verify that your fix works? Green from Travis here is sufficient?

MatrixFrog:
We should make sure that **/*.js doesn't include the contents of https://github.com/google/closure-compiler/tree/master/src/com/google/javascript/jscomp/js in the externs zip.

ChadKillingsworth:
@MatrixFrog It doesn't. It's based in the externs folder. I checked the generated externs.zip contents to make sure.

@dimvar I ran all the tests locally. This should only impact tests with Maven.

MatrixFrog:
Okay sgtm.

ChadKillingsworth:
The build error now is because it can't find `com.google.common.truth.Truth.assertWithMessage`, but it finds other methods in that library. Did the Truth distributable remove that?

MatrixFrog:
It may be that Maven is set up to use an older version of Truth than we're using.

ChadKillingsworth:
Yeah that seems like the issue.

PR #1096, opened 2015-08-21
Upgrade the truth library to 0.27 and fix compile errors in test.

ChadKillingsworth:
Problem identified in #1095

0.27 was the latest published version of Truth.


ChadKillingsworth:
Commit 828de39641b815b65964f0b3ed48ea3701d31786 caused the issue originally.

MatrixFrog:
This actually may be trickier to land than it ought to be. For now, I think I'll just get rid of the assertWithMessage call.

ChadKillingsworth:
@MatrixFrog Because of the binary I assume?

PR #1101, opened 2015-08-27
Allow simple alias assignments in externs for type inference.

ChadKillingsworth:
The compiler can correctly infer the types of aliases in externs when they are assigned and we should allow this pattern.

This removes the spurious warning that is generated for such simple assignments.

I can remove 400 lines from the jQuery externs with this change.

### DIFF
--- a/src/com/google/javascript/jscomp/VarCheck.java
+++ b/src/com/google/javascript/jscomp/VarCheck.java
@@ -17,7 +17,6 @@
 package com.google.javascript.jscomp;
 
 import com.google.common.base.Preconditions;
-import com.google.common.collect.ImmutableSet;
 import com.google.javascript.jscomp.NodeTraversal.AbstractPostOrderCallback;
 import com.google.javascript.jscomp.SyntacticScopeCreator.RedeclarationHandler;
 import com.google.javascript.rhino.IR;
@@ -102,9 +101,6 @@ class VarCheck extends AbstractPostOrderCallback implements
 
   // Whether extern checks emit error.
   private final boolean strictExternCheck;
-
-  private final Set allowedExternAssignmentTypes = ImmutableSet.of(
-      Token.VAR, Token.CONST, Token.LET);
 
   VarCheck(AbstractCompiler compiler) {
     this(compiler, false);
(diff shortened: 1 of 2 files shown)


src/com/google/javascript/jscomp/VarCheck.java
```diff
@@ -303,7 +303,12 @@ public void visit(NodeTraversal t, Node n, Node parent) {
             }
             // fall through
           default:
-            t.report(n, NAME_REFERENCE_IN_EXTERNS_ERROR, n.getString());
+            // Don't warn for simple var assignments "/** @const */ var foo = bar;"
+            // They are used to infer the types of namespace aliases.
+            if (parent.getType() != Token.NAME || parent.getParent() == null ||
+                parent.getParent().getType() != Token.VAR) {
```
MatrixFrog: May as well allow LET and CONST too as we'll probably want to allow ES6 in externs at some point (though we don't yet)

ChadKillingsworth: Good call. Fixed.

ChadKillingsworth:
All the tests passed locally. Not sure what's up with travis.

blickly:
Yeah, we've been getting a lot of flaky failures recently for testSpeed(com.google.debugging.sourcemap.Base64VLQTest) on travis. We should probably increase the timeout or just remove it.

src/com/google/javascript/jscomp/VarCheck.java
```diff
@@ -303,7 +307,12 @@ public void visit(NodeTraversal t, Node n, Node parent) {
             }
             // fall through
           default:
-            t.report(n, NAME_REFERENCE_IN_EXTERNS_ERROR, n.getString());
+            // Don't warn for simple var assignments "/** @const */ var foo = bar;"
+            // They are used to infer the types of namespace aliases.
+            if (parent.getType() != Token.NAME || parent.getParent() == null ||
+                !allowedExternAssignmentTypes.contains(parent.getParent().getType())) {
```
MatrixFrog: Sorry I should have said this before. We have NodeUtil.isNameDeclaration which does exactly this.

MatrixFrog:
otherwise lgtm

concavelenz:
nice and simple

ChadKillingsworth:
@MatrixFrog That does simplify things. Done.

ChadKillingsworth:
@MatrixFrog ping

PR #1117, opened 2015-09-06
Fix --js_module_root in CommandLineRunner.

### LINKED ISSUE #1116: Using --js_module_root raises java.lang.UnsupportedOperationException
test.js
```javascript
import foo from "foo";

foo();
```

lib/foo.js
```javascript
export default function foo() {
  console.log("foo");
}
```

compile
```
$ java -jar closure-compiler/build/compiler.jar --version
Closure Compiler (http://github.com/google/closure-compiler)
Version: v20150901-38-gb27aa3b
Built on: 2015/09/06 12:03

$ java -jar closure-compiler/build/compiler.jar --language_in=ES6_STRICT --language_out=ES5 --js_module_root lib/ lib/foo.js test.js
Exception in thread "main" java.lang.UnsupportedOperationException
	at com.google.common.collect.ImmutableCollection.add(ImmutableCollection.java:96)
	at org.kohsuke.args4j.spi.MultiValueFieldSetter.doAddValue(MultiValueFieldSetter.java:76)
	at org.kohsuke.args4j.spi.MultiValueFieldSetter.addValue(MultiValueFieldSetter.java:60)
	at org.kohsuke.args4j.spi.StringOptionHandler.parseArguments(StringOptionHandler.java:19)
	at org.kohsuke.args4j.CmdLineParser.parseArgument(CmdLineParser.java:508)
	at com.google.javascript.jscomp.CommandLineRunner$Flags.parse(CommandLineRunner.java:621)
	at com.google.javascript.jscomp.CommandLineRunner$Flags.access$300(CommandLineRunner.java:137)
	at com.google.javascript.jscomp.CommandLineRunner.initConfigFromFlags(CommandLineRunner.java:1000)
	at com.google.javascript.jscomp.CommandLineRunner.<init>(CommandLineRunner.java:870)
	at com.google.javascript.jscomp.CommandLineRunner.main(CommandLineRunner.java:1436)
```

I will post a pull request for fixing this issue.

rintaro:
Use ArrayList instead of ImmutableList.
Move ES6ModuleLoader.DEFAULT_FILENAME_PREFIX to the last.

Fixes #1116

### DIFF
--- a/src/com/google/javascript/jscomp/CommandLineRunner.java
+++ b/src/com/google/javascript/jscomp/CommandLineRunner.java
@@ -1055,8 +1055,11 @@ public class CommandLineRunner extends
 
       // For backwards compatibility, allow both commonJsPathPrefix and jsModuleRoot.
       List<String> moduleRoots = new ArrayList<>(flags.commonJsPathPrefix);
-      moduleRoots.addAll(flags.moduleRoot);
-      moduleRoots.add(ES6ModuleLoader.DEFAULT_FILENAME_PREFIX);
+      if (!flags.moduleRoot.isEmpty()) {
+        moduleRoots.addAll(flags.moduleRoot);
+      } else {
+        moduleRoots.add(ES6ModuleLoader.DEFAULT_FILENAME_PREFIX);
+      }
 
       getCommandLineConfig()
           .setPrintTree(flags.printTree)
--- a/src/com/google/javascript/jscomp/CommandLineRunner.java
+++ b/src/com/google/javascript/jscomp/CommandLineRunner.java
@@ -411,15 +411,14 @@ public class CommandLineRunner extends
       hidden = true,
       usage = "Path prefix to be removed from CommonJS module names."
     )
-    private List<String> commonJsPathPrefix =
-        ImmutableList.of(ES6ModuleLoader.DEFAULT_FILENAME_PREFIX);
+    private List<String> commonJsPathPrefix = new ArrayList<>();
 
     @Option(
       name = "--js_module_root",
       hidden = true,
       usage = "Path prefixes to be removed from ES6 & CommonJS modules."
     )
-    private List<String> moduleRoot = ImmutableList.of(ES6ModuleLoader.DEFAULT_FILENAME_PREFIX);
+    private List<String> moduleRoot = new ArrayList<>();
 
     @Option(name = "--common_js_entry_module",
         hidden = true,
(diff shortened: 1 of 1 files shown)


src/com/google/javascript/jscomp/CommandLineRunner.java
```diff
@@ -1064,6 +1063,7 @@ private void initConfigFromFlags(String[] args, PrintStream out, PrintStream err
       // For backwards compatibility, allow both commonJsPathPrefix and jsModuleRoot.
       List<String> moduleRoots = new ArrayList<>(flags.commonJsPathPrefix);
       moduleRoots.addAll(flags.moduleRoot);
+      moduleRoots.add(ES6ModuleLoader.DEFAULT_FILENAME_PREFIX);
```
MatrixFrog: If the user is actually passing flags on the command line, I would expect they would include the default path (`./`) explicitly. So I'm not sure we actually want this line, except in the case when flags.moduleRoot is empty.

rintaro: Actually, current `ES6ModuleLoader` loads modules from current working directory *regardless* `moduleRoots`  contains `./` or not.

MatrixFrog: In that case this line doesn't actually do anything?

rintaro: I think so.

MatrixFrog: Yeah, let's leave it out if it's actually a no-op.

rintaro: *If* we can remove it from here, `ES6ModuleLoader.DEFAULT_FILENAME_PREFIX` itself eventually can be removed from the entire project. I think I should look a little more deeply it.

rintaro: or @mprobst Do you have any thought on this?

mprobst: I'm not sure if that's correct.

Imagine you compile files `a/b.js` and `./c.js`. `ES6ModuleLoader` will, if given `./` as a module root, find `a/b` and `c` as modules. If you don't have `./` as a root, it will not normalize the `./c.js` and won't find it if code just write `import {something} from 'c';`.

So IMHO we do need `./`, and we should maybe actually think about also adding the absolute URI of the current working directory by default. E.g. if somebody specifies their inputs with absolute paths like `/home/user/foo.js` and `/home/user/bar.js`, then runs the compiler in `/home/user`, we maybe should find just `import {x} from 'bar';` relative to `foo.js`.

But let's fix that if/when somebody actually has the problem.

Dominator008: It's unfortunate that we don't have any specs for this.

mprobst:
@rintaro ping, it turns out I'm actually blocked by this, too ;-)

rintaro:
Sorry for late reply.
OK, as a quick fix for this immediate problem,  I will update this pull request just implements @MatrixFrog 's suggestion:
> So I'm not sure we actually want this line, except in the case when flags.moduleRoot is empty.

But, I'll be late tonight in Japan, may be up to 10 hours later from now.

mprobst:
Thanks @rintaro, and have a nice weekend!

PR #1123, opened 2015-09-09
New pass WhitespaceWrapGoogModules in specialist "always pre-check" phase (fixes #611) [rebased]

### LINKED ISSUE #611: Whitespace only mode breaks goog.module

supersteves:
Rebased version of https://github.com/google/closure-compiler/pull/1094 with redundant reverted commits removed.

Copied from original PR:

> I predict this will be rejected since it is arguable that the WHITESPACE_ONLY mode should mean just that. However my argument is that goog.modules are unusable without this (#611), therefore it is forgiveable.

> For a large codebase, you need to be able to develop rapidly without waiting for the ~ 40 second compile time we're currently seeing. Whitespace only mode drops this to ~2 seconds.

> Also I'm adding a new pass phase "always pre-check" (which happens always, even in whitespace only mode) immediately prior to the "checks" phase (which only happens in simple or advanced mode). Even more debatable! I first tried using a custom pass in the options (see reverted commits), but I don't have the Compiler instance yet (needed for compiler.reportCodeChange()) when I build the options in my external Java compiler front-end. If you don't like this, I could find a way to do it using custom passes, but that's not going to help other users as much.

> Currently WhitespaceWrapGoogModules is opt-in, but should be made default in WHITESPACE_ONLY once approved.


supersteves:
On retrospect, it may be better if I reword these options and passes as

supersteves:
Re-opening, mistakenly closed

supersteves:
On previous PR, @concavelenz wrote:
> I'm content with a change like this but I also want to expose the bundling logic (which would preserve comments, etc).

Presumably the parser + code generator won't do this easily, and you mean exposing ClosureBundler as an API as a separate idea for the future?

supersteves:
@blickly wrote:
> We actually already have a case where whitespace only mode runs the checks (when we are transpiling from ES6 down to ES3/5), so it's reasonable to just run check unconditionally, and add this logic to the beginning of getChecks in DefaultPassConfig.

(for https://github.com/visokio/closure-compiler/commit/48240823475e2fff8ec9e30e8aec4bb5cc99ceec)

I did see that you have:

* "do checks?" => doChecks() method
* "skip non transpilation passes" => on a case by case basis, disable checks and actually only do transpilation

However:

* Transpilation isn't "checks". It doesn't feel right having "checks" called when transpiling only.
* "goog.module" bundling in whitespace only mode isn't a check, nor is it transpilation. I'm using ES6 to ES6 with Chrome (now stable). (Which works beautifully.)  I don't want to run anything unless I have to, for fast refresh in development.

So this bundling doesn't fit into "do checks" or "do transpilation / skip non-transpilation". Hence I added a pass "always pre-checks", terrible name I admit, for this exclusive case of bundling.

It doesn't make sense to have to turn on checks or transpilation to get whitespace only mode to work with goog modules. Please give me some suggestions to adapt this so this PR could be accepted.

supersteves:
Are there any suggestions you can make so I can get this PR accepted, please? I am currently working in a permanent fork and it makes development of PRs and keeping up to date with google/master difficult. Happy to rework this in a different form, but whitespace mode for goog.modules is an essential part of our workflow. Thanks.

test/com/google/javascript/jscomp/WhitespaceWrapGoogModulesTest.java
```diff
@@ -0,0 +1,64 @@
...
+public class WhitespaceWrapGoogModulesTest extends CompilerTestCase {
+
+  private LanguageMode languageOut;
+
+  @Override
+  public void setUp() {
+    setAcceptedLanguage(LanguageMode.ECMASCRIPT6);
+    languageOut = LanguageMode.ECMASCRIPT6;
+    disableTypeCheck();
+    enableLineNumberCheck(false);
```
blickly: I think we want to make sure that the line numbers are copied over correctly.

test/com/google/javascript/jscomp/WhitespaceWrapGoogModulesTest.java
```diff
@@ -0,0 +1,64 @@
...
+    enableCompareAsTree(false); // otherwise the "use strict" pops to the top and breaks the test
```
blickly: I'm not sure I follow here

src/com/google/javascript/jscomp/WhitespaceWrapGoogModules.java
```diff
@@ -0,0 +1,73 @@
...
+        IR.getprop(IR.name("goog"), IR.string("loadModule")),
+        IR.function(IR.name(""), IR.paramList(IR.name("exports")), block)));
+    // ? .srcrefTree(srcref);
+
+    scriptRoot.removeChildren();
+    scriptRoot.addChildToBack(loadMod);
+    compiler.reportCodeChange();
+  }
+
+  // from ClosureRewriteModule, TODO move to common area
```
blickly: NodeUtil, perhaps?

src/com/google/javascript/jscomp/WhitespaceWrapGoogModules.java
```diff
@@ -0,0 +1,73 @@
...
+	          "\n" // terminate any trailing single line comment.
+	          + ";" // terminate any trailing expression.
+	          + "return exports;});\n");
+	      appendSourceUrl(out, Mode.NORMAL);
+     */
+
+    Node block = IR.block();
+    block.addChildToBack(IR.exprResult(IR.string("use strict"))); // needs to be explicit, to match ClosureBundler
+    for (Node c = scriptRoot.getFirstChild(); c != null; c = c.getNext()) {
+      block.addChildToBack(c.cloneTree());
```
blickly: I don't think we need to clone the entire trees here. Just detaching them should be enough.

src/com/google/javascript/jscomp/DefaultPassConfig.java
```diff
@@ -174,6 +174,17 @@ void maybeInitializePreprocessorSymbolTable(AbstractCompiler compiler) {
       }
     }
   }
+  
+  @Override
+  protected List<PassFactory> getAlwaysPreChecks() {
+    List<PassFactory> preChecks = new ArrayList<>();
+    
+    if (options.wrapGoogModulesForWhitespaceOnly) {
```
blickly: Seems a strange to me that wrapGoogModulesForWhitespaceOnly changes the behavior for non-whitespace only code.
My inclination is to just make this the default for whitespace-only when closurePass is set, and remove the option.

src/com/google/javascript/jscomp/WhitespaceWrapGoogModules.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+  }
+
+  @Override
+  public void hotSwapScript(Node scriptRoot, Node originalRoot) {
+    if (!isModuleFile(scriptRoot)) return;
+
+    // As per ClosureBundler:
+    /*
+	      // add the prefix on the first line so the line numbers aren't affected.
```
blickly: In general, we avoid tab characters

src/com/google/javascript/jscomp/DefaultPassConfig.java
supersteves: Happy to simplify, remove the option, and tie it to whitespace + closurePass mode. I only put it there to make it unambiguously an opt-in change given the debatable nature of doing changes in a so-called whitespace-only mode ;-)

src/com/google/javascript/jscomp/WhitespaceWrapGoogModules.java
supersteves: Me too. Oops

src/com/google/javascript/jscomp/WhitespaceWrapGoogModules.java
supersteves: Will do, and I'll make ClosureRewriteModule use it.

test/com/google/javascript/jscomp/WhitespaceWrapGoogModulesTest.java
supersteves: Done

test/com/google/javascript/jscomp/WhitespaceWrapGoogModulesTest.java
supersteves: I'm trying to assert that the output is identical to ClosureBundler, i.e. wrapping the entire script in:
``` javascript
goog.loadModule(function(exports) {'use strict';
...
;return exports;});
```
When I remove ```enableCompareAsTree(false);``` the unit fails because the expected output has been parsed and reformatted with "use strict" removed or moved elsewhere.

blickly: Where is the 'use strict' moved to? I wouldn't expect parsing and printing to change the AST.

supersteves: Here's what my test expects...
``` javascript
        "goog.loadModule(function(exports){"+
            "\"use strict\";"+
            "goog.module(\"test\");"+
            "var f=5;"+
            "exports=f;"+
            "return exports"+
        "})");
```
Test failure with that line removed:
```
Expected: goog.loadModule(function(exports){goog.module("test");var f=5;exports=f;return exports})
Result:   goog.loadModule(function(exports){"use strict";goog.module("test");var f=5;exports=f;return exports})
Node tree inequality:
Expected tree:
BLOCK [synthetic: 1]
    SCRIPT 1 [synthetic: 1] [source_file: expected0] [input_id: InputId: expected0]
        EXPR_RESULT 1 [source_file: expected0]
            CALL 1 [source_file: expected0]
                GETPROP 1 [source_file: expected0]
                    NAME goog 1 [source_file: expected0]
                    STRING loadModule 1 [source_file: expected0]
                FUNCTION  1 [source_file: expected0]
                    NAME  1 [source_file: expected0]
                    PARAM_LIST 1 [source_file: expected0]
                        NAME exports 1 [source_file: expected0]
                    BLOCK 1 [directives: [use strict]] [source_file: expected0]
                        EXPR_RESULT 1 [source_file: expected0]
                            CALL 1 [source_file: expected0]
                                GETPROP 1 [source_file: expected0]
                                    NAME goog 1 [source_file: expected0]
                                    STRING module 1 [source_file: expected0]
                                STRING test 1 [source_file: expected0]
                        VAR 1 [source_file: expected0]
                            NAME f 1 [source_file: expected0]
                                NUMBER 5.0 1 [source_file: expected0]
                        EXPR_RESULT 1 [source_file: expected0]
                            ASSIGN 1 [source_file: expected0]
                                NAME exports 1 [source_file: expected0]
                                NAME f 1 [source_file: expected0]
                        RETURN 1 [source_file: expected0]
                            NAME exports 1 [source_file: expected0]


Actual tree:
BLOCK [synthetic: 1]
    SCRIPT 1 [synthetic: 1] [source_file: testcode] [input_id: InputId: testcode]
        EXPR_RESULT 1 [source_file: testcode]
            CALL 1 [source_file: testcode]
                GETPROP 1 [source_file: testcode]
                    NAME goog 1 [source_file: testcode]
                    STRING loadModule 1 [source_file: testcode]
                FUNCTION  1 [source_file: testcode]
                    NAME  1 [source_file: testcode]
                    PARAM_LIST 1 [source_file: testcode]
                        NAME exports 1 [source_file: testcode]
                    BLOCK 1 [source_file: testcode]
                        EXPR_RESULT 1 [source_file: testcode]
                            STRING use strict 1 [source_file: testcode]
                        EXPR_RESULT 1 [source_file: testcode]
                            CALL 1 [source_file: testcode]
                                GETPROP 1 [source_file: testcode]
                                    NAME goog 1 [source_file: testcode]
                                    STRING module 1 [source_file: testcode]
                                STRING test 1 [source_file: testcode]
                        VAR 1 [source_file: testcode]
                            NAME f 1 [source_file: testcode]
                                NUMBER 5.0 1 [source_file: testcode]
                        EXPR_RESULT 1 [source_file: testcode]
                            ASSIGN 1 [source_file: testcode]
                                NAME exports 1 [source_file: testcode]
                                NAME f 1 [source_file: testcode]
                        RETURN 1 [source_file: testcode]
                            NAME exports 1 [source_file: testcode]


Expected subtree: BLOCK 1 [directives: [use strict]] [source_file: expected0]
    EXPR_RESULT 1 [source_file: expected0]
        CALL 1 [source_file: expected0]
            GETPROP 1 [source_file: expected0]
                NAME goog 1 [source_file: expected0]
                STRING module 1 [source_file: expected0]
            STRING test 1 [source_file: expected0]
    VAR 1 [source_file: expected0]
        NAME f 1 [source_file: expected0]
            NUMBER 5.0 1 [source_file: expected0]
    EXPR_RESULT 1 [source_file: expected0]
        ASSIGN 1 [source_file: expected0]
            NAME exports 1 [source_file: expected0]
            NAME f 1 [source_file: expected0]
    RETURN 1 [source_file: expected0]
        NAME exports 1 [source_file: expected0]


Actual subtree: BLOCK 1 [source_file: testcode]
    EXPR_RESULT 1 [source_file: testcode]
        STRING use strict 1 [source_file: testcode]
    EXPR_RESULT 1 [source_file: testcode]
        CALL 1 [source_file: testcode]
            GETPROP 1 [source_file: testcode]
                NAME goog 1 [source_file: testcode]
                STRING module 1 [source_file: testcode]
            STRING test 1 [source_file: testcode]
    VAR 1 [source_file: testcode]
        NAME f 1 [source_file: testcode]
            NUMBER 5.0 1 [source_file: testcode]
    EXPR_RESULT 1 [source_file: testcode]
        ASSIGN 1 [source_file: testcode]
            NAME exports 1 [source_file: testcode]
            NAME f 1 [source_file: testcode]
    RETURN 1 [source_file: testcode]
        NAME exports 1 [source_file: testcode]
```
It seems "use strict" is treated specially as a directive which breaks the comparison. So I changed the test to compare the source rather than the AST.

src/com/google/javascript/jscomp/DefaultPassConfig.java
supersteves: Actually I don't know what you mean about making this tied to closurePass. The point of whitespace only is to make it bundled with minimal changes and as fast as possible. Or at least that's my reason for using it. Options.closurePass is false in whitespace only mode.

I think you mean whitespace only mode can't work without this, so we should remove the opt-in / always enable this pass in whitespace-only mode. 

But whitespace only mode skips all non transpilation passes, and I'm not transpiling (why would I in normal development, now Chrome stable supports so many ES6 features?), so no passes are done. Hence the new "always pre checks" stage.

Incidentally why does the debugger disallow ES6 out (https://closure-compiler-debugger.appspot.com/#input0%3Dlet%2520x%2520%253D%25205%253B%250A%26input1%26conformanceConfig%26externs%26refasterjs-template%26includeDefaultExterns%3D1%26LANG_IN_IS_ES6%3D1)? I use this in Chrome 45 via the Java API as part of the lean and fast whitespace only mode (with this PR).

test/com/google/javascript/jscomp/WhitespaceWrapGoogModulesTest.java
supersteves: Note that the final output of whitespace only with this PR and without closurePass does indeed retain the "use strict" in the same place as my test assertion.

src/com/google/javascript/jscomp/DefaultPassConfig.java
blickly: In the debugger WHITESPACE_ONLY is called SKIP_NON_TRANSPILATION_PASSES.

Good point about not having closurePass. Maybe for now we should just check that we're in whitespace-only mode in addition to the compiler option, but eventually I'm ok with making it default to on (since using goog.modules in whitespace only mode doesn't really make sense without this option).

test/com/google/javascript/jscomp/WhitespaceWrapGoogModulesTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+  private LanguageMode languageOut;
+
+  @Override
+  public void setUp() {
+    setAcceptedLanguage(LanguageMode.ECMASCRIPT6);
+    languageOut = LanguageMode.ECMASCRIPT6;
+    disableTypeCheck();
+    
+    enableCompareAsTree(false);
```
blickly: Looks like the parser parses 'use strict' as a directive attached to the block node rather than as a string expression result.

supersteves: ... but it's ok for this test and this PR to workaround it by disabling tree comparison? The source based assertion should be ok, given this is a whitespace only pass.

src/com/google/javascript/jscomp/DefaultPassConfig.java
supersteves: @blickly I pushed a change that simplifies things somewhat. Please see TODO added to Compiler change.

src/com/google/javascript/jscomp/Compiler.java
```diff
@@ -672,6 +672,11 @@ private void compileInternal() {
     if (!precheck()) {
       return;
     }
+    
+    if (options.skipNonTranspilationPasses) { // TODO: add clause "|| options.closurePass"?
```
supersteves: that's meant to be:
```javascript
// TODO: add clause "|| !options.closurePass"?
```
I'm asking is there a case where skipNonTranspilationPasses=false but we still need this goog.module wrapping?

blickly: I don't think so.

src/com/google/javascript/jscomp/WhitespaceWrapGoogModules.java
```diff
@@ -0,0 +1,78 @@
...
+        IR.getprop(IR.name("goog"), IR.string("loadModule")),
+        IR.function(IR.name(""), IR.paramList(IR.name("exports")), block)))
+        .srcrefTree(scriptRoot);
+
+    if (scriptRoot.hasChildren()) {
+      block.addChildrenToBack(scriptRoot.removeChildren());
+    }
+    block.addChildToBack(IR.returnNode(IR.name("exports")).srcrefTree(scriptRoot));
+    
+    scriptRoot.removeChildren();
```
blickly: This looks redundant, since the children have already been removed.

test/com/google/javascript/jscomp/WhitespaceWrapGoogModulesTest.java
```diff
@@ -0,0 +1,67 @@
...
+  @Override
+  protected CompilerOptions getOptions() {
+    CompilerOptions options = super.getOptions();
+    options.setLanguageOut(languageOut);
+    return options;
+  }
+
+  @Override
+  protected int getNumRepetitions() {
+    return 1;
```
blickly: Since this pass is a no-op for non goog.module code, I'm guessing we don't actually need this?

supersteves: Removed, still passes.

supersteves:
@blickly thanks for all the feedback. I've addressed all of it, I think.

blickly:
The only remaining issue I see now, is that now we have a problem with transpiling in WHITESPACE_ONLY, in that goog.modes are rewritten twice. Once in this pass, and once by the old ClosureRewriteModule.

Which makes me think again, wouldn't it be OK to just rewrite modules in the ClosureRewriteModule style, rather than having two styles of module rewriting?

supersteves:
On the latter point, probably not. The point of WHITESPACE_ONLY is to do no rewriting at all, to rule out any compiler-caused issues, to facilitate debugging especially in older browsers, and for build performance. Or at least that's why we use it. With goog.modules, of course that's not possible. This new pass does the minimum AST change needed. ClosureRewriteModules has a lot more complexity in it.

Re the first point: I would think that the new pass should be used instead, when transpiling only, for the same reason - WHITESPACE_ONLY transpiled mode should make only the absolute minimum changes to make it compatible with older browsers.

If you agree I'll update the PR to treat Closure RewriteModules as a non transpilation pass, and skip it according to skipNonTranspilationPasses.

blickly:
People sometimes run WHITESPACE_ONLY transpilations as a way to do per-file compilation down to a version that can run in their browser, and then pass the result into a full ADVANCED optimized compile.
The issue I see now is that it looks like the output of this new pass will not be understood and optimized as well as it was previously.

Also note that ClosureRewriteModules does a fairly minimal rewrite compared to other closure passes (basically just rewrites the goog.module/goog.require, but keeps the code in the body the same).

supersteves:
I didn't look in depth, but that class seems several times larger. The original issue

Thinking in terminology, what does "closure" signify when you have "closure pass" option in the debugger and "closure rewrite modules" class? Doesn't seem to fit with whitespace only's description.

For myself I don't really mind if we discard my PR, I'd just like to get "closure" (sorry!) on having a blindingly fast whitespace only non transpiled mode that is goog module compatible. The reason I went down this route with my efforts was that it was the suggestion from @concavelenz here: https://github.com/google/closure-compiler/issues/611#issuecomment-63153872.

blickly:
I could be convinced either way on which way the modules should be rewritten in non-transpiled whitespace only mode, but I strongly agree that goog.module should be rewritten by default in whitespace only mode (because otherwise the code gets silently broken), and that when transpilation is also required, it should use the ClosureRewriteModules rewriting, because some people have a workflow that depends on that (doing per-file transpilation).

supersteves:
Curious why people sometimes run compilation twice, presumably with chained sourcemap generation. Why don't they just go straight for advanced compile from scratch? Is this to optimise the build process so you get a runnable version first, then reuse that bundling stage outputs to do the build, thus reducing I/O on many source files?

If you discounted this particular scenario for a second, looking at the compiler options and how they are described (and intended?), it would be better if whitespace only was as minimal in effect as possible.

E.g. when transforming this code:
``` javascript
goog.module('foo')
var bar = goog.require('bar');
exports = something;
```
ClosureRewriteModule produces:
``` javascript
goog.provide('foo');
goog.require('ns.bar');
goog.scope(function() {
  var bar = ns.bar;
  foo = something;
});
```
WhitespaceWrapGoogModules produces:
``` javascript
goog.loadModule(function(exports) {'use strict';
  goog.module('foo')
  var bar = goog.require('bar');
  exports = something;
;return exports;});
```

Isn't the latter closer to what a developer expects from a JS module system such as CommonJS and closure to how ES6 modules behave particularly regarding how required modules are referenced at runtime / when debugging?
Also, from ClosureRewriteModule:
``` javascript
// TODO(johnlenz): Don't use goog.scope as an intermediary; add type checker support instead.
```

How about:

* preserving whitespace only as you describe while also fixing rewrite modules in non-transpiled mode using ClosureRewriteModules (e.g. in case those same people find they can use ES6 feature X in browser Y and no longer need to transpile in development)?
* and introducing a new compilation level "MINIMAL_BUNDLING" which supports optional transpilation between ES levels, processes modules using WhitespaceWrapGoogModules, and nothing else, perhaps not even stripping whitespace/comments? 

Personally I feel a bit uncomfortable about the way Compiler.check() is and options interplay. It's difficult to follow the relationship between skipNonTranspilationPasses and lowerFromEs6() and seems somewhat random that a bunch of non-transpilation stuff happens precisely when only skipNonTranspilationPasses and lowerFromEs6() are both true (custom passes, strip code, checkJsDoc, closurePass, checkVariableReferences, etc. etc. etc.). Doesn't seem to fit with "whitespace only" goals.

blickly:
Sorry for letting this drop for so long.  It turns out that ClosureRewriteModules accepts code either in the original goog.module format, or in the rewritten bundled format with the leading:
   
    goog.loadModule(function(exports) {

See, for example, the unit tests at:
https://github.com/google/closure-compiler/blob/master/test/com/google/javascript/jscomp/ClosureRewriteModuleTest.java#L147-L211

So looks like this is safe to merge even if people are relying on the two-step workflow.

supersteves:
Yes, this is same old me. I did author these.

supersteves:
@blickly I rebased and have passes, see above.

supersteves:
@blickly I realise I broke the history for this PR (sorry) and have created a new replacement by cherry-picking. See new PR: https://github.com/google/closure-compiler/pull/1365

PR #1129, opened 2015-09-11
Rename private properties only compiler option

dchambers:
This supersedes #1087.

### DIFF
--- a/src/com/google/javascript/jscomp/CompilerOptions.java
+++ b/src/com/google/javascript/jscomp/CompilerOptions.java
@@ -554,6 +554,12 @@ public class CompilerOptions implements Serializable {
   String debugFunctionSideEffectsPath;
 
   /**
+   * Whether to skip the renaming of public properties as determined by the
+   * coding convention.
+   */
+  boolean renamePublicProperties = true;
+
+  /**
    * Rename private properties to disambiguate between unrelated fields based on
    * the coding convention.
    */
@@ -2018,6 +2024,21 @@ public class CompilerOptions implements Serializable {
 
   public void setDebugFunctionSideEffectsPath(String debugFunctionSideEffectsPath) {
     this.debugFunctionSideEffectsPath = debugFunctionSideEffectsPath;
+  }
+
+  /**
+   * @return Whether rename public properties is enabled.
+   */
+  public boolean isRenamePublicProperties() {
+    return renamePublicProperties;
+  }
+
+  /**
+   * @param value Whether to rename public properties based on
+   * the coding convention.
+   */
+  public void setRenamePublicProperties(boolean value) {
+    this.renamePublicProperties = value;
   }
 
   /**
--- a/src/com/google/javascript/jscomp/CompilerOptions.java
+++ b/src/com/google/javascript/jscomp/CompilerOptions.java
@@ -554,10 +554,10 @@ public class CompilerOptions implements Serializable {
   String debugFunctionSideEffectsPath;
 
   /**
-   * Whether to skip the renaming of public properties as determined by the
-   * coding convention.
+   * Whether to only rename properties determined as private by the coding
+   * convention.
    */
-  boolean renamePublicProperties = true;
+  boolean renamePrivatePropertiesOnly = false;
 
   /**
    * Rename private properties to disambiguate between unrelated fields based on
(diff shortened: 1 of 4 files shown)


MatrixFrog:
Can you elaborate on the use case here? Why would someone want to enable/disable this option? What would be the default? In general, the more options we add, the more difficult it is to use the compiler (and for us to support everyone's unique settings) so I'd like to avoid adding more options when possible. We already have a lot of options that control renaming (including annotations like `@export`) so I'd want to see a lot of justification for adding another.

dchambers:
Hi @MatrixFrog, this change allows developers to safely obfuscate their code even though not all of their code is delivered as a single bundle, or in cases where it's impossible to have prior knowledge of all the public methods that may be invoked ahead of time, for example:

  1. They use an app store type system where users are free to add disparately sourced components to a single page that all work together using well defined communication interfaces (e.g. a trading app where the components are produced by distinct business verticals).
  1. Where the serialization mechanism their apps uses to reform saved layouts involves invoking serialized public methods that might otherwise change name in the interim.
  1. They occasionally dynamically send code to the client to debug problems their clients may be having, as part of the telephone support they offer.

For such apps, it's still desirable to obfuscate the code, which involves obfuscating private methods, properties & member variables, in addition to local variables. This change makes these use cases possible.

MatrixFrog:
This seems like a good use case for using lots of `@export` annotations, or similar. Just deciding not to rename all public methods seems overkill. But I'll see what others think as well.

dchambers:
Okay, @MatrixFrog, much appreciated.

But out of interest, how do you propose that you can add lots of `@export` annotations in cases where you can''t have prior knowledge of what might be invoked. At best you will need to add `@export` annotations for the entire public API. This is a more convenient way of achieving that, and one which  uses convention over configuration, so is more DRY.

ChadKillingsworth:
There are problems with `@export` annotations - dependency on closure-library, requires a flag to enable and maintaining mutability.

However, renaming "public" methods would be a source of great confusion - primarily because of the various definitions for "public" (file level visibility, package level visibility, anything accessible from the global scope, etc). Also, because of the dependence on the coding convention, I do not think that this should be an exposed flag - it should require a custom build.

If a desire to have this as a bridge between SIMPLE and ADVANCED optimizations, then renaming private methods is easier to explain. Also, we'll need to make sure that it's clearly stated that the determination of public vs private properties is completely based on naming conventions and NOT type information. This will be another source of confusion. When and where type information is or is not used is already hard enough to explain.

dchambers:
>  I do not think that this should be an exposed flag - it should require a custom build.

It does require a custom build at present, and is not an exposed flag.

> If a desire to have this as a bridge between SIMPLE and ADVANCED optimizations, then renaming private methods is easier to explain.

Yes, I agree with you, but then I also keep flip-flopping on this issue because at the point you choose to enable `ALL_UNQUOTED` (`SIMPLE` doesn't do this by default) then it almost feels more natural to tone down `ALL_UNQUOTED` by saying `renamePublicProperties=false`. Apart from that, I agree it does make more sense to talk about renaming private properties, and what's more the coding convention class has a `isPrivateMethod()` method rather than a `isPublicMethod()` method.

With that said, let me know one way or another and I'll update the pull-request to suit.

dchambers:
> Yes, I agree with you, but then I also keep flip-flopping on this issue because at the point you choose to enable `ALL_UNQUOTED` (`SIMPLE` doesn't do this by default) then it almost feels more natural to tone down `ALL_UNQUOTED` by saying `renamePublicProperties=false`.

It occurred to me that if the property is called `onlyRenamePrivateProperties` then it still feels natural in combination with  `ALL_UNQUOTED`  *and* allows us to talk about private properties.

dchambers:
Hi @ChadKillingsworth, I've now changed things so we instead talk in terms of renaming private properties &mdash; using a `onlyRenamePrivateProperties` property.

ChadKillingsworth:
I'm ok with this change. While not a widely used feature, it seems like it adds very little in the way of code maintenance.

concavelenz:
As I mentioned elsewhere, renaming "private" properties is generally dangerous.  Subclasses private properties can conflict with superclass private properties, mixin properties can conflict.    "in part" property renaming is only "safe" if the object doesn't escape.  Whole world renaming is generally needed to know that properties don't conflict.

dchambers:
> As I mentioned elsewhere, renaming "private" properties is generally dangerous.

Hi @concavelenz, this is already the default when you use `ADVANCED` right now. I've just added an option that limits renaming to private properties only (i.e. it prevents properties that appear to be public properties, based on the naming convention, from being renamed), and so this doesn't, AFAIK, add any additional danger to what's already there.

concavelenz:
This encourages unsafe file-by-file property renaming, which should not be done.   It sounds like your use case is exactly this.  How are you structuring your code such that you avoid this?

ChadKillingsworth:
I think at this point we aren't going to merge this.

PR #1132, opened 2015-09-12
Speed Travis CI execution up + have it run protobuf-gen

ochafik:
This downloads & caches the right version of protoc, and cuts nearly a minute of execution in Travis CI (thanks to single mvn build & caching of local repo).

[Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1132)

### DIFF
--- a/get-protoc.sh
+++ b/get-protoc.sh
@@ -1,0 +1,43 @@
+#!/bin/bash
+# 
+# Gets the version of protoc needed to build Closure, and prints its path.
+# Downloads sources and builds them if needed, and caches the binary in ~/.protobuf.
+#
+# Note: the version of protobuf-compiler available through
+# Travis CI's addons.apt.packages mechanism is 2.4.1, and protoc's version really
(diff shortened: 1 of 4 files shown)


MatrixFrog:
Why do we need protoc at all, since the protoc-generated files are checked in under `gen/`?

get-protoc.sh
```diff
@@ -0,0 +1,43 @@
...
+
+# Where we'll install protoc (on Travis CI, will be cached).
+PROTOBUF_DIR=$HOME/.protobuf/$PROTOBUF_VERSION
+PROTOC=$PROTOBUF_DIR/bin/protoc
+
+function getAndBuildProtoc() {
+  local name=protobuf-$PROTOBUF_VERSION
+
+  rm -fR $name*
+  wget https://protobuf.googlecode.com/files/$name.tar.gz
```
MatrixFrog: (assuming we do need it) is this the right URL to get the latest version? I would imagine there is a dl.google.com URL that is more up to date.

PR #1133, opened 2015-09-12
Transpile ES6 super getprop & getelem calls (fixes #1089)

### LINKED ISSUE #1089: Implement ES6 transpilation of super.getter calls
Calls to super getters aren't supported yet.

``` javascript
class Foo {
  get x() { return 'Foo'; }
}
class Bar extends Foo {
  get x() { return super.x; }
}
console.log(new Foo().x);
console.log(new Bar().x);
```

Current error:

```
ERROR - ES6 transpilation of 'Only calls to super or to a method of super are supported.' is not yet implemented.
  get x() { return super.x; }
                   ^
```

This is needed to compile [DDC](https://github.com/dart-lang/dev_compiler)'s output with Closure.

ochafik:
(fixes https://github.com/google/closure-compiler/issues/1089, which is blocking https://github.com/dart-lang/dev_compiler/issues/312)

### DIFF
--- a/src/com/google/javascript/jscomp/Es6ConvertSuper.java
+++ b/src/com/google/javascript/jscomp/Es6ConvertSuper.java
@@ -62,7 +62,7 @@ public final class Es6ConvertSuper implements NodeTraversal.Callback, HotSwapCom
       }
       checkClassSuperReferences(n);
     } else if (isSuperGet(n)) {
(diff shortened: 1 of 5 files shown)


ochafik:
Fixed in https://github.com/google/closure-compiler/commit/3f54e473500ee54a4aa5c264dbe6f61d82586a8d

Issue #1136, opened 2015-09-13
ES6 template strings fail parsing, when there's another parse error


supersteves:
I'm using reserved word 'type' and this causes template strings parsing to generate false errors:

``` javascript
function f({type}) {

  var s = `translate(${this._position.x}, ${this._position.y})`);

};
```

```
input0:1: ERROR - Parse error. cannot use keyword 'type' here.
function f({type}) {
                ^

input0:3: ERROR - Parse error. primary expression expected
  var s = `translate(${this._position.x}, ${this._position.y})`);
                                         ^

input0:3: ERROR - Parse error. Semi-colon expected
  var s = `translate(${this._position.x}, ${this._position.y})`);
                                          ^

input0:3: ERROR - Parse error. Semi-colon expected
  var s = `translate(${this._position.x}, ${this._position.y})`);
                                           ^

input0:3: ERROR - Parse error. primary expression expected
  var s = `translate(${this._position.x}, ${this._position.y})`);
                                                              ^

input0:3: ERROR - Parse error. Unterminated template literal
  var s = `translate(${this._position.x}, ${this._position.y})`);
                                                               ^
```

https://closure-compiler-debugger.appspot.com/#input0%3Dfunction%2520f(%257Btype%257D)%2520%257B%250A%250A%2520%2520var%2520s%2520%253D%2520%2560translate(%2524%257Bthis._position.x%257D%252C%2520%2524%257Bthis._position.y%257D)%2560)%253B%250A%250A%257D%253B%250A%26input1%26conformanceConfig%26externs%26refasterjs-template%26includeDefaultExterns%3D1%26IDE_MODE%3D1%26LANG_IN_IS_ES6%3D1

If I turn off IDE mode, the 5 template string parse errors disappear when running locally, or crash using the debugger:

https://closure-compiler-debugger.appspot.com/#input0%3Dfunction%2520f(%257Btype%257D)%2520%257B%250A%250A%2520%2520var%2520s%2520%253D%2520%2560translate(%2524%257Bthis._position.x%257D%252C%2520%2524%257Bthis._position.y%257D)%2560)%253B%250A%250A%257D%253B%250A%26input1%26conformanceConfig%26externs%26refasterjs-template%26includeDefaultExterns%3D1%26LANG_IN_IS_ES6%3D1

```
java.lang.IllegalArgumentException: Unexpected control flow graph root BLOCK [synthetic: 1]
[...]

PR #1137, opened 2015-09-13
Fix spurious parse error on TypeScript keywords

### LINKED ISSUE #1135: ERROR - Parse error. cannot use keyword 'type' here.
When I tried to compile my ECMAScript 6 program using the Closure Compiler, I ran into this error.

Here is the smallest reproducible test case:

```
var { type } = { type: 1 };
```

And here is the error message I get:

```
test.js:1: ERROR - Parse error. cannot use keyword 'type' here.
var { type } = { type: 1 };
           ^
```

As far as I know, `type` is not a keyword or reserved word in any version of ECMAScript.

----

In addition, this code compiles perfectly fine:

```
var type = 1;
```

So it seems Closure is only treating `type` as a keyword in certain situations.

Dominator008:
TypeScript specific keywords should be allowed as object literal field names. Fixed the parser to not report errors on them.

Fixes #1135 and #1136

### DIFF
--- a/src/com/google/javascript/jscomp/parsing/parser/Keywords.java
+++ b/src/com/google/javascript/jscomp/parsing/parser/Keywords.java
@@ -120,6 +120,18 @@ public enum Keywords {
   public static boolean isKeyword(TokenType token) {
     return get(token) != null;
   }
+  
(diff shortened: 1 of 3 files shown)


blickly:
Great, thanks for the quick fix!

PR #1140, opened 2015-09-14
Add externs for the W3C's Touch Events Extensions specification

fredj:


MatrixFrog:
If this is (or is expected to be) an official w3c spec it can probably go in externs rather than contrib/externs. (It looks like Touch itself is defined in a file called iphone.js but it sounds like the API is actually implemented on Android browsers as well, so we should rename that at some point.)

ChadKillingsworth:
A quick scan of the iphone.js file leads me to think that almost everything in there should be moved to other files.

ChadKillingsworth:
@fredj If you'd like to move the Touch events out of iphone.js and into `/externs/browser/touchevents.js`, you can add  your changes to that.

Official spec for touch events: http://www.w3.org/TR/touch-events/

fredj:
I've moved out the touch events from iphone.js to w3c_touch_event.js and added the Touch Events Extensions stuff to that file.

I've also added the prefixed (`webkit`) Touch Events Extensions properties to iphone.js

PR #1154, opened 2015-09-19
ES6: Accept IdentiferName (including Keyword) in ExportSpecifier and ImportSpecifier.

rintaro:
see:
    http://www.ecma-international.org/ecma-262/6.0/index.html#sec-imports
    http://www.ecma-international.org/ecma-262/6.0/index.html#sec-exports

This fix is necessary for  implementing "package index" module. e.g.:

```javascript
// index.js: 

export {default as AClass} from "./AClass" // AClass.js: export default class AClass { ...}
export {default as BClass} from "./BClass" // BClass.js: export default class BClass { ...}
```

### DIFF
--- a/src/com/google/javascript/jscomp/parsing/parser/Parser.java
+++ b/src/com/google/javascript/jscomp/parsing/parser/Parser.java
@@ -402,7 +402,7 @@ public class Parser {
     ImmutableList.Builder<ParseTree> elements;
     elements = ImmutableList.builder();
     eat(TokenType.OPEN_CURLY);
-    while (peekId()) {
+    while (peekIdOrKeyword()) {
       elements.add(parseImportSpecifier());
       if (!peek(TokenType.CLOSE_CURLY)) {
         eat(TokenType.COMMA);
@@ -415,9 +415,10 @@ public class Parser {
   //  ImportSpecifier ::= Identifier ('as' Identifier)?
   private ParseTree parseImportSpecifier() {
     SourcePosition start = getTreeStartLocation();
-    IdentifierToken importedName = eatId();
+    boolean isKeyword = Keywords.isKeyword(peekType());
+    IdentifierToken importedName = eatIdOrKeywordAsId();
     IdentifierToken destinationName = null;
-    if (peekPredefinedString(PredefinedName.AS)) {
+    if (isKeyword || peekPredefinedString(PredefinedName.AS)) {
       eatPredefinedString(PredefinedName.AS);
       destinationName = eatId();
     }
@@ -537,7 +538,7 @@ public class Parser {
     ImmutableList.Builder<ParseTree> elements;
     elements = ImmutableList.builder();
     eat(TokenType.OPEN_CURLY);
-    while (peekId()) {
+    while (peekIdOrKeyword()) {
       elements.add(parseExportSpecifier());
       if (!peek(TokenType.CLOSE_CURLY)) {
         eat(TokenType.COMMA);
@@ -550,11 +551,11 @@ public class Parser {
   //  ExportSpecifier ::= Identifier ('as' Identifier)?
   private ParseTree parseExportSpecifier() {
     SourcePosition start = getTreeStartLocation();
-    IdentifierToken importedName = eatId();
+    IdentifierToken importedName = eatIdOrKeywordAsId();
     IdentifierToken destinationName = null;
     if (peekPredefinedString(PredefinedName.AS)) {
       eatPredefinedString(PredefinedName.AS);
-      destinationName = eatId();
+      destinationName = eatIdOrKeywordAsId();
     }
     return new ExportSpecifierTree(
         getTreeLocation(start), importedName, destinationName);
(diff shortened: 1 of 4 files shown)


MatrixFrog:
Is any keyword supposed to be valid? Can you do this (even though it's a bit weird)? `import {else as x}` or `import {switch as x}` etc.

concavelenz:
This doesn't sound right.  Is it just "default" you want handled?

rintaro:
*Yes*, any keyword should be accepted I think.
https://esdiscuss.org/topic/moduleimport#content-55

test/com/google/javascript/jscomp/parsing/NewParserTest.java
```diff
@@ -2648,9 +2648,34 @@ public void testImport() {
...
+
+    parse("export {x as x1, y as y1} from './someModule'");
+    parse("export {x as x1, y as y1, } from './someModule'");
+    parse("export {default as d} from './someModule'");
+    parse("export {x as default, y as y1} from './someModule'");
+    parse("export {default as default, y as y1} from './someModule'");
+    parse("export * from './someModule'");
+  }
+
   public void testShebang() {
```
blickly: Can we add some more tests of the error cases?
In particular, I think we need:

```javascript
import keyword from '...';
import {nonkeyword as keyword} from '...';
import {keyword} from '...';
export {keyword};````

Comment (rintaro) [inline test/com/google/javascript/jscomp/parsing/NewParserTest.java]: Done.

File: test/com/google/javascript/jscomp/ProcessEs6ModulesTest.java [hunk 2]
```diff
@@ -108,6 +108,10 @@ public void testImport() {
     testModules(
         "import x, {f as foo, b as bar} from 'other'; use(x);",
         LINE_JOINER.join("goog.require('module$other');", "use(module$other.default);"));
+
+    testModules(
+        "import {default as name} from 'other'; use(name);",
+        LINE_JOINER.join("goog.require('module$other');", "use(module$other.default);"));
```
MatrixFrog: The purpose of the line joiner is to put the strings one on each line so it reads more naturally. Either do that or just put it all on one line in one string.

test/com/google/javascript/jscomp/parsing/NewParserTest.java
```diff
@@ -2657,7 +2657,44 @@ public void testImport() {
...
     parse("import {x as x1, y as y1, } from './someModule'");
+    parse("import {default as d} from './someModule'");
+    parse("import d, {x as x1, y as y1} from './someModule'");
     parse("import * as sm from './someModule'");
+
+    parseError("import class from './someModule'",
+            "cannot use keyword 'class' here.");
+    parseError("import * as class from './someModule'",
+            "'identifier' expected");
+    parseError("import {a as class} from './someModule'",
```
MatrixFrog: If I understand correctly {a as class} is invalid (makes sense as it would put a var called 'class' into the module scope) but {class as a} is valid. I don't see tests of this valid case ({someKeyword as nonKeyword}) except for `default`.

test/com/google/javascript/jscomp/parsing/NewParserTest.java
```diff
@@ -2657,7 +2657,44 @@ public void testImport() {
...
+    parse("export var x = 1");
+    parse("export function f() {}");
+    parse("export class c {}");
+    parse("export {x, y}");
+    parse("export {x as x1}");
+    parse("export {x as default, y as y1}");
+
+    parseError("export {default as x}",
+        "cannot use keyword 'default' here.");
+    parseError("export {package as x}",
```
MatrixFrog: I take it back, that would be this case. So was I misunderstanding? Are you not allowed to export things under keyword names other than default?

blickly: Import and export are opposite.
In an export, the first name is an identifier and the second is a property.
In an import, the first name is a property and the second is an identifier.

MatrixFrog: Ah right. So then we should still have a test which says `import {class as a}` which is valid, as well as one that says `export {a as class}` which is also valid. Right?

src/com/google/javascript/jscomp/parsing/parser/Parser.java
```diff
@@ -3414,6 +3425,15 @@ private boolean peekId(int index) {
         || (!inStrictContext() && Keywords.isStrictKeyword(type));
   }
 
+  private boolean peekIdOrKeyword() {
+    return peekIdOrKeyword(0);
```
blickly: Let's remove the int index version since it doesn't look like we need it.

test/com/google/javascript/jscomp/parsing/NewParserTest.java
blickly: Seems pretty much the same as the 
`import {default as d}`
and
`export {x as default}`
cases that we already have. But I wouldn't object to adding them.

blickly:
Thanks for the fix!

PR #1164, opened 2015-09-23
Add Array.prototype.toLocaleString function in es3 externs

fredj:


MatrixFrog:
This just overrides Object.prototype.toLocaleString which we already have. I don't think we need a separate definition for the Array version.

concavelenz:
Reopen if there is a non-obvious need for this.

fredj:
Thanks for the explanation

PR #1166, opened 2015-09-24
Remove UTF-8 BOM from source files

ChadKillingsworth:
UTF8 Byte Order Marks interfere with goog.require/provide regular expression matching. This removes the BOM when the file is read from disk.

I thought this had already been fixed, but came across it on a [stackoverflow issue](http://stackoverflow.com/a/32747595/1211524).

### DIFF
--- a/src/com/google/javascript/jscomp/SourceFile.java
+++ b/src/com/google/javascript/jscomp/SourceFile.java
@@ -52,6 +52,7 @@ import java.util.zip.ZipFile;
  */
 public class SourceFile implements StaticSourceFile, Serializable {
   private static final long serialVersionUID = 1L;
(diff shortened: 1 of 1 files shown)


ChadKillingsworth:
This problem plagues windows machines as certain editors (such as Visual Studio) make it very difficult to save files without adding the byte order marker.

MatrixFrog:
I didn't think the compiler did any regex matching; I thought it was always looking at the parsed AST now. But this seems fine to do in any case.

ChadKillingsworth:
@MatrixFrog See the last paragraph https://github.com/google/closure-compiler/wiki/Manage-Closure-Dependencies#faq

MatrixFrog:
> There's a reason we did it this way. It wasn't 100% laziness....just 80-90% laziness. 

<3 okay

PR #1167, opened 2015-09-24
Revert "Remove UTF-8 BOM from source files"

blickly:
Reverts google/closure-compiler#1166

SourceFile is sometimes compiled without CommandLinerRunner, and String.format is not GWT-compatible.

### DIFF
--- a/src/com/google/javascript/jscomp/SourceFile.java
+++ b/src/com/google/javascript/jscomp/SourceFile.java
@@ -52,7 +52,6 @@ import java.util.zip.ZipFile;
  */
 public class SourceFile implements StaticSourceFile, Serializable {
   private static final long serialVersionUID = 1L;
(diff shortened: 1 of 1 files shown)


Dominator008:
Try SimpleFormat.format?

blickly:
Bingo. Maybe we can also add a test to the maven build that we can
GWT-compile successfully.

Dominator008:
We can, but we need to use GWT's trunk / nightly build until 2.8.0 is out.

ChadKillingsworth:
So reading between the lines, the issue was with the `String.format` command? On internal failures and reverts, I don't see the same things you do.

MatrixFrog:
That was part of it, and the other part is that SourceFile can't depend on CommandLineRunner because of the way the internal build is set up.

PR #1168, opened 2015-09-25
Add Document.prototype.readyState to the HTML5 externs

Dominator008:
This eliminates a new type inference warning in Closure Library's base.js

### DIFF
--- a/externs/browser/html5.js
+++ b/externs/browser/html5.js
@@ -735,6 +735,13 @@ Document.prototype.postMessage = function(message) {};
 Document.prototype.head;
 
 /**
+ * Describes the loading state of the document.
+ * @see https://html.spec.whatwg.org/multipage/dom.html#current-document-readiness
+ * @type {string}
+ */
+Document.prototype.readyState;
+
+/**
  * @see https://developer.apple.com/webapps/docs/documentation/AppleApplications/Reference/SafariJSRef/DOMApplicationCache/DOMApplicationCache.html
  * @constructor
  * @implements {EventTarget}
--- a/externs/browser/ie_dom.js
+++ b/externs/browser/ie_dom.js
@@ -834,12 +834,6 @@ Document.prototype.parentWindow;
 Document.prototype.protocol;
 
 /**
- * @type {string}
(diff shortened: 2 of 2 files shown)


Dominator008:
Also, I noticed that some properties in browser/gecko_dom.js and browser/ie_dom.js are now part of HTML5. We should move them to browser/html5.js instead.

MatrixFrog:
We already have this on subclasses HTMLDocument and XMLDocument. Let's remove it from those (if we're sure it is in fact true that all Documents have a readyState).

Dominator008:
The one on HTMLDOMDocument seems fine to remove, but the one on XMLDOMDocument is of type "number" :(

https://msdn.microsoft.com/en-us/library/ms753702(v=vs.85).aspx

MatrixFrog:
And the compiler doesn't complain that you're overriding a string field in Document to be a number field in XMLDOMDocument? I guess this is okay then but it's a bit strange.

Dominator008:
Not sure what our rules about overriding fields are, but I guess it's ok.

Dominator008:
Hmm.. I do get a warning

externs.zip//browser/ie_dom.js:79: WARNING - Invalid redeclaration of property readyState.
inherited type  : string
overriding type : number

XMLDOMDocument.prototype.readyState;

blickly:
Ok, looks like we should rollback for now.

Dominator008:
Yeah not sure what the best fix would be.

bicknellr:
Hi, I'm working on a project that uses `Document.prototype.readyState` in a few places and have warnings as a result. Why did [this non-standard extension](https://msdn.microsoft.com/en-us/library/ms756987(v=vs.85).aspx) take precedence here and prevent the standard `Document` from having warning-free use of `readyState`? If the type system doesn't support extensions un-subtyping themselves, maybe `XMLDOMDocument.prototype.readyState` should be removed from the built-in externs so that `Document` can properly declare it? If there's any way to suppress the "Invalid redeclaration of property" warning, then we can have our cake and eat it too.

PR #1170, opened 2015-09-25
Adding externs support for the Angular Material `md.$mdUtil` service.

dlochrie:
See the file this creates externs for [here](https://github.com/angular/material/blob/v0.11.0/src/core/util/util.js).

Some of the method signatures took some detective work to figure out the argument types, so feel free to point out any issues there.


contrib/externs/angular-material.js
```diff
@@ -364,3 +364,41 @@ md.$mdIconProvider.prototype.defaultIconSize = function(iconSize) {};
...
+ *   reconnectScope: function(!angular.Scope),
+ *   getClosest: function getClosest(!Element, string, boolean),
+ *   elementContains: function(!Element, !Element),
+ *   extractElementByName: function(!Element, string),
+ *   initOptionalProperties: function(!angular.Scope, !Object, Object),
+ *   nextTick: function(!Function, boolean),
+ *   processTemplate: function(string)
+ * }}
+ */
+md.$mdUtil;
```
mprobst: Could you write this in the `@constructor/prototype` style used above? That gives much better type checking in JSCompiler.

E.g.
```ts
/** @constructor */
md.$mdUtil = function() {};

/**
 * @param {!Element} el
 * ...
 * @return ...
 */
md.$mdUtil.prototype.clientRect = function(e, ...) {};
```

mprobst:
@jelbourn for Material Design knowledge.

dlochrie:
Yes, can do. Would that go along the lines of
https://github.com/google/closure-compiler/issues/157? If so, maybe it will
help in using this extern in my own code, since it seems like Closure
ignores typedef properties like:

/**
 * A debounce handler for toggling the open/close state of a $mdSidenav.
 * @type {!md.$mdUtil.debounce}
 */
this.debounceHandler;


I literally have to do things like:

var debounce = $mdUtil['debounce'];


To get my code to work with the typedef version of the extern.

mprobst:
@dlochrie: that possibly (I thought that was fixed?), but also that `@typedef` cannot be recursive, and the error messages with it tend to be rather verbose (just lists *all* the object properties...).

jelbourn:
$mdUtil is not part of the ng-material public api and shouldn't be in the externs.

dlochrie:
@mprobst: I agree, I will update the code to match that style (which is lot easier to read, as well).

@jelbourn: Just because $mdUtil is not in the public api, does not mean it should not be in the externs. It is still injectable, which therefore means it has-a/is-a type. 

For example, in the Angular Material docs for `$mdSidenav` (see [here](https://material.angularjs.org/latest/#/demo/material.components.sidenav)), the following code would require an extern for `$mdUtil`, if you were using Google Closure:

```
$scope.toggleLeft = buildToggler('left');
    $scope.toggleRight = buildToggler('right');
    /**
     * Build handler to open/close a SideNav; when animation finishes
     * report completion in console
     */
    function buildToggler(navID) {
      var debounceFn =  $mdUtil.debounce(function(){
            $mdSidenav(navID)
              .toggle()
              .then(function () {
                $log.debug("toggle " + navID + " is done");
              });
          },200);
      return debounceFn;
    }
```

In the above code, when you register a handler (here `buildToggler`), it returns an `$mdUtil.debounce` function - which Closure would need to know about, but does not. The code above, closurized, would look like this:

```
angular
  .module('sidenavDemo1', ['ngMaterial'])
  .controller('AppCtrl', AppCtrl);

/**
 * Some description.
 * @param {!angular.scope-type} $scope The current $scope.
 * @param {!angular.timeout-type} $timeout The Angular $timeout service.
 * @param {!md.$mdSidenav} $mdSidenav The $mdSidenav service.
 * @param {!md.$mdUtil} $mdUtil The $mdUtil factory (requires extern).
 * @param {!angular.timeout-type} $log The Angular $log service.
 */
function AppCtrl($scope, $timeout, $mdSidenav, $mdUtil, $log) {}
```

So, the externs would _need_ to be there for Closure to be aware of `$mdUtil`, and any method/property that belonged to it (such as `debounce`), in which case the extern logic I added should be updated so reflect the style that @mprobst suggested.

jelbourn:
The bug there is that our demo references `$mdUtil`. The end user for ng-material should never be aware that `$mdUtil` exists and would never call any of it's methods. Not being part of the public api means _exactly_ that it shouldn't be in the externs. The fact that it is injectable by client code is an unfortunate side effect of Angular's global dependency injection namespace (the service should also be prefixed with `$$` instead of `$`, indicating that it is private). 

And, most importantly, people shouldn't _want_ to use `$mdUtil`, since it is kind of a messy grab-bag of unrelated, super specific functions.

dlochrie:
Thanks for clarifying. I will close this PR - but I opened a PR in the
Angular Material repo so that it no longer references `$mdUtil`, in this
particular demo at least. PR is here
<https://github.com/angular/material/pull/4888>.

PR #1178, opened 2015-09-29
Check for name collisions in CollapseProperties (take 2)

Dominator008:
Use 1-base unique ids to avoid conflict with the special case of "$0".


blickly:
Good point, LGTM.

PR #1183, opened 2015-10-03
Mark the second argument to document.registerElement as optional

### LINKED ISSUE #1174: Document.prototype.registerElement extern should have an optional second argument
Currently in the extern file ```browser/html5.js``` annotation for ```registerElement``` does not return anything, however the specification says it should return constructor function. 

Unfortunately I am not sure how to annotate that in an extern file, but hopefully someone will know and it will be fixed so I can use named constructors for new custom elements.

ChadKillingsworth:
This fixes #1174.

The second argument for document.registerElement should be optional. Also, we are intentionally returning the unknown type from the method so that it can be properly annotated at the call site.

### DIFF
--- a/externs/browser/html5.js
+++ b/externs/browser/html5.js
@@ -2976,7 +2976,10 @@ Document.prototype.msHidden;
  * @see http://www.w3.org/TR/components-intro/
  * @see http://w3c.github.io/webcomponents/spec/custom/#extensions-to-document-interface-to-register
  * @param {string} type
- * @param {{extends: (string|undefined), prototype: (Object|undefined)}} options
+ * @param {{extends: (string|undefined), prototype: (Object|undefined)}=} options
+ * @return {!Function} a constructor for the new tag. A generic function is the best we
+ *     can do here as it allows the return value to be annotated properly
+ *     at the call site.
  */
 Document.prototype.registerElement;
 


externs/browser/html5.js
```diff
@@ -2976,7 +2976,10 @@ Document.prototype.msHidden;
  * @see http://www.w3.org/TR/components-intro/
  * @see http://w3c.github.io/webcomponents/spec/custom/#extensions-to-document-interface-to-register
  * @param {string} type
- * @param {{extends: (string|undefined), prototype: (Object|undefined)}} options
+ * @param {{extends: (string|undefined), prototype: (Object|undefined)}=} options
+ * @return {?} a constructor for the new tag. An unknown type is the best we
```
dimvar: A slightly better type is !Function, which says that the return type is a function and we don't know anything else about it. Still can be downcast at the call site, and prevents errors such as this:

http://closure-compiler-debugger.appspot.com/#input0%3D%252F**%2520%2540type%2520%257Bnumber%257D%2520*%252F%250Avar%2520n%2520%253D%2520f()%253B%26input1%26conformanceConfig%26externs%3D%252F**%2520%2540return%2520%257B!Function%257D%2520*%252F%250Afunction%2520f()%2520%257B%257D%26refasterjs-template%26includeDefaultExterns%3D1%26CHECK_SYMBOLS%3D1%26CHECK_TYPES%3D1%26MISSING_PROPERTIES%3D1

ChadKillingsworth:
Done. I considered this right after I pushed it, but I was concerned that it might break something internally. However, if this breaks something, it was probably a bug to begin with.

dimvar:
can you squash the commits too?

ChadKillingsworth:
Squashed

PR #1187, opened 2015-10-06
Add externs for Array.prototype.find and Array.prototype.findIndex

fredj:


externs/es6.js
```diff
@@ -1093,6 +1093,26 @@ Array.prototype.keys;
 Array.prototype.entries;
 
 
+/**
+ * @param {?function(this:S, T, number, !Array<T>): boolean} predicate
```
MatrixFrog: Can we make this non-null? It doesn't look like there's a use case for passing null here.

fredj: yes, fixed.

PR #1188, opened 2015-10-06
Mark function parameter as non nullable

fredj:
The function parameter of:
 * `Array.prototype.reduce`
 * `Array.prototype.reduceRight`
 * `Array.prototype.every`
 * `Array.prototype.filter`
 * `Array.prototype.forEach`
 * `Array.prototype.map`
 * `Array.prototype.some`

Must be not null

### DIFF
--- a/externs/es3.js
+++ b/externs/es3.js
@@ -596,7 +596,7 @@ Array.prototype.unshift = function(var_args) {};
  * Apply a function simultaneously against two values of the array (from
  * left-to-right) as to reduce it to a single value.
  *
- * @param {?function(?, T, number, !Array<T>) : R} callback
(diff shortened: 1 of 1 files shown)


fredj:
This will certainly break closure-library

concavelenz:
I'm pretty sure, this is "?function..." because of the breakage "!function..." caused.

MatrixFrog:
For most of these, I'm pretty sure there is no good reason to ever pass null. But fixing all the types in Closure Library (not to mention everywhere else at Google) may not be feasible :(

dimvar:
OK, seems like the google-internal cleanup isn't high priority for us to accept this PR. Closing.

PR #1190, opened 2015-10-07
Add @this type annotation for Array.prototype.{find|findIndex}

fredj:


externs/es6.js
```diff
@@ -1097,6 +1097,7 @@ Array.prototype.entries;
  * @param {!function(this:S, T, number, !Array<T>): boolean} predicate
  * @param {S=} opt_this
  * @return {T|undefined}
+ * @this {{length: number}|Array<T>|string}
```
blickly: Consider changing {length: number} to IArrayLike<T>

fredj: Changed to `IArrayLike<T>`

blickly:
Thanks!

Issue #1150, opened 2015-09-17
Undefined in typedef causes type mismatches to be silently ignored


supersteves:
Externs:
``` javascript
/** @typedef {{  sausages: (?string)  }} */
var GOOD;

/** @typedef {{  sausages: (string|null|undefined) }} */
var BAD;
```
Code:
``` javascript
/** @param {GOOD} obj */
var testGood = function(obj) {};

/** @param {BAD} obj */
var testBad = function(obj) {};

testGood({bacon: "mmm"}) // correctly fails 
testBad({bacon: "yuck"}) // should fail, doesn't
```

It appears changing from ```?string``` to ```string|null|undefined``` causes the entire typedef to be disregarded silently.

https://closure-compiler-debugger.appspot.com/#input0%3D%252F**%2520%2540param%2520%257BGOOD%257D%2520obj%2520*%252F%250Avar%2520testGood%2520%253D%2520function(obj)%2520%257B%257D%253B%250A%250A%252F**%2520%2540param%2520%257BBAD%257D%2520obj%2520*%252F%250Avar%2520testBad%2520%253D%2520function(obj)%2520%257B%257D%253B%250A%250AtestGood(%257Bbacon%253A%2520%2522mmm%2522%257D)%2520%252F%252F%2520correctly%2520fails%2520%250AtestBad(%257Bbacon%253A%2520%2522yuck%2522%257D)%2520%252F%252F%2520should%2520fail%252C%2520doesn't%26input1%26conformanceConfig%26externs%3D%252F**%2520%2540typedef%2520%257B%257B%2520%2520sausages%253A%2520(%253Fstring)%2520%2520%257D%257D%2520*%252F%250Avar%2520GOOD%253B%250A%250A%252F**%2520%2540typedef%2520%257B%257B%2520%2520sausages%253A%2520(string%257Cnull%257Cundefined)%2520%257D%257D%2520*%252F%250Avar%2520BAD%253B%26refasterjs-template%26CHECK_SYMBOLS%3D1%26CHECK_TYPES%3D1%26MISSING_PROPERTIES%3D1

Background:
I'm trying to make it easy for a developer to pass a clean and simple JS object literal into a method, and have Closure Compiler verify the object literal matches all expected properties, e.g.:
``` javascript
someFunc({
  propA: 5,
  propB: ["a", "b"]
  // propC optional and missing
});
```
[...]

Issue #152, opened 2014-04-29
More nuanced version of reportUnknownTypes


blickly:
This issue was imported from Closure Compiler's previous home at http://closure-compiler.googlecode.com

The link to the original issue is:
https://blickly.github.io/closure-compiler-issues/#1262

PR #1191, opened 2015-10-07
Better error message for REPORT_UNKNOWN_TYPES

supersteves:
Including the formatted source for the problematic expression, so it can
be identified when there are multiple expressions on the given line.

### DIFF
--- a/src/com/google/javascript/jscomp/TypeCheck.java
+++ b/src/com/google/javascript/jscomp/TypeCheck.java
@@ -205,7 +205,7 @@ public final class TypeCheck implements NodeTraversal.Callback, CompilerPass {
 
   static final DiagnosticType UNKNOWN_EXPR_TYPE =
       DiagnosticType.disabled("JSC_UNKNOWN_EXPR_TYPE",
-          "could not determine the type of this expression");
+          "could not determine the type of expression: {0}");
 
   static final DiagnosticType UNRESOLVED_TYPE =
       DiagnosticType.warning("JSC_UNRESOLVED_TYPE",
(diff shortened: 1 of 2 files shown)


MatrixFrog:
I'd rather not add more code to support reportUnknownTypes, which we don't really recommend anyway. (There is a conformance rule which serves largely the same purpose but works better.)

supersteves:
@MatrixFrog That's a shame, I've been using it to reasonable success, using a custom path based warnings guard to enable it for certain higher quality subtrees of my codebase.

This is a fairly trivial change - just an extra param in the error message, which helps one to understand which expression has an unknown type. Will you reconsider?

I want to know which types are unknown, so I can figure out if there is a problem with my type annotations or if my assumptions about inference are wrong, and so I can maintain a quality standard of 100% typed (except for those with explicit @suppress).

As for the conformance rule I've figured out how to construct a rule using the java APIs (which I use) - and am trying to find the closest equivalent. Presumably BanUnresolvedType is something else, and I actually need BanUnknownTypedClassPropsReferences etc.? Do these give me what I want?

MatrixFrog:
Yup, I think those are the ones. @concavelenz may have tips on how to use them.

ChadKillingsworth:
@supersteves reportUnknownTypes was introduced long ago to debug the compiler itself and wasn't meant to be a dev tool for code conformance. Our gut reaction is "don't use it" because in many cases it causes people to report type bugs for AST nodes that never actually need type information.

It *can* be useful on occasion, but in general isn't that helpful. I find that enabling and fixing the warnings generated by the disambiguateProperties pass to be of much more value and actually make a real difference in my code base.

supersteves:
@ChadKillingsworth Here's one of several similar cases I found yesterday that highlighted a compiler/default externs bug which meant parts of my code had no type checks:
``` javascript
someArray.find(arg => some_expression)
```
where ```arg```'s type was not inferred (and thus unknown) from ```someArray```'s element type within some_expression (already reported). Other cases were similar Array prototype methods where you typically use a fat arrow function to provide a test or a mapping (although longer function declarations have the same issues).

The above was picked up by reportUnknownTypes, but not by conformance rules. 

The problem is, especially when using ES6 and combined with inline vs long type annotation styles, is that the position and effect of JSDoc type annotations is somewhat unclear and has to be discovered through experimentation. Often a type annotation appears to work but in practice the compiler sees the type as unknown and silently passes potentially bad code.

One needs the security of being able to find out what types are working, in the absence of an "ignored type annotation" warning I've previously discussed and am contemplating developing. I think this goes beyond what conformance rules are intended for.

MatrixFrog:
These are good points but reportUnknownTypes as currently written is too picky to really support. I have an in-progress patch for https://github.com/google/closure-compiler/issues/152 from a long time ago but I probably won't have time to finish it anytime soon, though I'll try to get it exported to Github today.

supersteves:
@MatrixFrog Will look forward to seeing that. Curious what's the gripe about report unknown types, its implementation seems to make sense - if a node has an unknown type, report an error. 

I use Closure Compiler first for type safety, second for optimisation and obfuscation; clearly I want to know which nodes of my code's AST are failing inference and thus not type safe, and possibly needing explicit annotations. My experience over the last couple of years learning and using Closure Compiler is that the hardest part is understanding which type annotations are needed and actually take effect, compounded by the occasional weird compiler bug such as https://github.com/google/closure-compiler/issues/1150, and other cases where you think you're type-safe but upon introducing a deliberate error nothing fails.

I'll probably write my own custom conformance rule outside of the closure compiler codebase to do the same thing as report unknown types - not been getting much traction on PRs recently.

PR #1192, opened 2015-10-07
Suppress missing require from externs

supersteves:
This is to fix the case where you have React defined globally in externs:
``` javascript
// Externs:

var React = {};

/**
 * @constructor
 * @param {Object} props
 */

// Main code:

class X extends React.Component {
}
```
Without this PR you get the following warning, which is incorrect since externs don't need to be required as they are global:
```
'React.Component' used but not goog.require'd
```

If this is accepted, PR https://github.com/google/closure-compiler/pull/1147 for suppressing missing requires on class nodes can be deleted.

### DIFF
--- a/src/com/google/javascript/jscomp/CheckRequiresForConstructors.java
+++ b/src/com/google/javascript/jscomp/CheckRequiresForConstructors.java
@@ -52,14 +52,6 @@ class CheckRequiresForConstructors implements HotSwapCompilerPass, NodeTraversal
   private final AbstractCompiler compiler;
   private final CodingConvention codingConvention;
 
-  public static enum Mode {
-    // Looking at a single file. Externs are not present.
-    SINGLE_FILE,
-    // Used during a normal compilation. The entire program + externs are available.
-    FULL_COMPILE
-  };
-  private final Mode mode;
-
   private final Set<String> constructors = new HashSet<>();
   private final Map<String, Node> requires = new HashMap<>();
 
@@ -70,9 +62,6 @@ class CheckRequiresForConstructors implements HotSwapCompilerPass, NodeTraversal
   // require a goog.require, such as in a @type annotation. If the only usages of a name are
   // in weakUsages, don't give a missingRequire warning, nor an extraRequire warning.
   private final Map<String, Node> weakUsages = new HashMap<>();
-
-  // Whether the current file is an ES6 module.
-  private boolean isModule = false;
 
   // Warnings
   static final DiagnosticType MISSING_REQUIRE_WARNING =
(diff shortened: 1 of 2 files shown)


src/com/google/javascript/jscomp/CheckRequiresForConstructors.java
```diff
@@ -182,7 +183,8 @@ private void visitScriptNode(NodeTraversal t) {
             (requires == null || (!requires.containsKey(className)
                                   && !requires.containsKey(outermostClassName)
                                   && !requires.containsKey(parentNamespace)));
-        if (notProvidedByConstructors && notProvidedByRequires
+        boolean notProvidedByExterns = scopeVar==null || !scopeVar.isExtern();
```
blickly: nit: spaces around the ==

test/com/google/javascript/jscomp/MissingRequireTest.java
```diff
@@ -19,6 +19,8 @@
 import static com.google.common.truth.Truth.assertThat;
 import static com.google.javascript.jscomp.CheckRequiresForConstructors.MISSING_REQUIRE_WARNING;
 
+import java.util.List;
```
blickly: nit: usually we put the java language imports last

test/com/google/javascript/jscomp/MissingRequireTest.java
```diff
@@ -139,7 +141,31 @@ public void testFailEs6ClassExtends() {
...
+    List<SourceFile> externs = ImmutableList.of(SourceFile.fromCode("externs", 
+        "/** @constructor */ var SomethingInExterns;")); 
+    test(externs, js, js, null, null, null);
+  }
+  
+  public void testEs6ClassExtendsSomethingInExternsWithNS() {
+    setAcceptedLanguage(LanguageMode.ECMASCRIPT6);
+    String js = "var goog = {}; class SubClass extends MyExterns.SomethingInExterns {}";
+    List<SourceFile> externs = ImmutableList.of(SourceFile.fromCode("externs", 
+        "var MyExterns;\n" +
```
blickly: nit: we try to put binary operators at the beginning of lines

MatrixFrog:
I thought the existing var.isExtern() check on line 406 was supposed to make this work. We should figure out why it's not.

blickly:
Looks live visitNewNode doesn't call maybeAddUsage

supersteves:
I'm afraid I broke this PR when I rebased and resolved conflicts with a formatting change. I've committed a fix.

supersteves:
Closing, reworked here:
https://github.com/google/closure-compiler/pull/1366

PR #1193, opened 2015-10-07
Platform settings for line endings and encoding

supersteves:
So I develop with Eclipse Windows, and my last PR initially changed 100s of lines due to a line ending change. I had to undo the commit and fix the line endings. Tedious.

Not sure how this came about, it could be Eclipse or Git. I suspect Git.

Here are Eclipse settings to enforce the existing project style (UTF-8 encoding and UNIX line endings).

Also here is a .gitattributes file to fix (for now, just) .java files to use UNIX line endings, and thereby avoid any undesired conversion on checkout for Windows developers.
See https://help.github.com/articles/dealing-with-line-endings/
This change didn't affect Git's handling of any existing project files.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Always use UNIX line endings, even on Windows
+*.java eol=lf
--- a/.settings/org.eclipse.core.resources.prefs
+++ b/.settings/org.eclipse.core.resources.prefs
@@ -1,0 +1,2 @@
+eclipse.preferences.version=1
+encoding/<project>=UTF-8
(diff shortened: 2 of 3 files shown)


Dominator008:
FYI, try "git config --global core.autocrlf true".

supersteves:
@Dominator008 thanks but I prefer to work in UNIX line endings if that's the project standard. Don't like the sound of git trying to automatically convert line endings for me...

ChadKillingsworth:
@supersteves Git autoconverting line endings is a pretty standard setup. It works well across platforms - I develop on both windows and unix and have absolutely no problems with it.

Standard practice is to configure git globally to commit unix line endings and checkout platform specific. I haven't had a project yet that violated those rules.

supersteves:
Understood. In which case discard this. Still no idea how my workflow of refactoring two small methods between existing files in the CC sources resulted in all 600+ line endings converted to windows (and showing up in github as changed after committing and pushing).

PR #1199, opened 2015-10-08
Adding missing 'conformanceViolations' to @suppress values list

supersteves:
So I can use ```@suppress {conformanceViolations}``` within my code (which works)

### DIFF
--- a/src/com/google/javascript/jscomp/parsing/ParserConfig.properties
+++ b/src/com/google/javascript/jscomp/parsing/ParserConfig.properties
@@ -140,6 +140,7 @@ jsdoc.suppressions =\
     closureDepMethodUsageChecks,\
     const,\
     constantProperty,\
+    conformanceViolations,\
     deprecated,\
     duplicate,\
     es5Strict,\


MatrixFrog:
I think it's preferable to use the `whitelist` or `only_apply_to` fields in the conformance proto: https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/conformance.proto

In fact, this is one of the advantages of the conformance framework: Since it's controlled by a text proto instead of just flags, we have more control over the details of how it's used.

supersteves:
Yes, but I haven't a clue how to set up conformance proto's, and can't find any docs. In-file annotations work for me, it's what JS devs are used to with jslint / eslint etc.

Also:
https://github.com/google/closure-compiler/blob/3b716944657b764197824e6c898e2a31d9e7622c/src/com/google/javascript/jscomp/DiagnosticGroups.java#L91

MatrixFrog:
This is probably a bit out of date but should be enough to get you started https://docs.google.com/document/d/13Zx-p-GmgPIEig26dFV4Iy9u6wTOLem8o7ScaqT66lA/view#heading=h.dxfl76vthxkv

concavelenz:
This is deliberately not suppressible, suppression is done through the conformance rule.

supersteves:
Please. I like "@suppress", it helps me, and does it actually do any harm, this one line change? The public docs guide you down the @suppress route and conformance docs are more sparse.

MatrixFrog:
Sorry, I don't think we're going to support two different ways of suppressing the same warning. Every time we give users two different ways to do something, it ends up leading to more confusion for them and for us. We *should* improve the docs though. If you'd like to help out with that I'm sure we'd all be very grateful.

supersteves:
As I understand it conformance rules are defined by textproto files, and can have whitelists and regex which act like file-wide @suppress for those rules. (I haven't yet figured out what the naming conventions are for these or whether any and all textproto files are observed. I'm using the compiler from Java.)

When you have code that breaks the rule but cannot be easily fixed - such as for certain unknown types - perhaps there is an outstanding bug in the compiler - you need to suppress the error to get a passing compilation. I think you are suggesting I do this by creating different conformance rules.

1. I don't like the separation between the rule definition and the problematic code. To opt-out you need to mark the file up in a different file. Personally I prefer to keep things together; it's also less cruft.

2.  What if you want to suppress only a function? I had this only yesterday - one of the unknown type conformance rules was failing due to a compiler bug or limitation, can't recall exactly, possibly it was because template types are currently unbounded. But the rest of the source file was fine. Is there a way of suppressing a conformance rule for only part of a file? (I'm aiming for 100% typed, since the primary reason for using Closure Compiler is to type check my code).

So far I'm not using conformance rules the way you probably intended. I'm configuring them via a Java API and want a project-wide ruleset without needing localised textproto files. I can code my own equivalent of @suppress using warnings guards, or use a permanent fork, but thought this one-line change seemingly consistent with other suppress options would be useful to others who might be getting started, trying to figure it out, and would appreciate a bit more flexibility regarding usage patterns.

(last death rattle of this PR!)

PR #1206, opened 2015-10-13
Move w3c_pointer_events.js from contrib/externs to externs

fredj:
The spec is an now an official W3C spec and is available in IE11+, and soon to be implemented in firefox (https://bugzilla.mozilla.org/show_bug.cgi?id=822898) and chromium (https://code.google.com/p/chromium/issues/detail?id=471824)


dimvar:
File-renaming changes are easier to do internally, otherwise we get build failures, and errors when trying to push to github. Closing this and we'll consider the rename internally.

fredj:
done in 6de2ce61dc113fa5420e3f81aa3ddcda784cf771

Issue #1216, opened 2015-10-15
[ES6][Parser] getter/setter varaible declared with type cause parse error


rikuayanokozy:
```
class Foo {
  get bar() : number {
    return 1;
  }
  set bar(a:number) {}
}
```
compile with
```
$ java -jar /Users/riku/workspace/closure-compiler/build/compiler.jar --language_in ES6_TYPED --language_out ES5 --js tests/es6class.js --js_output_file tests/es6class_out.js
tests/es6class.js:2: ERROR - Parse error. '{' expected
  get bar() : number {
            ^

1 error(s), 0 warning(s)
```

Dominator008:
Looks like the parse error has been fixed.

Issue #1158, opened 2015-09-21
Transpiled ES6 getters / setters take a lot of space (& should not be enumerable)


ochafik:
```javascript
class A {
  get x() { return 1 }
}
console.log(new A().x);
```
Is currently [compiled](http://closure-compiler.appspot.com/home#code%3D%252F%252F%2520%253D%253DClosureCompiler%253D%253D%250A%252F%252F%2520%2540compilation_level%2520ADVANCED_OPTIMIZATIONS%250A%252F%252F%2520%2540output_file_name%2520default.js%250A%252F%252F%2520%2540formatting%2520pretty_print%250A%252F%252F%2520%2540language%2520ES6%250A%252F%252F%2520%253D%253D%252FClosureCompiler%253D%253D%250A%250Aclass%2520A%2520%257B%250A%2520%2520get%2520x()%2520%257B%2520return%25201%2520%257D%250A%257D%250Aconsole.log(new%2520A().x)%253B) to:
```javascript
function a() {
}
Object.defineProperties(a.prototype, {x:{configurable:!0, enumerable:!0, get() {
  return 1;
}}});
console.log((new a).x);
```
(note: ideally, that specific getter should just be inlined, see #1157; discussing here what should happen to non-inlineable getters / setters)

According to the spec ([ClassDefinitionEvaluation](http://www.ecma-international.org/ecma-262/6.0/#sec-runtime-semantics-classdefinitionevaluation)): `PropertyDefinitionEvaluation for m with arguments proto and false` (i.e. `enumerable = false` in [PropertyDefinitionEvaluation](http://www.ecma-international.org/ecma-262/6.0/#sec-method-definitions-runtime-semantics-propertydefinitionevaluation)).

Eliding the `enumerable:!0` for getters/setters would make the generated code shorter (note that respecting this for methods as well would produce larger code than the current `A.prototype.m = body` output and would make ES6 classes behave differently from old-style Closure classes).

Regarding `configurable`, it should indeed be `true` for class members but I'm not sure how much this conflicts with the general idea of [@struct](https://github.com/google/closure-compiler/wiki/@struct-and-@dict-Annotations), any thoughts?

Note that respect of `enumerable` & `configurable` semantics + smaller output could be achieved with something like the following (with its own drawbacks, like complexity / size / speed of the extend helper):
```javascript
$jscomp.extend(a.prototype, {
  get x() { .... }
})
```

rikuayanokozy:
[...]

PR #1217, opened 2015-10-16
fix issue #1216

rikuayanokozy:
fix issue [#1216](https://github.com/google/closure-compiler/issues/1216)

### DIFF
--- a/test/com/google/javascript/jscomp/parsing/TypeSyntaxTest.java
+++ b/test/com/google/javascript/jscomp/parsing/TypeSyntaxTest.java
@@ -111,6 +111,16 @@ public final class TypeSyntaxTest extends TestCase {
     parse("var /** string */ foo: string = 'hello';");
   }
 
+  public void testTypedGetterSetterDeclaration() {
+    Node n = parse("var x = {get a(): number {\n}};", LanguageMode.ECMASCRIPT6_TYPED);
+    assertDeclaredType("number type", numberType(),
+        n.getFirstChild().getFirstChild().getFirstChild().getFirstChild().getFirstChild());
+    n = parse("var x = {set a(v: number) {\n}};", LanguageMode.ECMASCRIPT6_TYPED);
+    assertDeclaredType("number type", numberType(),
+        n.getFirstChild().getFirstChild().getFirstChild().getFirstChild().getFirstChild().getChildAtIndex(1)
+            .getFirstChild());
+  }
+
   public void testFunctionParamDeclaration() {
     Node fn = parse("function foo(x: string) {\n}").getFirstChild();
     Node param = fn.getFirstChild().getNext().getFirstChild();
--- a/test/com/google/javascript/jscomp/parsing/TypeSyntaxTest.java
+++ b/test/com/google/javascript/jscomp/parsing/TypeSyntaxTest.java
@@ -121,6 +121,11 @@ public final class TypeSyntaxTest extends TestCase {
             .getFirstChild());
   }
 
+  public void testSetterDeclarationWithReturnType() {
+    expectErrors("Parse error. setter should not have any returns");
+    parse("var x = {set a(x): number {\n}};", LanguageMode.ECMASCRIPT6_TYPED);
+  }
+
   public void testFunctionParamDeclaration() {
     Node fn = parse("function foo(x: string) {\n}").getFirstChild();
     Node param = fn.getFirstChild().getNext().getFirstChild();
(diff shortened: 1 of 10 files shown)


test/com/google/javascript/jscomp/parsing/TypeSyntaxTest.java
```diff
@@ -111,6 +111,16 @@ public void testTypeInDocAndSyntax() {
     parse("var /** string */ foo: string = 'hello';");
   }
 
+  public void testTypedGetterSetterDeclaration() {
+    Node n = parse("var x = {get a(): number {\n}};", LanguageMode.ECMASCRIPT6_TYPED);
+    assertDeclaredType("number type", numberType(),
+        n.getFirstChild().getFirstChild().getFirstChild().getFirstChild().getFirstChild());
+    n = parse("var x = {set a(v: number) {\n}};", LanguageMode.ECMASCRIPT6_TYPED);
+    assertDeclaredType("number type", numberType(),
+        n.getFirstChild().getFirstChild().getFirstChild().getFirstChild().getFirstChild().getChildAtIndex(1)
```
blickly: Can we add one more test that we get a parse error for setters with a return?
e.g.
  var x = {set a(x): number {}}

Also, if it does work, consider adding a test to NewTypeInferenceES6TypedTest.java, such as the following:
    typeCheck(
        "var x = {get a(): number { return 'str'; }}",
        NewTypeInference.RETURN_NONDECLARED_TYPE);

(I consider this a more straightforward way of checking that the types are correctly parsed than traversing the AST)

mprobst: nit: I think there were some test utilities that allowed you to search a tree for a node of a particular type - that might be more readable than this tree navigation style.

mprobst:
Looks very good to me.

alexeagle:
Hmm, that's likely. We need a round-trip test in `test/com/google/javascript/jscomp/Es6TypedIntegrationTest.java`

I added
```
  public void testClassGetterSetter() throws Exception {
    test(createCompilerOptions(),
        "class C {" +
                " get x(): number {return 1;}" +
                " set x(n: number): void {} }", "");
  }
```
but since it fails to parse, I'm not sure if we will re-emit correctly.

rikuayanokozy:
@blickly 
I added a check to parse setter and it will throw an error when setters have returns declaration.
Two test case works now.

mprobst:
@rikuayanokozy could you add @alexeagle's suggested test?

rikuayanokozy:
I added Object.defineProperties to IntegrationTestCase and the test can be passed.
But output JSDoc is a bit strange:
```
Object.defineProperties(function(){}.prototype,{x:{configurable:!0,enumerable:!0,/**@return {number} @this {C} */get:/**@return {number} */function(){return 1},/**@this {C} */set:function(){}}})
```

mprobst:
That looks as if it tries to down-compile the getters and setters to `defineProperty`, which is correct as far as your code is concerned.

But then does not strip the comments, which it really should. @blickly any idea?

rikuayanokozy:
https://github.com/google/closure-compiler/issues/1158#issuecomment-148512603
I rewrote the getter/setter down-compile in Es6ToEs3Converter to provide better code size, but it still have bug on JSDoc generate.
Maybe we can make change based on this?

blickly:
I thought the non-transpiled parse-and-print test cases were in CodePrinterEs6TypedTest.java

Since this fix is a parser fix, it seems to me like checking the transpiled output (like in IntegrationTest) is out of scope for this PR.

rikuayanokozy:
Agree.
Should we merge this PR then open a new issue for that problem?

I thought the non-transpiled parse-and-print test cases were in
CodePrinterEs6TypedTest.java

Since this fix is a parser fix, it seems to me like checking the transpiled
output (like in IntegrationTest) is out of scope for this PR.

—
Reply to this email directly or view it on GitHub
<https://github.com/google/closure-compiler/pull/1217#issuecomment-153525396>
.

blickly:
Sorry for the conflicting advice, but I'd actually prefer to remove the test case in Es6TypedIntegrationTest.java, since it seems out of scope for this change.

I think that the tests in TypeSyntaxTest.java are already checking parse-and-printing anyway.

rikuayanokozy:
Test case testClassGetterSetter has been removed from Es6TypedIntegrationTest.java.

blickly:
Thanks!

PR #1221, opened 2015-10-20
Define history and location as global vars

ChadKillingsworth:
They will be auto declared on Window.

### DIFF
--- a/externs/browser/gecko_dom.js
+++ b/externs/browser/gecko_dom.js
@@ -106,7 +106,7 @@ Window.prototype.globalStorage;
  * @type {!History}
  * @see https://developer.mozilla.org/en/DOM/window.history
  */
-Window.prototype.history;
+var history;
 
 /**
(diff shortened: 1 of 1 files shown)


dimvar:
Sorry, ended up reverting this because it breaks projects internally. See:
https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/DeclaredGlobalExternsOnWindow.java#L116

ChadKillingsworth:
If I get rid of the exception in DeclaredGlobalExternsOnWindow, will it pass internally? If not, I'll just put a comment on the `Window.prototype.location` definition stating why we can't do this.

dimvar:
> If I get rid of the exception in DeclaredGlobalExternsOnWindow, will it pass internally?

I don't know, but given that I saw enough breakages and this cleanup is not super high priority, I'd rather not do it for now.

There was also one project using history in their externs.

Feel free to add a comment pointing to the exception in DeclaredGlobalExternsOnWindow.

PR #1223, opened 2015-10-21
Reapply pull request #1221 - but leave the Window.prototype.location reference

ChadKillingsworth:
Leave the Window.prototype.location reference, but add a comment as to why it has to be there.

### DIFF
--- a/externs/browser/gecko_dom.js
+++ b/externs/browser/gecko_dom.js
@@ -118,6 +118,10 @@ Window.prototype.history;
 Window.prototype.length;
 
 /**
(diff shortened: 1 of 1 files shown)


dimvar:
Actually there were some breakages with history as well. Do you mind just doing the extra comment in this PR? I will change history to a var internally and rerun the tests and see if I can land it without much work.
(That will avoid another external rollback if I can't.)

ChadKillingsworth:
Done. As an FYI, it was the history reference that originally prompted this - it's the one I actually need.

PR #1225, opened 2015-10-22
Report early errors for invalid increment/decrement target.

rintaro:
Reports something like:

```javascript
test.js:1: ERROR - Parse error. Invalid postfix increment operand.
12++;
^

test.js:2: ERROR - Parse error. Invalid postfix decrement operand.
(+foo)--;
 ^

test.js:3: ERROR - Parse error. Invalid prefix increment operand.
++this;
  ^

test.js:4: ERROR - Parse error. Invalid prefix decrement operand.
--(a + b);
   ^
```

### DIFF
--- a/src/com/google/javascript/jscomp/parsing/IRFactory.java
+++ b/src/com/google/javascript/jscomp/parsing/IRFactory.java
@@ -1792,6 +1792,8 @@ class IRFactory {
               msg,
               sourceName,
               operand.getLineno(), 0);
+        } else if (type == Token.INC || type == Token.DEC) {
+          return createIncrDecrNode(type, false, operand);
         }
 
         return newNode(type, operand);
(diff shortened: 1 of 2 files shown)


blickly:
Can you add a few unit tests to NewParserTest?

rintaro:
Done. Please tell me if squash needed.

src/com/google/javascript/jscomp/parsing/IRFactory.java
```diff
@@ -1792,6 +1792,16 @@ Node processUnaryExpression(UnaryExpressionTree exprNode) {
               msg,
               sourceName,
               operand.getLineno(), 0);
+        } else if ((type == Token.INC || type == Token.DEC)
```
Dominator008: These two snippets look pretty similar. Maybe we can factor them out into a separate method to avoid duplication?

rintaro: done.

MatrixFrog:
lgtm. @blickly do we still need (or want) squashing to happen?

dimvar:
Yes, it's better to have a clean history in the main branch, with each commit being a logical unit of work.

blickly:
Looks good. Thanks!

PR #1228, opened 2015-10-24
Added missing TypedArray functions

JaredWilliams:
TypedArray.of()
TypedArray.from()
TypedArray.prototype.copyWithin()

### DIFF
--- a/externs/es6.js
+++ b/externs/es6.js
@@ -395,7 +395,7 @@ function Int8Array(length, opt_byteOffset, opt_length) {}
 Int8Array.BYTES_PER_ELEMENT;
 
 /** 
-* @param {Array<number>} source
+* @param {!Array<number>} source
 * @param {function(this:S, number): number=} opt_mapFn
 * @param {S=} opt_this
 * @template S
@@ -466,7 +466,7 @@ function Uint8Array(length, opt_byteOffset, opt_length) {}
 Uint8Array.BYTES_PER_ELEMENT;
 
 /** 
-* @param {Array<number>} source
+* @param {!Array<number>} source
 * @param {function(this:S, number): number=} opt_mapFn
 * @param {S=} opt_this
 * @template S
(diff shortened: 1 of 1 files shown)


JaredWilliams:
I signed it!

JaredWilliams:
I signed it!

JaredWilliams:
I signed it!

externs/es6.js
```diff
@@ -394,6 +394,23 @@ function Int8Array(length, opt_byteOffset, opt_length) {}
 /** @type {number} */
 Int8Array.BYTES_PER_ELEMENT;
 
+/** 
+* @param {Array<number>} source
+* @param {function(this:S, number): number=} opt_mapFn
+* @param {S=} opt_this
+* @template S
+* @return {!Int8Array}
```
ChadKillingsworth: Shouldn't this be `Int8Array<number>`?

concavelenz: No, Int8Array isn't generic, it is always IArrayLike<number>

externs/es6.js
```diff
@@ -394,6 +394,23 @@ function Int8Array(length, opt_byteOffset, opt_length) {}
 /** @type {number} */
 Int8Array.BYTES_PER_ELEMENT;
 
+/** 
+* @param {Array<number>} source
```
MatrixFrog: If there is a good reason to pass null this should be {?Array<number>}. If not then make it {!Array<number>}

JaredWilliams: It does need to change, just sure to what. 

There are two call scenarios, one with a single argument, which would have to be something iterable returning numbers.

Second is with a source and map function, then the source can by anything iterable as the mapFn provides with numbers. 

Eg. Uint8Array.from(['a', 'aa', 'aaa'], v => v.length)

MatrixFrog: I admit I haven't looked at this part of the spec but it sounds like this should really be Iterable<number> not Array<number>? In any case I'm fine with submitting this as Array for now and then we can change it to Iterable (or ArrayLike, or whatever it's supposed to really be) later. But, let's make it non-nullable (add the `!`)

Dominator008:
Seems like this will need a rebase.

Dominator008:
Rebased in #1413.

PR #1246, opened 2015-11-06
Allow ES6 imports from namespaces.

mprobst:
This allows code that uses ES6 imports in a heterogeneous code base
that has ES6 and ES5-style Closure imports to consistently use ES6
import syntax, allowing a smoother migration path.

### DIFF
--- a/src/com/google/javascript/jscomp/ProcessEs6Modules.java
+++ b/src/com/google/javascript/jscomp/ProcessEs6Modules.java
@@ -30,8 +30,8 @@ import com.google.javascript.rhino.Token;
 import java.net.URI;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.LinkedHashMap;
-import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -55,6 +55,12 @@ public final class ProcessEs6Modules extends AbstractPostOrderCallback {
       DiagnosticType.warning(
           "JSC_USELESS_USE_STRICT_DIRECTIVE",
           "'use strict' is unnecessary in goog.module files.");
+
+  static final DiagnosticType NAMESPACE_IMPORT_CANNOT_USE_STAR =
+      DiagnosticType.error(
+          "JSC_NAMESPACE_IMPORT_CANNOT_USE_STAR",
+          "Namespace imports ('goog:some.Namespace') cannot use import * as. "
+              + "Did you mean to import {0} from '{1}';?");
 
   private static final ImmutableSet<String> USE_STRICT_ONLY = ImmutableSet.of("use strict");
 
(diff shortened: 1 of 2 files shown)


mprobst:
CC @concavelenz, this is the `import Foo from 'goog:ns.Foo';` we talked about a while back.

A question then was whether to use a prefix or not. This was actually rather easy to implement with the prefix/URL scheme. I like how `goog:` makes it obvious what you're doing from the call site, without having to have transitive knowledge of all available modules - for both the compiler and the programm. YMMV, happy to change this if needed.

rintaro:
I feel
```javascript
import * as ns from "goog:some.namespace";
use(ns.Foo);
```
is more natural than:
```javascript
import ns from "goog:some.namespace";
use(ns.Foo);
```

In ES6, `import NAME from` is just a syntax sugar of `import {default as NAME} from`.

mprobst:
@rintaro I think that's a very good point about the default imports.

I was thinking from a different angle though. From a user's perspective, a `goog.provide` or `goog.module` in most cases only has one export, the object that ends up in it. For example:

```js
goog.provide('my.funky.Class');

/** @constructor */
my.funky.Class = function() {};
my.funky.Class.prototype.foo = function() {};
// ... and so on.
```

For these cases, it'd be rather annoying always having to `import * as Class from 'goog:my.funky.Class';`. The namespace provided is the export itself, it doesn't usually have multiple exports in it. Most Closure code follows this pattern, so I think it's nicer to make user's life easier here.

ChadKillingsworth:
In cases where a Closure-Library file has multiple provides, we may want to split them up so that a goog.require can roughly become analogous to an ES6 import or a commonjs require statement.

ChadKillingsworth:
Oh and :+1: - this provides a missing piece of the puzzle for mixing module types in a single compilation.

mprobst:
@ChadKillingsworth do you mean if a file provides `ns.Foo` and `ns.Bar`, we should be able to `import {Foo, Bar} from 'goog:ns';`?

That might be a bit more difficult, as we might risk creating incomplete namespace references. I think it might be easier conceptually to just always treat the entire namespace as an opaque token - that's at least easy to understand and implement.

ChadKillingsworth:
@mprobst No I meant that we should only ALLOW one export per file and that any closure-library files that don't match that should be changed.

I'd like to add module location loading natively to the compiler so that you only provide entry points and it finds the dependent files.

mprobst:
@ChadKillingsworth ah I see, that makes sense. Essentially that would be roughly equivalent to migrating all Closure library files to `goog.module`, right? That seems like a good idea.

I'm not 100% convinced on letting the compiler automatically discover the imported files. I'm sure that has advantages for pruning, but it can also lead to very confusing behaviour if your compiler discovers or does not discover some files.

YMMV, both modes of operation are reasonable, I guess.

mprobst:
BTW, this has been submitted in an internal commit and should show up soon. Closing this PR.

Issue #905, opened 2015-04-12
Feature Request: Allow use of use_only_custom_externs and standard input


dhai:
use_only_custom_externs does not get picked up when pushing in files via standard input:

	java -jar compiler.jar --use_only_custom_externs
	console.log("XYZ"); // stdin
	console.log("XYZ"); // stdout

This should result in an error.  Also, it would be nice if we could pass both the javascript to be compiled, and the externs, all in one pass via standard input (a delimiter maybe?).

	// the code to be compiled
	
	/* @hereth-beith-the-google-closure-compiler-externs */
	
	// externs definitions

ChadKillingsworth:
You need to set your warning level or enable the proper checks:
```
echo 'console.log("XYZ");' | java -jar build/compiler.jar --use_only_custom_externs -W=VERBOSE
stdin:1: ERROR - variable console is undeclared
console.log("XYZ");
^
```

The magic annotation you are looking for is `@externs` and it operates at the `@fileoverview` block. See:  https://github.com/google/closure-compiler/blob/master/contrib/externs/jquery-1.9.js#L27

ChadKillingsworth:
Although I just realized, `@externs` won't help when you pass multiple files in a single stream. That's not supported.

I'm guessing you are using a build system that utilizes streams? Like Gulp?

dhai:
I've written my own precompiler that generates it's own externs.

You're right, I totally forgot the verbose flag.  doh!

Thanks for the note about the externs!

dhai:
Just to clarify, since the externs annotation happens at the fileoverview level, I assume there is no way to pass in both the externs and code to be compiled via stdin?

java -jar bin/compiler.jar -- warning_level=VERBOSE --compilation_level=ADVANCED_OPTIMIZATIONS --language_in=ECMASCRIPT6_STRICT --language_out=ECMASCRIPT5_STRICT --use_only_custom_externs < "test.js"

where test.js is:

	console.log("XYZ");
	
	/**
	 * 	@externs
	 */
	
	/**
	 *	@constructor
	 *	@extends {Object.<string,*>}
	 *	@return {void}
	 */
	function Console(){
	};
	
	/**
	 *	@param {...*} args
	 *	@return {void}
	 */
	Console.prototype.log=function(args){};
	
	/**
	 *	@type {Console}
	 *	@const
	 */
	var console;

output is:

	"use strict";

Where I'm assuming the console.log is ignored, because it's being seen as an extern ....

PR #1253, opened 2015-11-16
Support multiplexing files via stdin/out streams

ChadKillingsworth:
As discusses last meeting, this is the initial work to support sending multiple input files to the compiler via stdin. This is needed to support stream based build systems like [Gulp](http://gulpjs.com/).

This currently uses a new boolean flag `--json_streams`. Ideally we'd avoid this, but I haven't come up with another good solution.

When the `--json_streams` flag is true, standard input is expected to be an array of json encoded files of the form:

```JSON
[
  {
    "src": "file source",
    "path": "/absolute/path/of/file.js",
    "source_map": "optional. source of input sourcemap"
  }
]
```

When the compiled source is written to standard output, it will also use the same format.

Currently this only supports single binaries. I'll be working to add modules shortly.

### DIFF
--- a/src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
+++ b/src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
@@ -30,6 +30,9 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Maps;
 import com.google.common.io.ByteStreams;
+import com.google.gson.Gson;
+import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonWriter;
 import com.google.javascript.jscomp.CompilerOptions.TweakProcessing;
 import com.google.javascript.jscomp.deps.ClosureBundler;
 import com.google.javascript.jscomp.deps.SourceCodeEscapers;
@@ -45,6 +48,7 @@ import java.io.FileOutputStream;
 import java.io.Flushable;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.InputStreamReader;
 import java.io.OutputStream;
 import java.io.OutputStreamWriter;
 import java.io.PrintStream;
@@ -172,8 +176,12 @@ public abstract class AbstractCommandLineRunner<A extends Compiler,
 
   private Map<String, String> parsedModuleWrappers = null;
 
+  private final Gson gson;
+
   static final String OUTPUT_MARKER = "%output%";
   private static final String OUTPUT_MARKER_JS_STRING = "%output|jsstring%";
+
+  private final List<JsonFileSpec> filesToStreamOut = new ArrayList<>();
 
   AbstractCommandLineRunner() {
     this(System.out, System.err);
@@ -183,6 +191,7 @@ public abstract class AbstractCommandLineRunner<A extends Compiler,
     this.config = new CommandLineConfig();
     this.defaultJsOutput = Preconditions.checkNotNull(out);
     this.err = Preconditions.checkNotNull(err);
+    this.gson = new Gson();
   }
 
   /**
@@ -359,6 +368,8 @@ public abstract class AbstractCommandLineRunner<A extends Compiler,
 
     if (config.createSourceMap.length() > 0) {
       options.sourceMapOutputPath = config.createSourceMap;
+    } else if (config.useJsonStreams) {
+      options.sourceMapOutputPath = "%outname%";
     }
     options.sourceMapDetailLevel = config.sourceMapDetailLevel;
     options.sourceMapFormat = config.sourceMapFormat;
@@ -555,6 +566,19 @@ public abstract class AbstractCommandLineRunner<A extends Compiler,
     }
   }
 
+  public List<JsonFileSpec> parseJsonFilesFromInputStream() throws IOException {
+    List<JsonFileSpec> jsonFiles = new ArrayList<>();
+    JsonReader reader = new JsonReader(new InputStreamReader(System.in, inputCharset));
+    reader.beginArray();
+    while (reader.hasNext()) {
+      JsonFileSpec jsonFile = gson.fromJson(reader, JsonFileSpec.class);
+      jsonFiles.add(jsonFile);
+    }
+    reader.endArray();
+    reader.close();
+    return jsonFiles;
+  }
+
   /**
    * Creates inputs from a list of files.
    *
@@ -572,6 +596,21 @@ public abstract class AbstractCommandLineRunner<A extends Compiler,
   }
 
   /**
+   * Creates inputs from a list of source files and json files.
+   *
+   * Can be overridden by subclasses who want to pull files from different
+   * places.
+   *
+   * @param files A list of filenames.
+   * @param jsonFiles A list of json encoded files.
+   * @return An array of inputs
+   */
+  protected List<SourceFile> createInputs(List<String> files,
+      List<JsonFileSpec> jsonFiles) throws FlagUsageException, IOException {
+    return createInputs(files, new ArrayList<String>() /* zips */, jsonFiles, false);
+  }
+
+  /**
    * Creates inputs from a list of source files and zips.
    *
    * Can be overridden by subclasses who want to pull files from different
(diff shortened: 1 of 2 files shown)


dbkaplun:
+1 keep it up @ChadKillingsworth!

ChadKillingsworth:
This is now ready for review and is needed for gulp/grunt plugins. See https://github.com/ChadKillingsworth/closure-compiler-npm

I'd like to have this merged before the next release.

Once any review items are addressed, I'll rebase and squash the commits to the top of the tree.

dimvar:
OK. I can take a look tomorrow.

ChadKillingsworth:
I've not tested this on a windows machine as well as in projects that make use of the `--module` flags. Found a couple of bugs, but changes have now been pushed.

Overall it's working very well.

dimvar:
This will have to wait until Monday after all. Got caught up with other things.

ChadKillingsworth:
No worries. I just got the windows box testing completed this afternoon. I'm testing it on projects that pipe all of closure-library in this way and it performs very well.

src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
```diff
@@ -30,6 +30,9 @@
 import com.google.common.collect.Iterables;
```
dimvar: Can you add a unit test here for the new flag?

https://github.com/google/closure-compiler/blob/master/test/com/google/javascript/jscomp/CommandLineRunnerTest.java

src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
```diff
@@ -280,6 +289,7 @@ static DependencyOptions createDependencyOptions(
           .setMoocherDropping(false)
           .setEntryPoints(closureEntryPoints);
     }
+
```
dimvar: why the extra line?

src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
```diff
@@ -579,12 +604,47 @@ public FlagUsageException(String message) {
    *
    * @param files A list of filenames.
    * @param zips A list of zip filenames.
+   * @param jsonFiles A list of json encoded files.
+   * @return An array of inputs
+   */
+  protected List<SourceFile> createInputs(List<String> files,
+      List<String> zips, List<JsonFileSpec> jsonFiles) throws FlagUsageException, IOException {
```
dimvar: Do you need to combine json inputs with zips or files? If not, you can have a createInputs just for the jsons, and it populates inputs directly, it doesn't call into another createInputs method.

src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
```diff
@@ -2292,4 +2478,42 @@ protected void clear() {
...
+  private class JsonFileSpec {
+    private final String src;
+    private final String path;
+    private String source_map;
+
+    public JsonFileSpec(String src) {
+      this(src, null, null);
+    }
+
+    public JsonFileSpec(String src, String path) {
```
dimvar: As far as I can tell you're only using this constructor. Why do you need the other two?

src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
```diff
@@ -2292,4 +2478,42 @@ protected void clear() {
...
+    public String getPath() {
+      return this.path;
+    }
+
+    public String getSourceMap() { return this.source_map; }
+
+    public void setSourceMap(String map) {
+      this.source_map = map;
+    }
+
```
dimvar: remove empty line

src/com/google/javascript/jscomp/CommandLineRunner.java
```diff
@@ -599,6 +603,14 @@
             usage = "A file containing an instrumentation template.")
         private String instrumentationFile = "";
 
+    @Option(name = "--json_streams",
+        hidden = true,
+        handler = BooleanOptionHandler.class,
+        usage = "Specifies that standard input and output streams will be "
+            + "a JSON array of sources. Each source will be an object of the "
+            + "form {path: filename, src: file_contents, srcmap: srcmap_contents }. "
+            + "Intended for use by stream based build systems such as gulpjs.")
```
dimvar: stream based => stream-based

dimvar:
Hey Chad, I left a few comments, but Github's bad workflow for code reviews makes this hard to read. Do you mind squashing the commits? Then I can use a tool to bring in the PR internally and use the nicer internal tool to look at the diffs.

src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
ChadKillingsworth: I definitely combine the json inputs with regular --js flag files. I just did the zip file because I wanted the option to still be there. I'm trying to keep as much parity with the command line runner as possible.

See https://github.com/ChadKillingsworth/closure-compiler-npm/blob/master/test/gulp.js#L206

src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
ChadKillingsworth: We don't have any tests currently for input from standard in. That's why I didn't. I can, but it'll take some time and refactoring of AbstractCommandLineRunner to handle a provided input stream rather than the hardcoded "stdin" stream.

ChadKillingsworth:
Commits squashed. I have a full test suite for this at https://github.com/ChadKillingsworth/closure-compiler-npm/tree/master/test

ChadKillingsworth:
FYI this fixes the last of the issues with #905

src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
```diff
@@ -1130,35 +1267,58 @@ private void outputModuleBinaryAndSourceMaps(
...
+    if (!(shouldGenerateMapPerModule(options) || config.useJsonStreams)) {
+      mapFileOut = fileNameToOutputWriter2(
+          expandSourceMapPath(options, null));
     }
 
     for (JSModule m : modules) {
-      if (shouldGenerateMapPerModule(options)) {
-        mapOut = fileNameToOutputWriter2(expandSourceMapPath(options, m));
-      }
+      if (config.useJsonStreams) {
```
dimvar: here, the result of the IF is the same in all iterations of the loop. So the IF can move outside the loop. And the json-stream stuff can be extracted into a helper method to make this easier to read.

    if (config.useJsonStreams) {
      helper-method;
    } else {
      for (JSModule m : modules) {
        previous-stuff;
      }
    }

src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
```diff
@@ -1105,7 +1206,7 @@ int processResults(Result result, List<JSModule> modules, B options)
     return SourceCodeEscapers.javascriptEscaper().asFunction();
   }
 
-  void outputSingleBinary() throws IOException {
+  void outputSingleBinary(B options) throws IOException {
     Function<String, String> escaper = null;
```
dimvar: I'm a bit confused about this method. Do you want to output to files and to a json stream in the same compilation?

src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
```diff
@@ -971,6 +1040,30 @@ protected int doRun() throws FlagUsageException, IOException {
...
     List<String> moduleSpecs = config.module;
+    List<JsonFileSpec> jsonFiles = null;
+
+    if (config.useJsonStreams) {
+      jsonFiles = parseJsonFilesFromInputStream();
+
+      ImmutableMap.Builder<String, SourceMapInput> inputSourceMaps
+          = new ImmutableMap.Builder<>();
+
+      int jsonInputSourceMapCount = 0;
```
dimvar: You only testing for non-zero of this. You can replace it with a boolean foundJsonInputSourceMap

src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
dimvar: OK, then it's fine to combine json and files. But the zips are currently used only for some rare usecase internally, so for now I think it's safe to assume we don't need to combine zips and json inputs. So you can delete the zips argument.

src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
dimvar: OK, it's fine to skip it then.

dimvar:
OK, I finished going through the whole thing. I'll take another look after you address these comments.

src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
ChadKillingsworth: No - if using json_streams, all JS and Sourcemaps will be streamed out. I refactored this method in a similar fashion to the note above regarding the helper method in module outputing.

ChadKillingsworth:
I think I've address all the review comments.

src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
```diff
@@ -578,6 +602,21 @@ public FlagUsageException(String message) {
    * places.
    *
    * @param files A list of filenames.
+   * @param jsonFiles A list of json encoded files.
```
dimvar: Fix the first line of this comment to not mention zips (github won't let me add this on that line).

src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
```diff
@@ -585,6 +624,25 @@ public FlagUsageException(String message) {
    */
   protected List<SourceFile> createInputs(List<String> files,
       List<String> zips, boolean allowStdIn) throws FlagUsageException, IOException {
+    return createInputs(files, zips, null, allowStdIn);
+  }
+
+  /**
+   * Creates inputs from a list of source files and zips.
```
dimvar: update this to mention json

dimvar:
Looks good. Only a couple minor comments.

dimvar:
I'll wait for Ben to do the daily sync and then I'll merge this.

ChadKillingsworth:
Done!

PR #1258, opened 2015-11-17
Add pom-gwt.xml to use Maven for GWT build

Dominator008:
This makes use of Thomas Broyer's gwt-maven-plugin.

Shortened module name in gwt_demo.html to be "jscomp"

Also added extra super sources for GWT incompatible classes.

### DIFF
--- a/pom-gwt.xml
+++ b/pom-gwt.xml
@@ -1,0 +1,272 @@
+<!--
+ Copyright 2015 Google Inc.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+   xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd"
+  >
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>com.google.javascript</groupId>
+  <artifactId>closure-compiler-gwt</artifactId>
+  <packaging>gwt-app</packaging>
+
+  <name>Closure Compiler GWT App</name>
+  <version>1.0-SNAPSHOT</version>
+
+  <url>https://developers.google.com/closure/compiler/</url>
+  <description>
+    Closure Compiler is a JavaScript optimizing compiler. It parses your
+    JavaScript, analyzes it, removes dead code and rewrites and minimizes
+    what's left. It also checks syntax, variable references, and types, and
+    warns about common JavaScript pitfalls. It is used in many of Google's
+    JavaScript apps, including Gmail, Google Web Search, Google Maps, and
+    Google Docs.
+  </description>
+  <inceptionYear>2009</inceptionYear>
+
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <project.build.outputEncoding>UTF-8</project.build.outputEncoding>
+    <jdk.version>1.7</jdk.version>
+    <!-- for substitution in ParserConfig.properties -->
+    <compiler.date>${maven.build.timestamp}</compiler.date>
+    <maven.build.timestamp.format>yyyy-MM-dd HH\:mm</maven.build.timestamp.format>
+    <compiler.version>${project.version}</compiler.version>
+  </properties>
+
+  <parent>
+    <groupId>com.google.javascript</groupId>
+    <artifactId>closure-compiler-parent</artifactId>
+    <version>1.0-SNAPSHOT</version>
+    <relativePath>pom.xml</relativePath>
+  </parent>
+
+  <dependencies>
+    <dependency>
+      <groupId>com.google.javascript</groupId>
+      <artifactId>closure-compiler-externs</artifactId>
+      <version>1.0-SNAPSHOT</version>
+    </dependency>
+
+    <dependency>
+      <groupId>args4j</groupId>
+      <artifactId>args4j</artifactId>
+      <version>2.0.26</version>
+    </dependency>
+
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+      <version>19.0-rc2</version>
+    </dependency>
+
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava-gwt</artifactId>
+      <version>19.0-SNAPSHOT</version>
+    </dependency>
+
+    <dependency>
+      <groupId>com.google.protobuf</groupId>
+      <artifactId>protobuf-java</artifactId>
+      <version>2.5.0</version>
+    </dependency>
+
+    <dependency>
+      <groupId>com.google.code.gson</groupId>
+      <artifactId>gson</artifactId>
+      <version>2.2.4</version>
+    </dependency>
+
+    <dependency>
+      <groupId>com.google.code.findbugs</groupId>
+      <artifactId>jsr305</artifactId>
+      <version>1.3.9</version>
+    </dependency>
+
+    <dependency>
+      <groupId>com.google.truth</groupId>
+      <artifactId>truth</artifactId>
+      <version>0.24</version>
+      <scope>test</scope>
+    </dependency>
+
+    <!-- Ant is a provided scope as it is only needed to compile; ant will provide itself when using the jar -->
+    <dependency>
+      <groupId>org.apache.ant</groupId>
+      <artifactId>ant</artifactId>
+      <scope>provided</scope>
+      <version>1.9.4</version>
+    </dependency>
+
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>4.11</version>
+      <scope>test</scope>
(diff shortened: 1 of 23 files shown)


Dominator008:
This builds a .war archive that includes the 1.6 MB JavaScript library. I haven't figured out how to include the gwt_demo.html page yet.

blickly:
Great! I can also open-source the GWT debugger.

src/com/google/javascript/jscomp/gwt/super/com/google/debugging/sourcemap/SourceMapConsumerFactory.java
```diff
@@ -0,0 +1,21 @@
+/*
```
MatrixFrog: I thought "super" was a gwt term meaning "Instead of compiling from Java, use this JavaScript code in place of this class". So it seems strange for Java files (even stubs like this) to live under a .../gwt/super/ directory. Is this the normal GWT convention?

MatrixFrog:
Cool! We'll probably want to bring this in and submit it internally so that we can update BUILD rules / MOE configs / etc.

tbroyer:
> I haven't figured out how to include the gwt_demo.html page yet.

Given where the file and your module are, you have to declare an explicit `<public>` path.

pom-gwt.xml
```diff
@@ -0,0 +1,286 @@
...
+      <version>4.11</version>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>com.google.gwt</groupId>
+      <artifactId>gwt</artifactId>
+      <version>2.8.0-SNAPSHOT</version>
+      <type>pom</type>
+      <scope>import</scope>
```
Dominator008: If I put them under <dependencyManagement> I get "could not find or load main class com.google.gwt.dev.Compiler"

Without the versions I get
'dependencies.dependency.version' for com.google.gwt:gwt-dev:jar is missing

pom-gwt.xml
```diff
@@ -0,0 +1,286 @@
...
+            <exclude>**/testing/**</exclude>
+            <exclude>**/webservice/**</exclude>
+          </excludes>
+        </configuration>
+      </plugin>
+
+      <plugin>
+        <groupId>net.ltgt.gwt.maven</groupId>
+        <artifactId>gwt-maven-plugin</artifactId>
+        <version>1.0-SNAPSHOT</version>
```
Dominator008: No, 1.0-rc-4 seems fine

pom-gwt.xml
```diff
@@ -0,0 +1,286 @@
...
+        <extensions>true</extensions>
+        <configuration>
+	  <moduleName>com.google.JsComp</moduleName>
+	  <moduleTemplate>src/com/google/JsComp.gwt.xml</moduleTemplate>
```
Dominator008: Done

pom-gwt.xml
```diff
@@ -0,0 +1,286 @@
...
+	  <compilerArgs>
+	    <arg>-strict</arg>
```
Dominator008: Done

pom-gwt.xml
tbroyer: What I meant is you need to either:

 * put this in dependencyManagement and then you can add them gwt-user and gwt-dev without version

 * or remove it and only keep your gwt-user and gwt-dev with versions.

Dominator008: Picked the second option for now. Thanks.

Dominator008:
Thanks for the help @tbroyer!
@MatrixFrog We don't care about the ant build.xml right?

MatrixFrog:
Well we can't get rid of it but if the GWT version is buildable only with Maven that's fine.

Dominator008:
Cool. I just don't want to add extra GWT .jars to lib/.

pom-gwt.xml
```diff
@@ -0,0 +1,277 @@
...
+        </configuration>
+      </plugin>
+
+      <plugin>
+        <groupId>net.ltgt.gwt.maven</groupId>
+        <artifactId>gwt-maven-plugin</artifactId>
+        <version>1.0-rc-4</version>
+        <extensions>true</extensions>
+        <configuration>
+	  <moduleName>com.google.JsComp</moduleName>
```
tbroyer: You should add

```xml
<moduleShortName>jscomp</moduleShortName>
```

(if you want incremental builds to work – i.e. gwt-maven-plugin detecting whether the compiled GWT app is uptodate or stale and decide whether to recompile it – without `moduleShortName`, the plugin will look for a `com.google.JsComp/com.google.JsComp.nocache.js` instead of `jscomp/jscomp.nocache.js` and will end up unconditionally recompiling the gwt-app, as if you'd set [`gwt.forceCompilation`](https://tbroyer.github.io/gwt-maven-plugin/compile-mojo.html#forceCompilation) to `true`)

src/com/google/JsComp.gwt.xml
```diff
@@ -0,0 +1,12 @@
+<module rename-to="jscomp">
+  <inherits name="com.google.gwt.user.User" />
+  <inherits name="com.google.common.base.Base"/>
+  <inherits name="com.google.common.collect.Collect"/>
+  <source path="debugging/sourcemap" />
+  <source path="javascript/jscomp" excludes=".svn,.git,**/ant/**,**/refactoring/**,**/webservice/**,**/testing/**" />
+  <source path="javascript/rhino" excludes=".svn,.git,**/testing/**" />
+  <super-source path="javascript/jscomp/gwt/super" />
```
tbroyer: I don't know how this is done internally at Google (I suppose it exists) but it looks to me like it could be clearer to put that file in `src/com/google/javascript/jscomp/gwt/` that follows the naming conventions of `client`, `super` and `public`; and add other `*.gwt.xml` files with `<source path="" />` inside `src/com/google/debugging/sourcemap` and `src/com/google/javascript/rhino` (and add the appropriate `<inherits/>` inside `src/com/google/javascript/jscomp/gwt/JsComp.gwt.xml`)

pom-gwt.xml
tbroyer: Oh, BTW, I see in `gwt_demo.html` that the module is actually named `com.google.javascript.jscomp.GwtModule` (and without `moduleShortName` / `rename-to="jscomp"`)

pom-gwt.xml
```diff
@@ -0,0 +1,277 @@
...
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>4.11</version>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>com.google.gwt</groupId>
+      <artifactId>gwt-user</artifactId>
+      <version>2.8.0-SNAPSHOT</version>
```
tbroyer: BTW, any reason you use 2.8.0-SNAPSHOT rather than 2.7.0?

pom-gwt.xml
```diff
@@ -0,0 +1,277 @@
...
+
+    <dependency>
+      <groupId>com.google.gwt</groupId>
+      <artifactId>gwt-dev</artifactId>
+      <version>2.8.0-SNAPSHOT</version>
+    </dependency>
+
+    <dependency>
+      <groupId>com.google.gwt</groupId>
+      <artifactId>gwt-codeserver</artifactId>
```
tbroyer: gwt-codeserver is not needed actually; all the classes it contains are also into gwt-user (since GWT 2.7.0)

tbroyer:
Some small things can be improved, but otherwise SGTM.

pom-gwt.xml
Dominator008: The build needs a few JRE emulations I submitted to 2.8.0-SNAPSHOT. It won't build under 2.7.0.

src/com/google/JsComp.gwt.xml
Dominator008: The internal BUILD has rules to generate .gwt.xml files from dependency info. If possible I would like to get rid of this file and obtain something similar

tbroyer: Maybe better wait for J2Cl than invest time on gwt.xml generation 😉

With multiple modules, it would probably be "cleaner" to split super sources in separate folders, but not necessary.

pom-gwt.xml
Dominator008: Done

src/com/google/JsComp.gwt.xml
Dominator008: I was asking because the submodules couldn't seem to find their super-sources.
I guess I can't use a non-canonical path like
`<super-source path="../jscomp/gwt/super" />`?

concavelenz:
Tyler, how did you want to handle merging this?

MatrixFrog:
I tried to do it yesterday but had some MOE issues so I'll try again soon.

MatrixFrog:
Hoping to have this merged today. There is some internal stuff that has to happen first but it seems to be actually working now.

Dominator008:
Great, thanks Tyler!

pom-gwt.xml
```diff
@@ -0,0 +1,272 @@
...
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+      <version>19.0-SNAPSHOT</version>
+    </dependency>
+
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava-gwt</artifactId>
+      <version>19.0-SNAPSHOT</version>
```
MatrixFrog: Do you know if it will still work if we use 19.0-rc2 here?

Dominator008: Yes, it still works.

MatrixFrog: Bad news and good news: It looks like rc2 does not work :crying_cat_face: but depending on snapshots is actually totally fine :grinning: 

Sorry again for all the delays on this.

MatrixFrog: (The reason being that there was a particular commit that was reverted on the 19.0 release branch but not in the snapshot builds.)

Dominator008: Good to know :D

MatrixFrog:
Landed as 4a80baef7f4c9361c6b2d33a0dfeae7f0dba5fa8

PR #1266, opened 2015-11-23
Remove the ability to generate a single sourcemap file for multiple output modules

ChadKillingsworth:
If `--module` flags are specified along with a `--create_source_map` flag, the value of the `--create_source_map` determines whether the compiler will generate a sourcemap for each output file or a single sourcemap file.

After chatting with @concavelenz this morning, we determined that a single sourcemap is not valid and this is most likely is not what was intended by the user. This PR removes the ability to generate a single sourcemap for multiple modules and reports an error to the user in this case. 

As further support, every unit test for CommandLineRunner which tested sourcemaps for modules generated one sourcemap per module.

In addition, the `file` property of a sourcemap, while optional, is intended to point back to the file which loaded it. Currently we are only using the module name - not the full module filename.

So for `java -jar compiler.jar --js input.js --module base:1 --create_source_map="%outname%.map"` the compiler would generate:

```Javascript
   {
      "file": "base" // this should be base.js at a minimum - and ideally would include the full module path
     /// ...
```

This PR changes the file reference to be the full module output path.

The module flags are not widely used, but it is still possible that someone is depending on this functionality - broken as it is.

[Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1266)

### DIFF
--- a/src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
+++ b/src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
@@ -116,6 +116,10 @@ public abstract class AbstractCommandLineRunner<A extends Compiler,
   static final DiagnosticType NO_TREE_GENERATED_ERROR = DiagnosticType.error(
       "JSC_NO_TREE_GENERATED_ERROR",
       "Code contains errors. No tree was generated.");
+  static final DiagnosticType INVALID_MODULE_SOURCEMAP_PATTERN = DiagnosticType.error(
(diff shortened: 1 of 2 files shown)


ChadKillingsworth:
I've rebased this to the top of the tree and resolved the conflicts. This is ready to review.

dimvar:
@ChadKillingsworth I'm going thru the list of PRs to see what can be merged. Is this PR still relevant?

ChadKillingsworth:
@dimvar Yes - just nobody has reviewed it. I've rebased to the top of the tree to fix the merge conflicts.

PR #1267, opened 2015-11-24
Fix #1262: Add correct implementation of NumberToString

### LINKED ISSUE #1262: Wrong constant-folding of the number to string conversion
The constant-folder, or some other parts of GCC, wrongly handles the conversion from number to string. This happens both with Simple Opts and Advanced Opts. The following snippet reproduces the problem:
```javascript
function foldNumberToString() {
  alert('A number: ' + 1.2323919403474454e+21.toString());
  alert('A number: ' + 1.2323919403474454e+21);
}
foldNumberToString();
```
The first call is not constant-folded, and outputs the correct results, i.e., `1.2323919403474454e+21`. The second line, however, is constant-folded into
```javascript
  alert("A number: 1.2323919403474454E21");
```
which seems to indicate that a `.toString()` was used on the JVM to convert the number into a string.

ECMAScript [specifies pretty precisely](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-tostring-applied-to-the-number-type) how a number must be converted into a string, and the algorithm does *not* match what the JVM does.

paullepoulpe:
[Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1267)

### DIFF
--- a/test/com/google/javascript/rhino/dtoa/NumberToStringTest.java
+++ b/test/com/google/javascript/rhino/dtoa/NumberToStringTest.java
@@ -1,0 +1,102 @@
+/*
+ *
+ * ***** BEGIN LICENSE BLOCK *****
+ * Version: MPL 1.1/GPL 2.0
+ *
+ * The contents of this file are subject to the Mozilla Public License Version
+ * 1.1 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ * http://www.mozilla.org/MPL/
+ *
+ * Software distributed under the License is distributed on an "AS IS" basis,
+ * WITHOUT WARRANTY OF ANY KIND, either express or implied. See the License
+ * for the specific language governing rights and limitations under the
+ * License.
+ *
+ * The Original Code is Rhino code, released
+ * May 6, 1999.
+ *
+ * The Initial Developer of the Original Code is
+ * Netscape Communications Corporation.
+ * Portions created by the Initial Developer are Copyright (C) 1997-1999
+ * the Initial Developer. All Rights Reserved.
+ *
+ * Contributor(s):
+ *   Damien Engels
+ *
+ * Alternatively, the contents of this file may be used under the terms of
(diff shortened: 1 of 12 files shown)


Dominator008:
@MatrixFrog Will it be easyto merge this internally?

concavelenz:
Ben, what is happening with this?

blickly:
Sorry to spend so long on this.
I think that this is good in terms of functionality. The only issue I'm having are with tracking down the licensing of all the different files.

Is there any way we can use the slightly older versions of these files that still had the MPL v1 license for consistency with the rest of our Rhino codebase?
Also it would help if the file you wrote yourself also includes a license header (similar to the other files in that directory).

paullepoulpe:
The license version change was actually the last modification made to those files.

blickly:
Cool, thanks for the update!

test/com/google/javascript/rhino/dtoa/NumberToStringTest.java
```diff
@@ -0,0 +1,107 @@
...
+ * MPL, indicate your decision by deleting the provisions above and replacing
+ * them with the notice and other provisions required by the GPL. If you do
+ * not delete the provisions above, a recipient may use your version of this
+ * file under either the MPL or the GPL.
+ *
+ * ***** END LICENSE BLOCK ***** */
+
+package com.google.javascript.rhino.dtoa;
+
+import com.google.javascript.rhino.dtoa.DToA;
```
blickly: I think this import is superfluous.

test/com/google/javascript/rhino/dtoa/NumberToStringTest.java
```diff
@@ -0,0 +1,107 @@
...
+
+import junit.framework.TestCase;
+
+public class NumberToStringTest extends TestCase {
+
+  public NumberToStringTest() {
```
blickly: Same with this constructor.

paullepoulpe:
Bump

PR #1268, opened 2015-11-24
Update the .project and .classpath to use maven build paths. 

ChadKillingsworth:
Switch the eclipse build instructions and settings to use maven. We may want to consider officially deprecating ant builds and eventually remove the build.xml file as it will quickly get out of date.

[Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1268)

### DIFF
--- a/externs/pom.xml
+++ b/externs/pom.xml
@@ -68,5 +68,38 @@
         <directory>target/generated-resources//</directory>
       </resource>
     </resources>
+    <pluginManagement>
+    	<plugins>
+    		<!--This plugin's configuration is used to store Eclipse m2e settings only. It has no influence on the Maven build itself.-->
+    		<plugin>
+    			<groupId>org.eclipse.m2e</groupId>
+    			<artifactId>lifecycle-mapping</artifactId>
+    			<version>1.0.0</version>
+    			<configuration>
+    				<lifecycleMappingMetadata>
+    					<pluginExecutions>
+    						<pluginExecution>
+    							<pluginExecutionFilter>
+    								<groupId>
+    									org.apache.maven.plugins
+    								</groupId>
+    								<artifactId>
+    									maven-antrun-plugin
+    								</artifactId>
+    								<versionRange>[1.7,)</versionRange>
+    								<goals>
+    									<goal>run</goal>
+    								</goals>
+    							</pluginExecutionFilter>
+    							<action>
+    								<ignore></ignore>
+    							</action>
+    						</pluginExecution>
+    					</pluginExecutions>
+    				</lifecycleMappingMetadata>
+    			</configuration>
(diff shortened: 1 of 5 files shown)


MatrixFrog:
Definitely want to deprecate and delete the Ant build. I was working on getting Chrome to switch to the Maven build before I went on vacation, before officially claiming that the Maven one was the official way to build the compiler, but I haven't had time to get back to it lately.

externs/pom.xml
```diff
@@ -68,5 +68,38 @@
...
+    					<pluginExecutions>
+    						<pluginExecution>
+    							<pluginExecutionFilter>
+    								<groupId>
+    									org.apache.maven.plugins
+    								</groupId>
+    								<artifactId>
+    									maven-antrun-plugin
+    								</artifactId>
+    								<versionRange>[1.7,)</versionRange>
```
MatrixFrog: I think our internal infrastructure will likely work better if we specify a particular version rather than a range.

ChadKillingsworth: This whole section was auto-generated by eclipse - and I'm not great at maven. What should it be?

ChadKillingsworth:
Having mixed paths in .classpath file is screwing with intellij's ability to run tests. I have to go manually switch the build folder everytime I open the project which was the reason I submitted this in the first place.

externs/pom.xml
MatrixFrog: It's strange to me that Eclipse is modifying this file without you asking it to. I'll try pulling this in and see if it works.

ChadKillingsworth:
Sorry - should have been clearer. Eclipse gives warnings that it doesn't understand the antrun plugin. See https://www.eclipse.org/m2e/documentation/m2e-execution-not-covered.html#ignore-plugin-goal

The eclipse maven integration has an option to update the POM with that template to ignore the plugin.

Dominator008:
@ChadKillingsworth While fixing things for Eclipse, can you try running ScopedAliasesTest and see if some of them fail? I'm getting failures because "js/base.js" is not treated as a resource in Eclipse.

ChadKillingsworth:
@Dominator008 Right clicked on it, "Run As > JUnit Test". All tests passed. Tested on this branch.

Have to run a maven build first, but that's kind of expected.

ChadKillingsworth:
@Dominator008 But once you switch to [IntelliJ](https://www.jetbrains.com/idea/download/), you'll loathe debugging in Eclipse.

Dominator008:
@ChadKillingsworth Seems like the issue I mentioned has been fixed by #1423.

brad4d:
This PR is obsoleted by #1423 (already merged) and #1940 (merging today)

PR #1271, opened 2015-11-25
Switch the --json_streams flag to an enumeration

ChadKillingsworth:
I realize we just merged this feature yesterday, but it has worked so well for the gulp plugin that I also would like to use it for the grunt plugin as it completely mitigates problems with windows command length restrictions.

However, if both stdin and stdout are json streams, the plugin has to recreate much of the functionality of the CommandLineRunner in regards to smartly parsing sourcemap, module and output file flags. This is further complicated if a flagfile is used.

By allowing control of input/output streams independently, build systems can choose to allow the compiler to handle these tasks.

### DIFF
--- a/src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
+++ b/src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
@@ -33,6 +33,7 @@ import com.google.common.io.ByteStreams;
 import com.google.gson.Gson;
 import com.google.gson.stream.JsonReader;
 import com.google.gson.stream.JsonWriter;
+import com.google.javascript.jscomp.CompilerOptions.JsonStreamMode;
 import com.google.javascript.jscomp.CompilerOptions.TweakProcessing;
 import com.google.javascript.jscomp.deps.ClosureBundler;
 import com.google.javascript.jscomp.deps.SourceCodeEscapers;
@@ -227,6 +228,15 @@ public abstract class AbstractCommandLineRunner<A extends Compiler,
   }
 
   /**
+   * Returns whether output should be a JSON stream.
+   */
+  private boolean isOutputInJson() {
+    return config.jsonStreamMode == JsonStreamMode.OUT
+        || config.jsonStreamMode == JsonStreamMode.BOTH;
+  }
+
+
+  /**
    * Get the command line config, so that it can be initialized.
    */
   protected CommandLineConfig getCommandLineConfig() {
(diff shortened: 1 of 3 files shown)


src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
```diff
@@ -368,7 +368,8 @@ protected void setRunOptions(CompilerOptions options)
 
     if (config.createSourceMap.length() > 0) {
       options.sourceMapOutputPath = config.createSourceMap;
-    } else if (config.useJsonStreams) {
+    } else if (config.jsonStreamMode == CompilerOptions.JsonStreamMode.OUT ||
+        config.jsonStreamMode == CompilerOptions.JsonStreamMode.BOTH) {
```
dimvar: The out || both check is the most common. You can define a private helper method in this file:

    private boolean isOutputInJson() {
      return config.jsonStreamMode == CompilerOptions.JsonStreamMode.OUT
            || config.jsonStreamMode == CompilerOptions.JsonStreamMode.BOTH;
    }

src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
```diff
@@ -1041,7 +1042,8 @@ protected int doRun() throws FlagUsageException, IOException {
     List<String> moduleSpecs = config.module;
     List<JsonFileSpec> jsonFiles = null;
 
-    if (config.useJsonStreams) {
+    if (config.jsonStreamMode == CompilerOptions.JsonStreamMode.IN ||
```
dimvar: You can import JsonStreamMode to avoid wrting the CompilerOptions prefix every time you use the enum.

    import com.google.javascript.jscomp.CompilerOptions.JsonStreamMode;

src/com/google/javascript/jscomp/CompilerOptions.java
```diff
@@ -2639,4 +2639,30 @@ public void addAlias(String alias, String definition) {
      */
     CUSTOM
   }
+
+  /**
+   * Whether standard input or standard output should be an array of
+   * JSON encoded files
+   */
+  public static enum JsonStreamMode {
```
dimvar: can you make this package private?

ChadKillingsworth:
Done. Commits squashed.

src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
```diff
@@ -1171,7 +1182,7 @@ int processResults(Result result, List<JSModule> modules, B options)
         // Output the source map if requested.
         // If output files are being written to stdout as a JSON string,
         // outputSingleBinary will have added the sourcemap to the output file
-        if (!config.useJsonStreams) {
+        if (isOutputInJson()) {
```
dimvar: Did you forget a ! here?

ChadKillingsworth:
Sure did - nice catch.

PR #1275, opened 2015-11-30
jQuery.prototype.not - arg1 as a function can take 2 arguments and should return boolean

vobruba-martin:
From jQuery API:

function
Type: Function( Integer index, Element element ) => Boolean

A function used as a test for each element in the set. It accepts two arguments, index, which is the element's index in the jQuery collection, and element, which is the DOM element. Within the function, this refers to the current DOM element.

https://api.jquery.com/not/


Dominator008:
Looks good, thanks for the PR!

PR #1278, opened 2015-12-02
Mark angular.$timeout fn argument optional

fredj:
The fn argument to angular.$timeout service is optional, update the externs.
See https://code.angularjs.org/1.4.8/docs/api/ng/service/$timeout


Dominator008:
Looks good, thanks!

PR #1280, opened 2015-12-03
Update link to ES5 documentation

vobruba-martin:
Current link is dead.

### DIFF
--- a/externs/es5.js
+++ b/externs/es5.js
@@ -16,7 +16,7 @@
 
 /**
  * @fileoverview Definitions for ECMAScript 5.
- * @see http://www.ecma-international.org/publications/files/drafts/tc39-2009-025.pdf
+ * @see https://es5.github.io/
  * @externs
  */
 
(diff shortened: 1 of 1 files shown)


externs/es5.js
```diff
@@ -16,7 +16,7 @@
 
 /**
  * @fileoverview Definitions for ECMAScript 5.
- * @see http://www.ecma-international.org/publications/files/drafts/tc39-2009-025.pdf
+ * @see http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-262.pdf
```
MatrixFrog: That links to the latest edition which is currently the 6th. Although it's not as "official" I think it would be nice to link to https://es5.github.io/ since it has nice cross-reference links and such.

externs/es5.js
```diff
@@ -211,7 +211,7 @@ Object.isFrozen = function(obj) {};
 
 
 /**
- * As per ECMAScript 5, 15.12.3.
+ * As per ECMAScript 5, 24.3.2 and 24.3.2.1
```
MatrixFrog: It looks like these numbers are wrong for ES5, but I would be happy to just nuke this line anyway. If someone wants to look in the spec they'll probably just ctrl-F for `toJSON`.

vobruba-martin:
Link updated. Thanks!

PR #1281, opened 2015-12-03
Remove nosideeffects tag from JSON.parse and JSON.stringify

vobruba-martin:
Or does it make any sence to mix throw and nosideeffects tags?

[Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1281)


Dominator008:
@concavelenz I think removing @nosideeffects here makes sense?

dimvar:
@concavelenz can you chime in here?

concavelenz:
I think removing @nosideeffect is correct.

PR #1285, opened 2015-12-04
Added nosideeffects for more jQuery methods

vobruba-martin:
I went through the jQuery docs and source code and found out that these calls can be safely removed if their return value is not used.

Plus I've included small patch for jQuery.error annotation. Or do you prefer sending it as a new PR?


ChadKillingsworth:
One PR is fine - it's all the same file. Historically, we've been pretty conservative with the `@nosideeffects` with jQuery because the internal implementation can tricky, but these all look correct.

Since contrib/externs aren't compiled with the unit tests, I'll have to pull this down manually to check.

PR #1287, opened 2015-12-05
Support mixing ES6, CommonJS and goog.module systems

ChadKillingsworth:
When processing commonjs modules, the compiler assumed that all modules had a single export. This does not hold true for ES6 modules, so a "duplicate module detected" error was thrown. This change uses the unique set of modules found to build the module graph.

Also, when mixing module systems, the entry point can be specified as either a closure entry point or a commonjs entry point or a combination of the two. Keep the warning for the most common use case, but allow advanced users to specify a single entry point in whichever format they choose.

This allows full dependency pruning across multiple module types to properly function :100:

### DIFF
--- a/src/com/google/javascript/jscomp/CommandLineRunner.java
+++ b/src/com/google/javascript/jscomp/CommandLineRunner.java
@@ -428,7 +428,7 @@ public class CommandLineRunner extends
         hidden = true,
         usage = "Root of your common JS dependency hierarchy. " +
             "Your main script.")
-    private String commonJsEntryModule;
+    private List<String> commonJsEntryModule;
 
     @Option(name = "--transform_amd_modules",
         hidden = true,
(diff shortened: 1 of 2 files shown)


src/com/google/javascript/jscomp/CommandLineRunner.java
```diff
@@ -1043,11 +1043,19 @@ private void initConfigFromFlags(String[] args, PrintStream out, PrintStream err
     if (flags.processCommonJsModules) {
       flags.processClosurePrimitives = true;
       flags.manageClosureDependencies = true;
-      if (flags.commonJsEntryModule == null) {
+
+      // When module systems are mixed, the entry point can be either a commonJs module
```
Dominator008: s/commonJs/CommonJS

bfricka:
:+1:

ChadKillingsworth:
Ping. Can someone review this?

concavelenz:
I don't understand the premise.  ES6 modules have a single export object.

ChadKillingsworth:
Hmm I'll rerun the tests I was using. The culprit may have been the closure-library code where files can and do have multiple "provide" statements.

bfricka:
What I'd love to see: `import array from 'goog'`;

ChadKillingsworth:
@bfricka That new in yesterday's release:

`import EventTarget from 'goog:goog.events.EventTarget';`

@concavelenz It was the closure-library imports that triggered the issue. When I stepped through the code, it had [something like 1400 provided files](https://github.com/google/closure-compiler/pull/1287/files#diff-9d2224104bbe9103745c28cd598e1f75L1520) but only 900 unique ones. The [JSModuleGraph requires a unique set](https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/JSModuleGraph.java#L94).

bfricka:
@ChadKillingsworth Thanks so much for that news. This could not come at a better time for our team!

ChadKillingsworth:
I'm closing this in favor of #1329 which will handle all these cases and more. And have tests.

PR #1289, opened 2015-12-07
Enable GWT logging in JsComp.gwt.xml

Dominator008:
This enables printing of errors / warnings to the browser's console from
gwt_demo.html.


MatrixFrog:
I thought we were already printing them to a <div> on the page or something?

Dominator008:
Not yet in the open-source gwt_demo.html. Not sure about internal usage :)

PR #1292, opened 2015-12-11
Add --hide_warnings_for to suppress warnings by path.

Mazerus:
I'd love the ability to hide warnings for all code in a given path, such as for third-party projects that I don't control. This change attempts to support that.

If I should send a CL instead, let me know. This is my first PR, so be gentle (ish). =)

Thanks!


dimvar:
Yes, let's handle this through a CL. There is a similar flag that is internal only.

PR #1295, opened 2015-12-13
Remove unused field "acceptTypeSyntax" from CompilerOptions

Dominator008:


Dominator008:
Not sure if there's internal usage of this, but I guess there's none.

mprobst:
@Dominator008 yeah, I'm not sure where that came from, but I think ECMASCRIPT6_TYPED is now used for that purpose.

LGTM.

MatrixFrog:
Looks like we are actually using it in some code that is MOE-stripped out :(

mprobst:
Making open "easy" :-(

Maybe it's possible to recover in the sync CL?

PR #1297, opened 2015-12-13
Add missing extern for `Array.from`

ambar:

PR #1298, opened 2015-12-13
Add missing extern for `Array.from` (update of #1297)

Dominator008:


Dominator008:
@ambar I've updated some type annotations here.

ambar:
Thanks!

MatrixFrog:
We landed an Array.from extern recently and it broke some internal projects. We'll probably have to do this one internally first and then push it out.

MatrixFrog:
LGTM but it looks like it needs a rebase, which should also kick off a new Travis run.

Dominator008:
The "pull/1297" ref is read-only, reapplying this in #1335.

PR #1307, opened 2015-12-17
Add dependency-reduced-pom.xml to .gitignore

Dominator008:
This file is generated by the Maven Shade Plugin and should be ignored.


ChadKillingsworth:
:+1: Thanks. This has been bugging me.

PR #1310, opened 2015-12-18
Allow more characters to be used in output module names

ChadKillingsworth:
This has been asked for by others, but now I need it.

In theory, an output module name could be any valid file path. This PR just loosens the existing requirements slightly to allow "." and "-" as valid characters.

### DIFF
--- a/src/com/google/javascript/jscomp/CommandLineRunner.java
+++ b/src/com/google/javascript/jscomp/CommandLineRunner.java
@@ -27,6 +27,7 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.io.Files;
 import com.google.javascript.jscomp.SourceMap.LocationMapping;
+import com.google.javascript.rhino.TokenStream;
 import com.google.protobuf.TextFormat;
 
 import org.kohsuke.args4j.Argument;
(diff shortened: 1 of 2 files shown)


src/com/google/javascript/jscomp/CommandLineRunner.java
```diff
@@ -1156,6 +1160,15 @@ protected void addWhitelistWarningsGuard(
   }
 
   @Override
+  protected void checkModuleName(String name)
+      throws FlagUsageException {
+    if (!TokenStream.isJSIdentifier(
+        extraModuleNameChars.matcher(name).replaceAll("_"))) {
```
MatrixFrog: This looks like it would change "foo--bar.js" (with multiple -'s) to foo_bar.js (with only one _). Should the regex be `[-.]` instead of `[-.]+`?

ChadKillingsworth: It's only doing that to test if the identifier is valid.

ChadKillingsworth: I could - I think the difference would be the same though.

MatrixFrog: Let's add a test to be sure.

ChadKillingsworth: The easiest way to test this would be to utilize the `--json_streams` flag. So we'll need to merge #1313 first to support that.

ChadKillingsworth:
Tests added and rebased to HEAD.

MatrixFrog:
lgtm, just waiting for travis

ChadKillingsworth:
I like that we're over 10,000 tests now

ChadKillingsworth:
I just reran the tests locally - and they all passed.

PR #1312, opened 2015-12-18
Remove angular-specific references from jasmine-2.0 externs

ChadKillingsworth:
The Jasmine 2.0 externs have angular specific references. This requires jasmine users to include the angular externs for nothing more than type information.

I'm guessing this will cause issues internal to Google. I can make an angular-jasmine specific extern if that will help.


mprobst:
@ChadKillingsworth indeed. Could you extract these additional methods into a separate externs file?

ChadKillingsworth:
@mprobst done

mprobst:
LGTM.

PR #1329, opened 2015-12-22
Simplify dependency flags to use a common set for all module types.

### LINKED ISSUE #1319: Simplify Dependency Management Flags
Currently, we have multiple dependency flags:

### Closure
*  `--manage_closure_dependencies`
* `--only_closure_dependencies`
* `--closure_entry_point`

### CommonJS

* `--process_common_js_modules` - internally this enables `--only_closure_dependencies`
* `--common_js_entry_point` - until #1287 is merged, this only supports a single entry point

### ES6
We're missing dependencies flags here. You "can" use the closure flags, but you have to specify your entry points as the normalized module name `module$path$file`.

I'd like to get this down to a single set of flags that supports all 3.

### New Flags Proposal
These flags would replace all of the existing dependency flags mentioned above.

* `--process_common_js_modules` - this would only enable module rewriting.
* `--manage_dependencies` - This would have two modes (yet to be named) to allow users to control whether `dropMoochers` is enabled in DependencyOptions.
* `--entry_point` - this would use the ES6 and CommonJS file based syntax. For closure, it would use the module import syntax: `goog:namespace`

ChadKillingsworth:
Fixes #1319.

Removes the `--manage_closure_dependencies`, `--only_closure_dependencies`, `--closure_entry_point` and `--common_js_entry_module` flags in favor of simplified set:

* `--dependencies_mode` - Enum with `NONE`, `LOOSE` or `STRICT` as options. None is default. Loose is equivalent to the old `--manage_closure_dependencies`. Strict is equivalent to `--only_closure_dependencies`.
* `--entry_point` - for ES6 and CommonJS moduels, this is the file path without the extension. For closure, it uses the module import syntax: goog:namespace

This change also allows the `--dependency_mode` flag to fully control CommonJS modules. Previously, CommonJS module dependencies implied `--only_closure_dependencies` and supported only a single entry point.

### DIFF
--- a/src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
+++ b/src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
@@ -277,34 +277,26 @@ public abstract class AbstractCommandLineRunner<A extends Compiler,
    * A helper function for creating the dependency options object.
    */
   static DependencyOptions createDependencyOptions(
-      boolean manageClosureDependencies,
-      boolean onlyClosureDependencies,
-      boolean processCommonJSModules,
-      List<String> closureEntryPoints)
+      CompilerOptions.DependencyMode dependencyMode,
+      List<DependencyOptions.ModuleIdentifier> entryPoints)
       throws FlagUsageException {
-    if (onlyClosureDependencies) {
-      if (closureEntryPoints.isEmpty()) {
-        throw new FlagUsageException("When only_closure_dependencies is "
-          + "on, you must specify at least one closure_entry_point");
+    if (dependencyMode == CompilerOptions.DependencyMode.STRICT) {
+      if (entryPoints.isEmpty()) {
+        throw new FlagUsageException("When dependency_mode=STRICT, you must "
+            + "specify at least one entry_point");
       }
 
       return new DependencyOptions()
           .setDependencyPruning(true)
           .setDependencySorting(true)
           .setMoocherDropping(true)
-          .setEntryPoints(closureEntryPoints);
-    } else if (processCommonJSModules) {
-      return new DependencyOptions()
-        .setDependencyPruning(false)
-        .setDependencySorting(true)
-        .setMoocherDropping(false)
-        .setEntryPoints(closureEntryPoints);
-    } else if (manageClosureDependencies || !closureEntryPoints.isEmpty()) {
+          .setEntryPoints(entryPoints);
+    } else if (dependencyMode == CompilerOptions.DependencyMode.LOOSE || !entryPoints.isEmpty()) {
       return new DependencyOptions()
           .setDependencyPruning(true)
           .setDependencySorting(true)
           .setMoocherDropping(false)
-          .setEntryPoints(closureEntryPoints);
+          .setEntryPoints(entryPoints);
     }
     return null;
   }
@@ -355,10 +347,8 @@ public abstract class AbstractCommandLineRunner<A extends Compiler,
     createDefineOrTweakReplacements(config.tweak, options, true);
 
     DependencyOptions depOptions = createDependencyOptions(
-        config.manageClosureDependencies,
-        config.onlyClosureDependencies,
-        config.processCommonJSModules,
-        config.closureEntryPoints);
+        config.dependencyMode,
+        config.entryPoints);
     if (depOptions != null) {
       options.setDependencyOptions(depOptions);
     }
@@ -1771,8 +1761,8 @@ public abstract class AbstractCommandLineRunner<A extends Compiler,
   }
 
   /**
-   * Writes the manifest or bundle of all compiler input files that survived
-   * manage_closure_dependencies, if requested.
+   * Writes the manifest or bundle of all compiler input files that were included
+   * as controlled by --dependency_mode, if requested.
    */
   private void outputManifestOrBundle(List<String> outputFiles,
       boolean isManifest) throws IOException {
@@ -2287,38 +2277,26 @@ public abstract class AbstractCommandLineRunner<A extends Compiler,
       return this;
     }
 
-    private boolean manageClosureDependencies = false;
+    private CompilerOptions.DependencyMode dependencyMode = CompilerOptions.DependencyMode.NONE;
 
     /**
      * Sets whether to sort files by their goog.provide/require deps,
      * and prune inputs that are not required.
      */
-    CommandLineConfig setManageClosureDependencies(boolean newVal) {
-      this.manageClosureDependencies = newVal;
+    CommandLineConfig setDependencyMode(CompilerOptions.DependencyMode newVal) {
+      this.dependencyMode = newVal;
       return this;
     }
 
-    private boolean onlyClosureDependencies = false;
-
-    /**
-     * Sets whether to sort files by their goog.provide/require deps,
-     * and prune inputs that are not required, and drop all non-closure
-     * files.
-     */
-    CommandLineConfig setOnlyClosureDependencies(boolean newVal) {
-      this.onlyClosureDependencies = newVal;
-      return this;
-    }
-
-    private List<String> closureEntryPoints = ImmutableList.of();
+    private List<DependencyOptions.ModuleIdentifier> entryPoints = ImmutableList.of();
 
     /**
      * Set closure entry points, which makes the compiler only include
      * those files and sort them in dependency order.
      */
-    CommandLineConfig setClosureEntryPoints(List<String> entryPoints) {
+    CommandLineConfig setEntryPoints(List<DependencyOptions.ModuleIdentifier> entryPoints) {
       Preconditions.checkNotNull(entryPoints);
-      this.closureEntryPoints = entryPoints;
+      this.entryPoints = entryPoints;
       return this;
     }
 
--- a/src/com/google/javascript/jscomp/CommandLineRunner.java
+++ b/src/com/google/javascript/jscomp/CommandLineRunner.java
@@ -425,8 +425,7 @@ public class CommandLineRunner extends
     @Option(
       name = "--common_js_module_path_prefix",
       hidden = true,
-      usage = "Path prefix to be removed from CommonJS module names."
-    )
+      usage = "Deprecated: use --js_module_root.")
     private List<String> commonJsPathPrefix = new ArrayList<>();
 
     @Option(
@@ -438,8 +437,7 @@ public class CommandLineRunner extends
 
     @Option(name = "--common_js_entry_module",
         hidden = true,
-        usage = "Root of your common JS dependency hierarchy. " +
-            "Your main script.")
+        usage = "Deprecated: use --entry_point.")
     private String commonJsEntryModule;
 
     @Option(name = "--transform_amd_modules",
@@ -459,31 +457,18 @@ public class CommandLineRunner extends
     @Option(name = "--manage_closure_dependencies",
         hidden = true,
         handler = BooleanOptionHandler.class,
-        usage = "Automatically sort dependencies so that a file that "
-        + "goog.provides symbol X will always come before a file that "
-        + "goog.requires symbol X. If an input provides symbols, and "
-        + "those symbols are never required, then that input will not "
-        + "be included in the compilation.")
+        usage = "Deprecated: use --dependency_mode=LOOSE.")
     private boolean manageClosureDependencies = false;
 
     @Option(name = "--only_closure_dependencies",
         hidden = true,
         handler = BooleanOptionHandler.class,
-        usage = "Only include files in the transitive dependency of the "
-        + "entry points (specified by closure_entry_point). Files that do "
-        + "not provide dependencies will be removed. This supersedes "
-        + "manage_closure_dependencies")
+        usage = "Deprecated: use --dependency_mode=STRICT.")
     private boolean onlyClosureDependencies = false;
 
     @Option(name = "--closure_entry_point",
         hidden = true,
-        usage = "Entry points to the program. Must be goog.provide'd "
-        + "symbols. Any goog.provide'd symbols that are not a transitive "
-        + "dependency of the entry points will be removed. Files without "
-        + "goog.provides, and their dependencies, will always be left in. "
-        + "If any entry points are specified, then the "
-        + "manage_closure_dependencies option will be set to true and "
-        + "all files will be sorted in dependency order.")
+        usage = "Deprecated: use --entry_point.")
     private List<String> closureEntryPoint = new ArrayList<>();
 
     @Option(name = "--process_jquery_primitives",
@@ -521,7 +506,7 @@ public class CommandLineRunner extends
     @Option(name = "--output_manifest",
         hidden = true,
         usage = "Prints out a list of all the files in the compilation. "
-        + "If --manage_closure_dependencies is on, this will not include "
+        + "If --dependency_mode=STRICT or LOOSE is specified, this will not include "
         + "files that got dropped because they were not required. "
         + "The %outname% placeholder expands to the JS output file. "
         + "If you're using modularization, using %outname% will create "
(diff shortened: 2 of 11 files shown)


ChadKillingsworth:
This branch is now fully functional - just waiting on #1341 to be merged. After that, I'll rebase this one for review. I'd also like to ping @nicks to make sure these changes work for his use as well.

src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
```diff
@@ -278,13 +278,12 @@ protected DiagnosticGroups getDiagnosticGroups() {
   static DependencyOptions createDependencyOptions(
       boolean manageClosureDependencies,
       boolean onlyClosureDependencies,
-      boolean processCommonJSModules,
       List<String> closureEntryPoints)
```
mprobst: rename all occurences of `closureEntryPoint[s]` to just `entryPoints`?

mprobst: AFAIK this used to be a list of symbols (e.g. `foo.bar.myapp`), now it's a list of files, right? If so, would it make sense to change this to a `List<File>`, so we can also catch old code locations that work off incorrect assumptions?

src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
```diff
@@ -1771,8 +1763,8 @@ private void outputBundle() throws IOException {
   }
 
   /**
-   * Writes the manifest or bundle of all compiler input files that survived
-   * manage_closure_dependencies, if requested.
+   * Writes the manifest or bundle of all compiler input files that were used
```
mprobst: nit: maybe `s/used/included/`?

src/com/google/javascript/jscomp/CommandLineRunner.java
```diff
@@ -642,6 +599,28 @@
         usage = "Prevent injecting the named runtime libraries.")
     private List<String> noinjectLibrary = new ArrayList<>();
 
+    @Option(name = "--dependency_mode",
+        hidden = true,
+        usage = "Specifies how the compiler should determine the set and order "
+            + "of files for a compilation. Options: NONE the compiler will include all "
+            + "src files in the order listed, STRICT files will be included and sorted by "
+            + "starting from namespaces or files listed by the --entry_point flag - files "
```
mprobst: An interesting question with dependency management is the *order* in which files get included. This might need some documentation - as a user I'd expect the effective order to be a pre-order traversal of the dependency tree. Is that what happens here? Maybe spell it out, possible even with an example somewhere?

src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
```diff
@@ -292,12 +291,6 @@ static DependencyOptions createDependencyOptions(
           .setDependencySorting(true)
           .setMoocherDropping(true)
           .setEntryPoints(closureEntryPoints);
-    } else if (processCommonJSModules) {
-      return new DependencyOptions()
-        .setDependencyPruning(false)
-        .setDependencySorting(true)
-        .setMoocherDropping(false)
-        .setEntryPoints(closureEntryPoints);
```
mprobst: Could you update `DependencyOptions` and document the new meaning of `EntryPoints`?

test/com/google/javascript/jscomp/CommandLineRunnerTest.java
```diff
@@ -1459,6 +1458,123 @@ public void testProcessCJSWithClosureRequires() {
...
+
+  public void testES6ImportOfCJS() {
+    args.add("--process_common_js_modules");
+    args.add("--entry_point=app");
+    args.add("--dependency_mode=STRICT");
+    args.add("--language_in=ECMASCRIPT6");
+    setFilename(0, "foo.js");
+    setFilename(1, "app.js");
+    test(new String[] {
+            "/** @constructor */ function Foo () {}" +
```
mprobst: maybe pull out a `join` method that takes varargs `String...` and joins them with a line break (using `Joiner`)? That'd make both this code more readable and the error messages the code might produce.

I vaguely remember writing something like that in another file, but no longer remember where exactly - maybe look for `Joiner` in our test files.

src/com/google/javascript/jscomp/CommandLineRunner.java
```diff
@@ -1116,13 +1089,28 @@ private void initConfigFromFlags(String[] args, PrintStream out, PrintStream err
...
       } else {
         moduleRoots.add(ES6ModuleLoader.DEFAULT_FILENAME_PREFIX);
       }
 
+      List<String> normalizedEntryPoints = new ArrayList<>();
+      for(String entryPoint : flags.entryPoints) {
+        if (entryPoint.startsWith("goog:")) {
+          normalizedEntryPoints.add(entryPoint.substring(5));
+        } else {
+          normalizedEntryPoints.add(ES6ModuleLoader.toModuleName(URI.create(entryPoint)));
```
mprobst: Nice. One question though - what's the error message the user will see when the expected entry point is missing? We should make sure it doesn't contain `$` all over the place, but rather reflects the original file name.

To make that simple, maybe we shouldn't be passing around `String`s, but some simple `ModuleIdentifier` that has both the original name (file or goog: symbol) and the mangled file name?

test/com/google/javascript/jscomp/ProcessCommonJSModulesTest.java
```diff
@@ -251,38 +251,4 @@ public void testEs6ObjectShorthand() {
         "goog.provide('module$test');"
             + "module$test = { prop: 'value', foo() { console.log('bar'); }}" );
   }
-
-  public void testSortInputs() throws Exception {
-    SourceFile a = SourceFile.fromCode("a.js", "require('b');require('c')");
```
mprobst: Is this test really obsolete? Do we have other code testing the pre-order traversal sort here?

ChadKillingsworth:
I believe all the review items have been addressed. I did not want to define the behavior when the old flags and these new flags are used together, so I marked the old flags with a `forbids` option. However that requires updating our args4j version so #1350 will need merged first.

From @mprobst 
> We should clearly communicate what order of files users can expect, and maybe also make sure that we get repeatable file-ordering results between different runs, even if there's no defined total order of files (e.g. unrelated siblings).

I agree - but this may need to be handled on the wiki as it's a bit hard to document in a flag. We have an entire page that will need updated related to the manage closure dependencies flags.  Order is even trickier with `--dependency_mode=LOOSE` as files that do not have dependency information should retain their relative order. We haven't documented this well to date.

test/com/google/javascript/jscomp/ProcessCommonJSModulesTest.java
ChadKillingsworth: I believe we do - we great test coverage for sorting: https://github.com/google/closure-compiler/blob/master/test/com/google/javascript/jscomp/JSModuleGraphTest.java

The CommonJS modules created their own module graph and this test was for that. I removed that code as it didn't work well with other modules and as far as I can tell was completely unnecessary.

src/com/google/javascript/jscomp/DependencyOptions.java
```diff
@@ -151,7 +151,34 @@ boolean shouldDropMoochers() {
     return pruneDependencies && dropMoochers;
   }
 
-  Collection<String> getEntryPoints() {
+  Collection<ModuleIdentifier> getEntryPoints() {
     return entryPoints;
   }
+
+  public static final class ModuleIdentifier {
+    private final String name;
```
mprobst: Could you document what each of these are?

src/com/google/javascript/jscomp/DependencyOptions.java
```diff
@@ -151,7 +151,34 @@ boolean shouldDropMoochers() {
...
+    private final String normalizedModuleName;
+
+    public ModuleIdentifier(String name) {
```
mprobst: Maybe give this factory methods to make code more obvious, e.g. `ModuleIdentifier.forClosureModule("goog:foo.bar")` and `ModuleIdentifier.forFile("foo/bar.js")`?

src/com/google/javascript/jscomp/JSModuleGraph.java
```diff
@@ -452,26 +452,30 @@ public void coalesceDuplicateFiles() {
...
         CompilerInput entryPointInput = null;
-        if (splitPoint != -1) {
-          String moduleName = entryPoint.substring(0, splitPoint);
-          inputName = entryPoint.substring(
-              Math.min(splitPoint + 1, entryPoint.length() - 1));
-          JSModule module = modulesByName.get(moduleName);
-          if (module == null) {
-            throw new MissingModuleException(moduleName);
+        try {
+          if (splitPoint != -1) {
```
mprobst: optional: maybe move this parsing logic into the module identifier?

src/com/google/javascript/jscomp/JSModuleGraph.java
```diff
@@ -452,26 +452,30 @@ public void coalesceDuplicateFiles() {
...
+              entryPointInput.overrideModule(module);
+            }
           } else {
             entryPointInput = sorter.getInputProviding(inputName);
-            entryPointInput.overrideModule(module);
           }
-        } else {
-          entryPointInput = sorter.getInputProviding(inputName);
+        } catch (MissingProvideException e) {
+          throw new MissingProvideException(entryPoint.getName());
```
mprobst: please make sure to include the original exception (and stack trace!) as the cause, it makes debugging much easier.

test/com/google/javascript/jscomp/CommandLineRunnerTest.java
```diff
@@ -1406,37 +1405,37 @@ public void testProcessCJSWithModuleOutput() {
...
-           "    return 4 + 4;\n" +
-           "  }\n" +
-           "}\n" +
-           "module.exports = Baz;",
-
-           "var Baz = require('./Baz');\n" +
-           "var baz = new Baz();\n" +
-           "console.log(baz.baz());\n" +
-           "console.log(baz.bar());\n"
+           Joiner.on('\n').join("/** @provideGoog */",
```
mprobst: pull the `Joiner.on('\n').join(...)` out into a static method?

src/com/google/javascript/jscomp/JSModuleGraph.java
ChadKillingsworth: This makes a whole lot of sense. I'll tackle that.

src/com/google/javascript/jscomp/JSModuleGraph.java
ChadKillingsworth: I just had to go add a new constructor method to handle that case. Done now.

ChadKillingsworth:
Latest round of review items have now been addressed.

src/com/google/javascript/jscomp/DependencyOptions.java
```diff
@@ -151,7 +152,87 @@ boolean shouldDropMoochers() {
...
+      String moduleName = normalizedName;
+      int splitPoint = normalizedName.indexOf(':');
+      if (splitPoint != -1) {
+        moduleName = normalizedName.substring(0, splitPoint);
+        namespace = normalizedName.substring(Math.min(splitPoint + 1, normalizedName.length() - 1));
+      }
+
+      return new ModuleIdentifier(normalizedName, namespace, moduleName);
+    }
+    /**
```
mprobst: nit: newline before

src/com/google/javascript/jscomp/deps/SortedDependencies.java
```diff
@@ -71,8 +71,12 @@
   }
 
   public static class MissingProvideException extends Exception {
-    MissingProvideException(String provide) {
+    public MissingProvideException(String provide) {
```
mprobst: nit: I presume you can delete this ctor?

src/com/google/javascript/jscomp/deps/SortedDependencies.java
```diff
@@ -71,8 +71,12 @@
...
       super(provide);
     }
+
+    public MissingProvideException(String provide, Exception e) {
+      super(provide, e);
```
mprobst: Are we displaying this to the user? It might be nicer to have a message like `could not find some/file.js` instead of just `some/file.js`

test/com/google/javascript/jscomp/CommandLineRunnerTest.java
mprobst: Ping :-)

test/com/google/javascript/jscomp/CommandLineRunnerTest.java
```diff
@@ -1406,37 +1405,37 @@ public void testProcessCJSWithModuleOutput() {
...
-           "    return 4 + 4;\n" +
-           "  }\n" +
-           "}\n" +
-           "module.exports = Baz;",
-
-           "var Baz = require('./Baz');\n" +
-           "var baz = new Baz();\n" +
-           "console.log(baz.baz());\n" +
-           "console.log(baz.bar());\n"
+           joinStrings("/** @provideGoog */",
```
mprobst: Ah I see, if test() always takes an array, the `joinStrings` suggestion is a lot less useful than I expected. Feel free to revert to the previous style, but if you keep `joinStrings`, I'd suggest using it consistently across the whole file.

mprobst:
Some nits, LGTM otherwise.

src/com/google/javascript/jscomp/deps/SortedDependencies.java
ChadKillingsworth: No - this version is used in other places.

src/com/google/javascript/jscomp/deps/SortedDependencies.java
ChadKillingsworth: It's caught in [Compiler.java](https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/Compiler.java#L1338) and translated to a user friendly message.

test/com/google/javascript/jscomp/CommandLineRunnerTest.java
ChadKillingsworth: I try to avoid making code style changes in the same PR as a feature change - especially in tests. We have a lot of tests which don't quite meat current style guides.

ChadKillingsworth:
Review items fixed and rebased to the top of the tree for Travis fixes. Should be gtg.

Also, @nicks weighed in via email:

> Malte designed the Closure Compiler / CommonJS  system many years ago. It was very exploratory/experimental, and had very different goals than the goals we have now. (I may be misremembering, but i think he wanted the Closure Compiler to preserve the module structure -- so that if there were N inputs, there'd be N outputs, but with dead code removed and code moved into later modules when possible). I think it'd be reasonable to change the design.

ChadKillingsworth:
Also, since #1350 isn't mergable, it no longer requires args4j to be updated first.

ChadKillingsworth:
Rebased and fixed tests now that #1348 was merged.

src/com/google/javascript/jscomp/DependencyOptions.java
```diff
@@ -151,7 +152,88 @@ boolean shouldDropMoochers() {
...
+     *
+     * Closure entry points may also be formatted as:
+     *     'moduleName:name.space'
+     * which specifies that the module name and provided namespace
+     * are different
+     */
+    public static ModuleIdentifier forClosure(String name) {
+      String normalizedName = name;
+      if (normalizedName.startsWith("goog:")) {
+        normalizedName = normalizedName.substring(5);
```
MatrixFrog: nit: "goog:".length()

src/com/google/javascript/jscomp/DependencyOptions.java
```diff
@@ -151,7 +152,88 @@ boolean shouldDropMoochers() {
...
+      }
+      return moduleName + ":" + closureNamespace;
+    }
+
+    /**
+     * @param name Closure namespace used as an entry point. May start
+     *     "goog:" when provided as a flag from the command line.
+     *
+     * Closure entry points may also be formatted as:
+     *     'moduleName:name.space'
```
MatrixFrog: This is after the "goog:" right? I think the comment would be clearer if it said 'goog:moduleName:name.space'

test/com/google/javascript/jscomp/CommandLineRunnerTest.java
MatrixFrog: +1, I really find tests easier to read when they use LINE_JOINER.join

src/com/google/javascript/jscomp/DependencyOptions.java
ChadKillingsworth: done

src/com/google/javascript/jscomp/DependencyOptions.java
ChadKillingsworth: done

test/com/google/javascript/jscomp/CommandLineRunnerTest.java
ChadKillingsworth: @MatrixFrog I was proposing we convert the styles of the tests as a separate PR as this one is pretty large already.

supersteves:
Awesome!!!

ChadKillingsworth:
I've updated the documentation to reflect this merge: https://github.com/google/closure-compiler/wiki/Managing-Dependencies

blickly:
Hi Chad. Unfortunately, we rely on setClosureEntryPoints internally. I may rollback now and then rollforward once we get this sorted out.

Is there a straightforward mapping from the namespace name to the new entry point format?

ChadKillingsworth:
@blickly I was worried about just this when @mprobst suggested it. Yes there is a very simple mapping - let me get a PR ready.

mprobst:
Sorry if I caused trouble here, I wasn't aware of that.

MatrixFrog:
No problem. The separate command line runners cause some friction occasionally. MOE is improving so it should be easier to deal with this kind of thing soon.

ChadKillingsworth:
@mprobst I think it was the right thing to do. I'm re-adding the `setClosureEntryPoints` method with the hope that it can eventually be removed.

ChadKillingsworth:
@blickly #1415 should fix this. But I don't have a good way to test it myself.

blickly:
Thanks Chad! I'll rollforward a version of this PR that includes that change.

PR #1333, opened 2015-12-22
Revert "Remove angular-specific references from jasmine-2.0 externs"

blickly:
Reverts google/closure-compiler#1312

This should probably be merged internally first so we can fix up people using both externs files.


ChadKillingsworth:
@blickly Can this be rolled forward?

PR #1341, opened 2015-12-24
Update the CommonJS module rewriting to match an ES6 default export

ChadKillingsworth:
This allows CommonJS and ES6 modules greater compatibility. Previously, the only way to import a CommonJS module using ES6 syntax was:

```js
import * as Foo from './bar';
```

Using default property exports makes interoperability much cleaner. See the added tests for examples.


ChadKillingsworth:
These changes also allow CommonJS modules to benefit from the early inlining pass for improved typing. On my projects I received new (and valid) warnings within the CommonJS modules.

ChadKillingsworth:
After doing substantial research into other transpilers and module interop, I no longer think this is the correct approach: http://www.2ality.com/2015/12/babel-commonjs.html

I'll issue a separate PR to add the `@const` annotations to CommonJS exports so that early inlining can be utilized.

ChadKillingsworth:
Replaced by #1348

PR #1343, opened 2015-12-28
Switch cases should end with an unconditional "break" statement

kirill-vlasov:
This pull request is focused on resolving occurrences of Sonar rule squid:S128 - Switch cases should end with an unconditional "break" statement
You can find more information about the issue here:
https://dev.eclipse.org/sonar/coding_rules#q=squid:S128
Please let me know if you have any questions.
Kirill Vlasov

### DIFF
--- a/src/com/google/javascript/jscomp/CollapseProperties.java
+++ b/src/com/google/javascript/jscomp/CollapseProperties.java
@@ -233,7 +233,7 @@ class CollapseProperties implements CompilerPass {
         for (Ref ref : refs) {
           switch (ref.type) {
             case SET_FROM_GLOBAL:
-              continue;
+              break;
             case DIRECT_GET:
             case ALIASING_GET:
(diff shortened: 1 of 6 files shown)


MatrixFrog:
I'm hesitant to do something a static analysis tool recommends, unless we're going to decide as a team that we want to use that tool on all future commits as well. Otherwise issues will keep coming up indefinitely and there will be a lot of code churn..

kirill-vlasov:
I signed it!

dimvar:
This looks fine to me, since it only changes some continues to breaks without changing semantics; it's not touching any fall-throughs.

In any case, let's not merge until after we resolve the Travis issue and sync the internal and external repos, they have diverged a bit over the holidays.

blickly:
I would consider this a shortcoming in the static analysis tool that it doesn't treat continue the same as return/throw/break. They all prevent unintended fallthrough.

PR #1344, opened 2015-12-28
"equals(Object obj)" should test argument type

kirill-vlasov:
This pull request is focused on resolving occurrences of Sonar rule squid:S2097 - "equals(Object obj)" should test argument type
You can find more information about the issue here:
https://dev.eclipse.org/sonar/coding_rules#q=squid:S2097
Please let me know if you have any questions.
Kirill Vlasov


kirill-vlasov:
I signed it!

dimvar:
This changes semantics. The equals method explicitly wants to fail if it gets an object from an unexpected class.

PR #1348, opened 2015-12-28
Add const annotations to CommonJS module exports

ChadKillingsworth:
To help get CommonJS to parity with ES6 modules, this changes the CommonJS module rewriting to add `@const` annotations if we can reasonably be certain that the export property was assigned exactly once.

### DIFF
--- a/src/com/google/javascript/jscomp/ProcessCommonJSModules.java
+++ b/src/com/google/javascript/jscomp/ProcessCommonJSModules.java
@@ -17,6 +17,8 @@ package com.google.javascript.jscomp;
 
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Iterables;
+import com.google.common.collect.HashMultiset;
+import com.google.common.collect.Multiset;
 import com.google.javascript.jscomp.NodeTraversal.AbstractPostOrderCallback;
 import com.google.javascript.jscomp.NodeTraversal.AbstractPreOrderCallback;
 import com.google.javascript.rhino.IR;
@@ -26,8 +28,6 @@ import com.google.javascript.rhino.Node;
 
 import java.net.URI;
 import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.Map;
 import java.util.List;
 
 /**
@@ -182,7 +182,7 @@ public final class ProcessCommonJSModules implements CompilerPass {
     private int scriptNodeCount = 0;
     private List<Node> moduleExportRefs = new ArrayList<>();
     private List<Node> exportRefs = new ArrayList<>();
-    Map<String, Integer> propertyExportRefCount = new HashMap<>();
+    Multiset<String> propertyExportRefCount = HashMultiset.create();
 
     @Override
     public void visit(NodeTraversal t, Node n, Node parent) {
@@ -289,12 +289,7 @@ public final class ProcessCommonJSModules implements CompilerPass {
         return;
       }
 
-      int assignCount = 1;
-      if (propertyExportRefCount.containsKey(qName)) {
-        assignCount = propertyExportRefCount.get(qName) + 1;
-        propertyExportRefCount.remove(qName);
-      }
-      propertyExportRefCount.put(qName, assignCount);
+      propertyExportRefCount.add(qName);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


ChadKillingsworth:
Merge conflicts resolved. Can this be reviewed now?

src/com/google/javascript/jscomp/ProcessCommonJSModules.java
```diff
@@ -250,10 +255,48 @@ public void visit(NodeTraversal t, Node n, Node parent) {
...
+    private void maybeAddReferenceCount(Node n) {
+      Node refParent = getBaseQualifiedNameNode(n);
+
+      if (refParent == null) {
+        return;
+      }
+
+      String qName = refParent.getQualifiedName();
+      if (qName.startsWith("module.exports.")) {
+        qName = qName.substring(15);
```
MatrixFrog: nit: use `"module.exports.".length` so it's clear where that number comes from (likewise 2 lines down)

ChadKillingsworth: Done

src/com/google/javascript/jscomp/ProcessCommonJSModules.java
```diff
@@ -377,57 +420,36 @@ private void processExports(Node script, String moduleName) {
...
-          }
+          addConstToObjLitKeys(rhsValue);
         }
 
         Node assign = ref.getParent();
-        Node exprResult = assign.getParent();
-        script.replaceChild(exprResult, newExprResult);
+        assign.replaceChild(ref, newName);
+        JSDocInfoBuilder builder = new JSDocInfoBuilder(true);
+        builder.recordConstancy();
```
MatrixFrog: I would expect to see these `@const` annotations in the tests, but I guess we don't have the compareJsDoc option enabled for this test? We should enable it but I don't think we need to block this PR on it.

src/com/google/javascript/jscomp/ProcessCommonJSModules.java
```diff
@@ -179,6 +182,7 @@ public boolean shouldTraverse(NodeTraversal nodeTraversal, Node n, Node parent)
     private int scriptNodeCount = 0;
     private List<Node> moduleExportRefs = new ArrayList<>();
     private List<Node> exportRefs = new ArrayList<>();
+    Map<String, Integer> propertyExportRefCount = new HashMap<>();
```
MatrixFrog: You can often replace a `Map<String, Integer>` with a `Multiset<Key>` and I think this is a case where that would work well, but I'm not sure. Take a look and if it switching to a Multiset is a lot of work then nevermind but if it turns out to be easy let's do it.

http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/collect/Multiset.html

MatrixFrog:
otherwise lgtm

src/com/google/javascript/jscomp/ProcessCommonJSModules.java
ChadKillingsworth: Simple change - I just don't use Guava much.

src/com/google/javascript/jscomp/ProcessCommonJSModules.java
ChadKillingsworth: It's definitely disabled - and enabling looks to be a non-trivial amount of work.

ChadKillingsworth:
Toggled to force travis to run.

src/com/google/javascript/jscomp/ProcessCommonJSModules.java
Dominator008: Maybe add a test to ProcessCommonJSModulesTest after #1395 is merged?

ChadKillingsworth: I actually think the existing tests will have to be updated and then it should be covered. I could pull #1395 into here, but they are based off different master commits so it makes a mess of the tree.

Dominator008: We can merge this first :)

PR #1350, opened 2015-12-29
Update args4j version to add support for the "forbids" option.

ChadKillingsworth:
This should simplify some logic in the CommandLineRunner.

I pulled in the [maven branch](https://github.com/MatrixFrog/closure-compiler/commit/123dfed7653c5aa15e7beeba0679a9621a76d9c1) from the fork by @MatrixFrog to test the maven builds.

I have no idea why GitHub is failing to display the diff.


MatrixFrog:
Using 'forbids' would still break the Ant build so we'll need to get rid of the Ant build (which I've been wanting to do anyway but can never seem to find the time for...)

ChadKillingsworth:
Why would it break the ant build? I tested it with the ant build because Maven is broken ...

MatrixFrog:
I assumed the Ant build was using an older version. Nevermind then.

ChadKillingsworth:
It uses the binary that's part of this commit.

ChadKillingsworth:
@MatrixFrog Can we merge this?

MatrixFrog:
Probably not, unfortunately. Let's discuss at the next meeting.

ChadKillingsworth:
Then I'm going to switch #1329 to throw or report errors. It's not as nice of a user experience, but without the "forbids" option that's what I'm stuck with.

PR #1353, opened 2015-12-30
Added FileReader state constants

vobruba-martin:
FileReader state constants should be defined not only in prototype but in constructor object itself.


vobruba-martin:
Sorry, something went wrong with my commit. It should have changed externs/browser/fileapi.js only...

PR #1354, opened 2015-12-30
Added FileReader state constants

vobruba-martin:
FileReader state constants should be defined not only in prototype but in constructor object itself.


ChadKillingsworth:
The spec doesn't state that the static versions of these constants should exist (at least to my reading). But they most definitely exist in Chrome.

vobruba-martin:
I think it is the same as `FileError` https://github.com/google/closure-compiler/blob/master/externs/browser/fileapi.js#L338

PR #1359, opened 2016-01-01
add missing @return tags for Array.prototype.sort() and Array.prototype.reverse()

adufilie:


adufilie:
I signed it.

externs/es3.js
```diff
@@ -513,6 +513,7 @@ Array.prototype.push = function(var_args) {};
  * Transposes the elements of an array in place: the first array element becomes the
  * last and the last becomes the first.
  *
+ * @return {!Array<T>}
```
Dominator008: This will need a "@template T"

externs/es3.js
```diff
@@ -513,6 +513,7 @@ Array.prototype.push = function(var_args) {};
...
  * @this {{length: number}}
  * @modifies {this}
  * @see http://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reverse
```
Dominator008: While we are here, can you fix this URL to be http://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice ?

Dominator008: While we are here, can you fix this URL to be http://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice ?

adufilie:
I've added the template T, but regarding the URLs they seem to be correct.

Dominator008:
@adufilie I misread the URL on my phone :(

dimvar:
FYI, these changes will probably require cleaning up code internally, and if the cleanup required ends up being big, the PR may get reverted.

Dominator008:
@dimvar Acknowledged.

blickly:
In order to get back to a passing state with both repositories in sync so we can rerelease, I'm reverting this change for now.

Dominator008:
@blickly Thanks!

blickly:
It looks like a good type-checking improvement, but unfortunately that makes it more difficult to land today.

PR #1362, opened 2016-01-04
Rewrite of jQueryAjaxSettings type to avoid circular typedef

vobruba-martin:
Current jQuery extern file shows warning "Circular typedefs/enums are not allowed" when using NTI.

### DIFF
--- a/contrib/externs/jquery-1.9.js
+++ b/contrib/externs/jquery-1.9.js
@@ -40,43 +40,109 @@ var jQuerySelector;
 /** @typedef {function(...)|Array<function(...)>} */
 var jQueryCallback;
 
-/** @typedef {
-              {
-               accepts: (Object<string, string>|undefined),
-               async: (?boolean|undefined),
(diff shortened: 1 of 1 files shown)


dimvar:
Can you switch @interface to @record? Also, can you leave a blank line before each property definition?

vobruba-martin:
Updated. Where can I find documentation for the @record tag? I'm using https://developers.google.com/closure/compiler/docs/js-for-compiler but it seems outdated.

blickly:
Currently we have a wiki page:
https://github.com/google/closure-compiler/wiki/Structural-Interfaces-in-Closure-Compiler

The hope is to eventually make @interface structural as well.

dimvar:
LGTM. Can you squash the commits before we merge the PR?

ChadKillingsworth:
Looks like every reference of `jQueryAjaxSettings` is a union with `Object<string, *>`, so this shouldn't break anything. Otherwise we would need more testing.

vobruba-martin:
@dimvar Is it ok now?

PR #1366, opened 2016-01-05
Suppress missing require from externs (reworked)

supersteves:
This is to fix the case where you have React defined globally in externs:
``` javascript
// Externs:

var React = {};

/**
 * @constructor
 * @param {Object} props
 */

// Main code:

class X extends React.Component {
}
```
Without this PR you get the following warning, which is incorrect since externs don't need to be required as they are global:
```
'React.Component' used but not goog.require'd
```

This is a rework of PR https://github.com/google/closure-compiler/pull/1192 which follows the behaviour of the rest of the class more closely. The check now happens before adding to usages, in visitClassNode, rather than in visitScriptNode, as per @MatrixFrog comment https://github.com/google/closure-compiler/pull/1192#issuecomment-146266314

### DIFF
--- a/src/com/google/javascript/jscomp/CheckRequiresForConstructors.java
+++ b/src/com/google/javascript/jscomp/CheckRequiresForConstructors.java
@@ -329,23 +329,22 @@ class CheckRequiresForConstructors implements HotSwapCompilerPass, NodeTraversal
     Node extendClass = classNode.getSecondChild();
     
     // If the superclass is something other than a qualified name, ignore it.
-    if (extendClass.isQualifiedName()) {
-      // Grab the root superclass namespace.
-      Node root = NodeUtil.getRootOfQualifiedName(extendClass);
-      
-      // It should always be a name. Extending this.something or 
-      // super.something is unlikely.
-      // We only consider programmer-defined superclasses that are
-      // global variables, or are defined on global variables.
-      if (root.isName()) {
-        String rootName = root.getString();
-        Var var = t.getScope().getVar(rootName);
-        if (var != null && (var.isLocal() || var.isExtern())) {
-          // "require" not needed for these
-        } else {
-          usages.put(extendClass.getQualifiedName(), extendClass);
-        }
-      }
+    if (!extendClass.isQualifiedName()) return;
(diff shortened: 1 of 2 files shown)


supersteves:
@blickly wrote in https://github.com/google/closure-compiler/pull/1192#issuecomment-146269906

> Looks live visitNewNode doesn't call maybeAddUsage

@MatrixFrog wrote in https://github.com/google/closure-compiler/pull/1192#issuecomment-146266314

> I thought the existing var.isExtern() check on line 406 was supposed to make this work. We should figure out why it's not.

In fact because this is an ES6 class, the problem was in ```visitClassNode```, which was unreservedly adding a usage for the extends clause.

Line 406 no longer exists, but I think this was ```maybeAddUsage``` which is specific to types mentioned in JSDoc, and not the extends clause.

This fix was done sympathetically to ```visitNewNode```; the unit test is unchanged from the original PR.

ChadKillingsworth:
This makes sense. I'm surprised I haven't hit this yet.

src/com/google/javascript/jscomp/CheckRequiresForConstructors.java
```diff
@@ -320,14 +320,32 @@ private void visitNewNode(NodeTraversal t, Node newNode) {
...
-  private void visitClassNode(Node classNode) {
+  private void visitClassNode(NodeTraversal t, Node classNode) {
     String name = NodeUtil.getName(classNode);
     if (name != null) {
       constructors.add(name);
     }
+    
     Node extendClass = classNode.getSecondChild();
+    
+    // If the superclass is something other than a qualified name, ignore it.
```
MatrixFrog: nit: Make the code structure follow this comment with an early return:

if (!extendClass.isQualifiedName()) {
  return;
}

// Grab the root superclass namespace.
Node root = ...

In fact then we can probably drop the comment too.

MatrixFrog:
lgtm other than that. Thank you.

supersteves:
@MatrixFrog Nits done, thanks

supersteves:
Thanks!

PR #1378, opened 2016-01-09
Allow TypeScript syntax in arrow function parameters and return types

Dominator008:
[Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1378)

### DIFF
--- a/src/com/google/javascript/jscomp/parsing/parser/Parser.java
+++ b/src/com/google/javascript/jscomp/parsing/parser/Parser.java
@@ -123,7 +123,9 @@ import com.google.javascript.jscomp.parsing.parser.util.Timer;
 import java.util.ArrayDeque;
 import java.util.EnumSet;
 import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Set;
 
 /**
  * Parses a javascript file.
@@ -172,6 +174,7 @@ public class Parser {
   private final CommentRecorder commentRecorder = new CommentRecorder();
   private final ArrayDeque<Boolean> inGeneratorContext = new ArrayDeque<>();
   private FeatureSet features = FeatureSet.ES3;
+  private final Set<Token> colonTypeTokens = new LinkedHashSet<>();
 
   public Parser(
       Config config, ErrorReporter errorReporter,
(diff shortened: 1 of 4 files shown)


Dominator008:
Not quite working yet :(

src/com/google/javascript/jscomp/parsing/parser/Parser.java
```diff
@@ -2620,16 +2620,29 @@ private boolean peekAssignmentExpression() {
...
+  }
+
+  private ParseTree parseAssignment(Expression expressionIn, boolean mayHaveColonType) {
     if (peek(TokenType.YIELD) && inGeneratorContext()) {
       return parseYield(expressionIn);
     }
 
     SourcePosition start = getTreeStartLocation();
-    // TODO(blickly): Allow TypeScript syntax in arrow function parameters
+    GenericTypeListTree generics = maybeParseGenericTypes();
```
blickly: Isn't this allowing parsing of generics outside arrow functions?

Let's add some tests that generics in random places are forbidden.
e.g.
var x  = <X>(5 + 7);

Dominator008: Done.

blickly: I think there may be a similar issue for COLON below.

Dominator008: Done.

blickly:
One more test case:

```javascript
function f(cond) {
  return cond ?
        (v = 7, f = ()=>v) :
        n=>2*n;
}
```

Can we check that this still parses correctly as an ?: (since this is valid non TS ES6 code).

Dominator008:
@blickly Done adding that test case, which is already parsed correctly.

Dominator008:
@blickly PTAL?

blickly:
Looks good. I'll pull it in.

Dominator008:
@blickly Do you have a test case for the parse error in `switch` statements?

blickly:
I think it was parsing the COLON of the case as a type expression.
```javascript
switch (EXPR1) {
   case EXPR2:
       break;
}
```
The particular `EXPR2` that was broken was a comma expression, I think, but since `EXPR2` can be an arbitrary expression, we probably need better test coverage of case expressions.

nicks:
lol, sorry, github's trying to be too smart

MatrixFrog:
Can you try sending this internally if you still want this?

PR #1380, opened 2016-01-09
CommonJS Module relocation feature

lgeorgieff:
Adds the CLI argument --relocate_common_js_module which takes a path to a module and a mapped path to a replacement for this module. Multiple occurrences are allowed. For example the option --relocate_common_js_module=./proj/libs/1.0.0/=./proj/libs/2.0.0/ will map the require statement "require('./libs/1.0.0/someModule')" to the folder with the version 2.0.0 in the file ./proj/src/main.js.

This allows to compile different versions of a project focused on the referenced module versions without modifying the source code or the file system. (The relocated files must be passed to the compiler, e.g. via --js).

In addition this can be used as workaround for node.js modules that are stored in a node_modules folder and are referenced without a correct relative part.

[Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1380)

### DIFF
--- a/src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
+++ b/src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
@@ -468,6 +468,7 @@ public abstract class AbstractCommandLineRunner<A extends Compiler,
     options.tracer = config.tracerMode;
     options.setNewTypeInference(config.useNewTypeInference);
     options.instrumentationTemplateFile = config.instrumentationTemplateFile;
+    options.commonJSModuleRelocations = config.relocateCommonJSModule;
   }
 
   protected final A getCompiler() {
@@ -2424,6 +2425,17 @@ public abstract class AbstractCommandLineRunner<A extends Compiler,
      */
     CommandLineConfig setModuleRoots(List<String> jsModuleRoots) {
       this.moduleRoots = jsModuleRoots;
+      return this;
+    }
+
+    private Map<String, String> relocateCommonJSModule;
+
+    CommandLineConfig setRelocateCommonJSModule(Map<String, String> moduleRelocations) {
+      if (moduleRelocations == null) {
+        this.relocateCommonJSModule = new HashMap<>();
+      } else {
+        this.relocateCommonJSModule = moduleRelocations;
+      }
       return this;
     }
 
--- a/src/com/google/javascript/jscomp/CommandLineRunner.java
+++ b/src/com/google/javascript/jscomp/CommandLineRunner.java
@@ -62,6 +62,7 @@ import java.nio.file.attribute.BasicFileAttributes;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
@@ -421,6 +422,18 @@ public class CommandLineRunner extends
       usage = "Path prefix to be removed from CommonJS module names."
     )
     private List<String> commonJsPathPrefix = new ArrayList<>();
+
+    @Option(
+      name = "--relocate_common_js_module",
+      hidden = true,
+      usage = "Mapping for required CommonJS modules. You may specify " +
+          "multiple. This allows the compiler to map the path of the " +
+          "required module to the given mapping and to load the mapped " +
+          "module instead of the original module, e.g. " +
+          "--relocate_common_js_module=./lib/1.0.0/module.js=" +
+          "./lib/2.0.0/module.js"
+    )
+    private Map<String, String> comonJSModuleRelocations = new HashMap<>();
 
     @Option(
       name = "--js_module_root",
@@ -1077,6 +1090,10 @@ public class CommandLineRunner extends
           ImmutableList.of(ES6ModuleLoader.toModuleName(URI.create(flags.commonJsEntryModule)));
     }
 
+    if (!flags.processCommonJsModules && flags.comonJSModuleRelocations.size() != 0) {
+      reportError("Error - --relocate_common_js_module must be used together with --process_common_js_modules");
+    }
+
     if (flags.outputWrapperFile != null && !flags.outputWrapperFile.isEmpty()) {
       flags.outputWrapper = "";
       try {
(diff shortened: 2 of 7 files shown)


ChadKillingsworth:
This is definitely a problem that needs solved, but I don't think this is the way we want to do it. Any solution needs to work equally well with both CommonJS and ES6 modules. See #1329 for how the flags are merging.

Also, particularly with Node, the compiler doesn't have to be the place this is solved. Mapping a `require` statement to the actual code is quite complex and there is no need to re-write this code in the compiler when we could leverage existing Node/NPM work.

Just as much as we need mapping ability, we need the ability to prevent the compiler from rewriting some `require` statements altogether. Namely, core NodeJS modules or modules that should be loaded at runtime.

We need a solid plan and direction before we add new flags - which would then have to be supported long term.

supersteves:
@ChadKillingsworth I also would like to know how to solve compiling of code that references node_modules without proper relative paths - in this case I want to compile them (rather than use the external require trick to avoid compilation and do runtime loading).

lgeorgieff:
OK I understand your comments.

1)
> Also, particularly with Node, the compiler doesn't have to be the place this is solved.
> Mapping a require statement to the actual code is quite complex and there is no need
> to re-write this code in the compiler when we could leverage existing Node/NPM work.
How would look like a proper solution from your point of view to solve this issue?

2)
I also found the design document https://gist.github.com/Kegsay/2cb999a78dbd2b8bda95. Is there any progress on this discussion? I would be willing to work on a potential solution for this problem.

In my opinion the closure compiler is a great tool. But for node-focused environments currently I have the feeling that a lot of workarounds have to be done until it works properly. So I am very motivated to help in improving this.

ChadKillingsworth:
We've made two critical improvements lately that help these issues:

 1. We have [officially distributed node/npm code](https://github.com/chadkillingsworth/closure-compiler-npm) which for the first time provides the option to solve some of these issues in JS rather than Java.
 2. Once #1329 lands, all 3 module/dependency systems can coexist easily in a single compilation.

**External Require Work**
I've created issue #1382 for this case. It's recognized as needing a flag/option, but no work has been done on it yet. If you want to help with this jump on in. I suggest posting the proposed flag name to the issue first then begin work on the implementation. All of my current work has been blockers to this, but that's almost complete.

**Mapping Non-Path Imports**
There have been several discussions regarding this via emails between long-time devs that work with Node. I briefly explored the option of simply passing an entry point to the compiler and letting it "discover" all of the related source files. However that path didn't receive consensus and the more I think through it isn't really compatible with the current architecture of the compiler.

I think the best steps forward are:

 1. Find or write a NodeJS package that will do the discovery work. There are multiple module packagers that have to do this work, so at least something should already exist. This is also the place where any rewriting for non-js sources should occur because `require` statements can import JSON.
 2. Decide on the best plan to pass these files into the compiler. The `--json_streams` flag already lets us provide arbitrary mappings between paths and source (and supports sourcemaps), so my thinking is we should just leverage that.

These steps are also free to jump in on. We should keep the PRs into the smallest possible steps. The discovery work should be PR'd to https://github.com/chadkillingsworth/closure-compiler-npm

ChadKillingsworth:
Also, I recommend writing a small proof-of-concept for some of these solutions and posting an issue. At that point I'll make sure all the appropriate devs see it and are given a chance to weigh in. Keep in mind that face-to-face meetings of the team occur on a bi-weekly schedule so depending on the solution it may take a couple of weeks before I can get the appropriate feedback.

supersteves:
From https://gist.github.com/Kegsay/2cb999a78dbd2b8bda95, the ability to locate node_modules automatically seemed reasonable but are you saying this is unlikely to happen in Java, i.e. requiring compilation of a node_modules based project to use a JS (gulp/etc.) based build process?

Presumably the only alternative is to have ```require("myNodeModule")``` to be edited manually and changed to ```require("node_modules/myNodeModule")```, putting yourself at the mercy of npm's recently changed rules for organising module locations?

Or to write my own compiler pass which implements node_modules location resolution and conversion to relative paths, at compile time? I'm quite happy and capable to do the latter, but more so if it is in the pipeline or likely to get accepted as a PR.

ChadKillingsworth:
> From https://gist.github.com/Kegsay/2cb999a78dbd2b8bda95, the ability to locate node_modules automatically seemed reasonable but are you saying this is unlikely to happen in Java, i.e. requiring compilation of a node_modules based project to use a JS (gulp/etc.) based build process?

Yes - at least any officially supported or recommended version. In practice it doesn't matter, but should not be part of the compiler.

> Presumably the only alternative is to have require("myNodeModule") to be edited manually and changed to require("node_modules/myNodeModule"), putting yourself at the mercy of npm's recently changed rules for organising module locations?

This can be automated as part of a build process. It's what I do now. There isn't a manual edit step.

> Or to write my own compiler pass which implements node_modules location resolution and conversion to relative paths, at compile time? I'm quite happy and capable to do the latter, but more so if it is in the pipeline or likely to get accepted as a PR.

This would only work in the most simplistic cases. The CommandLineRunner part of the compiler gathers all sources then begins compilation. Compilation is where parsing and module rewriting occur. Once the compilation process begins, the compiler drops sources, but doesn't add them. Changing that would require a large architectural shift and is not something that I see being accepted.

Also, the [node module location algorithm](https://nodejs.org/api/modules.html#modules_all_together) is far from trivial to duplicate and requires consuming `package.json` files.

supersteves:
So I think to retain our Java based build process in tandem with node_modules we'd be best rearranging our project to have all node_modules below the root dir, and to include all of them (with that only-closure-deps flag as true), and changing (manually or otherwise) all requires to use relative paths.

Incidentally how do you do that rewrite in your project? Some custom gulp stream task?

ChadKillingsworth:
> So I think to retain our Java based build process in tandem with node_modules we'd be best rearranging our project to have all node_modules below the root dir, and to include all of them (with that only-closure-deps flag as true), and changing (manually or otherwise) all requires to use relative paths.

Good luck. I don't see this as a maintainable strategy.

> Incidentally how do you do that rewrite in your project? Some custom gulp stream task?

There's not that much custom about it. It's not really that hard:

```js
var replace = require('gulp-replace');
/**
 * Move the node_module dependencies to the temp build folder
 * and correct the require statements to have the new path
 */
gulp.task('rewrite-node-deps', function () {
  return gulp.src('./node_modules/modulename/lib.js', {base: './node_modules'})
      .pipe(
      //Change the node/require style path references to be project specific for module imports
      replace(/require\((['"])([^'"]+)(['"])\)/g,
          'require($1./$2.js$3)'))
      .pipe(gulp.dest('./build'));
});
```

supersteves:
Thanks Chad.

ChadKillingsworth:
After spending some time looking, I believe the browserify tool pipeline handles the vast majority of this. The only thing needed would be to write a custom tool that uses the browserify api, removes the `pack` and `wrap` pipeline steps.

We would then simply need a way to indicate to the compiler the `require("foo")` should reference `require("/node_modules/foo/lib/foo.js")`.

dimvar:
@ChadKillingsworth can this PR be deleted?

ChadKillingsworth:
Closing this. We're still discussing the full design, but the features covered by this PR are in process.

PR #1385, opened 2016-01-12
Fix annotations in nodejs/fs.js extern file

maio:
These functions actually have side-effects.

Other extern files in nodejs directory look OK.

### DIFF
--- a/contrib/nodejs/fs.js
+++ b/contrib/nodejs/fs.js
@@ -356,7 +356,6 @@ fs.utimes;
  * @param {number|Date} atime
  * @param {number|Date} mtime
  * @return {void}
- * @nosideeffects
  */
 fs.utimesSync;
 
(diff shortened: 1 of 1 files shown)


dimvar:
I'm not familiar with these externs. I looked here:
https://nodejs.org/api/fs.html
and I couldn't tell why these functions have side effects. Can you briefly explain (or add a comment in the file) ?

maio:
Thanks for looking at this.

Side-effect is that these functions [change timestamp of a file on a file-system](https://nodejs.org/api/fs.html#fs_fs_utimes_path_atime_mtime_callback).

Also their return type is void (which is also defined in externs file), therefore there must be some side-effect otherwise there would be no point in calling such function. `@return {void}` together with `@nosideeffects` just doesn't make sense.

ChadKillingsworth:
This makes complete since. Thanks!

PR #1387, opened 2016-01-13
Preserve constancy and type annotation when tranforming loop closures

Dominator008:


test/com/google/javascript/jscomp/Es6RewriteBlockScopedDeclarationTest.java
```diff
@@ -408,6 +408,28 @@ public void testLoopClosure() {
...
+            "var $jscomp$loop$0 = {x: undefined};",
+            "for (;;$jscomp$loop$0 = {x: $jscomp$loop$0.x}) {",
+            "  /** @type {number} */ $jscomp$loop$0.x = 3;",
+            "  var f = function($jscomp$loop$0) {",
+            "    return function() { return $jscomp$loop$0.x}",
+            "  }($jscomp$loop$0);",
+            "}"));
+
+    // Preserve inline type annotation and constancy
+    test("for (;;){ const /** number */ x = 3; var f = function() { return x; } }",
```
MatrixFrog: Let's add one more test where x is a "var" -- I just want to make sure we won't accidentally try to remove something from letConsts that isn't in it.

Dominator008: vars don't need transpilation I guess :)

MatrixFrog: So then it would probably be a testSame()

Dominator008: Done.

MatrixFrog:
Looks like this fixes the handful of NPEs we were seeing. Thank you!

PR #1389, opened 2016-01-13
Remove XMLHttpRequest.prototype.mozResponseArrayBuffer

vobruba-martin:
This property was introduced in Firefox 4 (Gecko 2.0) and it was removed in Firefox 6 (Gecko 6.0). It has been removed from the [documentation] (https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Using_XMLHttpRequest) almost [4 years ago] (https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Using_XMLHttpRequest$compare?to=2362&from=2361).

### DIFF
--- a/externs/browser/html5.js
+++ b/externs/browser/html5.js
@@ -2402,6 +2402,9 @@ XMLHttpRequest.prototype.response;
  * Implemented as a draft spec in Firefox 4 as the way to get a requested array
  * buffer from an XMLHttpRequest.
  * @see https://developer.mozilla.org/En/Using_XMLHttpRequest#Receiving_binary_data_using_JavaScript_typed_arrays
+ * 
+ * This property is not used anymore and should be removed.
+ * @see https://github.com/google/closure-compiler/pull/1389
  */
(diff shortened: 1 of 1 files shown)


ChadKillingsworth:
It's possible that this is used internally at Google, but if not this is fine with me.

dimvar:
Yes unfortunately this property is used internally, and doing the cleanup required to land this PR is not a high priority. Can you change it and instead of removing the property, just add a comment in the jsdoc mentioning that it's not used anymore and should be removed?

MatrixFrog:
We could start by removing it from Closure, but I agree it's not a high priority, at least for the compiler team. https://github.com/google/closure-library/search?utf8=%E2%9C%93&q=mozResponseArrayBuffer

vobruba-martin:
I've added a comment to the property. Thanks

dimvar:
Let's leave it in this file for now, even moving it requires cleanups internally.

PR #1390, opened 2016-01-13
Ensure the source map file property references the full module name

ChadKillingsworth:
I fixed most of these instances when I added the `--json_streams` flag, however one case slipped by. When using `--module` flags, the source map `file` property did not reference the full module file name (it was missing the `.js` extension).

This property is used by 3rd party tools such as [Mozilla's JS Source Map Library](https://github.com/mozilla/source-map).


ChadKillingsworth:
I forgot that this is actually covered by #1266

PR #1398, opened 2016-01-14
Be less aggressive with default parameters in VariableReferenceCheck

### LINKED ISSUE #1397: ES6 function with namespaced default parameter fails to compile
```javascript
goog.provide('project.Dog');
project.Dog = class {};
var f = function(testing = new project.Dog()) {
  console.log(testing);
};
f();
```

This fails with the following error on line 3:
    Illegal variable reference before declaration: testing

Dominator008:
Be less aggressive with default parameters in VariableReferenceCheck

When checking default parameters, only report early reference error
if the name refers to a variable defined in the function body scope.

This will prevent reporting early reference error on variables
possibly defined in the outer scope.

Add test cases about undefined variables in default parameters to
VarCheckTest.

Fixes #1397

[Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1398)

### DIFF
--- a/test/com/google/javascript/jscomp/Es6VariableReferenceCheckTest.java
+++ b/test/com/google/javascript/jscomp/Es6VariableReferenceCheckTest.java
@@ -327,15 +327,20 @@ public final class Es6VariableReferenceCheckTest extends CompilerTestCase {
    * basic checks.
    */
   public void testDefaultParam() {
-    assertEarlyReferenceError("function f(x=a) {}");
     assertEarlyReferenceError("function f(x=a) { let a; }");
+    assertEarlyReferenceError(LINE_JOINER.join(
+        "function f(x=a) { let a; }",
+        "function g(x=1) { var a; }"));
     assertEarlyReferenceError("function f(x=a) { var a; }");
     assertEarlyReferenceError("function f(x=a()) { function a() {} }");
     assertEarlyReferenceError("function f(x=[a]) { var a; }");
     assertEarlyReferenceError("function f(x=y, y=2) {}");
+    assertNoWarning("function f(x=a) {}");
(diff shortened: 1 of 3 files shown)


test/com/google/javascript/jscomp/Es6VariableReferenceCheckTest.java
```diff
@@ -327,12 +327,15 @@ public void testObjectPattern_defaultValue() {
...
-    assertEarlyReferenceError("function f(x=a) {}");
     assertEarlyReferenceError("function f(x=a) { let a; }");
+    assertEarlyReferenceError(LINE_JOINER.join(
+        "function f(x=a) { let a; }",
+        "function g(x=1) { var a; }"));
     assertEarlyReferenceError("function f(x=a) { var a; }");
     assertEarlyReferenceError("function f(x=a()) { function a() {} }");
     assertEarlyReferenceError("function f(x=[a]) { var a; }");
     assertEarlyReferenceError("function f(x=y, y=2) {}");
+    assertNoWarning("function f(x=a) {}");
```
MatrixFrog: The VarCheck pass would give us a warning here right? Just not the Es6VariableReferenceCheck pass?

Dominator008: Right.

MatrixFrog:
Should we add a test that looks like the snippet in #1397 ? `function f(x = new foo.bar()) {}` where foo is defined, and then another where foo isn't defined, I guess.

Dominator008:
Done. Added test cases like `function f(x = new foo.bar()) {}`

alexeagle:
ping @MatrixFrog - we would like to get this change in, need it for angular2

Issue #1393, opened 2016-01-13
--generate_exports crashes sanityVarCheck when used with closurePass


Dominator008:
http://closure-compiler-debugger.appspot.com/#input0%3Dgoog.provide('test')%253B%250A%250A%252F**%250A%2520*%2520%2540constructor%250A%2520*%2520%2540export%250A%2520*%252F%250Atest%2520%253D%2520function()%2520%257B%257D%253B%26input1%26conformanceConfig%26externs%26refasterjs-template%26includeDefaultExterns%3D1%26CLOSURE_PASS%3D1%26GENERATE_EXPORTS%3D1

MatrixFrog:
Is this only if you don't include closure's base.js?

Dominator008:
Yes. Seems like --generate_exports requires base.js to be passed in as a source?
In that case, it would be good to remind the user instead of just crashing.

Dominator008:
Updated minimal test case.

solominsn:
Hello,
I've spent a lot of time to understand why my code  crash on building with new version of the compiler. The root cause of my problem is this bug. All works correctly with compiler 20180402.0.0, but next versions required only one function from base.js
```js
goog.exportProperty = function (object, publicName, symbol) {
  object[publicName] = symbol;
};
```
Suppose it can be done without base.js (user should add whole library when require only one function).
My workaround solution is create own base.js file with one declaration exportProperty function.

EatingW:
Created Google internal issue http://b/124762892

Aqours:
@MatrixFrog  How to include Closure Library in ES module?
```js
import '../node_modules/google-closure-library/closure/goog/base.js';
// other code
```
I will encounter some errors when I using above code.
```text
ERROR in type.js:345 (originally at ./node_modules/google-closure-library/closure/goog/base.js:213) from closure-compiler: Closure primitive methods (goog.provide, goog.require, goog.define, etc) must be called  at file scope.
goog.define('goog.DEBUG', true);
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

ERROR in type.js:368 (originally at ./node_modules/google-closure-library/closure/goog/base.js:236) from closure-compiler: Closure primitive methods (goog.provide, goog.require, goog.define, etc) must be called  at file scope.
goog.define('goog.LOCALE', 'en');  // default to en
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

[...]

PR #1408, opened 2016-01-20
Make sure CommandLineRunnerTest checks the exit code

Dominator008:
In CommandLineRunnerTest, check that the exit code is zero when a
compilation is expected to succeed.
    
Add "var goog;" to testGenerateExports() to avoid an exception.
    
GenerateExports expects base.js to be included. Specifically, the
variable "goog" should exist in the global scope.

### DIFF
--- a/test/com/google/javascript/jscomp/CommandLineRunnerTest.java
+++ b/test/com/google/javascript/jscomp/CommandLineRunnerTest.java
@@ -1300,8 +1300,8 @@ public final class CommandLineRunnerTest extends TestCase {
 
   public void testGenerateExports() {
     args.add("--generate_exports=true");
-    test("/** @export */ foo.prototype.x = function() {};",
-        "foo.prototype.x=function(){};" +
+    test("var goog; /** @export */ foo.prototype.x = function() {};",
+        "var goog; foo.prototype.x=function(){};" +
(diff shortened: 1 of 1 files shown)


Dominator008:
This is related to #1393

MatrixFrog:
Why was this passing before this change?

Dominator008:
Because a test() method in CommandLineRunnerTest wasn't checking the exit code.

test/com/google/javascript/jscomp/CommandLineRunnerTest.java
```diff
@@ -1678,6 +1678,9 @@ private void test(String[] original, String[] compiled,
                     DiagnosticType warning) {
     Compiler compiler = compile(original);
 
+    assertThat(exitCodes).isNotEmpty();
+    assertThat(exitCodes.get(exitCodes.size() - 1)).isEqualTo(0);
```
MatrixFrog: I would think that exitCodes would always have length 1 so then you could just write

`assertThat(exitCodes).containsExactly(0)`

(if I'm remembering the Truth methods right) but I may be totally wrong.

Dominator008: That's nifty. Done.

Dominator008:
@MatrixFrog This is ready.

PR #1415, opened 2016-01-20
Re-add the setClosureEntryPoints method

ChadKillingsworth:
setClosureEntryPoints is deprecated, but used internally at Google. This re-adds the method to AbstractCommandLineRunner to aid in migration to the new setEntryPoints signature.

### DIFF
--- a/src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
+++ b/src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
@@ -301,6 +301,34 @@ public abstract class AbstractCommandLineRunner<A extends Compiler,
     return null;
   }
 
+  /**
+   * Helper method to convert the manage closure dependecy options to the new
+   * DependencyMode enum value
+   */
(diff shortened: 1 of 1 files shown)


ChadKillingsworth:
@blickly I just noticed that the list is Immutable - so this won't quite work as-is.

ChadKillingsworth:
Ok that should do it.

blickly:
Thanks for the quick response!
I just noticed that we also have usages of the old createDependencyOptions, so now I'm thinking it might be easier to just have a couple helpers here:
  List<String> -> List<ModuleIdentifier>
and
  (boolean, boolean) -> DependencyMode
and then I can fix up the callers using those in our internal code.
I'm pretty sure this will be cleaner than trying to support all of the old API usages that we have internally.

ChadKillingsworth:
@blickly I think this is what you had in mind. Let me know if you want something different.

blickly:
Great, perfect! Thanks Chad

PR #1433, opened 2016-01-24
Allow code to @export a list of tokens.

mprobst:
E.g.: `/** @export */ ['a', 'b'];`

This is useful for applications that use a reflection based templating language
to tell JSCompiler which tokens are ever used in a template expression. The
intention is to e.g. generate a list of tokens from Angular templates as part of
the compilation process of an app.

[Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1433)

### DIFF
--- a/src/com/google/javascript/jscomp/FindExportableNodes.java
+++ b/src/com/google/javascript/jscomp/FindExportableNodes.java
@@ -21,7 +21,9 @@ import com.google.javascript.rhino.JSDocInfo;
 import com.google.javascript.rhino.Node;
 import com.google.javascript.rhino.Token;
 
+import java.util.ArrayList;
 import java.util.LinkedHashMap;
+import java.util.List;
 
 /**
  * Records all of the symbols and properties that should be exported.
@@ -50,6 +52,14 @@ class FindExportableNodes extends AbstractPostOrderCallback {
   static final DiagnosticType EXPORT_ANNOTATION_NOT_ALLOWED =
       DiagnosticType.error("JSC_EXPORT_ANNOTATION_NOT_ALLOWED",
           "@export is not supported on this expression.");
+
+  static final DiagnosticType EXPORT_ARRAY_LITERAL_NOT_EXPRESSION =
+      DiagnosticType.error("JSC_EXPORT_NOT_STRING_LITERAL",
+          "An @export array literal's value must be a standalone expression.");
+
+  static final DiagnosticType EXPORT_NOT_STRING_LITERAL =
+      DiagnosticType.error("JSC_EXPORT_NOT_STRING_LITERAL",
+          "All members of an @export array literal must be string literals.");
 
   private final AbstractCompiler compiler;
 
@@ -83,6 +93,7 @@ class FindExportableNodes extends AbstractPostOrderCallback {
       }
 
       String export = null;
+      List<String> exportsList = null;
       GenerateNodeContext context = null;
 
       switch (n.getType()) {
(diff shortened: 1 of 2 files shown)


mprobst:
@concavelenz I think this is not particularly Angular-specific, and thus better to keep with the `GenerateExports` functionality.

MatrixFrog:
I'm hesitant to expand the meaning of `@export` since we already have a lot of confusion about it.

The template is essentially code that is external to the JS the compiler is seeing. So maybe there's a way we can extract an externs file from the template showing the `@interface` the template expects, and then the code passed to the compiler can have an `@implements` annotation.

mprobst:
@MatrixFrog the problem is that Angular1 templates cannot be statically analyzed - it's not clear what the expected interface is. On top of that, templates can access transitive properties, e.g. `foo.bar.baz` - so we'd have to make not only the controller implement something, but potentially arbitrary other classes too. In particular for protobufs, that'd be tricky.

I think this code (`/** @export */ [tokens];`) falls in line with the semantics of `/** @export */ this.foo;`. Or would you know another feature that matches the semantics/meaning better?

Adding yet another flag is of course also possible, but OTOH JSCompiler has quite a bunch of flags already, and this matches both the use case and the semantics of the "allow non-global exports" field. Users would also want to enable these under the same circumstances, i.e. having one of the flags on would always imply having the other on, too.

Your call obviously, but formulating this as an extension of that behaviour makes more sense to me.

MatrixFrog:
Is this PR intended to help with Angular 1 or Angular 2?

mprobst:
Discussed offline: this is for Angular 1 users, Angular 2 doesn't have the problem. However Angular 2 is still a while off for GA, and will take even longer before everyone is updated, so this is still a useful fix for the time beign.

dimvar:
Is it possible to achieve the same thing without new syntax, by generating a bunch of `goog.exportSymbol` calls for the new symbols? If so, we would prefer that approach.

mprobst:
@dimvar: agreed, it'd be preferable to implement this in terms of the existing primitives. However I don't think that's possible, for the same reasons that having an `@interface` in externs will not work.

AngularJS 1.x templates look roughly like this:

```html
<div>
  <my-directive some-attr="value.in.attr" class="some.other.value">{{body expression}}</my-directive>
</div>
```

The problem is that you cannot statically tell whether `some.other.value` is just some value of a DOM attribute, or whether it will be evaluated (reflectively) as a JavaScript expression. You also cannot tell statically what the context of the template is, or what `some`, `other`, etc points to.

For `goog.exportSymbol`, `@interface` in externs, or `goog.exportProperty` to work, we'd need to know the exact contexts of all these expressions (AFAICT), which we don't.

concavelenz:
I don't think /** @export */ on an array is the right thing (also "goog.exportSymbol" is the wrong thing as well).   As it currently works, @export can be used like this:

(function() {
function throwThisAway(a) {};
var UNUSED = 1;
throwThisAway({
  /** @export */
  propName: UNUSED,
  ...
}))();

The key here is an object literal with the property names you care about with an @export annotation, for the rest there are infinite variations.

For reference:

https://github.com/google/closure-compiler/blob/80084bd8b2f91b2acb21814bb97f4f68f391d958/src/com/google/javascript/jscomp/FindExportableNodes.java#L142

For the way we currently handle exports, this is sufficient.  We are unlikely to have enough context on Object to do "better" in a way that will break this.

Will this work?

mprobst:
@concavelenz I think that could work. However I'm a bit uneasy on the *unlikely to break* part. It seems better to me to explicitly convey intent with `/** @export */ ['tokens'];` than relying on some side effect of an unused declaration together with an empty function call. Your call though.

dimvar:
OK, let's do John's solution. The problem with `goog.exportSymbol` isn't the dynamism of the created names (you need to know these names to create any of the other forms anyway), but `goog.exportSymbol` won't work because it's not purely static, it changes the runtime behavior (by adding some properties to `window`).

mprobst:
OK, let's see if John's proposed code actually works, and keep this open for the time being.

ChadKillingsworth:
I had hoped to one day add an `@norename` annotation. However doing so is not trivial and impacts a large number of passes. In theory it would allow renaming to be blocked in cases like this without impacting disambiguation for all similarly named properties.

This is very low on my list however.

mprobst:
chad: I think `@export` on a local property is pretty close. You also need
both no renaming and no dead code removal, which is not quite the same
thing.

Chad Killingsworth <notifications@github.com> schrieb am Mo., 25. Jan. 2016
um 16:55 Uhr:

ChadKillingsworth:
In my plans, it was. There is already a mechanism in place to block dead code elimination. It's used for the suspicious code check. The compiler adds a sink value early in the process for any symbols and removes it at the very end.

Also, `@norename` would imply `@nocollapse`

concavelenz:
We decided not to do this. Let me know if you feel there is something to still discuss.

PR #1438, opened 2016-01-25
Be more strict in es3.js extern

vobruba-martin:
Is there some internal reason for `{*}` annotations or am I missing something?

When I compare `String.prototype.match` and `String.prototype.search` methods. I would say that `regexp` param should be the same type for both but it isn't.

And I expect this code to throw some warning:

``` javascript
var r = new RegExp();
r.test([]);
```

[Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1438)

### DIFF
--- a/externs/es3.js
+++ b/externs/es3.js
@@ -1730,7 +1730,7 @@ String.prototype.localeCompare = function(compareString, locales, options) {};
  * expression.
  *
  * @this {String|string}
- * @param {*} regexp
+ * @param {RegExp|string} regexp
  * @return {Array<string>} This should really return an Array with a few
  *     special properties, but we do not have a good way to model this in
  *     our type system. Also see Regexp.prototype.exec.
@@ -1792,7 +1792,7 @@ String.prototype.small = function() {};
 
 /**
  * @this {String|string}
- * @param {*=} opt_separator
+ * @param {(RegExp|string)=} opt_separator
  * @param {number=} opt_limit
  * @return {!Array<string>}
  * @nosideeffects
@@ -1906,8 +1906,8 @@ String.prototype.length;
 
 /**
  * @constructor
- * @param {*=} opt_pattern
- * @param {*=} opt_flags
+ * @param {(RegExp|string)=} opt_pattern
+ * @param {string=} opt_flags
  * @return {!RegExp}
  * @nosideeffects
  * @see http://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp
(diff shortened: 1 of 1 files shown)


MatrixFrog:
Most likely there is some code somewhere within Google that is inadvertently taking advantage of the looser types, and thus would break if we land this. We should bring it in and give it a try though, I would be happy to be wrong.

dimvar:
I was gonna add that generally making changes to these externs files always finds projects that don't follow the types. So, it's hard to land these changes, because the cleanup required (for internal projects) isn't very high priority.

concavelenz:
We should clean these up.  We have to a couple like this so we setup a script to add implicit casts to fix these up (Math.* have similar issues).

externs/es3.js
```diff
@@ -1906,8 +1906,8 @@ String.prototype.length;
 
 /**
  * @constructor
- * @param {*=} opt_pattern
- * @param {*=} opt_flags
+ * @param {(RegExp|string)=} opt_pattern
+ * @param {string=} opt_flags
```
MatrixFrog: these changes don't work, unfortunately. The type of RegExp is hard-coded in the compiler, and since some people use the head externs with the released compiler we can't easily update both at once. There is a way to do this (I don't remember what it is), but for now I'll drop this and apply the rest of the change.

vobruba-martin:
Do you want me to drop RegExp prototype changes?

MatrixFrog:
It will probably be a while before we have time to really push this through anyway, to be honest.

brad4d:
I'll take this since @MatrixFrog left the team.

imported, resolved conflict
running internal tests
I expect there will be problems landing this, though.

vobruba-martin:
Let me know if there is something I can help with!

brad4d:
If I remove the update to the `RegExp` constructor itself, there seem to be just a handful of errors.
I'm doing a test to see if that can be submitted.

brad4d:
FYI, I'm spending a bit more time on testing for this today.

brad4d:
I did a sampling test run (1000 test targets) in Google's internal code base on
a version of this PR with the especially-tricky changes to the Regexp constructor type removed.
It turned up 13 unique errors in just that very small sample.

I'm afraid that the cost of landing this PR greatly outweighs the benefit of landing it AFAICT,
so I'm going to have to reject it.

Here's what I can suggest, though.
We might be able to get there eventually by only modifying one method at a time to keep the
stuff that needs to be fixed to a more manageable level.

In my sample run all of the errors were with `Regexp.prototype.test` and `Regexp.prototype.exec`.
If you did a PR to, for example, just update the type info for `String.prototype.match`, that might be possible to land without too much difficulty.

PR #1457, opened 2016-01-27
Remove angular-specific references from jasmine-2.0 externs

ChadKillingsworth:
This is a roll-forward of #1333. It was rolled back right before the holidays but was never re-addressed.

Will require fixup internal to Google for projects using the jasmine-2.0 externs file with angular.

[Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1457)

### DIFF
--- a/contrib/externs/angular-jasmine.js
+++ b/contrib/externs/angular-jasmine.js
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2015 The Closure Compiler Authors.
+ *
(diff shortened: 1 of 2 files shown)


dimvar:
This may require cleanup internally. I'll bring it in and test it before merging.

Dominator008:
Closed by https://github.com/google/closure-compiler/commit/1c3245607a8c2bce83a68722403b468ce2684bac

PR #1463, opened 2016-01-28
Add missing ValidityState properties

vobruba-martin:
Here is the [specification](https://www.w3.org/TR/html5/forms.html#validitystate)


ChadKillingsworth:
lgtm

PR #1467, opened 2016-01-29
Added missing HTMLFieldSetElement properties

vobruba-martin:
[This commit](https://github.com/google/closure-compiler/commit/f98608a07dc43fb3e950a730ca59906179386ffb) added `disabled` property for `HTMLFieldSetElement` but this property is not in [DOM2 specification](https://www.w3.org/TR/DOM-Level-2-HTML/html.html#ID-7365882). I've moved it into `html5.js` extern and added other missing properties according to the [HTML5 specification](https://www.w3.org/TR/html5/forms.html#the-fieldset-element).

[Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1467)

### DIFF
--- a/externs/browser/html5.js
+++ b/externs/browser/html5.js
@@ -3426,3 +3426,53 @@ HTMLTemplateElement.prototype.content;
  * @see http://www.w3.org/TR/html-imports/#interface-import
  */
 HTMLLinkElement.prototype.import;
+
+
+/**
+ * @return {boolean}
+ * @see https://www.w3.org/TR/html5/forms.html#dom-fieldset-elements
+ */
+HTMLFieldSetElement.prototype.checkValidity = function() {};
+
+/**
+ * @type {HTMLCollection}
+ * @see https://www.w3.org/TR/html5/forms.html#dom-fieldset-elements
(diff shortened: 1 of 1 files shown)


MatrixFrog:
Things are all over the place with our externs, honestly. Unless someone is going to take on a huge project to reorganize them, I don't think this is worth it.

vobruba-martin:
Why do you think that adding missing properties isn't worth it? Or you want these properties to go under the `w3c_dom2.js` extern?

MatrixFrog:
Adding missing properties is fine but moving things from one file another is just churn.

vobruba-martin:
So, if I move these new properties under the `w3c_dom2.js` extern, will you merge it?

dimvar:
> So, if I move these new properties under the w3c_dom2.js extern, will you merge it?

We'll try. Can you move the definition of HTMLFieldSetElement.prototype.disabled back to its original place?

vobruba-martin:
Updated

dimvar:
AFAICT, there are no custom externs defined for these properties internally, so this should be safe.

Issue #1471, opened 2016-01-30
Update readme to say guava 19


supersteves:
I believe guava was updated to v19, readme still says 18.

Dominator008:
@supersteves Care to send a PR?

supersteves:
@Dominator008 (ahem) sure... I just tried to and found that pom-main has 19 and pom-gwt has 20, is this a problem, or should I document both?

ChadKillingsworth:
Just list both (comma separated) in the version field.

Dominator008:
@supersteves Once Guava 20 is released we will switch both to 20.

PR #1473, opened 2016-01-30
Updating readme to reference Guava v 19 and 20 to match pom-main and …

supersteves:
…pom-gwt

Fixes https://github.com/google/closure-compiler/issues/1471


Dominator008:
Thanks!

PR #1484, opened 2016-02-01
Revert "add missing @return tags for Array.prototype.sort() and Array.prototype.reverse()"

blickly:
Reverts google/closure-compiler#1359


MatrixFrog:
yeah, not surprising :(

PR #1486, opened 2016-02-02
Fix Maven build and test

Dominator008:
"mvn test" is failing. We cannot use both Guava 19.0 and 20.0-SNAPSHOT after
https://github.com/google/guava/commit/1ef638f369964c1d1f56bd0ae588535c6abd99a4
See https://github.com/google/guava/issues/2380.

Fix by bumping Guava to 20.0-SNAPSHOT in pom-main.xml.

Remove reference to Guava 19 from README.md.

Remove conflicting import statement and duplicate testGlobJs*() methods
from CommandLineRunnerTest.java to fix the Maven build.

### DIFF
--- a/README.md
+++ b/README.md
@@ -241,7 +241,7 @@ options/arguments in your CUI application.</td>
 
   <tr>
     <td>Version</td>
-    <td>19.0, 20.0</td>
+    <td>20.0</td>
   </tr>
 
   <tr>
--- a/pom-main.xml
+++ b/pom-main.xml
@@ -70,7 +70,7 @@
     <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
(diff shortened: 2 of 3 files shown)


MatrixFrog:
Let's not "fix" it by just not compiling. If 20.0 is available maybe that will work.

Dominator008:
@MatrixFrog The GWT compiler will still be invoke anyway with "mvn".  just disabled the default maven compile / test which uses the java compiler and does the same work as in pom-main.xml.

But yeah, let's hope Guava will fix this soon.

Dominator008:
@MatrixFrog Ok seems like we have to use Guava 20.0-SNAPSHOT in both pom-main.xml and pom-gwt.xml for now.

MatrixFrog:
ok lgtm once travis is done

PR #1488, opened 2016-02-03
Fix scanner for Unicode point escape with curly braces

kang-y:
Right curly braces "}" was unconditionally parsed as the same token if there is a Unicode escape string before it. For example, in the following JS code:

`if(foo){foo=\u03b5}else bar()`

"\u03b5}" was parsed as a single token. In this pull request, my code checks if there is already a left curly brace "{" in this token (and if this token contains Unicode escape string) before appending "}".

[Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1488)

### DIFF
--- a/src/com/google/javascript/jscomp/parsing/parser/Scanner.java
+++ b/src/com/google/javascript/jscomp/parsing/parser/Scanner.java
@@ -651,15 +651,32 @@ public class Scanner {
     valueBuilder.append(ch);
 
     boolean containsUnicodeEscape = ch == '\\';
+    boolean bracedUnicodeEscape = false;
+    int unicodeEscapeLen = containsUnicodeEscape ? 1 : 0;
 
     ch = peekChar();
     while (isIdentifierPart(ch)
         || ch == '\\'
-        || (ch == '{' && containsUnicodeEscape)
-        || (ch == '}' && containsUnicodeEscape)) {
+        || (ch == '{' && unicodeEscapeLen == 2)
+        || (ch == '}' && bracedUnicodeEscape)) {
       if (ch == '\\') {
         containsUnicodeEscape = true;
       }
+      // Update length of current Unicode escape.
+      if (ch == '\\' || unicodeEscapeLen > 0) {
+        unicodeEscapeLen ++;
+      }
+      // Enter Unicode point escape.
+      if (ch == '{') {
+        bracedUnicodeEscape = true;
+      }
+      // Exit Unicode escape
+      if (ch == '}' || (unicodeEscapeLen >= 6 && !bracedUnicodeEscape)) {
+        bracedUnicodeEscape = false;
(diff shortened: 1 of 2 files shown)


dimvar:
Can you add a test for `var \u{0043}}}};` as well? I think the current PR probably doesn't handle it.

kang-y:
OK. I'll find a spare time and do it.

kang-y:
@dimvar It works for `var \u{0043}}}}` now.

src/com/google/javascript/jscomp/parsing/parser/Scanner.java
```diff
@@ -651,15 +651,32 @@ private Token scanIdentifierOrKeyword(int beginToken, char ch) {
     valueBuilder.append(ch);
 
     boolean containsUnicodeEscape = ch == '\\';
+    boolean bracedUnicodeEscape = false;
+    int unicodeEscapeLen = (ch == '\\' ? 1 : 0);
```
Dominator008:
````
int unicodeEscapeLen = containsUnicodeEscape ? 1 : 0;
````

File: src/com/google/javascript/jscomp/parsing/parser/Scanner.java [hunk 2]
```diff
@@ -651,15 +651,32 @@ private Token scanIdentifierOrKeyword(int beginToken, char ch) {
     valueBuilder.append(ch);
 
     boolean containsUnicodeEscape = ch == '\\';
```
Dominator008: Actually, `containsUnicodeEscape` is unused now, so I guess you can just inline it?

src/com/google/javascript/jscomp/parsing/parser/Scanner.java
```diff
@@ -651,15 +651,32 @@ private Token scanIdentifierOrKeyword(int beginToken, char ch) {
...
 
     ch = peekChar();
     while (isIdentifierPart(ch)
         || ch == '\\'
-        || (ch == '{' && containsUnicodeEscape)
-        || (ch == '}' && containsUnicodeEscape)) {
+        || (ch == '{' && unicodeEscapeLen == 2)
+        || (ch == '}' && bracedUnicodeEscape)) {
       if (ch == '\\') {
         containsUnicodeEscape = true;
```
Dominator008: This if statement is unused now.

src/com/google/javascript/jscomp/parsing/parser/Scanner.java
kang-y: It's used below, on line 687.

Dominator008: My bad :( Confused by Github UI..

dimvar:
Looks good. Can you squash the commits into one? Then we can merge this.

PR #1369, opened 2016-01-06
TypedCodeGenerator temp fix for crash when printing Closure Library

supersteves:
Temporary fix and demonstrative unit test for ```TypedCodeGenerator.getFunctionAnnotation()``` Preconditions failure in ```NodeUtil.getFunctionParameters(fnNode)``` when code printing e.g.
``` javascript
goog.removeHashCode = goog.removeUid;
```
from ```base.js``` in Closure Library.

See 4 examples in code below.

I suspect there is some other proper fix that may be apparent to whoever introduced this problem, but this highlights the problem and works around it by printing without type annotations for these cases. I appreciate this PR probably won't be accepted in full but may be in part, so I've submitted as a PR rather than an issue.

This used to work in around June-September 2015 and was noticed to fail in Jan 2016.

PR #1490, opened 2016-02-03
Miscs improvements to TypedCodeGenerator

Dominator008:
Addresses several crashes and failures mentioned in #1369.


blickly:
Note that there are two ways of printing out JSDoc in the output. This is one, but the other way (which is actually used more often and thus better tested) is the regular CodeGenerator with the preserveTypeAnnotaitons option set to true.

That said, I don't see any reason not to merge this.

Dominator008:
@blickly Cool, thanks!

PR #1497, opened 2016-02-04
Add externs for angular.$locale service

Joel-Kornbluh:
[Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1497)

### DIFF
--- a/contrib/externs/angular-1.5.js
+++ b/contrib/externs/angular-1.5.js
@@ -22,7 +22,6 @@
  *     $cookieStore
  *     $document
  *     $httpBackend
- *     $locale
  *     $rootElement
  *     $rootScope
  *     $rootScopeProvider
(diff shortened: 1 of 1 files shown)


contrib/externs/angular-1.5.js
```diff
@@ -2468,3 +2467,41 @@ angular.$timeout_.cancel = function(promise) {};
...
 /** @typedef {!Window} */
 angular.$window;
+
+/******************************************************************************
+ * $locale Service
+ *****************************************************************************/
+
+ /**
+ * @typedef {{
+ *   DATETIME_FORMATS: {
```
mprobst: According to the Angular docs, the only public API is the id field:
https://docs.angularjs.org/api/ng/service/$locale

Do you really need all those nested fields? They are supposed to be an implementation detail.

```js
/** @interface */
angular.$locale = function() {};
/** @type {string} */
angular.$locale.prototype.id;
```

ChadKillingsworth: I think a better question is do we want to encourage their use by including it in official externs?

mprobst: We don't :-)

Dominator008:
Per the comment by @mprobst, these should not be included in the externs.

PR #1499, opened 2016-02-05
New lint check: CheckUselessBlocks

### LINKED ISSUE #1458: Warn about "useless blocks"
Inspired by some conversations with @concavelenz and tweets by @cramforce, we should warn about an unnecessary use of a block. A block is considered useful if it is part of an if / else / while / switch / etc. or if it contains any block-scoped variables (let, const, class, or function declarations). Otherwise there is no reason to use it so it is likely a mistake. This would catch the classic error

```javascript
return
    {foo: 'bar'};
```
as well as the contrived but plausible "backdoor" example at https://twitter.com/cramforce/status/692197974957527040

```javascript
if (denied) {
  showAccessDenied()
} {
  grantAccess()
}
```

It might also catch some cases where a user expected a `var`-declared variable to go out of scope because they mistakenly believed that `var`-declared variables are block-scoped.

Dominator008:
Check for useless blocks. A block is considered useful if it is part of a
control structure like if / else / while / switch / etc. or if it contains
any block-scoped variables (let, const, class, or function declarations).
Otherwise there is no reason to use it so it is likely a mistake. This would
catch the classic error:

```javascript
return
  {foo: 'bar'};
```

or more contrived cases like:

```javascript
if (denied) {
  showAccessDenied();
} {
  grantAccess();
}
```

Inspired by ESLint (https://github.com/eslint/eslint/blob/master/lib/rules/no-lone-blocks.js)

Fixes #1458.

[Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1499)

### DIFF
--- a/src/com/google/javascript/jscomp/lint/CheckUselessBlocks.java
+++ b/src/com/google/javascript/jscomp/lint/CheckUselessBlocks.java
@@ -1,0 +1,131 @@
+/*
+ * Copyright 2016 The Closure Compiler Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.javascript.jscomp.lint;
+
+import com.google.javascript.jscomp.AbstractCompiler;
+import com.google.javascript.jscomp.DiagnosticType;
+import com.google.javascript.jscomp.HotSwapCompilerPass;
+import com.google.javascript.jscomp.NodeTraversal;
+import com.google.javascript.jscomp.NodeTraversal.Callback;
+import com.google.javascript.jscomp.NodeUtil;
+import com.google.javascript.rhino.Node;
+import com.google.javascript.rhino.Token;
+
+import java.util.ArrayDeque;
+import java.util.Deque;
+
+/**
+ * Check for useless blocks. A block is considered useful if it is part of a
+ * control structure like if / else / while / switch / etc. or if it contains
+ * any block-scoped variables (let, const, class, or function declarations).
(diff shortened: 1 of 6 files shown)


test/com/google/javascript/jscomp/lint/CheckUselessBlocksTest.java
```diff
@@ -0,0 +1,76 @@
...
+    // TODO(moz): For block-scoped function declaration, we should technically
+    // warn if we are in non-strict mode and the language mode is ES5 or below.
+    testSame("{ function foo() {} }");
+    testSameEs6("let x = 1;");
+    testSameEs6("{ let x = 1; }");
+    testSameEs6("if (true) { let x = 1; }");
+    testSameEs6("{ const y = 1; }");
+    testSameEs6("{ class Foo {} }");
+  }
+
```
blickly: What's your intention for "useless" blocks in case statements?

e.g.
switch (x) {
  case 1: {
    return 5;
  }
}

I don't have a strong preference, but let's at least add a test case to document the behavior.

src/com/google/javascript/jscomp/lint/CheckUselessBlocks.java
```diff
@@ -0,0 +1,131 @@
...
+  public final boolean shouldTraverse(NodeTraversal t, Node n, Node parent) {
+    switch (n.getType()) {
+      case Token.BLOCK:
+        if (isLoneBlock(n)) {
+          loneBlocks.push(n);
+        }
+        break;
+      case Token.LET:
+      case Token.CONST:
+        allowLoneBlock(n);
```
blickly: I think it would be simpler to change these to:
  allowLoneBlock(parent)

MatrixFrog:
What Ben said. lgtm other than that.

src/com/google/javascript/jscomp/lint/CheckUselessBlocks.java
Dominator008: Done.

test/com/google/javascript/jscomp/lint/CheckUselessBlocksTest.java
Dominator008: I won't consider it useless for better readability. The check won't consider it useless either because the BLOCK is under a CASE node :) Added a test to document the behavior.

Dominator008:
Reviewed 4 of 6 files at r1, 2 of 2 files at r2.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/google/closure-compiler/1499)*

PR #1500, opened 2016-02-05
Add @SafeVarargs to methods that cause warnings

Dominator008:
This removes multiple warnings about potentially unsafe operations
on varargs.

[Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1500)


MatrixFrog:
What do these warnings look like and where do they come from?

Dominator008:
I'm using Eclipse. At the compileFiles() method declaration:
"Type safety: Potential heap pollution via varargs parameter entries"

At the various compileFiles() call sites:
"Type safety: A generic array of AbstractCommandLineRunner.FlagEntry<AbstractCommandLineRunner.JsSourceType> is created for a varargs parameter"

MatrixFrog:
I've read a few stackoverflow answers and I don't quite get what this warning is about but it seems clear we're not doing anything weird so this seems fine.

PR #1510, opened 2016-02-09
New lint check: CheckDuplicateCase

Dominator008:
Check for duplicate case labels in a switch statement
Eg:
```javascript
switch (foo) {
  case 1:
  case 1:
}
```
This is normally an indication of a programmer error.

Inspired by ESLint (https://github.com/eslint/eslint/blob/master/lib/rules/no-duplicate-case.js)

[Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1510)

### DIFF
--- a/src/com/google/javascript/jscomp/lint/CheckDuplicateCase.java
+++ b/src/com/google/javascript/jscomp/lint/CheckDuplicateCase.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2016 The Closure Compiler Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.javascript.jscomp.lint;
+
+import com.google.javascript.jscomp.AbstractCompiler;
+import com.google.javascript.jscomp.DiagnosticType;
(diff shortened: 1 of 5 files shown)


ChadKillingsworth:
Why defer this to a lint check? Shouldn't this be a suspicious code type warning? Where would this ever be valid?

Dominator008:
@ChadKillingsworth I don't know. People might put weird logic in their code.

@MatrixFrog What do you think?

ChadKillingsworth:
According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch a duplicate label would never be matched - only the first case would be. It would only be available via fall-through.

I think the likelihood of this being an error is very high and a warning would be appropriate. As long as it is suppressible.

Dominator008:
@ChadKillingsworth I don't have a preference. I can move this into CheckSuspiciousCode but there is a possibility that some internal code will be broken  :)

MatrixFrog:
Let's keep it as a lint check so it won't break anyone, with a TODO to make it part of the suspicious-code group.

MatrixFrog:
otherwise lgtm

Dominator008:
@MatrixFrog Done.

MatrixFrog:
I'll bring this in internally first since it will require a little bit of BUILD changes.

src/com/google/javascript/jscomp/lint/CheckDuplicateCase.java
```diff
@@ -0,0 +1,77 @@
...
+  @Override
+  public void hotSwapScript(Node scriptRoot, Node originalRoot) {
+    NodeTraversal.traverseEs6(compiler, scriptRoot, this);
+  }
+
+  @Override
+  public void visit(NodeTraversal t, Node n, Node parent) {
+    if (n.isSwitch()) {
+      Set<String> cases = new HashSet<>();
+      for (Node curr = n.getSecondChild(); curr != null; curr = curr.getNext()) {
```
Dominator008: @MatrixFrog Just an FYI, I changed this from getFirstChild().getNext() to getSecondChild().

MatrixFrog:
This is landed internally and will be MOE-pushed out soon (or it already was maybe?)

Issue #1396, opened 2016-01-14
Turn compareJsDoc on in all unit tests


Dominator008:
`test/com/google/javascript/jscomp/AmbiguatePropertiesTest.java:    compareJsDoc = false;
test/com/google/javascript/jscomp/AngularPassTest.java:    compareJsDoc = false;
test/com/google/javascript/jscomp/CheckAccessControlsTest.java:    compareJsDoc = false;
test/com/google/javascript/jscomp/CheckAccessControlsTest.java:    compareJsDoc = false;
test/com/google/javascript/jscomp/CheckAccessControlsTest.java:    compareJsDoc = false;
test/com/google/javascript/jscomp/CollapsePropertiesTest.java:    compareJsDoc = false;
test/com/google/javascript/jscomp/CollapseVariableDeclarationsTest.java:    compareJsDoc = false;
test/com/google/javascript/jscomp/DevirtualizePrototypeMethodsTest.java:    compareJsDoc = false;
test/com/google/javascript/jscomp/Es6InlineTypesTest.java:    compareJsDoc = false;
test/com/google/javascript/jscomp/GenerateExportsTest.java:    compareJsDoc = false;
test/com/google/javascript/jscomp/InferJSDocInfoTest.java:    compareJsDoc = false;
test/com/google/javascript/jscomp/InlinePropertiesTest.java:    compareJsDoc = false;
test/com/google/javascript/jscomp/InlineVariablesTest.java:    compareJsDoc = false;
test/com/google/javascript/jscomp/JsdocToEs6TypedConverterTest.java:    compareJsDoc = false;
test/com/google/javascript/jscomp/NormalizeTest.java:    compareJsDoc = false;
test/com/google/javascript/jscomp/OptimizeParametersTest.java:    compareJsDoc = false;
test/com/google/javascript/jscomp/ProcessDefinesTest.java:    compareJsDoc = false;
test/com/google/javascript/jscomp/RemoveUnusedVarsTest.java:    compareJsDoc = false;
test/com/google/javascript/jscomp/RenameVarsTest.java:    compareJsDoc = false;
test/com/google/javascript/jscomp/ReplaceIdGeneratorsTest.java:    compareJsDoc = false;
test/com/google/javascript/jscomp/ReplaceMessagesForChromeTest.java:    compareJsDoc = false;
test/com/google/javascript/jscomp/ReplaceMessagesTest.java:    compareJsDoc = false;
[...]

PR #1511, opened 2016-02-10
Turn on compareJsDoc in InferJSDocInfoTest

Dominator008:
This requires fixing JSDocInfo.areEquivalent().

Part of #1396.

[Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1511)

### DIFF
--- a/src/com/google/javascript/rhino/JSDocInfo.java
+++ b/src/com/google/javascript/rhino/JSDocInfo.java
@@ -51,6 +51,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


src/com/google/javascript/rhino/JSDocInfo.java
```diff
@@ -223,6 +224,21 @@ private int getMaskForBitIndex(int bitIndex) {
    * with a string.
    */
   public static class StringPosition extends SourcePosition<String> {
+    static boolean areEquivalent(StringPosition p1, StringPosition p2) {
```
MatrixFrog: Why static areEquivalent methods instead of a non-static equals()? Then it seems like we could have the equals() method in SourcePosition and not have to duplicate as much.

Dominator008: That would work for SourcePosition<String> but not SourcePosition<Node>. We have to call node1.isEquivalentTo(node2) instead of equals() on Node because the source files are different :(

MatrixFrog:
Hm, okay.

PR #1514, opened 2016-02-10
Simplify Array<T>|IArrayLike<T> to IArrayLike<T> in externs/es6.js

Dominator008:
After 9d52fed2c7da2a9e3807477c852c60b56edd2497, Array implements IArrayLike, so the union types can be simplified.

[Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1514)


MatrixFrog:
I think there are a bunch of these in Closure Library too. We should probably write a general check for redundant unions.

Dominator008:
@MatrixFrog Suprisingly there aren't that many in this particular case :)
https://github.com/google/closure-library/pull/652/files

But yeah, a general check would be good.

PR #1528, opened 2016-02-11
Cleanup TODOs in PeepholeIntegrationTest

Dominator008:
We no longer need to do string comparison, so cleanup all related
code and TODOs.

[Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1528)

### DIFF
--- a/test/com/google/javascript/jscomp/PeepholeIntegrationTest.java
+++ b/test/com/google/javascript/jscomp/PeepholeIntegrationTest.java
@@ -23,23 +23,10 @@ public class PeepholeIntegrationTest extends CompilerTestCase {
 
   private boolean late = true;
 
-  // TODO(user): Remove this when we no longer need to do string comparison.
-  private PeepholeIntegrationTest(boolean compareAsTree) {
-    super("", compareAsTree);
-  }
-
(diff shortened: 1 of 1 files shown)


test/com/google/javascript/jscomp/PeepholeIntegrationTest.java
```diff
@@ -132,18 +109,12 @@ public void testFoldOneChildBlocksIntegration() {
   }
 
   public void testFoldOneChildBlocksStringCompare() {
-    // The expected parse tree has a BLOCK structure around the true branch.
-    assertResultString("if(x){if(y){var x;}}else{var z;}",
-        "if(x){if(y)var x}else var z");
+    fold("if(x){if(y){var x;}}else{var z;}", "if(x){if(y)var x}else var z");
```
MatrixFrog: I assume you mean we're now doing AST comparison instead of comparing the literal bytes of the JS. In that case we can afford a little whitespace (`if (x) {` instead of `if(x){` etc.)

Dominator008: Done.

MatrixFrog:
lgtm

PR #1539, opened 2016-02-13
Use a better heuristic when backing off in FlowSensitiveInlineVariables

### LINKED ISSUE #250: FlowSensitiveInlineVariables backed off on string concatenation
This issue was imported from Closure Compiler's previous home at http://closure-compiler.googlecode.com

The original discussion is archived at:
http://blickly.github.io/closure-compiler-issues/#993

Dominator008:
bc3e3a494c406098064ee63487f9bfafcf3354f3 caused
FlowSensitiveInlineVariables to back off in some cases like the one
reported in #250. In retrospect, the back off strategy was being too
conservative.

If candidate "c" has dependencies, then inlining it may have introduced
new dependencies for our other inlining candidates.
MustBeReachingVariableDef uses a dependency graph in its analysis.
Generating a new dependency graph will need another CFG computation.
Ideally we should iterate to a fixed point, but that can be costly.
Therefore, we use a conservative heuristic here: For each candidate
"other", we back off if its set of dependencies cannot contain all of
"c"'s dependencies.

This should fix #250 with a much smaller potential performance impact
than iterating to a fixed point.

Fixes #250.

[Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1539)

### DIFF
--- a/src/com/google/javascript/jscomp/FlowSensitiveInlineVariables.java
+++ b/src/com/google/javascript/jscomp/FlowSensitiveInlineVariables.java
@@ -31,9 +31,8 @@ import com.google.javascript.rhino.Node;
 import com.google.javascript.rhino.Token;
 
 import java.util.Collection;
-import java.util.HashSet;
-import java.util.LinkedList;
-import java.util.List;
+import java.util.Iterator;
+import java.util.LinkedHashSet;
 import java.util.Set;
 
 /**
@@ -71,12 +70,11 @@ class FlowSensitiveInlineVariables extends AbstractPostOrderCallback
    * need two separate dataflow result.
    */
   private final AbstractCompiler compiler;
-  private final Set<Var> inlinedNewDependencies = new HashSet<>();
 
   // These two pieces of data is persistent in the whole execution of enter
   // scope.
   private ControlFlowGraph<Node> cfg;
-  private List<Candidate> candidates;
+  private Set<Candidate> candidates;
   private MustBeReachingVariableDef reachingDef;
   private MaybeReachingVariableUse reachingUses;
 
@@ -138,7 +136,7 @@ class FlowSensitiveInlineVariables extends AbstractPostOrderCallback
     cfg = cfa.getCfg();
     reachingDef = new MustBeReachingVariableDef(cfg, t.getScope(), compiler);
     reachingDef.analyze();
-    candidates = new LinkedList<>();
+    candidates = new LinkedHashSet<>();
 
     // Using the forward reaching definition search to find all the inline
     // candidates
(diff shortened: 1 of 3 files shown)


concavelenz:
We will need to a regression speed test, I think.

Dominator008:
@concavelenz I'd expect the impact to be low.

dimvar:
I'm bringing this in to try it on the regression tests.

dimvar:
OK, I tried this and it doesn't offer enough benefit. It improves large apps by less than 50 bytes.

Dominator008:
This was reported again at https://groups.google.com/forum/#!topic/closure-compiler-discuss/hV-hM8G_F_M. @dimvar Was there a significant increase in compilation time?

Dominator008:
@nicks @dimvar I changed this to not iterate to a fixed point, but rather use a better heuristic when backing off.

dimvar:
@Dominator008 I don't see why the new patch is related to the new unit test. Can you explain? In the unit test, x has no dependencies, so shouldn't have it been handled before?

Also, I'm not sure that your new heuristic is better than the fixpoint calculation. It is quadratic in the number of candidates. I'd have to run some tests to see if it's worth it. (It'd be helpful if we could find some large JS apps to use as our external regression tests. Right now, if I review a performance-related PR, I have to do the perf evaluation myself.)

Dominator008:
@dimvar All the definitions of `x` (the `x += 1` lines, etc.) have a dependency on the variable `x`. Before, once we inline one of them, the other candidates will view `x` as a "inlined new dependency" and back off. This patch prevents that from happening.

It is indeed quadratic in the number of candidates, but my intuition is that the number of candidates is not that big, and a large portion of them will fail the new heuristic check and get removed from the workset when the first candidate is inlined.

With iterating to a fixpoint and given the old heuristic, we need *n+1* iterations for *n* inlining operations that cause back-offs. For the unit test, that would be 8 (or 9? You get the idea). This means 8 CFG generations / CFAs, which should be pretty slow.

I don't know of any open-source app that is large enough and uses us to serve as the regression test. I have one that is only 5K LOC (https://github.com/google/wicked-good-xpath) and takes ~9 seconds to compile on my laptop.

Closure Library might be a good candidate? https://github.com/google/closure-library/blob/master/scripts/ci/compile_closure.sh This script usually takes 1 min to finish on my laptop.

Dominator008:
Sample run of https://github.com/google/closure-library/blob/master/scripts/ci/compile_closure.sh on my laptop.
Before this PR:
real	0m58.772s
user	2m18.932s
sys	0m0.912s

After this PR:
real	0m57.278s
user	2m14.936s
sys	0m1.308s

I didn't see a significant difference across multiple trials.

dimvar:
OK, thanks for the explanation.

Perf should be OK, but I haven't run any tests yet.

Another idea to consider. Maybe you can change every `x += expr` to `x = x + expr` in `Normalize`, and then the optimizations don't have to handle =+ specially. This will likely enable some more optimizations. And then you can change it back in `Denormalize`. Let's try this out first, since it's more general than this PR. If it's a bad idea for some reason, then we can merge this PR.

dimvar:
(Note: you would only change += when the lvalue is a variable, not when it's a property access.)

Dominator008:
@dimvar OK I tried your suggestion. The result is interesting: https://github.com/google/closure-compiler/pull/1663/files#diff-a41f00968b9e5a04015c0b2ecaacfcffR3450 Looks like the inlining stopped at "1234".

dimvar:
Sorry for the delay. Can you rebase and ping this thread? I'll bring the PR in today.

Dominator008:
@dimvar Done rebasing, thanks.

Dominator008:
Closed by https://github.com/google/closure-compiler/commit/58d892bbddb92bc64e7fbf42494000560b4bc0c3

PR #1547, opened 2016-02-15
Add externs for Angular 1.5 component

gberaudo:
As described in https://docs.angularjs.org/api/ng/provider/$compileProvider#component.

[Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1547)

### DIFF
--- a/contrib/externs/angular-1.5.js
+++ b/contrib/externs/angular-1.5.js
@@ -458,6 +458,24 @@ angular.LinkingFunctions.post = function(scope, iElement, iAttrs, controller) {
  */
 angular.Directive;
 
+
+/**
+ * @typedef {{
+ *   bindings: (Object.<string, string>|undefined),
+ *   controller: (angular.Injectable|string|undefined),
+ *   controllerAs: (string|undefined),
+ *   template: (string|
+ *       function(!angular.JQLite=,!angular.Attributes=): string|
+ *       undefined),
+ *   templateUrl: (string|
+ *       function(!angular.JQLite=,!angular.Attributes=)|
+ *       undefined),
+ *   transclude: (boolean|undefined)
+ *   }}
+ */
+angular.Component;
+
+
 /**
  * @typedef {(Function|Array.<string|Function>)}
  */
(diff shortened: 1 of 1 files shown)


dimvar:
@mprobst Can you take a look at this?

contrib/externs/angular-1.5.js
```diff
@@ -717,6 +735,15 @@ angular.Module = function() {};
 angular.Module.prototype.animation = function(name, animationFactory) {};
 
 /**
+ * Intended to be called with a name string and a component definition.
+ *
+ * @param {string} name
+ * @param {Object<string, *>} componentDefinition
```
mprobst: Shouldn't this reference `!angular.Component` ?

mprobst:
One nit, LGTM otherwise.

contrib/externs/angular-1.5.js
gberaudo: Unfortunately, using `!angular.Component` would require a cast in user code to handle [optional properties](https://groups.google.com/forum/#!topic/closure-compiler-discuss/3qhcDfa9uX4) or to handle `"custom properties"` (`$... – {function()=} – additional annotations to provide to the directive factory function.`).

Directives are handled with a `angular.Injectable=` param which is also a loose type.

I added the missing `!`; if you still prefer `!angular.Component` I can update the PR.

mprobst:
Shouldn't the `typedef` with a record type allow passing objects that are lacking the optional fields?

```js
/** @typedef {{foo: string, bar: (number|undefined)}} */
var Foo;

/** @param {!Foo} foo */
function useFoo(foo) {}

useFoo({foo: 'asd'}); // no 'bar'
```

gberaudo:
Thank you @mprobst, I updated the PR.

With a compiler from October and the `--jscomp_error "*"` flag, I had an `initializing variable` error.
With a version from December it now passes, which is very great!

Is it a new feature?

gberaudo:
I have [a pile of Travis builds](https://travis-ci.org/google/closure-compiler/pull_requests) stuck in `creation mode`. I cannot cancel jobs; could someone try to cancel the older ones and restart the last one?

Dominator008:
@gberaudo Travis just takes a while during US business hours... Hang in there :)

gberaudo:
OK, will try to be patient ^^.

mprobst:
@gberaudo: my understanding was that this was always the intended behaviour, but this aspect of Closure Compiler seems like a less well-trodden path.

PR #1549, opened 2016-02-15
Preserve file order when using '--js <files>' with CommandlineRunner

### LINKED ISSUE #1548: File order not maintained
I call the closure compiler like:

    java.exe -jar dev\jars\closure\compiler.jar --compilation_level WHITESPACE_ONLY --formatting PRETTY_PRINT --formatting PRINT_INPUT_DELIMITER --formatting SINGLE_QUOTES --js_output_file jQuery_a52d83c3.min.js --js jquery.js ui\core.js ui\widget.js ui\mouse.js ui\draggable.js ui\droppable.js ui\resizable.js ui\selectable.js ui\sortable.js ui\button.js ui\position.js

When I use the `compiler.jar` from github binary downloads `compiler-20151216.tar.gz` everything is fine. When I use `compiler-20160208.tar.gz` the file originial order is not maintained in the compressed file which breaks the code, as the js files have to be included in a specific order.
I tried the new parameter `--dependency_mode NONE` but it changed nothing. According to the docs should NONE be the default.

I'm on Windows 10, 64 bit JAVA 8 as provided by Oracle.

Dominator008:
Use LinkedHashSet instead of TreeSet when we are dealing with cases
like '--js test3.js test2.js test1.js'.

For cases like '--js=*.js', keep using a TreeSet to preserve
alphabetical order by filename.

Fixes #1548.

[Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1549)

### DIFF
--- a/src/com/google/javascript/jscomp/CommandLineRunner.java
+++ b/src/com/google/javascript/jscomp/CommandLineRunner.java
@@ -62,6 +62,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -757,7 +758,7 @@ public class CommandLineRunner extends
             mixedSources.remove(new FlagEntry<>(JsSourceType.JS, filename));
           }
         } else {
-          for (String filename : findJsFiles(Collections.singletonList(source.value))) {
+          for (String filename : findJsFiles(Collections.singletonList(source.value), true)) {
             if (!excludes.contains(filename)) {
               mixedSources.add(new FlagEntry<>(JsSourceType.JS, filename));
             }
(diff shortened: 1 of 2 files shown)


src/com/google/javascript/jscomp/CommandLineRunner.java
```diff
@@ -1503,7 +1505,11 @@ private static ConformanceConfig loadConformanceConfig(String configFile)
    * within the directory and sub-directories.
    */
   public static List<String> findJsFiles(Collection<String> patterns) throws IOException {
-    Set<String> allJsInputs = new TreeSet<>();
+    return findJsFiles(patterns, new LinkedHashSet<String>());
+  }
+
+  private static List<String> findJsFiles(Collection<String> patterns, Set<String> allJsInputs)
```
MatrixFrog: This should probably have some JavaDoc, particularly since we're adding things to a collection that is passed as a param.

Dominator008: Done. Changed to take in a boolean `sortAlphabetically` and added JavaDoc.

MatrixFrog:
I'm not necessarily opposed to a function taking in a set and then adding things to it, but the new way is clearer and simpler in this case IMO. Thanks. lgtm once travis is done

Issue #1546, opened 2016-02-14
Class identifier not properly scoped in class expression


jleyba:
``` javascript
var Clazz = class Foo {};
class Foo {}
```

For the above, the first `Foo` should only be visible within the class expression. The compiler appears to be pulling this out during transpilation, producing

```
ERROR - Illegal redeclared variable: Foo
class Foo {}
      ^

1 error(s), 0 warning(s)
```

This should be handled similarly to function declaration/expressions:

http://closure-compiler-debugger.appspot.com/#input0%3Dvar%2520Clazz%2520%253D%2520class%2520Foo%2520%257B%257D%253B%250Aclass%2520Foo%2520%257B%257D%250A%250Avar%2520one%2520%253D%2520function%2520two()%2520%257B%257D%253B%250Afunction%2520two()%2520%257B%257D%26input1%26conformanceConfig%26externs%26refasterjs-template%26includeDefaultExterns%3D1%26CHECK_SYMBOLS%3D1%26CHECK_TYPES%3D1%26LANG_IN_IS_ES6%3D1%26MISSING_PROPERTIES%3D1%26PRESERVE_TYPE_ANNOTATIONS%3D1%26PRETTY_PRINT%3D1%26TRANSPILE%3D1

Dominator008:
Thanks for the report. The error you are seeing should be fixed by #1551, but we will also need to change our transpilation to support referring to the class name in the methods.

brad4d:
yep, I believe I fixed it and failed to realize this issue could be closed.
Thanks for pointing this out @teppeis

PR #1551, opened 2016-02-15
Fix class name not properly scoped to CLASS

Dominator008:
We need to treat CLASS as a root for a block scope possibly
containing just the "class name" declaration.

See "classScope" in:
http://www.ecma-international.org/ecma-262/6.0/index.html#sec-runtime-semantics-classdefinitionevaluation

Add a few unit tests regarding class names and function names.

Note that our transpilation doesn't support referring to the class name
in a method body yet. Eg:
```js
var Clazz = class Foo {
  bar() {
    return new Foo;
  }
};
```
is currently transpiled to:
```js
/**@constructor @struct */var Clazz = function() {
};
Clazz.prototype.bar = function() {
  return new Foo;
};
```
which does not work at runtime. This will be fixed later.

Part of the fix to #1546.

[Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1551)

### DIFF
--- a/src/com/google/javascript/jscomp/Es6SyntacticScopeCreator.java
+++ b/src/com/google/javascript/jscomp/Es6SyntacticScopeCreator.java
@@ -98,15 +98,28 @@ class Es6SyntacticScopeCreator implements ScopeCreator {
           declareLHS(scope, a);
         }
       }
-
       // Since we create a separate scope for body, stop scanning here
+
+    } else if (n.isClass()) {
+      if (scope.getParent() != null) {
+        inputId = NodeUtil.getInputId(n);
+      }
+
+      final Node classNameNode = n.getFirstChild();
+      // Bleed the class name into the scope, if it hasn't
+      // been declared in the outer scope.
+      if (!classNameNode.isEmpty()) {
+        if (NodeUtil.isClassExpression(n)) {
+          declareVar(classNameNode);
+        }
+      }
     } else if (n.isBlock() || n.isFor() || n.isForOf() || n.isSwitch()) {
       if (scope.getParent() != null) {
         inputId = NodeUtil.getInputId(n);
       }
       scanVars(n);
     } else {
-      // It's the global block
+      // n is the global SCRIPT node
       Preconditions.checkState(scope.getParent() == null);
       scanVars(n);
     }
@@ -262,8 +275,8 @@ class Es6SyntacticScopeCreator implements ScopeCreator {
   private boolean isNodeAtCurrentLexicalScope(Node n) {
     Node parent = n.getParent();
     Node grandparent = parent.getParent();
-    Preconditions.checkState(parent.isBlock() || parent.isFor()
-        || parent.isForOf() || parent.isScript() || parent.isLabel());
+    Preconditions.checkState(parent.isBlock() || parent.isFor() || parent.isForOf()
+        || parent.isScript() || parent.isLabel(), parent);
 
     if (parent.isSyntheticBlock()
         && grandparent != null && (grandparent.isCase() || grandparent.isDefaultCase())) {
--- a/src/com/google/javascript/jscomp/NodeTraversal.java
+++ b/src/com/google/javascript/jscomp/NodeTraversal.java
@@ -605,6 +605,8 @@ public class NodeTraversal {
 
     if (type == Token.FUNCTION) {
       traverseFunction(n, parent);
+    } else if (type == Token.CLASS) {
+      traverseClass(n, parent);
     } else if (useBlockScope && NodeUtil.createsBlockScope(n)) {
       traverseBlockScope(n);
     } else {
@@ -631,7 +633,7 @@ public class NodeTraversal {
         && NodeUtil.isFunctionExpression(n);
 
     if (!isFunctionExpression) {
-      // Functions declarations are in the scope containing the declaration.
+      // Function declarations are in the scope containing the declaration.
       traverseBranch(fnName, n);
     }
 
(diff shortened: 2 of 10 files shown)


test/com/google/javascript/jscomp/Es6VariableReferenceCheckTest.java
```diff
@@ -277,6 +277,10 @@ public void testClass() {
     assertNoWarning("class A { f() { return 1729; } }");
   }
 
+  public void testRedeclareClassName() {
+    assertNoWarning("var Clazz = class Foo {}; var Foo = 3;");
```
dimvar: @Dominator008 I would propose to just ban the pattern of using two different names for the same class. What's a good reason to support this?

Dominator008: I think the ES6 specs was trying to be consistent with how `function`s look like, i.e. `var func = function f() {}`.

@jleyba Do you have a real-world example showing the usage of this?

jleyba: I originally encountered this with a custom Promise implementation. This bug is easy to work around, but the behavior is inconsistent and surprising considering the compiler properly handles scoping on [function names](http://closure-compiler-debugger.appspot.com/#input0%3Dvar%2520Clazz%2520%253D%2520class%2520Foo%2520%257B%257D%253B%250Aclass%2520Foo%2520%257B%257D%250A%250Aclass%2520Bar%2520%257B%257D%250Avar%2520Baz%2520%253D%2520class%2520Bar%2520%257B%257D%250A%250Avar%2520one%2520%253D%2520function%2520two()%2520%257B%257D%253B%250Afunction%2520two()%2520%257B%257D%250A%250Afunction%2520three()%2520%257B%257D%250Avar%2520four%2520%253D%2520function%2520three()%2520%257B%257D%26input1%26conformanceConfig%26externs%26refasterjs-template%26includeDefaultExterns%3D1%26CHECK_SYMBOLS%3D1%26CHECK_TYPES%3D1%26LANG_IN_IS_ES6%3D1%26MISSING_PROPERTIES%3D1%26PRESERVE_TYPE_ANNOTATIONS%3D1%26PRETTY_PRINT%3D1%26TRANSPILE%3D1)

MatrixFrog: I agree with @dimvar that we should ban (or at least discourage) this pattern, but I also think we should support it since third party code is already doing this.

jleyba: Is it feasible to introduce a more descriptive error message?

Dominator008: I think we should support it in the compiler, but discourage it in the style guide.

Dominator008:
I've rebased and resolved the merge conflicts brought by https://github.com/google/closure-compiler/commit/b490b587458f9e26962f32e5e0d5019eaf84b56c.

src/com/google/javascript/jscomp/Es6SyntacticScopeCreator.java
```diff
@@ -100,13 +100,27 @@ private void scanRoot(Node n) {
       }
 
       // Since we create a separate scope for body, stop scanning here
```
blickly: Looks like this comment is stale

src/com/google/javascript/jscomp/Es6SyntacticScopeCreator.java
```diff
@@ -100,13 +100,27 @@ private void scanRoot(Node n) {
...
+      if (scope.getParent() != null) {
+        inputId = NodeUtil.getInputId(n);
+      }
+
+      final Node classNameNode = n.getFirstChild();
+      // Bleed the class name into the scope, if it hasn't
+      // been declared in the outer scope.
+      if (!classNameNode.isEmpty()) {
+        String className = classNameNode.getString();
+        if (!className.isEmpty() && NodeUtil.isClassExpression(n)) {
```
blickly: When class has no name, it has an EMPTY node rather than an empty string like function nodes.

src/com/google/javascript/jscomp/VariableReferenceCheck.java
```diff
@@ -227,11 +227,14 @@ private void checkVar(Var v, List<Reference> references) {
                 referenceNode, v);
         boolean letConstShadowsVar = v.getParentNode().isVar()
             && (reference.isLetDeclaration() || reference.isConstDeclaration());
```
blickly: Since the changes to this file look like just refactoring, can we split them into a separate PR?

src/com/google/javascript/jscomp/NodeTraversal.java
```diff
@@ -657,6 +659,41 @@ private void traverseFunction(Node n, Node parent) {
     popScope();
   }
 
+  /** Traverses a class. */
+  private void traverseClass(Node n, Node parent) {
+    Preconditions.checkState(n.getChildCount() == 3, n);
+    Preconditions.checkState(n.isClass());
+
+    final Node className = n.getFirstChild();
+    boolean isClassExpression = (parent != null)
```
blickly: How does this occur? (CLASS node with null parent)

I'd think that would be impossible in a full AST

dimvar:
Sorry, I've dropped the ball on this. Super busy this past week. Reassigning to Ben.

src/com/google/javascript/jscomp/Es6SyntacticScopeCreator.java
Dominator008: It was there since the file was created :) I think it's still right though.

src/com/google/javascript/jscomp/NodeTraversal.java
Dominator008: Done. Removed `(parent != null)`. This was copied from `traverseFunction()`, although I'm not sure when `FUNCTION` will have null parent either.

src/com/google/javascript/jscomp/Es6SyntacticScopeCreator.java
Dominator008: Done. Removed the unnecessary check.

src/com/google/javascript/jscomp/VariableReferenceCheck.java
Dominator008: Done.

src/com/google/javascript/jscomp/NodeTraversal.java
```diff
@@ -657,6 +659,40 @@ private void traverseFunction(Node n, Node parent) {
...
+    if (!isClassExpression) {
+      // Class declarations are in the scope containing the declaration.
+      traverseBranch(className, n);
+    }
+
+    curNode = n;
+    pushScope(n);
+
+    if (isClassExpression) {
+      // Class expression names are only accessible within the function
```
blickly: function => class

Dominator008: Done.

test/com/google/javascript/jscomp/VarCheckTest.java
```diff
@@ -173,6 +173,7 @@ public void testNamedClass() {
     testSameEs6("class x {}");
     testSameEs6("var x = class x {};");
     testSameEs6("var y = class x {};");
+    testSameEs6("var y = class x { foo() { return new x; } };");
```
blickly: Let's add a test to demonstrate that the class expression name is in scope inside the extends clause.

Dominator008: Done. We can't transpile this code yet though but that's fine.

Issue #1162, opened 2015-09-22
Array.isArray vs goog.isArray


fredj:
The following code compiles without any error:
```js
// ==ClosureCompiler==
// @compilation_level ADVANCED_OPTIMIZATIONS
// @output_file_name default.js
// @use_closure_library true
// ==/ClosureCompiler==

/**
 * @param {Array.<string>|string} arg
 * @return {Array.<string>}
 */
function f(arg) {
  if (goog.isArray(arg)) {
    return arg;
  } else {
    return [arg];
  }
}
```

But if `goog.isArray` is replaced with `Array.isArray`:
```js
// ==ClosureCompiler==
// @compilation_level ADVANCED_OPTIMIZATIONS
// @output_file_name default.js
// @use_closure_library true
// ==/ClosureCompiler==

/**
 * @param {Array.<string>|string} arg
 * @return {Array.<string>}
 */
function f(arg) {
  if (Array.isArray(arg)) {
    return arg;
  } else {
    return [arg];
  }
}
```
The following warning is returned:
```
JSC_TYPE_MISMATCH: inconsistent return type
found   : (Array<string>|null|string)
required: (Array<string>|null) at line 7 character 11
    return arg;
```

Would it be possible to instruct the compiler about `Array.isArray`?

MatrixFrog:
It looks like the compiler really understands goog.isArray, from here: https://github.com/google/closure-compiler/blob/0c09d82ab71f5518d6334692d279615e59c2b455/src/com/google/javascript/jscomp/type/ClosureReverseAbstractInterpreter.java#L185 for the old type inference. So we'd need to make sure there's a way it can understand non-`goog` things and then we could probably add in `Array.isArray` as well

fredj:
[Number.isInteger()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger) is an other good example

Dominator008:
This is fixed in OTI and NTI: #1451 and #1452.
The following code:
```
/**
 * @param {!Array.<string>|string} arg
 * @return {Array.<string>}
 */
function f(arg) {
  if (Array.isArray(arg)) {
    return arg;
  } else {
    return [arg];
  }
}
```
compiles with NTI without warning. Note that `arg` should be declared as non-nullable.

PR #1545, opened 2016-02-14
Fix Number.isInteger(), etc. in NewTypeInference

Dominator008:
As suggested in #1162.

[Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1545)

PR #1552, opened 2016-02-15
Fix Number.isInteger(), etc. in old type inference

Dominator008:
As suggested in #1162.

[Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1552)

### DIFF
--- a/test/com/google/javascript/jscomp/TypeInferenceTest.java
+++ b/test/com/google/javascript/jscomp/TypeInferenceTest.java
@@ -1516,6 +1516,40 @@ public final class TypeInferenceTest extends TestCase {
     verify("x", NUMBER_TYPE);
   }
 
+  public void testIsNumber() {
+    assuming("x", createUnionType(ARRAY_TYPE, NUMBER_TYPE));
+    inFunction("goog.asserts.assert(Number.isFinite(x));");
+    verify("x", NUMBER_TYPE);
+
+    assuming("x", createUnionType(ARRAY_TYPE, NUMBER_TYPE));
+    inFunction("goog.asserts.assert(Number.isInteger(x));");
+    verify("x", NUMBER_TYPE);
+
+    assuming("x", createUnionType(ARRAY_TYPE, NUMBER_TYPE));
+    inFunction("goog.asserts.assert(Number.isNaN(x));");
+    verify("x", NUMBER_TYPE);
+
+    assuming("x", createUnionType(ARRAY_TYPE, NUMBER_TYPE));
+    inFunction("goog.asserts.assert(Number.isSafeInteger(x));");
+    verify("x", NUMBER_TYPE);
+
+    assuming("x", createUnionType(ARRAY_TYPE, NUMBER_TYPE));
+    inFunction("goog.asserts.assert(!Number.isFinite(x));");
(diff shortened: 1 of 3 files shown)


dimvar:
Assigning to self. Depending on what we decide for #1545, we'll revisit this as well.

dimvar:
So, turns out OTI doesn't restrict the type of an untyped variable if it's passed to a typed function. Since it generally doesn't do this, we shouldn't just special-case these functions to do it.

http://closure-compiler-debugger.appspot.com/#input0%3Dfunction%2520f(%252F**%2520number%2520*%252F%2520x)%2520%257B%257D%250A%250Afunction%2520g(x)%2520%257B%250A%2520%2520f(x)%253B%250A%2520%2520var%2520%252F**%2520string%2520*%252F%2520s%2520%253D%2520x%253B%250A%257D%26input1%26conformanceConfig%26externs%26refasterjs-template%26includeDefaultExterns%3D1%26CHECK_SYMBOLS%3D1%26CHECK_TYPES%3D1%26LANG_IN_IS_ES6%3D1%26MISSING_PROPERTIES%3D1%26PRESERVE_TYPE_ANNOTATIONS%3D1%26PRETTY_PRINT%3D1%26TRANSPILE%3D1

concavelenz:
Dimvar@ we do this with a number of specific cases: goog.asserts methods, goog.isX method.  isInteger is a natural extensions of this

Dominator008:
@concavelenz LG?

concavelenz:
LGTM

test/com/google/javascript/jscomp/TypeInferenceTest.java
```diff
@@ -1516,6 +1516,40 @@ public void testNotIsArray() {
...
+    assuming("x", createUnionType(ARRAY_TYPE, NUMBER_TYPE));
+    inFunction("goog.asserts.assert(Number.isSafeInteger(x));");
+    verify("x", NUMBER_TYPE);
+
+    assuming("x", createUnionType(ARRAY_TYPE, NUMBER_TYPE));
+    inFunction("goog.asserts.assert(!Number.isFinite(x));");
+    verify("x", NUMBER_TYPE);
+
+    assuming("x", createUnionType(ARRAY_TYPE, NUMBER_TYPE));
+    inFunction("goog.asserts.assert(!Number.isInteger(x));");
```
blickly: Not sure I follow here. If something is not an integer, how do we know it's a number type?

Dominator008: I think the intention is that `x` should always be a number type: https://github.com/google/closure-compiler/blob/master/externs/es6.js#L1525-L1531.

blickly: Then what makes the Numer.isXXX methods different than any other extern definitions?

Dominator008: I wrote this before seeing that the externs are already stricter than the specs. I think this change is unnecessary then.

Dominator008:
I wrote this before seeing that the externs are already stricter than the specs. I think this change is unnecessary then.

PR #1570, opened 2016-02-19
New lint check: CheckRedundantBooleanCast

Dominator008:
Check for redundant casts to boolean via double negation or a Boolean call.
Eg:
```js
var foo = !!!bar;
var foo = Boolean(!!bar);
if (!!foo) {}
if (Boolean(foo)) {}
```
Inspired by ESLint
  (https://github.com/eslint/eslint/blob/master/lib/rules/no-extra-boolean-cast.js)

[Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1570)


MatrixFrog:
I'm not sure this one is worth it. Seems like the kind of thing that is easy to spot in review.

Dominator008:
@MatrixFrog It might be worth trying. I **guess** there will be code already doing this...

dimvar:
But this is purely a style issue. Explicitly converting to a boolean in a context that auto-converts to boolean doesn't introduce a bug.

Dominator008:
@dimvar Yeah so it kind of makes sense to introduce it as a lint check (think unnecessary cast), although our other lint checks catch things that are more likely to be errors.

PR #1571, opened 2016-02-19
Add checkNegatedLeftOperandOfInOperator to CheckSuspiciousCode

Dominator008:
Check for negated left operand of the "in" operator
Eg:
```js
if (!a in b) {}
```
should almost always be
```js
if (!(a in b)) {}
```
or
```js
if (('' + !a) in b) {}
```
instead.

Negated left operand of the "in" operator usually indicates a programmer error.

Inspired by ESLint (https://github.com/eslint/eslint/blob/master/lib/rules/no-negated-in-lhs.js)

[Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1571)


MatrixFrog:
This is very similar to SUSPICIOUS_INSTANCEOF_LEFT_OPERAND so maybe we should put it in the CheckSuspiciousCode pass.

Dominator008:
Done.

MatrixFrog:
lgtm. An inexact but probably-good-enough code search suggests this will not find any errors within Google so I think we can just merge it.

PR #1573, opened 2016-02-19
Allow file based modules as entry points

ChadKillingsworth:
ES6 module rewriting only issues `goog.provide` statements when an `export` keyword is encountered. It is common for an ES6 module entry point to only `import` statements. This PR supports using file based modules as entry points.

[Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1573)

### DIFF
--- a/src/com/google/javascript/jscomp/JSModuleGraph.java
+++ b/src/com/google/javascript/jscomp/JSModuleGraph.java
@@ -453,7 +453,12 @@ public final class JSModuleGraph {
         CompilerInput entryPointInput = null;
         try {
           if (entryPoint.getClosureNamespace().equals(entryPoint.getModuleName())) {
-            entryPointInput = sorter.getInputProviding(entryPoint.getClosureNamespace());
+            entryPointInput = sorter.maybeGetInputProviding(entryPoint.getClosureNamespace());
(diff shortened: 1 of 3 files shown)


ChadKillingsworth:
What happened with Travis on this PR? I'm going to close and reopen to see if that will trigger it.

Dominator008:
@ChadKillingsworth Travis just takes a while during the day...

ChadKillingsworth:
@Dominator008 In this case it didn't even trigger as one of the checks though. The only check was the CLA.

MatrixFrog:
This will be a bit tricky to land because of the way our Blaze targets are arranged. Will try to get that resolved next week.

MatrixFrog:
resetting the ETA on this to two weeks from now :(

MatrixFrog:
Updating this https://github.com/google/closure-compiler/blob/819c6099c1dc27c0a8612695f372c0abc4fa82a6/src/com/google/javascript/jscomp/ModuleIdentifier.java#L97 to not call ES6ModuleLoader may make it easier to land this sooner. Need to investigate.

ChadKillingsworth:
@MatrixFrog I've refactored the `ES6SortedDependencies` class to no longer import or use either the `ModuleIdentifier` or `ES6ModuleLoader` classes.

PR #1575, opened 2016-02-20
Allow "=" in the "this" and "new" type of functional types

Dominator008:
Eg:
```js
/** @type {function(this:Foo=)} */
```
Note that JSTypeRegistry, which is used by OTI, will ignore the optionality and resolve this to "function(this:Foo)". NTI won't ignore it.

[Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1575)

### DIFF
--- a/src/com/google/javascript/jscomp/parsing/JsDocInfoParser.java
+++ b/src/com/google/javascript/jscomp/parsing/JsDocInfoParser.java
@@ -2153,9 +2153,14 @@ public final class JsDocInfoParser {
           if (match(JsDocToken.COLON)) {
             next();
             skipEOLs();
+            Node type = parseContextTypeExpression(next());
+            if (match(JsDocToken.EQUALS)) {
+              next();
+              skipEOLs();
+              type = wrapNode(Token.EQUALS, type);
+            }
             Node contextType = wrapNode(
-                isThis ? Token.THIS : Token.NEW,
-                parseContextTypeExpression(next()));
+                isThis ? Token.THIS : Token.NEW, type);
             if (contextType == null) {
               return null;
             }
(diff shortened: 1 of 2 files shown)


MatrixFrog:
What does new:Foo= mean? That it might return a Foo if called with `new` and it might not?

Dominator008:
It might return `(Foo|null|undefined)`.

MatrixFrog:
I don't think it's possible to write a constructor that returns null or undefined. Even if it were, I'd prefer we write that as `function(new:(Foo|null|undefined))`. Likewise for `this:`.

Dominator008:
It's impossible with `new`, but we don't check for that in OTI.
```js
/** @param {function(new:undefined)} x */
function f(x) {
  var /** string */ s = new x;
}
```
gives no warning.
http://closure-compiler-debugger.appspot.com/#input0%3D%252F**%2520%2540param%2520%257Bfunction(new%253Aundefined)%257D%2520x%2520*%252F%250Afunction%2520f(x)%2520%257B%250A%2520%2520var%2520%252F**%2520string%2520*%252F%2520s%2520%253D%2520new%2520x%253B%250A%257D%26input1%26conformanceConfig%26externs%26refasterjs-template%26includeDefaultExterns%3D1%26CHECK_SYMBOLS%3D1%26CHECK_TYPES%3D1%26LANG_IN_IS_ES6%3D1%26MISSING_PROPERTIES%3D1%26PRESERVE_TYPE_ANNOTATIONS%3D1%26PRETTY_PRINT%3D1%26TRANSPILE%3D1

dimvar:
I don't think we should support this syntax. Using the `=` means that a type can be omitted. In params, if the first param is optional but the second is not, you get a warning. So if you want to support optional `this`, you'd need more than the changes here, you'd have to go and check that all params are also optional. This is more work for not too much benefit. Most of the time, `this` should be a "classy" type, so it's not a big burden to explicitly type `|undefined` in the few cases where it's needed.

Dominator008:
@dimvar, ah that makes sense.

Issue #1589, opened 2016-02-26
[NTI] Failed to infer instanceof in nested conditions.


Dominator008:
```js
/** @constructor */
function Foo() {}

/** @constructor */
function Bar() {}

/**
 * @param {Foo|Bar} x
 * @param {Foo|Bar} y
 */
function f(x, y) {
  if ((x instanceof Foo) || (y instanceof Foo)) {
    var foo;
    if (x instanceof Foo) {
      foo = x;
    } else {
      foo = y;
    }
    var /** Foo */ foo2 = foo;
  }
}
```
gives
```
input0:19: WARNING - The right side in the assignment is not a subtype of the left side.
left side  : Foo|null
right side : Bar|Foo|null

    var /** Foo */ foo2 = foo;
                          ^

0 error(s), 1 warning(s), 100.0% typed
```
I guess the `y` was not inferred to be a `Foo` in the `else` branch.

[...]

PR #1590, opened 2016-02-26
[NTI] Improve handling of instanceof

Dominator008:
If we know for sure that `x` is not of type `Foo`, we can infer that
`x instanceof Foo` is of `FALSE_TYPE`. On the other hand, if we know
for sure that `x` is of type `Foo` already, we can infer that
`x instanceof Foo` is of `TRUE_TYPE`.

Part of the fix to #1589.

[Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1590)

### DIFF
--- a/src/com/google/javascript/jscomp/NewTypeInference.java
+++ b/src/com/google/javascript/jscomp/NewTypeInference.java
@@ -1607,7 +1607,14 @@ final class NewTypeInference implements CompilerPass {
       objPair = analyzeExprFwd(obj, inEnv, JSType.UNKNOWN, instanceSpecType);
       ctorPair = analyzeExprFwd(ctor, objPair.env, commonTypes.topFunction());
     }
-    ctorPair.type = JSType.BOOLEAN;
+    if (instanceSpecType.isBottom()
+        || (specializedType.isFalseOrFalsy() && objType.equals(instanceSpecType))) {
+      ctorPair.type = JSType.FALSE_TYPE;
+    } else if (specializedType.isTrueOrTruthy() && objType.equals(instanceSpecType)) {
+      ctorPair.type = JSType.TRUE_TYPE;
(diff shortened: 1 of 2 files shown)


Dominator008:
@dimvar Not sure if this makes sense, but I think we could do better.

dimvar:
This won't help with #1589. Here is what is happening in #1589 in more detail. When analyzing a statement of the form `if (A || B) { ... }`, we analyze A for the case when it's true and for the case when it's false, and get two type environments out as a result. We analyze B with the Afalse environment. We analyze the then branch with the join of the Atrue and the AfalseBtrue environment. So, for #1589,

Atrue: x is a Foo
Afalse: x is a Bar (we don't have a way to represent "not a Foo")
Btrue: y is a Foo
AfalseBtrue: x is a Bar, y is a Foo
Atrue join AfalseBtrue: x is Foo|Bar, y is a Foo

See that in the last environment (and in all environments in general), a variable is mapped to a type and that's all we can express, we can't express relations between the types of variables such as, y is a Foo only when x is a Bar.

This is a fundamental limitation of the way NTI works (it's a good trade-off to not have this, separate discussion), it's not something that can be fixed locally. Even if you somehow special-case the handling of #1589 to fix that particular case, NTI won't be able to express such reasoning in the general case. And that's fine. It's not possible to infer everything, our current approach is a good trade-off between precision and having a scalable type checker.

dimvar:
Btw, thanks for your NTI-related patches and PRs, it's nice to have more people dig into it and know what's going on. FYI, if you haven't seen this already, it's helpful to set this to `true` when you're debugging NTI stuff, it shows you the type environments after every step.
https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/NewTypeInference.java#L353

PR #1591, opened 2016-02-26
Rewrite the logic for folding comparisons in PeepholeFoldConstants

Dominator008:
evaluateComparison() is now based on the three Abstract Operations
defined in the ES6 Specs: Abstract Relational Comparison [1],
Abstract Equality Comparison [2] and Strict Equality Comparison [3].

This fixes a handful of foldable cases that were not folded previously:

Cases like `0 < null` and `'hi' < null` are now folded.

String-string comparisons and string-number comparisions
like `'a' < 'b'` and `1 < '2'` are now folded.

Cases involving `NaN` are now folded.

As a bonus, a test case in testOptimizeSwitch() in
PeepholeRemoveDeadCodeTest now folds.

[1] http://www.ecma-international.org/ecma-262/6.0/#sec-abstract-relational-comparison
[2] http://www.ecma-international.org/ecma-262/6.0/#sec-abstract-equality-comparison
[3] http://www.ecma-international.org/ecma-262/6.0/#sec-strict-equality-comparison

[Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1591)

### DIFF
--- a/src/com/google/javascript/jscomp/PeepholeFoldConstants.java
+++ b/src/com/google/javascript/jscomp/PeepholeFoldConstants.java
@@ -942,7 +942,6 @@ class PeepholeFoldConstants extends AbstractPeepholeOptimization {
   /**
    * Try to fold comparison nodes, e.g ==
    */
-  @SuppressWarnings("fallthrough")
   private Node tryFoldComparison(Node n, Node left, Node right) {
     TernaryValue result = evaluateComparison(n.getType(), left, right, shouldUseTypes);
     if (result == TernaryValue.UNKNOWN) {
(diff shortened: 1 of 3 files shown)


Dominator008:
@concavelenz.

src/com/google/javascript/jscomp/PeepholeFoldConstants.java
```diff
@@ -1075,12 +1075,26 @@ static TernaryValue evaluateComparison(int op, Node left, Node right, boolean us
               return areStringsEqual(left.getString(), right.getString()).not();
           }
         }
+        switch (op) {
+          case Token.GE:
+          case Token.LE:
+          case Token.GT:
+          case Token.LT:
+            return compareAsNumbers(op, left, right, useTypes);
```
blickly: I don't think this is a correct optimization. Consider, for example: `'a'  < 'b'` vs. `+'a' < +'b'`

Dominator008: Good catch! Need an `else`.

dimvar:
LGTM. I'll merge this internally.

dimvar:
Undo LGTM. Ben is not sure that this is correct, so I'm reassigning to him and he can take another look and suggest changes.

src/com/google/javascript/jscomp/PeepholeFoldConstants.java
```diff
@@ -1080,6 +1081,15 @@ static TernaryValue evaluateComparison(int op, Node left, Node right, boolean us
       case Token.NUMBER:
         if (right.isNumber()) {
           return compareAsNumbers(op, left, right, useTypes);
+        } else if (right.isString()) {
+          return compareAsNumbers(op, left, IR.number(NodeUtil.getNumberValue(right)), useTypes);
```
blickly: This branch is not correct for SHEQ and SHNE, since, e.g. `'1' !== 1`.
I think that those cases are handled are handled earlier so that this branch isn't reachable for SHEQ and SHNE, but I'd rather that the code be more explicit about what cases are covered and what the ES semantics are. Links into the relevant parts of the spec would also be helpful.

Dominator008: I added tests for these and made the code explicit. Added a couple of links to the ES6 specs.

src/com/google/javascript/jscomp/PeepholeFoldConstants.java
```diff
@@ -1064,15 +1064,24 @@ static TernaryValue evaluateComparison(int op, Node left, Node right, boolean us
...
-            case Token.EQ:
-              return areStringsEqual(left.getString(), right.getString());
-
+              return TernaryValue.FALSE;
             case Token.SHNE:
-            case Token.NE:
-              return areStringsEqual(left.getString(), right.getString()).not();
+              return TernaryValue.TRUE;
+          }
+          return compareAsNumbers(op, IR.number(NodeUtil.getNumberValue(left)), right, useTypes);
```
blickly: Isn't this whole
`IR.number(NodeUtil.getNumberValue(x))`
extraneous with what compareAsNumbers does anyway?

src/com/google/javascript/jscomp/PeepholeFoldConstants.java
```diff
@@ -1064,15 +1064,24 @@ static TernaryValue evaluateComparison(int op, Node left, Node right, boolean us
 
       case Token.STRING:
         if (right.isString()) {
-          // Only eval if they are the same type
+          return compareStrings(op, left, right);
+        } else if (right.isNumber()) {
+          // See http://www.ecma-international.org/ecma-262/6.0/#sec-strict-equality-comparison
           switch (op) {
```
blickly: I'd prefer a switch (and links to the spec) for the cases that can be compared as numbers, rather than the cases that can't.

In fact, can't we merge this case with the following one?

src/com/google/javascript/jscomp/PeepholeFoldConstants.java
Dominator008: Yes. `compareAsNumbers` was being too restrictive about `EQ` and `NE`. I removed the precondition checks that require both operands to be numbers for `EQ` and `NE`.

src/com/google/javascript/jscomp/PeepholeFoldConstants.java
Dominator008: Done merging. Added a `TODO` saying we could do better, but it requires more changes to the other cases of the left hand side. Will investigate later.

src/com/google/javascript/jscomp/PeepholeFoldConstants.java
```diff
@@ -1080,6 +1087,22 @@ static TernaryValue evaluateComparison(int op, Node left, Node right, boolean us
...
+          // See http://www.ecma-international.org/ecma-262/6.0/#sec-abstract-relational-comparison
+          switch (op) {
+            case Token.GE:
+            case Token.LE:
+            case Token.GT:
+            case Token.LT:
+              return compareAsNumbers(op, left, right, useTypes);
+            case Token.EQ:
+            case Token.NE:
+              // TODO(moz): We can fold more cases, but testCommutativeOperators() will fail
```
blickly: I'm not sure I follow this TODO. If we did this for non-strings, it would be wrong.
e.g.
0 == null would be folded to true instead of false.

src/com/google/javascript/jscomp/PeepholeFoldConstants.java
```diff
@@ -1064,15 +1064,22 @@ static TernaryValue evaluateComparison(int op, Node left, Node right, boolean us
...
+            case Token.GT:
+            case Token.LT:
+              return compareAsNumbers(op, left, right, useTypes);
             case Token.EQ:
-              return areStringsEqual(left.getString(), right.getString());
-
-            case Token.SHNE:
             case Token.NE:
-              return areStringsEqual(left.getString(), right.getString()).not();
+              // TODO(moz): We can fold more cases, but testCommutativeOperators() will fail
```
blickly: I'm not sure I follow this TODO. If we did this for non-numbers, it would be wrong.
e.g.
'' == null
would be folded to true instead of false.

src/com/google/javascript/jscomp/PeepholeFoldConstants.java
```diff
@@ -1151,16 +1174,21 @@ private static TernaryValue compareAsNumbers(int op, Node left, Node right, bool
...
         Preconditions.checkState(
             left.isNumber() && right.isNumber());
         return TernaryValue.forBoolean(lv == rv);
       case Token.SHNE:
-      case Token.NE:
         Preconditions.checkState(
             left.isNumber() && right.isNumber());
         return TernaryValue.forBoolean(lv != rv);
+      // See http://www.ecma-international.org/ecma-262/6.0/#sec-abstract-equality-comparison
+      case Token.EQ:
```
blickly: I'm a little bit wary of making this method more permissive. Before, if you called it on `0 == null`, you would get a Preconditions failure, but now you'll get the (incorrect) value of true.

src/com/google/javascript/jscomp/PeepholeFoldConstants.java
```diff
@@ -1064,15 +1064,22 @@ static TernaryValue evaluateComparison(int op, Node left, Node right, boolean us
...
+          return compareStrings(op, left, right);
+        } else {
+          // See http://www.ecma-international.org/ecma-262/6.0/#sec-abstract-relational-comparison
           switch (op) {
-            case Token.SHEQ:
+            case Token.GE:
+            case Token.LE:
+            case Token.GT:
+            case Token.LT:
+              return compareAsNumbers(op, left, right, useTypes);
```
blickly: I dislike this pattern in that comparing a number to (something else) doesn't necessarily mean you should compare them as numbers.

Consider, for example,
`'1' < '0' + '2'`

If we (correctly) improved our implementation of Node.getNumberValue to be able to handle string addition, it would cause this code to become incorrect.

src/com/google/javascript/jscomp/PeepholeFoldConstants.java
Dominator008: By "more cases" I meant just booleans (`true` or `false`).

src/com/google/javascript/jscomp/PeepholeFoldConstants.java
Dominator008: By "more cases" I meant just booleans (`true` or `false`).

src/com/google/javascript/jscomp/PeepholeFoldConstants.java
Dominator008: I agree. I've done a large rewrite that is based on the ES6 specs. Hopefully the code structure is better...

blickly:
Overall, we need more tests. A couple of interesting cases:

1) Things with known type, but unknown value, like `+x`, `'' + x`,  or `!x`
2) Things with side effects, like `NaN === foo()`;

Dominator008:
@blickly 
1. I added test cases with known types but unknown values.
2. I added one test with side effects. Note that the first thing `evaluateComparison()` does is bailing out if `left` or `right` may have side effect: https://github.com/Dominator008/closure-compiler/blob/fold/src/com/google/javascript/jscomp/PeepholeFoldConstants.java#L1113-L1116.

src/com/google/javascript/jscomp/PeepholeFoldConstants.java
```diff
@@ -956,222 +955,185 @@ private Node tryFoldComparison(Node n, Node left, Node right) {
...
+            return TernaryValue.FALSE;
+          }
+          if (NodeUtil.isNaN(right)) {
+            return TernaryValue.FALSE;
+          }
+          Double lv = NodeUtil.getNumberValue(left, useTypes);
+          Double rv = NodeUtil.getNumberValue(right, useTypes);
+          if (lv != null && rv != null) {
+            return TernaryValue.forBoolean(lv.doubleValue() == rv.doubleValue());
           }
```
blickly: Missing a break here, I think

Dominator008: Done.

PR #1602, opened 2016-03-01
Don't rename @define's in ProcessEs6Modules

### LINKED ISSUE #1601: Cannot use -D for @define constant from an ES6 module
Repo case:
```
cat > a.js <<EOF
/** @define {string} */ var FOO = 'default foo';
export function test() { console.log(FOO); }
EOF

cat > b.js <<EOF
import * as a from './a';
a.test();
EOF

java -jar compiler.jar --js *.js \
    --language_in=ECMASCRIPT6_STRICT \
    --language_out=ECMASCRIPT5_STRICT \
    -D "FOO='hi'";

WARNING - unknown @define variable FOO

0 error(s), 1 warning(s)
'use strict';var module$a={},FOO$$module$a="default foo";function test$$module$a(){console.log(FOO$$module$a)}module$a.test=test$$module$a;module$a.test();
```

Things work if you use the mangled name:
```
java -jar compiler.jar --js *.js \
    --language_in=ECMASCRIPT6_STRICT \
    --language_out=ECMASCRIPT5_STRICT \
    -D "FOO\$\$module\$a='hi'";

'use strict';var module$a={},FOO$$module$a="hi";function test$$module$a(){console.log(FOO$$module$a)}module$a.test=test$$module$a;module$a.test();
```

Dominator008:
Treat @define's as truly global definitions and don't mangle the names.

Fixes #1601.

[Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1602)

### DIFF
--- a/src/com/google/javascript/jscomp/ProcessEs6Modules.java
+++ b/src/com/google/javascript/jscomp/ProcessEs6Modules.java
@@ -501,6 +501,11 @@ public final class ProcessEs6Modules extends AbstractPostOrderCallback {
 
         Var var = t.getScope().getVar(name);
         if (var != null && var.isGlobal()) {
+          JSDocInfo varInfo = NodeUtil.getBestJSDocInfo(var.getNameNode());
+          // Assume @define's are truly global and don't mangle them.
+          if (varInfo != null && varInfo.isDefine()) {
+            return;
+          }
           // Avoid polluting the global namespace.
           String newName = name + "$$" + suffix;
           if (isShorthandObjLitKey) {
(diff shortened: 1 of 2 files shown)


ChadKillingsworth:
I actually think we should go with the other option and scope `@define` to the module. Treating them as global means that you couldn't have 2 versions of the same module with the same define.

`--define='/path/to/foo.bar=false'`

dimvar:
@concavelenz has a proposal for scoping defines to a module.

concavelenz:
I just roughed it out, it is basically what Chad has here.  The key thing is that the compiler knows where to substitute the value.  goog.define is the other half of the puzzle (for uncompiled code)

concavelenz:
Nothing new here, we aren't going this way.

PR #1608, opened 2016-03-03
Add GwtRunner, the GWT equivalent of CommandlineRunner

Dominator008:
Changed JsComp.gwt.xml to use the SingleScriptLinker so that the GWT compiler only outputs a single JavaScript file (jscomp.nocache.js).

Changed JsCompGwtMain to GwtRunner with "JSCompiler" exported as the JavaScript name through JsInterop. Currently, GwtRunner supports a limited set of compiler flags, including compilationLevel,
warningLevel, languageIn and languageOut among others.

To use the GWT version of the compiler, do a Maven build and include "jscomp.nocache.js" from
target/closure-compiler-gwt-1.0-SNAPSHOT/jscomp/, Eg.:

```html
<script src="jscomp.nocache.js"></script>
<script>
  var output = new JSCompiler().compile("var x;", {
      compilationLevel: 'ADVANCED',
      warningLevel: 'VERBOSE',
      languageIn: 'ECMASCRIPT5_STRICT',
      languageOut: 'ECMASCRIPT3'
  });
  console.log(output);
</script>
```

Note that GwtRunner doesn't load the default externs yet, so ES6 is not currently supported.

As an aside, the Maven build for closure-compiler-gwt is much faster now that I removed most dependencies to gwt-user.jar.

[Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1608)


blickly:
The ResorceLoader/builtin externs part of this PR is hard to land internally, since our externs are located at different paths than externally.
Can we split off the GwtRunner changes without the default externs part to be landed first?

Dominator008:
@blickly Done.

PR #1610, opened 2016-03-03
Rename "--checks-only" to "--checks_only"

Dominator008:
The hyphen is inconsistent with all other compiler flags. Keep "--checks-only" as an alias to avoid breakage for now.

Also fix the help message to say "optimization passes" instead of "compiler passes".

[Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1610)


Dominator008:
@blickly I'd like to remove the `--checks-only` alias when there's no internal usage of it.

PR #1619, opened 2016-03-08
Fix glob patterns with mixing relative and absolute paths.

Dominator008:
See https://groups.google.com/forum/#!topic/closure-compiler-discuss/vIkk0XImMpg.

Also, skip a file or directory if visiting it failed.

Wrap all renameTo() calls with assertTrue() in CommandLineRunnerTest.

[Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1619)

### DIFF
--- a/test/com/google/javascript/jscomp/CommandLineRunnerTest.java
+++ b/test/com/google/javascript/jscomp/CommandLineRunnerTest.java
@@ -1161,8 +1161,8 @@ public final class CommandLineRunnerTest extends TestCase {
     FlagEntry<JsSourceType> jsFile2 = createJsFile("test2", "var b;");
     // Move test2 to the same directory as test1, also make the filename of test2
     // lexicographically larger than test1
-    new File(jsFile2.value).renameTo(new File(
-        new File(jsFile1.value).getParentFile() + File.separator + "utest2.js"));
+    assertTrue(new File(jsFile2.value).renameTo(new File(
+        new File(jsFile1.value).getParentFile() + File.separator + "utest2.js")));
     String glob = new File(jsFile1.value).getParent() + File.separator + "**.js";
     compileFiles(
         "var a;var b;", new FlagEntry<>(JsSourceType.JS, glob));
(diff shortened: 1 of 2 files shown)


Dominator008:
It's kind of hard to write a unit test for this that works internally :(

MatrixFrog:
If you write one that only works externally then we can do some MOE trickery so that it doesn't run internally. Or if it makes more sense, maybe the test can just start with essentially `if (!WINDOWS) {return}`.

Dominator008:
@MatrixFrog It's not Windows specific. I added two tests with `// NOTE(moz): This might not work on Google's internal build system.`.

MatrixFrog:
Okay, well I'll give it a try.

Dominator008:
@MatrixFrog IIRC @blickly said creating files in the current directory worked internally, but not creating directories. I modified the tests to not create directories so they might actually work internally.

test/com/google/javascript/jscomp/CommandLineRunnerTest.java
```diff
@@ -1213,6 +1214,57 @@ public void testGlobJs5() throws IOException, FlagUsageException {
...
+    // Make sure patterns like "!**\./ignored**.js" work
+    String glob1 = "!**\\." + File.separator + "ignored**.js";
+    String glob2 = new File(jsFile1.value).getParent() + File.separator + "**.js";
+    compileFiles(
+        "var a;", new FlagEntry<>(JsSourceType.JS, glob1),
+        new FlagEntry<>(JsSourceType.JS, glob2));
+    ignoredJs.delete();
+  }
+
+  // NOTE(moz): This might not work on Google's internal build system.
```
MatrixFrog: It seems to work. Will review internally.

Dominator008: Nice, thanks :)

MatrixFrog:
Oops, I was wrong. testGlobJs7 is failing but I don't see why. Anyway we'll figure it out.

Dominator008:
@MatrixFrog Might be a permission issue, but in that case testGlobJs6() should also fail.

MatrixFrog:
It is a permissions issue, and if you wrap all your `renameTo` calls in an `assertTrue` then the other test fails too. https://docs.oracle.com/javase/7/docs/api/java/io/File.html#renameTo(java.io.File) recommends this.

Dominator008:
@MatrixFrog Alright.. I wrapped all of the `renameTo` calls with `assertTrue`. We'd have to do some MOE trickery with `testGlobJs6` and `testGlobJs7` then :(

MatrixFrog:
Ran into another problem: This breaks the use of whitelist files. The whitelist file will generally use a path that is relative to the root of the project, but the error is reported using the absolute path, so the whitelist fails to suppress the error.

Dominator008:
@MatrixFrog OK I changed to check the normalized absolute paths for exclusion, but return the original paths. Hopefully it should work now.

PR #1621, opened 2016-03-10
Add the "trailingUnderscore" DiagnosticGroup

Dominator008:
This makes the name more descriptive.

Mark "underscore" as deprecated for now.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1621)


Dominator008:
@SLaks

MatrixFrog:
I agree, but there is a bunch of existing code that uses `@suppress {underscore}` (some in Closure, but a fair amount elsewhere).

Dominator008:
@MatrixFrog OK... Let's mark "underscore" as deprecated for now?

MatrixFrog:
I'd rather just leave it as is. If we have two different groups our users will believe they do different things, even if our docs say otherwise.

PR #1622, opened 2016-03-10
Return early when already found in FindGoogProvideOrGoogModule.

Dominator008:
FindGoogProvideOrGoogModule is now essentially an
AbstractShallowStatementCallback with an extra early return condition.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1622)

### DIFF
--- a/src/com/google/javascript/jscomp/ProcessCommonJSModules.java
+++ b/src/com/google/javascript/jscomp/ProcessCommonJSModules.java
@@ -16,8 +16,8 @@
 package com.google.javascript.jscomp;
 
 import com.google.common.base.Preconditions;
-import com.google.common.collect.Iterables;
 import com.google.common.collect.HashMultiset;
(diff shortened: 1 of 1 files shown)


MatrixFrog:
My intuition is that it's not worth it. All its doing is setting a flag. The particular callback can set its own flag just as easily -- and give it a more descriptive name, in this case `found`.

concavelenz:
I agree with Tyler here, in the few cases where an early termination useful, also NodeUtil.has might be a better match in those cases.

Dominator008:
@MatrixFrog @concavelenz OK I've reverted the NodeTraversal changes.

MatrixFrog:
Thanks, lgtm, will bring it in tomorrow.

MatrixFrog:
Submitted internally, will be MOE'd out soon.

Issue #1615, opened 2016-03-05
Allow more kinds of @suppress on non-function nodes.


Dominator008:
Most of the lint warnings from Closure Library are "@suppress is not allowed here", and many of the use cases seem reasonable to me. Eg.:
```js
/** @suppress {unusedLocalVariables} */
var dummy = 3;
```

We also need to document where @suppress is actually allowed, pending TODO by @MatrixFrog in https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/lint/CheckJSDocStyle.java#L45-L48.

dimvar:
I wrote a quick wiki page to say where @suppress is currently supported. https://github.com/google/closure-compiler/wiki/@suppress-annotations

MatrixFrog:
I'd like to keep the rules for where `@suppress` is allowed simple enough that people can keep them in their head. I don't want to get into a state where you have to remember that some suppresses are valid in a bunch of places where others are only on functions, and you get mixed up about which is which, all the time. But stretching the rules beyond where they currently are, a little bit, is fine.

Dominator008:
@MatrixFrog There are currently 187 lint warnings from Closure Library, and 121 of them are "`@suppress` annotation not allowed here". Most of those `@suppress`'s that I saw were pretty reasonable, which makes me wonder if CheckJSDocStyle is being too strict...

MatrixFrog:
They are probably in places where gjslint understood them but our linter does not. Getting our linter to understand most/all of the gjslint places is a reasonable request.

PR #1624, opened 2016-03-10
Add "extraProvide" to allowed suppressions.

Dominator008:
We still need to implement a proper lint check for "extraProvide", but allow this for now to unblock fixes to Closure Library.

Related to #1615.

---

This change is [Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1624)


MatrixFrog:
I'm not sure what the point of the extraProvide check would be, to be honest. Does it happen often that someone goog.provides a namespace but then doesn't actually define it? It seems like something you'd only do on purpose, for a file that has side effects, and users would goog.require it just to get the side effects. @concavelenz maybe you know the history of that gjslint check?

Dominator008:
Most of the `extraProvide`s I see in Closure Library are in test files, and have the pattern like:

```js
/** @suppress {extraProvide} */
goog.provide('goog.crypt.CbcTest');
...
goog.setTestOnly('goog.crypt.CbcTest');
```
I guess the internal testing framework needs the `goog.provide`. According to http://google.github.io/closure-library/api/namespace_goog.html#setTestOnly, gjslint should already ignore that extra provide so the suppression isn't necessary.

There are a few other `extraProvide`s in non-test code.

concavelenz:
Looks like this was already merged.

PR #1476, opened 2016-02-01
Force rewriting of ES6 or CommonJS modules which are imported

ChadKillingsworth:
A file is considered a module if it is imported (whether it exports anything or not). This forces module rewriting to occur on files which are imported, but were otherwise unrecognized as a module.

This is accomplished by assuming that after module rewriting, any input that has no provide statements is a potential module. The normalized module identifier for each of these inputs is calculated. If any input attempts to require the normalized module name for one of these files, the file is treated as an ES6 module and rewriting is forced. Since the module didn't export anything, it doesn't matter which type of module it is rewritten as.

This PR also removes the `USELESS_USE_STRICT_DIRECTIVE` warning for ES6 modules. As a file can now be treated as a script or a module. This caused warnings when such code was imported. In addition, "use strict" directives are often inserted by IDE tools automatically and are already removed by the compiler.

Fixes #1220

[Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1476)

PR #1626, opened 2016-03-10
Only re-write a file as a CommonJS module if it exports a symbol

ChadKillingsworth:
Full module rewriting of a CommonJS module should only occur if the module exports a symbol. `require` can be used within other module types as a module loader and its presence should not infer that the entire file is a CommonJS module.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1626)

### DIFF
--- a/src/com/google/javascript/jscomp/ProcessCommonJSModules.java
+++ b/src/com/google/javascript/jscomp/ProcessCommonJSModules.java
@@ -81,11 +81,8 @@ public final class ProcessCommonJSModules implements CompilerPass {
   public void process(Node externs, Node root) {
     FindGoogProvideOrGoogModule finder = new FindGoogProvideOrGoogModule();
     NodeTraversal.traverseEs6(compiler, root, finder);
-    if (finder.found) {
-      return;
-    }
     NodeTraversal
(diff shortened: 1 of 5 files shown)


src/com/google/javascript/jscomp/ProcessCommonJSModules.java
```diff
@@ -329,9 +331,18 @@ private void visitScript(NodeTraversal t, Node script) {
 
       String moduleName = inputToModuleName(t.getInput());
 
-      // Rename vars to not conflict in global scope.
-      NodeTraversal.traverseEs6(compiler, script, new SuffixVarsCallback(
-          moduleName));
+      boolean hasExports = !(moduleExportRefs.isEmpty() && exportRefs.isEmpty());
+
+      // Rename vars to not conflict in global scope - but only if the script exports something.
```
Dominator008: So if two files both have
```js
var name = require('name');
```
but neither of them exports anything, would it be a name conflict?

ChadKillingsworth: Yes - they would both exist in the global scope. Unless, it's actually imported by something else: see #1476

Issue #1625, opened 2016-03-10
Rollforward PR #1266, #1476 and #1576.


ChadKillingsworth:
#1476 and #1266 were rolled back with 641cf61b99280b8bb4b62ecce873e51208501fec

What needs to happen to re-apply those PRs?

ChadKillingsworth:
Yes - I'll rebase all 3 to the top of the tree and resubmit.

Dominator008:
All three PRs have been rolled forward \o/

Issue #1220, opened 2015-10-19
When using ECMAScript 6 modules, cannot import a module that has no exports


Pauan:
Consider these two modules:

```
// foo.js
alert("Hi");
```

```
// bar.js
import "./foo";
```

----

When compiling with `--language_in ECMASCRIPT6` I get the following error:

```
bar.js:1: ERROR - required "module$foo" namespace never provided
import "./foo";
^
```

However, if I add an export to `foo.js`, then it works correctly:

```
// foo.js
alert("Hi");

export const foo = 1;
```

----

In my application, I'm using `import "..."` to import a module solely for its side effects, so of course the module doesn't export anything, which is how I found this error.

dimvar:
The compiler only considers a file to be an ES6 module if it has an import or an export statement. See:
https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/ProcessEs6Modules.java#L41

I'm not sure if we plan to support your usecase or not.

Pauan:
What about treating *all* input JS files as modules, when using `--language_in ECMASCRIPT6`?

bfricka:
@Pauan You have to `export` something for it to be `import`able, thus 'foo.js' is not considered a module.

http://www.2ality.com/2015/07/es6-module-exports.html

concavelenz:
"export default undefined;" would make it clear that it is intended to be a module.  Do you have use case where you explicitly don't want to included any export.  Otherwise, "--language_in ECMASCRIPT6" is likely going to mix and match es6 modules for a long time.  I think a different command-line flag like: --js-module= ... in addition to --js would be a reasonable addition

Pauan:
@bfricka In older versions of the ES6 specification, whether a file is a module or not was determined by heuristics (such as whether an `import` or `export` statement exists in the file).

However, in the current (and final) version of the specification, that is no longer the case.

Instead, what determines whether a file is a module or not is based upon how it is imported.

If you have two files `foo.js` and `bar.js`:

```
// foo.js
import "./bar";
```

```
// bar.js
alert("hi!");
```

Then the file `bar.js` is treated as a module, because it is imported by `foo.js`

If you instead load `bar.js` by using a `<script>` tag, then it is treated as a script (and not a module).

That means you cannot use the *contents* of the file to determine if it is an ES6 module, you must instead determine it by *how the file is being imported*.

[...]

PR #1628, opened 2016-03-10
Rollforward of #1476

ChadKillingsworth:
Rollforward #1476 which was rolled back as part of 641cf61

Partially addresses #1625

This PR may have been the biggest reason for the rollback. In particular, it may be subject to the same restrictions as #1573 which caused a circular dependency problem with internal Google builds. If so, I can quite easily refactor this in the same manner to avoid that issue.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1628)

### DIFF
--- a/src/com/google/javascript/jscomp/ProcessEs6Modules.java
+++ b/src/com/google/javascript/jscomp/ProcessEs6Modules.java
@@ -324,7 +324,7 @@ public final class ProcessEs6Modules extends AbstractPostOrderCallback {
       return;
     }
 
-    setStrictModeDirective(t, script);
+    setStrictModeDirective(script);
 
     Preconditions.checkArgument(scriptNodeCount == 1,
         "ProcessEs6Modules supports only one invocation per "
@@ -402,7 +402,7 @@ public final class ProcessEs6Modules extends AbstractPostOrderCallback {
     compiler.reportCodeChange();
   }
 
-  private static void setStrictModeDirective(NodeTraversal t, Node n) {
+  private static void setStrictModeDirective(Node n) {
     Preconditions.checkState(n.isScript(), n);
     Set<String> directives = n.getDirectives();
     if (directives != null && directives.contains("use strict")) {
--- a/src/com/google/javascript/jscomp/ProcessEs6Modules.java
+++ b/src/com/google/javascript/jscomp/ProcessEs6Modules.java
@@ -50,11 +50,6 @@ public final class ProcessEs6Modules extends AbstractPostOrderCallback {
           "JSC_LHS_OF_GOOG_REQUIRE_MUST_BE_CONST",
           "The left side of a goog.require() must use ''const'' (not ''let'' or ''var'')");
 
-  static final DiagnosticType USELESS_USE_STRICT_DIRECTIVE =
-      DiagnosticType.warning(
-          "JSC_USELESS_USE_STRICT_DIRECTIVE",
-          "'use strict' is unnecessary in ES6 modules.");
-
   static final DiagnosticType NAMESPACE_IMPORT_CANNOT_USE_STAR =
       DiagnosticType.error(
           "JSC_NAMESPACE_IMPORT_CANNOT_USE_STAR",
@@ -88,6 +83,7 @@ public final class ProcessEs6Modules extends AbstractPostOrderCallback {
   private Set<String> alreadyRequired = new HashSet<>();
 
   private boolean isEs6Module;
+  private boolean forceRewrite;
 
   private boolean reportDependencies;
 
@@ -110,13 +106,18 @@ public final class ProcessEs6Modules extends AbstractPostOrderCallback {
     this.reportDependencies = reportDependencies;
   }
 
-  public void processFile(Node root) {
+  /**
+   * If a file contains an ES6 "import" or "export" statement, or the forceRewrite
+   * option is true, rewrite the source as a module.
+   */
+  public void processFile(Node root, boolean forceRewrite) {
     FindGoogProvideOrGoogModule finder = new FindGoogProvideOrGoogModule();
     NodeTraversal.traverseEs6(compiler, root, finder);
     if (finder.isFound()) {
       return;
     }
-    isEs6Module = false;
+    this.forceRewrite = forceRewrite;
+    isEs6Module = forceRewrite;
     NodeTraversal.traverseEs6(compiler, root, this);
   }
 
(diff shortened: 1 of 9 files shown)


ChadKillingsworth:
Rebased again to fix the merge conflict.

ChadKillingsworth:
Rebased again to fix the merge conflict. Keeping up with a rapidly moving target is difficult.

blickly:
One issue I noticed with this is that it causes us to call JsFileParser (in particular the setIncludeGoogBase method) more often, which is not yet implemented in the GWT version of the compiler.

ChadKillingsworth:
@Dominator008 Do you know why this isn't implemented for GWT?

https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/gwt/super/com/google/javascript/jscomp/deps/JsFileParser.java#L27

That's the line that's failing.

Dominator008:
@ChadKillingsworth I think that particular method can be implemented.

ChadKillingsworth:
@blickly @Dominator008 Requesting the provided/required symbol lists was re-parsing the files. But module re-writing had already notified the compiler input class of those symbols. I added some new methods to CompilerInput to prevent a full regeneration of dependency information when all we really want was the information it already knew about.

@blickly If this seems to work, I'll squash the commits and rebase to the top of the tree.

Dominator008:
@ChadKillingsworth https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/gwt/super/com/google/javascript/jscomp/deps/JsFileParser.java#L30 Will this `parseFile()` be called? This will be more work to implement since it uses regexes.

ChadKillingsworth:
@Dominator008 No - I completely sidestepped the entire `JsFileParser` usage with this change. When `input.getProvides()` is called, it creates a `JsFileParser` instance to scan the code to find all `goog.require` and `goog.provide` statements. That is completely un-necessary since module rewriting already reports those as they are generated. See https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/ProcessCommonJSModules.java#L358-L359

Dominator008:
@ChadKillingsworth I see, so we don't have to provide a GWT version of `setIncludeGoogBase()` for now.

ChadKillingsworth:
@Dominator008 Correct - I found this out after my question earlier today. I had to trace through the code to discover how this all worked.

blickly:
Found another issue here with creating URLs with invalid characters:

Here's an example stack:
```
java.lang.IllegalArgumentException: Illegal character in scheme name at index 9: synthetic_module_$48:mod:Stub
		at java.net.URI.create(URI.java:852)
		at com.google.javascript.jscomp.ModuleIdentifier.forFile(ModuleIdentifier.java:97)
		at com.google.javascript.jscomp.Compiler.parseInputs(Compiler.java:1369)
		at com.google.javascript.jscomp.Compiler.parse(Compiler.java:730)
		at com.google.javascript.jscomp.JSCompiler.parse(JSCompiler.java:156)
		at com.google.javascript.jscomp.Compiler.compileInternal(Compiler.java:691)
		at com.google.javascript.jscomp.Compiler.access$000(Compiler.java:85)
		at com.google.javascript.jscomp.Compiler$2.call(Compiler.java:662)
		at com.google.javascript.jscomp.Compiler$2.call(Compiler.java:659)
		at com.google.javascript.jscomp.CompilerExecutor$2.call(CompilerExecutor.java:93)
		at java.util.concurrent.FutureTask.run(FutureTask.java:266)
		at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
		at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
		at java.lang.Thread.run(Thread.java:745)
```

MatrixFrog:
I hit a similar issue when merging another PR. Look at the comments on cl/116700722

ChadKillingsworth:
[ES6ModuleLoader.java](https://github.com/google/closure-compiler/pull/1628/files#diff-51277464fa4940394ed596618d6d0b25R97) already had an implementation for invalid characters. If this still isn't enough, I can simply add a try/catch block around this call and skip any files whose name won't convert to a URI.

I rebased this to the top of the tree to handle the merge conflicts. I also discovered that the tests were treating this like an optimization/checks pass. I switched the tests to mirror the ES6 module tests so that the rewriting occurs first thing.

src/com/google/javascript/jscomp/ProcessEs6Modules.java
```diff
@@ -401,11 +402,11 @@ private void visitScript(NodeTraversal t, Node script) {
     compiler.reportCodeChange();
   }
 
-  private static void checkStrictModeDirective(NodeTraversal t, Node n) {
+  private static void setStrictModeDirective(NodeTraversal t, Node n) {
```
blickly: Looks like the NodeTraversal argument is no longer needed here

MatrixFrog:
This is a bit weird because ES6 transpilation will now behave differently on the same file when you run it as part of a full compilation, vs. in a file-by-file way.

We need to make sure we know what the implications of that are. Maybe we can just say that if your workflow requires file-by-file transpilation, and you need to import a module that exports nothing, then you have to add `export default undefined;` or something like that. But it would be nice not to have such a requirement.

ChadKillingsworth:
@MatrixFrog Currently we already mention that you do have to export something for a module to be treated as such. There is no way to determine if a file is a module unless you know that:

 1. It exports something or
 2. something else imports it

The file-by-file transpilation concept is incompatible with modules who are imported for side effects only. I know of no way around that. See #1220

src/com/google/javascript/jscomp/ProcessEs6Modules.java
ChadKillingsworth: Fixed

MatrixFrog:
Another option we've considered a few times is some kind of explicit annotation in the `@fileoverview` block to tell the compiler "treat this file as a module" but that's nearly as hacky as `export default undefined`. So I guess I'm okay with this as long as it's documented.

Dominator008:
@MatrixFrog @ChadKillingsworth Will something like an `'isEs6Module` directive at the beginning of the file work? That's also hacky but clearer I guess.

ChadKillingsworth:
@Dominator008 At that point there is no difference from exporting undefined. You lose the ability to treat the file as both script or module.

PR #1629, opened 2016-03-10
Rollforward #1266

ChadKillingsworth:
Rollforward #1266 which was rolled back as part of 641cf61

Partially addresses #1625

---
This change is [Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1629)

### DIFF
--- a/src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
+++ b/src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
@@ -119,6 +119,10 @@ public abstract class AbstractCommandLineRunner<A extends Compiler,
   static final DiagnosticType NO_TREE_GENERATED_ERROR = DiagnosticType.error(
       "JSC_NO_TREE_GENERATED_ERROR",
       "Code contains errors. No tree was generated.");
+  static final DiagnosticType INVALID_MODULE_SOURCEMAP_PATTERN = DiagnosticType.error(
+      "JSC_INVALID_MODULE_SOURCEMAP_PATTERN",
+      "When using --module flags, the --create_source_map flag must contain "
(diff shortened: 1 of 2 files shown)


src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
```diff
@@ -1288,9 +1297,11 @@ private void outputModuleBinaryAndSourceMaps(List<JSModule> modules, B options)
 
     // When the json_streams flag is specified, sourcemaps are always generated
     // per module
-    if (!(shouldGenerateMapPerModule(options) || isOutputInJson())) {
-      mapFileOut = fileNameToOutputWriter2(
-          expandSourceMapPath(options, null));
+    if (!(shouldGenerateMapPerModule(options) || options.sourceMapOutputPath == null ||
```
Dominator008: "||" should be on the next line.

src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
```diff
@@ -1275,8 +1283,9 @@ void outputJsonStream() throws IOException {
     }
   }
 
-  private void outputModuleBinaryAndSourceMaps(List<JSModule> modules, B options)
-      throws IOException {
+  private DiagnosticType outputModuleBinaryAndSourceMaps(
```
Dominator008: I think it would be easier to read if this method returns an int (0 or 1), and just reports the error instead of returning it.

src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
ChadKillingsworth: Done

PR #1630, opened 2016-03-10
Rewrite CommonJS require.ensure calls to be an IIFE

ChadKillingsworth:
`require.ensure` is defined by the CommonJS spec: http://wiki.commonjs.org/wiki/Modules/Async/A. It is implemented by the [InjectJS module loader](http://www.injectjs.com/docs/0.7.x/cjs/require.ensure.html) and is also used by WebPack to support async loading of CommonJS modules.

It's a key component to allowing CommonJS modules to work in a browser uncompiled.

After compilation, it serves no use as the modules are all bundled. This removes the calls to `require.ensure` during module rewriting.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1630)

### DIFF
--- a/src/com/google/javascript/jscomp/ProcessCommonJSModules.java
+++ b/src/com/google/javascript/jscomp/ProcessCommonJSModules.java
@@ -198,6 +198,27 @@ public final class ProcessCommonJSModules implements CompilerPass {
         visitRequireCall(t, n, parent);
       }
 
+      // Finds calls to require.ensure which is a method to allow async loading of
+      // CommonJS modules:
+      //
+      // require.ensure(['/path/to/module1', '/path/to/module2'], function(require) {
+      //    var module1 = require('/path/to/module1');
+      //    var module2 = require('/path/to/module2');
+      // });
+      //
+      // will be rewritten as an IIFE
+      //
+      // (function() {
+      //   var module1 = require('/path/to/module1');
+      //   var module2 = require('/path/to/module2');
+      // })()
+      //
+      // See http://wiki.commonjs.org/wiki/Modules/Async/A
+      //     http://www.injectjs.com/docs/0.7.x/cjs/require.ensure.html
+      if (n.isCall() && n.getChildCount() == 3
+          && n.getFirstChild().matchesQualifiedName("require.ensure")) {
+        visitRequireEnsureCall(n);
+      }
 
       // Detects UMD pattern, by checking for CommonJS exports and AMD define
       // statements in if-conditions and rewrites the if-then-else block as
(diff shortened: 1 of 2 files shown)


src/com/google/javascript/jscomp/ProcessCommonJSModules.java
```diff
@@ -192,6 +192,28 @@ public void visit(NodeTraversal t, Node n, Node parent) {
...
+      // will be rewritten as an IIFE
+      //
+      // (function() {
+      //   var module1 = require('/path/to/module1');
+      //   var module2 = require('/path/to/module2');
+      // })()
+      //
+      // See http://wiki.commonjs.org/wiki/Modules/Async/A
+      //     http://www.injectjs.com/docs/0.7.x/cjs/require.ensure.html
+      if (n.isCall() && n.getChildCount() == 3 && n.getFirstChild().isGetProp() &&
```
Dominator008: "&&" should be on the next line. Also, check out matchesQualifiedName() in Node.

ChadKillingsworth: Done

src/com/google/javascript/jscomp/ProcessCommonJSModules.java
```diff
@@ -192,6 +192,28 @@ public void visit(NodeTraversal t, Node n, Node parent) {
...
+      // will be rewritten as an IIFE
+      //
+      // (function() {
+      //   var module1 = require('/path/to/module1');
+      //   var module2 = require('/path/to/module2');
+      // })()
+      //
+      // See http://wiki.commonjs.org/wiki/Modules/Async/A
+      //     http://www.injectjs.com/docs/0.7.x/cjs/require.ensure.html
+      if (n.isCall() && n.getChildCount() == 3 && n.getFirstChild().isGetProp()
```
Dominator008: I think this isGetProp() is covered by matchesQualifiedName()?

src/com/google/javascript/jscomp/ProcessCommonJSModules.java
```diff
@@ -192,6 +192,28 @@ public void visit(NodeTraversal t, Node n, Node parent) {
...
+          && n.getFirstChild().matchesQualifiedName("require.ensure")
+          && n.getSecondChild().isArrayLit() && n.getChildAtIndex(2).isFunction()) {
```
Dominator008: Are these checks (isArrayLit() and isFunction()) necessary?

src/com/google/javascript/jscomp/ProcessCommonJSModules.java
ChadKillingsworth: Looks like it is. Fixed.

src/com/google/javascript/jscomp/ProcessCommonJSModules.java
ChadKillingsworth: In the future, I plan on only removing this call if all of the required modules are provided as source. Once we have the concept of external modules, then this call would need to be left at least partially intact if at least one of the module references in the array literal was external. We won't be able to rewrite calls where the first argument is anything but an array literal.

Also, the second argument must be a function with exactly one parameter named `require`. Aliasing `require` is not something I wanted to tackle right now.

Most of these checks could be deferred to the implementation of `visitRequireEnsureCall` though.

Dominator008: Yeah, I'd prefer checking here up to `matchesQualifiedName()`, moving the rest to `visitRequireEnsureCall`, and updating the comment about supported pattern there.

ChadKillingsworth: Done

concavelenz:
So this is the CommonJS version of AMD modules?  I hadn't run across it.

ChadKillingsworth:
@concavelenz Exactly. AMD modules are generally considered over-engineered and difficult to work with. I'm seeing more and more code using CommonJS for browser based projects. Most require rewriting to execute in the browser, but a couple of projects used this signature. I've switched my new projects over from ES6 modules to CommonJS with this signature. The InjectJS module loader works very well and is only needed for uncompiled mode support.

ChadKillingsworth:
Rebased to fix merge conflicts.

PR #1634, opened 2016-03-11
Make static methods in NodeUtil public for people writing custom compiler passes

### LINKED ISSUE #1632: NodeUtils methods are not publicly accessible.
The NodeUtils class has many utility methods that could be very useful to custom compiler passes, but most of them are not marked public, restricting their use. Could you make these accessible?

StefanLiebenberg:
Fixes #1632.

This change is [Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1634)


StefanLiebenberg:
I signed it!

Dominator008:
Thanks! This will be reviewed and submitted internally at Google, then pushed out to the GitHub repo.

dimvar:
I'd rather not make every single method public, it makes it harder for us to change things. Can you only make the few methods you need public?

StefanLiebenberg:
Its not just a few methods I need public, I want access to most of them. I have changed some of the less interesting ones back, but mostly these are very useful methods and given how long it takes for changes here to get to maven central, I'd really rather get as many as I can in one go.

MatrixFrog:
I'd recommend working from head in that case, and then you can make whatever local modifications you want.

concavelenz:
Generally, we need to carefully review the API and method names.  Some of the methods are badly named or some such

dimvar:
I'm hesitant to approve a PR that switches almost 150 methods. Please send smaller PRs for the methods you need at the moment, otherwise we likely won't approve them. Can you tell us more about your use case?

Also, as Tyler mentioned, if you are writing custom passes, it's better to build the compiler from head, so that you don't wait for the maven releases.

StefanLiebenberg:
I use maven (and sometimes gradle) and I want to stay up-to-date on the changes here and I don't want to get caught up in maintaining a fork, so I don't feel that using the compiler 'from head' is a viable solution for my situation. I'll try again to drop down the amount of public methods to the bare minimum that will cover our next iteration of changes, which should be about ~10 of them.

concavelenz:
Maybe something you could contribute so you don't need to maintain it externally?

dimvar:
> I'll try again to drop down the amount of public methods to the bare minimum that will cover our next iteration of changes, which should be about ~10 of them.

Sounds good. Just reply on the thread when you do this and I'll accept the PR.

And sorry for the hassle, but having many things be public makes it hard for us to make changes. (In fact, I can't guarantee that whatever methods you make public now will have the same signature indefinitely.)

StefanLiebenberg:
> Just reply on the thread when you do this

done

dimvar:
LGTM. I will bring this internally and submit it.

Dominator008:
Closed by https://github.com/google/closure-compiler/commit/11a746d780752cb3746d919f42ec14185f42e717.

PR #1641, opened 2016-03-11
Changed the GWT ResourceLoader to include most of the built-in externs.

Dominator008:
Modify GwtRunner to load the built-in externs and change the default
languageIn to ES6.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1641)

### DIFF
--- a/pom-gwt.xml
+++ b/pom-gwt.xml
@@ -209,8 +209,8 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-resources-plugin</artifactId>
 
-        <!-- Move Messages.properties where ScriptRuntime.java expects it. -->
         <executions>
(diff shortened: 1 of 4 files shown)


Dominator008:
This requires internal BUILD change.

blickly:
Just letting you know, that we have a slightly different approach to GWT resources like these internally that @shicks is working on that will conflict with this PR (but I think will be easier to support for both internal and external use)

Dominator008:
@blickly OK sounds good!

blickly:
Closing this, as we're planning on including these resources in a different way, as started in https://github.com/google/closure-compiler/commit/59383e4bebcee4bc5ea48ea5d4a0bc94c0f03621

PR #1645, opened 2016-03-12
Simplify IObject<(string|number),T> to IObject<string,T> in the externs

Dominator008:
These constructors already implement `IArrayLike<T>`, and `IArrayLike<T>` extends `IObject<number,T>`.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1645)

### DIFF
--- a/externs/browser/gecko_dom.js
+++ b/externs/browser/gecko_dom.js
@@ -956,7 +956,7 @@ Navigator.prototype.javaEnabled = function() {};
 
 /**
  * @constructor
- * @implements {IObject<(string|number),!Plugin>}
+ * @implements {IObject<string, !Plugin>}
  * @implements {IArrayLike<!Plugin>}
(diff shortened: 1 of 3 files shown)


concavelenz:
Why do you think this is an improvement?

Dominator008:
@concavelenz For consistency? People might write both `IObject<string,T>` and `IObject<(string|number),T>` in different places.

concavelenz:
IObject<(string|number), T> is appropriate here because the index can be either the node id or the index in the node list.  I'm not sure why you feel changing it would be more consistent, can you explain in more detail?

Dominator008:
@concavelenz I felt that since `IArrayLike<T>` already extends `IObject<number,T>`, having `IObject<(string|number), T>` is kind of redundant, but I didn't convince myself 100%.

concavelenz:
I'd like to keep it as is.

PR #1646, opened 2016-03-13
Remove the special case for interfaces in isSubTypeOf() in FunctionType

### LINKED ISSUE #1643: Clearer error message when an ES6 class extends an interface
https://jscomp-debug.googleplex.com/#input0%3D%252F**%2520%2540interface%2520*%252F%250Aclass%2520MyInterface%2520%257B%257D%250A%250Aclass%2520MyClass%2520extends%2520MyInterface%2520%257B%257D%26input1%26conformanceConfig%26externs%26refasterjs-template%26includeDefaultExterns%3D1%26CHECK_SYMBOLS%3D1%26CHECK_TYPES%3D1%26LANG_IN_IS_ES6%3D1%26MISSING_PROPERTIES%3D1%26PRESERVE_TYPE_ANNOTATIONS%3D1%26PRETTY_PRINT%3D1%26TRANSPILE%3D1

Currently this gives two different errors, and one of them mentions `$jscomp.inherits` which is totally unhelpful and meaningless to the user. We should:
1. Only report one error
2. Make sure it's clear
3. (bonus) Suggest that the user probably meant to include an `@implements` annotation in their JSDoc, rather than writing `extends`

Dominator008:
The special case doesn't quite make sense and it's untested. it causes TypeValidator to generate an obscure warning on this code:
```js
/** @interface */ function Foo() {};
var /** Function */ x = Foo;
```
```
input0:3: WARNING - initializing variable
found   : function (this:Foo): ?
required: (Function|null)
var /** Function */ x = Foo;
                            ^
```
It also causes TypeValidator to generate an extra warning when a class accidentally extends an interface.

NTI doesn't have this special handling.

Fixes #1643.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1646)

### DIFF
--- a/src/com/google/javascript/rhino/jstype/FunctionType.java
+++ b/src/com/google/javascript/rhino/jstype/FunctionType.java
@@ -1069,15 +1069,6 @@ public class FunctionType extends PrototypeObjectType implements FunctionTypeI {
 
     if (that.isFunctionType()) {
       FunctionType other = that.toMaybeFunctionType();
-      if (other.isInterface()) {
-        // Any function can be assigned to an interface function.
-        return true;
-      }
-      if (isInterface()) {
-        // An interface function cannot be assigned to anything.
-        return false;
-      }
-
       return treatThisTypesAsCovariant(other, implicitImplCache)
           && this.call.isSubtype(other.call, implicitImplCache);
     }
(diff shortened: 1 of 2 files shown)


blickly:
With this change, would the type checker allow:
```javascript
/** @interface */ function Foo() {};
var /** Function */ x = Foo;
window.x = x();
```
b/c this is something that we probably do want to help people avoid (so that they can be sure that interfaces get stripped completely, which is usually what people want).

blickly:
Closing this PR in favor of making the NTI stricter instead, which is being tracked by:
https://github.com/google/closure-compiler/issues/1680

PR #1648, opened 2016-03-13
Remove @nosideeffects from functions that @throws in externs

Dominator008:
By definition, throwing an exception is a side effect, so these two
annotations are contradictory.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1648)


Dominator008:
@concavelenz Ping. I think we should also add a check in `CheckJSDoc` for these incompatibilities.

PR #1651, opened 2016-03-14
Update versions for Maven plugins

Dominator008:
The current versions are a bit outdated. In particular,
maven-compiler-plugin is pinned at version 3.1, which has a known bug
with incremental compilation \[1\], causing all code to be compiled in
every Maven build. Unfortunately it hasn't been fixed yet.  

\[1\]: https://issues.apache.org/jira/browse/MCOMPILER-209


MatrixFrog:
Easier to do internally, so closing this.

PR #1652, opened 2016-03-14
Update the versions of external libraries for Ant and Maven

Dominator008:
Many of these are quite outdated.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/google/closure-compiler/1652)


MatrixFrog:
Obsolete since Ant support is gone. If you still want to update anything please send it internally.

PR #1658, opened 2016-03-15
Support class name alias in Es6ToEs3Converter

### LINKED ISSUE #1546: Class identifier not properly scoped in class expression

Dominator008:
Define the class name alias in member functions, eg.:
Before conversion:

``` js
var C = class Clazz {
  foo() {}
};
```

After conversion:

``` js
/** @constructor @struct */
var C = function() {
  var Clazz = C;
}
C.prototype.foo = function() {
  var Clazz = C;
}
```

Fixes #1546.

---

This change is [Reviewable](https://reviewable.io/reviews/google/closure-compiler/1658)

### DIFF
--- a/src/com/google/javascript/jscomp/Es6ToEs3Converter.java
+++ b/src/com/google/javascript/jscomp/Es6ToEs3Converter.java
@@ -154,7 +154,7 @@ public final class Es6ToEs3Converter implements NodeTraversal.Callback, HotSwapC
         visitStringKey(n);
         break;
       case Token.CLASS:
-        visitClass(n, parent);
+        visitClass(t, n, parent);
         break;
       case Token.ARRAYLIT:
       case Token.NEW:
@@ -527,7 +527,7 @@ public final class Es6ToEs3Converter implements NodeTraversal.Callback, HotSwapC
    *   <li>The constructor is built.
    * </ol>
    */
-  private void visitClass(final Node classNode, final Node parent) {
+  private void visitClass(NodeTraversal t, final Node classNode, final Node parent) {
     checkClassReassignment(classNode);
     // Collect Metadata
     ClassDeclarationMetadata metadata = ClassDeclarationMetadata.create(classNode, parent);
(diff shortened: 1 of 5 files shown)


blickly:
I was thinking that we would synthesize a class name. Otherwise how can we support changes to the var, or class expressions with no var/let/const statement?

Dominator008:
@blickly Can you give some examples?

blickly:
Yeah, that's one case.
Another is, e.g.:
```javascript
var Foo = class InnerName {
    copy() { return new InnerName; }
}
Foo = class {};
```

If not for cases like these (mutable name and anonymous classes), what's the utility of using the InnerName at all?

Dominator008:
Oh I see, I was only thinking about short aliases, like:
```
goog.foo.bar.Baz = class B {
    copy() { return new B; }
}
```

blickly:
As long as the type of the second class's constructor is `function(new:Foo)`, I would think it would be valid. That's how mocks usually work.

concavelenz:
Shall we close this?

PR #1663, opened 2016-03-15
Normalize assignment shorthand like "x += 1" to "x = x + 1"

Dominator008:
Normalize these in Normalize to free up potential optimizations.
Denormalize them in Denormalize later.

Related to #250.

---
This change is [Reviewable](https://reviewable.io/reviews/google/closure-compiler/1663)

### DIFF
--- a/src/com/google/javascript/jscomp/Denormalize.java
+++ b/src/com/google/javascript/jscomp/Denormalize.java
@@ -26,11 +26,14 @@ import com.google.javascript.rhino.Token;
  * normalization pass that are not handled by other passes (such as
  * CollapseVariableDeclarations) to avoid making the resulting code larger.
  *
- * Currently this pass only does one thing pushing statements into for-loop
- * initializer. This:
+ * Currently this pass only does two things:
+ *
+ * 1. Push statements into for-loop initializer. This:
  *   var a = 0; for(;a<0;a++) {}
  * becomes:
  *   for(var a = 0;a<0;a++) {}
+ *
+ * 2. Fold assignments like x = x + 1 into x += 1
  *
  * @author johnlenz@google.com (johnlenz)
  */
@@ -55,6 +58,7 @@ class Denormalize implements CompilerPass, Callback {
   @Override
   public void visit(NodeTraversal t, Node n, Node parent) {
     maybeCollapseIntoForStatements(n, parent);
+    maybeCollapseAssignShorthand(n, parent);
   }
 
   /**
@@ -122,6 +126,22 @@ class Denormalize implements CompilerPass, Callback {
       forNode.replaceChild(oldInitializer, newInitializer);
 
       compiler.reportCodeChange();
+    }
+  }
+
+  private void maybeCollapseAssignShorthand(Node n, Node parent) {
+    if (n.isAssign() && n.getFirstChild().isName()
+        && NodeUtil.hasCorrespondingAssignmentOp(n.getLastChild())
+        && n.getLastChild().getFirstChild().isName()) {
+      Node op = n.getLastChild();
+      int assignOp = NodeUtil.getAssignOpFromOp(op);
+      if (n.getFirstChild().getString().equals(op.getFirstChild().getString())) {
+        op.setType(assignOp);
+        Node opDetached = op.detachFromParent();
+        opDetached.setJSDocInfo(n.getJSDocInfo());
+        parent.replaceChild(n, opDetached);
+        compiler.reportCodeChange();
+      }
     }
   }
 }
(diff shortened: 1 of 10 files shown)


dimvar:
> Looks like the inlining stopped at "1234".

That's super weird, there is no reason to stop at that particular point. I'll take a look.

Dominator008:
@dimvar I think the issue is that FlowSensitiveInlineVariables is still backing off. Since this is in IntegrationTest, the fact that we run the pass twice in DefaultPassConfig after https://github.com/google/closure-compiler/commit/8143ba9cfea599365365ff0834716b3c38e4efa6 made the result look interesting.

After the first run, the pass inlined a little bit and the output becomes (after denormalization):
```js
window.a = function() { var b; b = '12'; b += '3'; b += '4'; b += '5'; b += '6'; return b += '7'; }
```
After the second run, it inlined a little bit more and the output becomes (after denormalization):
```js
window.a = function() { var b; b = '1234'; b += '5'; b += '6'; return b += '7'; }
```
So normalization helped a little bit, but not much :)

dimvar:
I tried this and it saves a few hundred bytes on large apps. So, nothing major, but not negligible either. I think we should merge it. A couple of comments inline.

src/com/google/javascript/jscomp/Denormalize.java
```diff
@@ -124,4 +128,48 @@ private void maybeCollapseIntoForStatements(Node n, Node parent) {
...
+      case Token.SUB:
+        return Token.ASSIGN_SUB;
+      case Token.MUL:
+        return Token.ASSIGN_MUL;
+      case Token.DIV:
+        return Token.ASSIGN_DIV;
+      case Token.MOD:
+        return Token.ASSIGN_MOD;
+      default:
+        return Token.ERROR;
```
dimvar: Don't use Token.ERROR, it's a bad token, I will be deleting it. Throw an illegal argument exception here.

src/com/google/javascript/jscomp/Denormalize.java
```diff
@@ -124,4 +128,48 @@ private void maybeCollapseIntoForStatements(Node n, Node parent) {
...
+    }
+  }
+
+  private void maybeCollapseAssignShorthand(Node n, Node parent) {
+    if (n.isAssign() && n.getFirstChild().isName() && n.getLastChild().hasChildren()
```
dimvar: This can create an invalid AST that contains the error token. Instead of hasChildren, you need to check that the last child is an operator that can be converted to an assignment operator, eg, using a helper method like this:

    static boolean hasCorrespondingAssignmentOp(Node n) {
      switch (n.getType()) {
        case Token.BITOR:
        case Token.BITXOR:
        case Token.BITAND:
        case Token.LSH:
        case Token.RSH:
        case Token.URSH:
        case Token.ADD:
        case Token.SUB:
        case Token.MUL:
        case Token.DIV:
        case Token.MOD:
          return true;
        default:
          return false;
      }
    }

src/com/google/javascript/jscomp/Denormalize.java
Dominator008: Done.

test/com/google/javascript/jscomp/PeepholeSubstituteAlternateSyntaxTest.java
```diff
@@ -335,7 +335,8 @@ public void testUndefined() {
     foldSame("try {} catch(undefined) {}");
     foldSame("for (undefined in {}) {}");
     foldSame("undefined++;");
-    fold("undefined += undefined;", "undefined += void 0;");
+    disableNormalize();
+    foldSame("undefined += undefined;");
```
dimvar: why did the result here change?

src/com/google/javascript/jscomp/Denormalize.java
```diff
@@ -124,4 +128,67 @@ private void maybeCollapseIntoForStatements(Node n, Node parent) {
...
+      case Token.ADD:
+      case Token.SUB:
+      case Token.MUL:
+      case Token.DIV:
+      case Token.MOD:
+        return true;
+      default:
+        return false;
+    }
+  }
```
dimvar: can you move the static methods to nodeutil (and make them package private)?

test/com/google/javascript/jscomp/PeepholeSubstituteAlternateSyntaxTest.java
Dominator008: https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/PeepholeSubstituteAlternateSyntax.java#L326
`isASTNormalized()` is now false in this case.

src/com/google/javascript/jscomp/Denormalize.java
Dominator008: Done.

dimvar:
LGTM. Can you resolve the merge conflicts and ping this thread?

test/com/google/javascript/jscomp/PeepholeSubstituteAlternateSyntaxTest.java
Dominator008: I added a test with `enableNormalize()` to make sure we are folding to `void 0` in that case

Dominator008:
@dimvar Done.

Dominator008:
Closed by https://github.com/google/closure-compiler/commit/a1768bac5e4145321539faf97f6575098b8951ba

PR #1666, opened 2016-03-16
Make getMessage public so that people can write their own message ext…

nicks:
…ractors

(it looks like this was just an oversight; all the similar accessors in this class are public)

---
This change is [Reviewable](https://reviewable.io/reviews/google/closure-compiler/1666)


Dominator008:
LGTM

PR #1683, opened 2016-03-22
Add hasOwnProperty as a global function

ChadKillingsworth:
I've come across more than one project that uses `hasOwnProperty` as a global function reference. Example:

```js
hasOwnProperty.call(foo, 'bar');
```

---
This change is [Reviewable](https://reviewable.io/reviews/google/closure-compiler/1683)


dimvar:
Will bring this in and see if it causes any issues.

Dominator008:
Closed by https://github.com/google/closure-compiler/commit/c71fd80067ccb4f017d27027402d7d47897e4217.

PR #1686, opened 2016-03-24
Mark CanvasRenderingContext2D.prototype.canvas as non-nullable

fredj:
From the spec: https://www.w3.org/TR/2dcontext/#dom-context-2d-canvas
"The canvas attribute must return the canvas element that the context paints on."

Does the **must** implies that the canvas is always defined?

---
This change is [Reviewable](https://reviewable.io/reviews/google/closure-compiler/1686)

### DIFF
--- a/externs/browser/html5.js
+++ b/externs/browser/html5.js
@@ -79,7 +79,7 @@ HTMLCanvasElement.prototype.getContext = function(contextId, opt_args) {};
  */
 function CanvasRenderingContext2D() {}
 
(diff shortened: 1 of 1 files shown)


MatrixFrog:
I imagine there's some weird case where the canvas has been deleted or something and it can be null, but it's probably rare enough not to worry about. I'll try and bring this in soon.

fredj:
thanks @MatrixFrog

PR #1690, opened 2016-03-27
Add possibleMissingProperty to DiagnosticGroup

selkhateeb:
This new flag allows for disabling missing properties for unknown types.
This behaviour is desired when you have mixed typed and untyped objects.
In this case you still want the compiler to complain about missing
properties for known types but not for the untyped objects.

Example when this flag is disabled using
```
'--jscomp_error=missingProperty'
'--jscomp_off=possibleMissingProperty':
```
```js
// Typed variable example
/** @constructor */
var A = function() {};
new A().foo; // This will emit an error about 'foo' undefined on 'A'.

// Untyped variable example
/** @type {?} */
var a = {};
a.foo; // No error is emitted.
```

---
This change is [Reviewable](https://reviewable.io/reviews/google/closure-compiler/1690)

### DIFF
--- a/src/com/google/javascript/jscomp/DiagnosticGroups.java
+++ b/src/com/google/javascript/jscomp/DiagnosticGroups.java
@@ -198,6 +198,10 @@ public class DiagnosticGroups {
           TypeCheck.INEXISTENT_PROPERTY_WITH_SUGGESTION,
           TypeCheck.POSSIBLE_INEXISTENT_PROPERTY);
 
+  public static final DiagnosticGroup POSSIBLE_MISSING_PROPERTIES =
+      DiagnosticGroups.registerGroup("possibleMissingProperties",
+          TypeCheck.POSSIBLE_INEXISTENT_PROPERTY);
+
   public static final DiagnosticGroup MISSING_RETURN =
       DiagnosticGroups.registerGroup("missingReturn",
           CheckMissingReturn.MISSING_RETURN_STATEMENT);
--- a/test/com/google/javascript/jscomp/CommandLineRunnerTest.java
+++ b/test/com/google/javascript/jscomp/CommandLineRunnerTest.java
@@ -417,6 +417,20 @@ public final class CommandLineRunnerTest extends TestCase {
         TypeCheck.INEXISTENT_PROPERTY);
   }
 
+  public void testCheckPossibleUndefinedProperties1() {
+    args.add("--warning_level=VERBOSE");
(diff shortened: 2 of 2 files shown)


concavelenz:
Why?

selkhateeb:
I want the ability to use the compiler's type checking feature for our NodeJS applications. Using the `--check-only` flag allows me to do just that without having to execute the compilation phase.
All of the code we write follows the `goog.module(...); goog.require(...)` module strategy. However, since we rely on third-party libraries, we use NodeJS's module strategy with `require(...)`.

Here is an example of using this:
```javascript
// In an externs file named globals.js
/** 
 * @param {!string} path
 * @return {?}
 */
var require = function(path) {};
// End of globals.js.
// ---
// Database.js:
goog.module('Database');
/** @constructor */
var Database = function() {};
Database.prototype.connect = function() {};
exports = Database; // To satisfy google closure compiler.
module.exports = Database; // To satisfy NodeJs module.
// End of Database.js
// ---
// Application.js:
goog.module('Application');
const Database = goog.require('Database');
const log = require('log4js').getLogger();

var db = new Database();
db.connect();
db.foo(); // Error missing property.
log.info('Done.'); // Error possible missing property.
// End of Application.js
```

The desired behavior when compiling this code is to complain about 'db.foo' as it is of a 'known type'. However, the actual output of the compilation complains about 'db.foo' and 'log.info'. Hence, the need for this `possibleMissingProperty`.

With this new flag, I can distinguish between the `missingProperty` errors set by `jscomp_error=missingProperty` and `possibleMissingProperty` warnings (set by `jscomp_error=possibleMissingProperty`. Those `possibleMissingProperty` warnings can be fixed eventually in extern files or not care about them at all. (catch those errors during unittests).

This will allow for mixing in third-party libraries with my code. My code follows the 'goog.module' conventions for defining types. Third-party libraries don't and won't. I want the compiler to catch my mistakes and not care about third-party libraries.

selkhateeb:
@dimvar Is there anything I can help with in testing or verifying to make this pull request go in as smooth as possible?

dimvar:
The compiler only warns for a missing property on unknown, when this property is not defined on any type in the program. That is a very loose check, and it's useful for catching typos in property names. If you want to avoid these warnings, it's easy to do so by defining them on some type in the externs.

    /** @type {?} */
    SomeRandomType.prototype.info;

I'd rather not add a new flag for this. The compiler has too many flags already.

selkhateeb:
Sounds good. Thanks for the update.

PR #1715, opened 2016-04-08
Add EventListenerOptions to EventTarget.add/removeEventListener.

dtapuska:
Chrome 51 has introduced support for EventListenerOptions on the
3rd argument. Change EventTarget.h to match the currently
published DOM specification; modifying addEventListener to have
an optional 3rd argument.

---
This change is [Reviewable](https://reviewable.io/reviews/google/closure-compiler/1715)

### DIFF
--- a/externs/browser/w3c_event.js
+++ b/externs/browser/w3c_event.js
@@ -25,6 +25,15 @@
 
 
 /**
+ * @typedef {{
+ *   passive: (boolean|undefined),
+ *   capture: (boolean|undefined)
+ * }}
+ * @see https://dom.spec.whatwg.org/#dictdef-eventlisteneroptions
+ */
+var EventListenerOptions;
+
+/**
  * @interface
  */
 function EventTarget() {}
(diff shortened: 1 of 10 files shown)


externs/browser/w3c_event.js
```diff
@@ -25,27 +25,35 @@
 
 
 /**
+ * @typedef {{
+ *   passive: (boolean|undefined),
+ *   capture: (boolean|undefined)
+ * }}
+ */
```
Dominator008: Can you add a `@see` JsDoc and link to the part of the specs that defines this?

dtapuska: Done; sorry I was slow.

Dominator008:
@dtapuska Is this no longer necessary?

dtapuska:
It is but I've abandoned it. Being a Googler I was asked to make the change on the internal version instead and ran into complications with goog.events depending on it. I've abandoned it and will let the closure team do it themselves. (As per discussion with MatrixFrog)

Dominator008:
@dtapuska Oh ok. Yeah this one did look like it would break things internally.

PR #1717, opened 2016-04-08
Add USELESS_USE_STRICT_DIRECTIVE to the "lintChecks" DiagnosticGroup

Dominator008:
I'm inclined to just remove that warning, but let's make it
suppressible at least.

---
This change is [Reviewable](https://reviewable.io/reviews/google/closure-compiler/1717)


blickly:
If people find this to be a useful lint check, we could add it back to ES6 modules.

Dominator008:
@blickly As pointed out by @ChadKillingsworth, ES6 modules can be standalone scripts themselves so this check might give false positives.

blickly:
Those with import or export can't be.

Dominator008:
@blickly Oh right.

blickly:
Not sure it's a high priority, though, if people have already learned that ES6 modules are strict mode code.

Dominator008:
@blickly Seems like we can remove some duplicate code by adding it back. See https://github.com/google/closure-compiler/pull/1718.

PR #1718, opened 2016-04-08
Add the check for unnecessary 'use strict' back to ProcessEs6Modules

Dominator008:
Rename checkStrictModeDirective() to checkAndSetStrictModeDirective() in
ClosureRewriteModule and reuse that method.

Also modify CompilerTestCase to fail tests if parse warnings are
expected but not reported.

---
This change is [Reviewable](https://reviewable.io/reviews/google/closure-compiler/1718)

### DIFF
--- a/src/com/google/javascript/jscomp/ClosureRewriteModule.java
+++ b/src/com/google/javascript/jscomp/ClosureRewriteModule.java
@@ -235,7 +235,7 @@ final class ClosureRewriteModule implements HotSwapCompilerPass {
     @Override
     public boolean shouldTraverse(NodeTraversal t, Node n, Node parent) {
       if (NodeUtil.isModuleFile(n)) {
-        checkStrictModeDirective(t, n);
+        checkAndSetStrictModeDirective(t, n);
       }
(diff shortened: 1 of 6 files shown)


blickly:
Nice!

MatrixFrog:
Do we have some tests that show this *doesn't* warn on files that could also be a script?

ChadKillingsworth:
> Do we have some tests that show this doesn't warn on files that could also be a script?
 Show all checks

How are you going to determine whether it's a script or a module? I still view this check as:

 1. Extra code/work we have to maintain
 2. Serving no real purpose (I don't count developer education)

Dominator008:
@ChadKillingsworth It's [disabled](https://github.com/google/closure-compiler/blob/587c05ee3db3dc373b10f19908aa9b0d9f2605ab/src/com/google/javascript/jscomp/ClosureRewriteModule.java#L128) by default now unless you turn "lintChecks" on.

Dominator008:
@MatrixFrog I just added the tests.

PR #1728, opened 2016-04-15
Optimize switch with default case / return / throw / continue

### LINKED ISSUE #1722: Dead code is not removed from switch statements
I am trying to compile a separate version of my web app for each language.
Therefore, I want to include only the translations for that language, using the goog.LOCALE parameter I set during compilation.

However, this leads to this kind of output under advanced mode (example with `goog.LOCALE = "nl"`):
```js
var bs = function(){
  switch("nl") { 
    case "nl": return {"al": "mijn", "Nederlandstalige": "vertalingen"};
    case "fr": return {"tous": "mes", "traductions": "françaises"};
    default: return {"all": "my", "english": "translations"}
  }
}();
```
_(I added the whitespace for readability, the object with the translations is obviously also different in reality)_

The first case of this switch statement will always be true, so I don't understand why it is not removed by the Closure Compiler.

Am I doing something wrong or are these optimization not yet supported?

Dominator008:
A default case that is the last case should not prevent optimizing the
other cases.

Also add optimizations for cases that end with return, throw or
continue.

Fixes #1722.

---
This change is [Reviewable](https://reviewable.io/reviews/google/closure-compiler/1728)

### DIFF
--- a/src/com/google/javascript/jscomp/PeepholeRemoveDeadCode.java
+++ b/src/com/google/javascript/jscomp/PeepholeRemoveDeadCode.java
@@ -296,6 +296,55 @@ class PeepholeRemoveDeadCode extends AbstractPeepholeOptimization {
 
   static final Predicate<Node> MATCH_UNNAMED_BREAK = new MatchUnnamedBreak();
 
+  private Node tryRemoveSwitchWithSingleCase(Node n, boolean shouldIncludeCondition) {
+    Node caseBlock = n.getLastChild().getLastChild();
+    // Drop statements that are dead code
+    boolean shouldDrop = false;
+    for (Node statement: caseBlock.children()) {
+      if (shouldDrop) {
+        statement.detachFromParent();
+      } else if (isExit(statement)) {
+        if (statement.isBreak() && !statement.hasChildren()) {
+          statement.detachFromParent();
+        }
+        shouldDrop = true;
+      }
+    }
+
+    // Back off if the switch contains statements like "if (a) { break; }"
+    if (NodeUtil.has(caseBlock, MATCH_UNNAMED_BREAK, NodeUtil.MATCH_NOT_FUNCTION)) {
+      if (shouldDrop && !isExit(caseBlock.getLastChild())) {
+        reportCodeChange();
+      }
+      return n;
+    } else {
+      // TODO(moz): This needs to change when we optimize for ES6. The block might need to be
+      // preserved in the presence of block-scoped declarations.
+      if (shouldIncludeCondition) {
+        caseBlock.addChildToFront(IR.exprResult(n.removeFirstChild()).srcref(n));
+      }
+      caseBlock.setIsSyntheticBlock(false);
+      n.getParent().replaceChild(n, caseBlock.detachFromParent());
+      reportCodeChange();
+      return caseBlock;
+    }
+  }
+
+  private Node tryRemoveSwitch(Node n) {
+    if (n.hasOneChild()) {
+      // Remove the switch if there are no remaining cases
+      Node condition = n.removeFirstChild();
+      Node replacement = IR.exprResult(condition).srcref(n);
+      n.getParent().replaceChild(n, replacement);
+      reportCodeChange();
+      return replacement;
+    } else if (n.getChildCount() == 2 && n.getLastChild().isDefaultCase()) {
+      return tryRemoveSwitchWithSingleCase(n, true);
+    } else {
+      return n;
+    }
+  }
+
   /**
    * Remove useless switches and cases.
    */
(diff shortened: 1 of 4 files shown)


Dominator008:
@concavelenz

blickly:
Travis is still showing a failure:

Dominator008:
@blickly Yeah the code had a different assumption about default cases. I'm still working on it.

Dominator008:
@blickly Ok the tests are passing now.

test/com/google/javascript/jscomp/PeepholeRemoveDeadCodeTest.java
```diff
@@ -478,6 +481,110 @@ public void testOptimizeSwitch2() {
...
+        "    bar();",
+        "    break;",
+        "}"),
+        "bar();");
+
+    test(LINE_JOINER.join(
+        "switch (a()) {",
+        "  default:",
+        "    bar();",
+        "    break;",
```
blickly: What if this were a labeled break? In that case I think we need to backoff.

blickly:
Nice, I like this optimization!

test/com/google/javascript/jscomp/PeepholeRemoveDeadCodeTest.java
Dominator008: Yeah I fixed it. I think we can still optimize it.

blickly:
What will happen if there are break statements that are not the last statement in the default block?

Dominator008:
@blickly Added test cases for those.

test/com/google/javascript/jscomp/PeepholeRemoveDeadCodeTest.java
```diff
@@ -478,6 +481,163 @@ public void testOptimizeSwitch2() {
...
+        "}"),
+        "a(); bar();");
+
+    test(LINE_JOINER.join(
+        "switch (a()) {",
+        "  default:",
+        "    break;",
```
blickly: I think a more realistic example would be:
   default:
       if (condition) { break; }
       doStuff();

Dominator008: Done.

test/com/google/javascript/jscomp/PeepholeRemoveDeadCodeTest.java
```diff
@@ -478,6 +481,226 @@ public void testOptimizeSwitch2() {
...
+        "  switch('x') {",
+        "    case 'x': var x = 1; break; var y = 2;",
+        "    case 'y': break;",
+        "  }",
+        "}"),
+        "function f() { var x = 1; }");
+
+    testSame(LINE_JOINER.join(
+        "function f() {",
+        "  switch(x) {",
```
blickly: Note, that this is optimizable to
```javascript
if (x == 'y') {
  // case block
} else {
  // default block
}
```
which is generally more optimizable by other passes.

src/com/google/javascript/jscomp/PeepholeRemoveDeadCode.java
```diff
@@ -296,6 +296,55 @@ public boolean apply(Node n) {
 
   static final Predicate<Node> MATCH_UNNAMED_BREAK = new MatchUnnamedBreak();
 
+  private Node tryRemoveSwitchWithSingleCase(Node n, boolean shouldIncludeCondition) {
+    Node caseBlock = n.getLastChild().getLastChild();
+    // Drop statements that are dead code
+    boolean shouldDrop = false;
+    for (Node statement: caseBlock.children()) {
```
blickly: Since this particular code pattern seems pretty unrealistic for people to write (obviously dead code in the middle of a case), I'd probably skip this optimization it if it simplifies our implementation code.

test/com/google/javascript/jscomp/PeepholeRemoveDeadCodeTest.java
Dominator008: Is there a reference for why the if statement would be more optimizable?

src/com/google/javascript/jscomp/PeepholeRemoveDeadCode.java
Dominator008: I think since tackling the general case actually makes the implementation simpler?

test/com/google/javascript/jscomp/PeepholeRemoveDeadCodeTest.java
ChadKillingsworth: For one, it can be converted to a hook `x == 'y' ? /* case block */ : /* else block */`. It can also use `&&` and `||` shorthands.

Dominator008: I've added a TODO for this.

blickly:
Does the "In Progress" label mean that I should hold off on reviewing this?

Dominator008:
@blickly Actually it should be "In Review" instead :)

src/com/google/javascript/jscomp/PeepholeRemoveDeadCode.java
blickly: Hmm.. I'll show you the simplifications I had in mind when I pull this in.

PR #1739, opened 2016-04-19
Fix duplicate block scoped declaration not reported in the same scope

Dominator008:
Duplicate let / const / class declaration is a hard error in any scope,
but VarCheck was only checking this in the global scope.

---
This change is [Reviewable](https://reviewable.io/reviews/google/closure-compiler/1739)

### DIFF
--- a/src/com/google/javascript/jscomp/VarCheck.java
+++ b/src/com/google/javascript/jscomp/VarCheck.java
@@ -80,10 +80,10 @@ class VarCheck extends AbstractPostOrderCallback implements
         "JSC_VAR_ARGUMENTS_SHADOWED_ERROR",
         "Shadowing \"arguments\" is not allowed");
 
-  static final DiagnosticType LET_CONST_MULTIPLY_DECLARED_ERROR =
+  static final DiagnosticType LET_CONST_CLASS_MULTIPLY_DECLARED_ERROR =
       DiagnosticType.error(
-          "JSC_LET_CONST_MULTIPLY_DECLARED_ERROR",
-          "Duplicate let / const declaration in the same scope is not allowed.");
+          "JSC_LET_CONST_CLASS_MULTIPLY_DECLARED_ERROR",
+          "Duplicate let / const / class declaration in the same scope is not allowed.");
 
   // The arguments variable is special, in that it's declared in every local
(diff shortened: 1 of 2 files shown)


dimvar:
Let's not merge this just yet. I'm in the process of making a related change to VariableReferenceCheck, and it requires cleanup internally. And we're trying to do a release as well. After the release and my change have landed, then we can come back to this.

Dominator008:
@dimvar This PR fixes a crash in the rewriting pass: https://closure-compiler-debugger.appspot.com/#input0%3Dfunction%2520f%28x%29%2520%257B%250A%2520%2520switch%2520%28x%29%2520%257B%250A%2520%2520%2520%2520case%2520%27a%27%253A%250A%2520%2520%2520%2520%2520%2520let%2520z%2520%253D%2520123%253B%250A%2520%2520%2520%2520%2520%2520break%253B%250A%2520%2520%2520%2520case%2520%27b%27%253A%250A%2520%2520%2520%2520%2520%2520let%2520z%2520%253D%2520234%253B%250A%2520%2520%2520%2520%2520%2520break%253B%250A%2520%2520%257D%250A%257D%26input1%26conformanceConfig%26externs%26refasterjs-template%26includeDefaultExterns%3D1%26CHECK_SYMBOLS%3D1%26CHECK_TYPES%3D1%26LANG_IN_IS_ES6%3D1%26MISSING_PROPERTIES%3D1%26PRESERVE_TYPE_ANNOTATIONS%3D1%26PRETTY_PRINT%3D1%26TRANSPILE%3D1

Will your change in VariableReferenceCheck catch this error?

MatrixFrog:
@dimvar This is the crash you reported internally as b/28248708 . Sorry for the internal/external bug tracker/pull request confusion. I'm fine with holding off on landing this for a little while though, since it should be pretty easy for people to work around.

dimvar:
@Dominator008 Do you expect that this PR just turns crashes into errors, or do you also expect new errors that were not being detected before? In the latter case, we may have to delay the fix.

Dominator008:
@dimvar It should just turn crashes into errors, but I can't guarantee that. My guess is that this kind of error is usually easy to spot.

dimvar:
sgtm

PR #1746, opened 2016-04-20
Remove assignment to properties of return values of no-side-effect calls

### LINKED ISSUE #1745: Object created with Object.create not dead code eliminated
When creating an object through `Object.create` and adding a property to it, it will not be removed in ADVANCED mode, even if nothing references it.

**Not** removed:
```javascript
/** @type {!Object<string, number>} */
const foo = Object.create(null); // or Object.create();
foo.prop = 2;
```

Removed:
```javascript
/** @type {!Object<string, number>} */
const bar = {};
bar.prop = 2;
```

Dominator008:
Assignments like "Object.create({}).prop = 1" is dead code and should be
removed.

Closes #1745.

---

This change is [Reviewable](https://reviewable.io/reviews/google/closure-compiler/1746)

### DIFF
--- a/test/com/google/javascript/jscomp/PeepholeRemoveDeadCodeTest.java
+++ b/test/com/google/javascript/jscomp/PeepholeRemoveDeadCodeTest.java
@@ -25,13 +25,21 @@ import com.google.javascript.rhino.Node;
 
 public final class PeepholeRemoveDeadCodeTest extends CompilerTestCase {
 
-  private static final String MATH =
-      "/** @const */ var Math = {};" +
-      "/** @nosideeffects */ Math.random = function(){};" +
-      "/** @nosideeffects */ Math.sin = function(){};";
+  private static final String MATH = LINE_JOINER.join(
+      "/** @const */ var Math = {};",
+      "/** @nosideeffects */ Math.random = function() {};",
+      "/** @nosideeffects */ Math.sin = function() {};");
+
+  private static final String OBJECT =
+      "/** @nosideeffects */ Object.create = function(proto, opt_properties) {};";
+
+  private static final String JQUERY = LINE_JOINER.join(
+      "/** @constructor */ function $() {};",
+      "/** @nosideeffects */ $.prototype.add = function() {};",
+      "$.prototype.addClass = function() {};");
 
   public PeepholeRemoveDeadCodeTest() {
-    super(MATH);
+    super(MATH + OBJECT + JQUERY);
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


Dominator008:
@concavelenz

SLaks:
Actually, this is not a safe change.
An `@nosideeffects` function can still return a shared mutable object, and such mutations must not be removed.

Dominator008:
@SLaks I thought about that, but convinced myself that returning a shared mutable object is a side effect by definition.

ChadKillingsworth:
Do you have an example snippet of a shared mutable object?

SLaks:
No.

For example:

```js
document.getElementById('foo').innerHTML = 'blah';
```

Dominator008:
@SLaks While I think the `@nosideeffects` annotation is blurry, I agree with you that we need another annotation to make this optimization safe.

test/com/google/javascript/jscomp/PeepholeRemoveDeadCodeTest.java
```diff
@@ -732,6 +740,17 @@ public void testFoldAssign() {
...
     test("y=1 + (x=x)", "y=1 + x");
+
+    // GitHub issue #1745: https://github.com/google/closure-compiler/issues/1745
+    test("Object.create({}).prop = 1", "");
+    test("Object.create(proto).foo.bar.baz = 1", "");
+
+    // add() is side-effect free, but addClass() is not
+    test("var jQuery = new $; jQuery.add().add().bar = 1", "var jQuery = new $;");
+    test("var jQuery = new $; jQuery.add().bar.add().baz = 1", "var jQuery = new $;");
+    test("var jQuery = new $; jQuery.bar.add().baz = 1", "var jQuery = new $;");
```
dimvar: This doesn't seem right to me. Even if the add method is a no-op, this could be adding a new baz property to jQuery.bar, so we shouldn't be removing the assignment.

AFAICT, we can only remove the assignment if the leftmost receiver is an object created by a non-side-effect call. If it's a pre-existing object, we can't remove the assignment.

Or am I missing something?

dimvar:
Left one comment. Can you also rebase to head?

dimvar:
Any update on this PR?

Dominator008:
@dimvar No I haven't looked at this yet. I will ping you once I update this.

test/com/google/javascript/jscomp/PeepholeRemoveDeadCodeTest.java
Dominator008: I think you are right. I'm not sure how to determine what the "leftmost receiver" is? For example, in:
```js
Object.create({}).prop = 1;
```
Should `Object` be considered a pre-existing object? We'd like to remove the assignment though.

concavelenz:
@Dominator008 

Returning mutable state is not in itself a side-effect. This is removable.

function fn() { return window; }
fn();

However, Object.createObject({}) should be treated the same way as "new Blah"/({})/"[]" which is treated as creating a new unescaped object.

concavelenz:
Not going to merge.

PR #1750, opened 2016-04-21
Escape a single quote in a warning message  in PeepholeFoldConstants

Dominator008:
This change is [Reviewable](https://reviewable.io/reviews/google/closure-compiler/1750)


MatrixFrog:
Maybe someday we can switch to a more intuitive message formatter.

PR #1767, opened 2016-05-03
Make LintPassConfig public

jart:
I'm planning on using this class in the Closure Rules repository
because there currently exists no flags for the Linter class.

---
This change is [Reviewable](https://reviewable.io/reviews/google/closure-compiler/1767)


jart:
n/m

PR #1768, opened 2016-05-03
Fix closure linter for private members

jart:
I'm getting errors like:

```
closure/templates/test/greeter.js:35: WARNING - Private property name_ should end with '_'
  this.name_ = name;
```

Because of this:

https://github.com/google/closure-compiler/blob/c0f6720d0b58f2a84d7a9cfb6adc26b5a9da5d62/src/com/google/javascript/jscomp/lint/CheckJSDocStyle.java#L202

This is because the Closure convention always returns false for isPrivate. Now it's better.

---
This change is [Reviewable](https://reviewable.io/reviews/google/closure-compiler/1768)

### DIFF
--- a/src/com/google/javascript/jscomp/ClosureCodingConvention.java
+++ b/src/com/google/javascript/jscomp/ClosureCodingConvention.java
@@ -391,7 +391,7 @@ public final class ClosureCodingConvention extends CodingConventions.Proxy {
 
   @Override
   public boolean isPrivate(String name) {
-    return false;
+    return name.endsWith("_");
   }
 
   @Override


MatrixFrog:
I guess currently the convention of private members ending in `_` is considered a Google one, not a Closure one. But since we use that convention in Closure maybe it makes sense to move it. I'll see what others think.

jart:
If `ClosureCodingConvention` doesn't enforce go/closurestyle, then what does it do?

jart:
Also keep in mind that according to go/closurestyle, Closure Style is a superset of Google Java Style.

MatrixFrog:
@concavelenz probably has more of the historical context on that. By the way, if you can send changes internally we would appreciate it as it eliminates some steps for us.

jart:
Sure thing.

PR #1774, opened 2016-05-05
Adding support for ES6 modules with non-JS extensions:

supersteves:
existing:
```js
import "./MyJsFile" -> MyJsFile.js
```
and now:
```js
import "./MyJsxFile.jsx" -> MyJsFile.jsx
```

@ChadKillingsworth might want to glance at this.

---
This change is [Reviewable](https://reviewable.io/reviews/google/closure-compiler/1774)

### DIFF
--- a/src/com/google/javascript/jscomp/deps/ModuleLoader.java
+++ b/src/com/google/javascript/jscomp/deps/ModuleLoader.java
@@ -148,7 +148,10 @@ public final class ModuleLoader {
      * @return The normalized module URI, or {@code null} if not found.
      */
     public ModuleUri resolveEs6Module(String moduleName) {
-      URI resolved = locateNoCheck(moduleName + ".js");
+      // Typically module names don't have the ".js" extension. 
+      // But tolerate this, and support explicit ".jsx" extension:
+      if (!moduleName.endsWith(".js") && !moduleName.endsWith(".jsx")) moduleName += ".js";
+      URI resolved = locateNoCheck(moduleName);
       if (!moduleUris.contains(resolved) && errorHandler != null) {
         errorHandler.report(CheckLevel.WARNING, JSError.make(LOAD_WARNING, moduleName));
       }
--- a/test/com/google/javascript/jscomp/deps/ModuleLoaderTest.java
+++ b/test/com/google/javascript/jscomp/deps/ModuleLoaderTest.java
@@ -33,6 +33,14 @@ public final class ModuleLoaderTest extends TestCase {
     assertUri("js/a.js", loader.resolve("js\\a.js"));
     assertUri("js/b.js", loader.resolve("js\\a.js").resolveEs6Module("./b"));
   }
+  
+  public void testJsxExtension() {
+	  ModuleLoader loader =
+        new ModuleLoader(null, ImmutableList.of("."), inputs("js/a.js", "js/b.js", "js/c.jsx"));
+    assertUri("js/a.js", loader.resolve("js/a.js"));
+    assertUri("js/b.js", loader.resolve("js/a.js").resolveEs6Module("./b"));
+    assertUri("js/c.jsx", loader.resolve("js/a.js").resolveEs6Module("./c.jsx"));
+  }
 
   public void testLocateCommonJs() throws Exception {
     ModuleLoader loader = new ModuleLoader(


ChadKillingsworth:
I'm fine with this concept - but I'd like to keep all the module types in sync. We'll need to do the same thing for common js.

ChadKillingsworth:
Actually - now that I say that. We can do this temporarily, but it will be superseded when I implement the [Node module resolution algorithm](https://nodejs.org/api/modules.html#modules_all_together).

supersteves:
Can't wait!

supersteves:
I've rebased this PR and will squash the commits if you're happy with the revised implementation, following @MatrixFrog's conflicting edit https://github.com/google/closure-compiler/commit/3d03f42df9fcff0ee9f1e7d481b1c98c1fe830c2.

This is a small stop-gap to allow us to use ".jsx" files with ES6 modules. I can look at using ".js" extension for jsx files although it would hurt my own build process, and ideally CompilerOptions would include more control over implicit and explicit extensions, but meanwhile this fix shouldn't harm any existing ES6 module use with Closure Compiler.

MatrixFrog:
Wait, what? How would this work? We don't support the JSX syntax so we wouldn't even be able to parse the files.

supersteves:
@MatrixFrog I've preprocessed them using Babel's JSX conversion. Ideally I'd build this conversion in Java but adapting the CC parser to support JSX is way beyond me. It's just a pure JSX conversion, with none of the other transpilation enabled.

MatrixFrog:
Does Babel not output files with the `.js` extension?

dimvar:
> Ideally I'd build this conversion in Java but adapting the CC parser to support JSX

Actually, it's good that you're not working on this :) There is a strict policy at Google about introducing new language extensions, and we don't have any plan to support JSX at the moment.

supersteves:
In my sources I have a mix of js and jsx files. This is common practice. I use an IDE like webstorm, and browserify+babel to process require() and .jsx files. It's normal to have js and jsx files together, and to omit the .js extension (or to keep it) but to always have the .jsx extension when require'ing a module. The same goes for ES6 modules.

So yes I could theoretically output the jsx files as js (or rename them in-stream) before processing them but then my module references would need to also refer to modules without extensions, and my IDE would lose track of them. Also you couldn't have two files both called the same name .js and .jsx (not that you'd probably want that). I don't fancy transforming the module references on the fly.

(In fact I do something more obscure in that I use Closure Compiler from Java and have my own Java pre+post processing + AST caching going on, including invoking the Babel process to do the JSX alone, thereby allowing me to patch the JSX conversion to reinstate type casts it loses - https://github.com/google/closure-compiler/issues/1782... but I try to minimise the custom steps.)

My aim is to use a standardised approach which works with babel/browserify toolstack but optionally can slot into a Closure Compiler toolstack. So I don't want to use a module naming convention that would only work with Closure Compiler. ES6 modules, my current successful experiment with Closure Compiler, allows this to work seamlessly (albeit with this PR for .jsx support within the module name/filename).

In summary this tiny PR is trying to make common patterns of CJS/ES6 module names (implicit .js + explicit .jsx) work the same way in Closure Compiler.

MatrixFrog:
> In my sources I have a mix of js and jsx files. This is common practice.

Having the jsx files alongside the js files is common practice, sure. With you so far.

> It's normal to [...] have the .jsx extension when require'ing a module.

This is where I start to be more confused. I doubt that browsers will ever support `import * from 'something.jsx'`, so I'm worried that we'll end up with a syntax that works in Closure but doesn't actually work when you just load the JS directly in the browser. We want to make sure that our users are writing code that will run unmodified in the browser (which is part of the reason we don't support JSX/TypeScript/etc -- we are a JavaScript compiler so we only want to accept valid spec-compliant JavaScript)

Anyway maybe I'm misunderstanding things, and if this small change will make things a lot easier for certain workflows, we can consider merging it. But my reluctance is because it seems like we're diverging from JS here and starting to accept "code that is mostly JS but with a few minor things that aren't quite JS" and I want to be very careful that we don't go very far down that road.

supersteves:
While I'd love it if JSX support was possible, I completely am not asking for this and understand the project goals (though TS syntax seems partly supported in contrary to that... I'm waiting for unification of flow and TS in some future ES standard but guess it is very far away).

Here's an example talking about the same problem in the webpack toolchain: 
http://stackoverflow.com/a/34678598

They suggest either an explicit extension (could be something other than js or jsx and not necessarily indicative of a language extension) in the module ref.
Or by configuring "resolve" extensions in the tool config.

The latter is clearly more favourable as a generic solution in CC but it's not a 2 line PR whereas the former brings support for a common workflow out of the box very cheaply. Making it easier to interop between toolchains and migrate to CC.

Dominator008:
@supersteves A subset of TypeScript syntax was added for three reasons:
1. The Angular team experimented with AtScript and added the initial TypeScript syntax support, but then they switched to just using TypeScript.
2. We were hoping that some future standard (whether it be in ECMAScript itself) would allow optional TypeScript-style colon typing in JavaScript that runs in the browsers, but that has not happened yet.
3. Even without browser support, being able to use the existing large collection of TypeScript `.d.ts` type definition files as externs files would be nice. They are less verbose and people already wrote a bunch of them.

3 is already supported by using ``--language_in=ECMASCRIPT6_TYPED`, but we currently we don't have a separate language mode for externs so users would be able to pass in a subset of TypeScript in actual code.

JSX is even less likely to be supported in the browsers because it's React specific, but if this PR is small and it greatly facilitates some workflow, I'm fine with it. Some documentation would be required though.

supersteves:
Interesting history of TS syntax support - thanks. I'd love to see that TS standardised day come for its brevity but am not holding on. But I never thought of TS support as potentially opening up a new route to getting better externs E.g. for React. Something I'll experiment with.

Docs in code, or wiki/elsewhere?

Dominator008:
@supersteves The unit tests and integration tests might be the best docs at the moment:
https://github.com/google/closure-compiler/blob/bfaf7180678604011a06580e695dca7c3bddf9e2/test/com/google/javascript/jscomp/Es6TypedToEs6ConverterTest.java
and
https://github.com/google/closure-compiler/blob/871694f79e8bf45c2706b9fdf247909f367703bf/test/com/google/javascript/jscomp/Es6TypedIntegrationTest.java

If you try it on a random `.d.ts` file, you will likely get some errors / warnings due to things currently not supported in Closure. Last time we tried, about 200 files from https://github.com/DefinitelyTyped/DefinitelyTyped got converted without error / warning.

I will continue working on this feature. There are also improvements coming up in Closure's type system, such as bounded generics, that will allow us to support a large subset of TypeScript definition files.

blickly:
I think it's pretty reasonable if you are using the existing .d.ts library
to transpile the an externs definition for that library.  For libraries
that we don't maintain externs for, this is much better than starting from
scratch, and if you need to edit a few lines of the .d.ts file for it to
work, it still provides a benefit.

supersteves:
Thanks for the info, I'll play with that.

My "Docs in code, or wiki/elsewhere?" referred to @Dominator008's:

> Some documentation would be required though.

i.e. what's needed for this PR.

supersteves:
reminder note to update https://github.com/google/closure-compiler/wiki/JS-Modules with info on resolving file extensions after whatever comes out of this PR

ChadKillingsworth:
Whether we use this exact PR or not, I do plan on enabling this functionality. It's specifically part of the [Node module resolution algorithm](https://nodejs.org/api/modules.html#modules_all_together). Node/CommonJS module loading uses this functionality to support importing JSON files.

supersteves:
Gets my vote!

MatrixFrog:
Per discussion on https://github.com/google/closure-compiler/issues/1897 I'll bring this in but without the `jsx` support (just `js`, at least for now).

supersteves:
I.e. just making the .js extension optional. 

Separately I propose adding a CompilerOptions member to list the set of known extensions, thereby allowing users to specify .jsx and other extensions...?

I considered skipping adding .js if any .* was found, but it would conflict with filenames containing dots in the name part.

MatrixFrog:
I don't want to get ahead of the browsers here. It seems likely browsers are only going to support `.js` but we'll keep an eye on it.

supersteves:
@MatrixFrog It's not that I propose explicit support for .jsx. I just want to be able to tell the compiler "Don't add a .js" if it already has a whitelisted extension (default ".js", configurable as ".,js,.jsx").

MatrixFrog:
You can send a separate PR for that but I'm not sure I would be in favor of it. We are generally trying to make the compiler less configurable, not more.

joscha:
@supersteves Did you ever get around to handing in a PR or making this work? We have a mixed codebase with `.js` and `.es6` extensions and I run into the same import problem.

supersteves:
@joscha No, I'd be happy to submit a proper configurable "known extensions option" PR, but the last response didn't fill me with confidence it would be accepted. We have our own fork with a bunch of minor changes including a low-grade version of this. Supporting explicit .es6 extensions (hard-coded) would be a 1 line fix if you were to fork it yourself: https://github.com/visokio/closure-compiler/commit/ec8a81ac1aeadbbc74e06c2c191920e6e71bb5d2

ChadKillingsworth:
It's covered by #2130

ChadKillingsworth:
Closed by #2130 - requires use the `--module_resolution` flag to be either `BROWSER` or `NODE`.

PR #1790, opened 2016-05-12
jQuery.noop returns undefined, not a function

tdelmas:
See https://api.jquery.com/jquery.noop/ and https://github.com/jquery/jquery/blob/master/src/core.js#L207

---
This change is [Reviewable](https://reviewable.io/reviews/google/closure-compiler/1790)

### DIFF
--- a/contrib/externs/jquery-1.4.4.js
+++ b/contrib/externs/jquery-1.4.4.js
@@ -290,7 +290,7 @@ jQuery.merge = function(first, second) {};
 jQuery.noConflict = function(removeAll) {};
 
 /**
- * @return {function()}
+ * @return {undefined}
  * @nosideeffects
  */
 jQuery.noop = function() {};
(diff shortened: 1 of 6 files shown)


Dominator008:
@tdelmas Looks good, thanks!

MatrixFrog:
Presumably they meant to write `@type {function()}` not `@return {function()}`. Good catch, will bring this in soon.

tdelmas:
I catch it because I had something like `var f = $.noop; var res = f(); if ( res === undefined ) {...}` and a warning saying that my `res === undefined` is always `false` because `res` is a function.

I think the `@return {undefined}` annotation is missing for angular.noop too. Should I open a Issue/Pull request ?

Dominator008:
@tdelmas Yes, feel free to open another pull request for `angular.noop`.

PR #1792, opened 2016-05-13
angular.noop returns undefined

tdelmas:
See https://docs.angularjs.org/api/ng/function/angular.noop and https://github.com/angular/angular.js/pull/14597

Follow up on https://github.com/google/closure-compiler/pull/1790

---
This change is [Reviewable](https://reviewable.io/reviews/google/closure-compiler/1792)

### DIFF
--- a/contrib/externs/angular-1.0.js
+++ b/contrib/externs/angular-1.0.js
@@ -201,6 +201,9 @@ angular.mock = {};
  */
 angular.module = function(name, opt_requires, opt_configFn) {};
 
+/**
+ * @return {undefined}
+ */
 angular.noop = function() {};
 
 /**
--- a/contrib/externs/angular-1.2.js
+++ b/contrib/externs/angular-1.2.js
@@ -197,6 +197,9 @@ angular.mock = {};
  */
 angular.module = function(name, opt_requires, opt_configFn) {};
 
+/**
+ * @return {undefined}
+ */
 angular.noop = function() {};
 
 /**
--- a/contrib/externs/angular-1.3.js
+++ b/contrib/externs/angular-1.3.js
@@ -202,6 +202,9 @@ angular.mock = {};
  */
 angular.module = function(name, opt_requires, opt_configFn) {};
 
+/**
+ * @return {undefined}
+ */
 angular.noop = function() {};
 
 /**
--- a/contrib/externs/angular-1.4.js
+++ b/contrib/externs/angular-1.4.js
@@ -194,6 +194,9 @@ angular.mock = {};
  */
 angular.module = function(name, opt_requires, opt_configFn) {};
 
+/**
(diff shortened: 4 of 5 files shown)


Dominator008:
Looks good, thanks!

mprobst:
Not sure I understand what this fixes. `angular.noop` indeed doesn't return a value and thus its return is `undefined`, but that's exactly what you get without specifying a `@return`. We don't annotate other `void` functions with `@return {undefined}`.

Is there any symptom you're trying to fix?

Dominator008:
@mprobst This is similar to https://github.com/google/closure-compiler/pull/1790.

I think by default, the Compiler treats unannotated functions as returning `?`, not `undefined`:
http://closure-compiler-debugger.appspot.com/#input0%3Dvar%2520%252F**%2520number%2520*%252F%2520a%2520%253D%2520f()%253B%26input1%26conformanceConfig%26externs%3Dfunction%2520f()%2520%257B%257D%26refasterjs-template%26includeDefaultExterns%3D1%26CHECK_SYMBOLS%3D1%26CHECK_TYPES%3D1%26LANG_IN_IS_ES6%3D1%26MISSING_PROPERTIES%3D1%26PRESERVE_TYPE_ANNOTATIONS%3D1%26PRETTY_PRINT%3D1%26TRANSPILE%3D1

So this should improve type checking.

tdelmas:
@mprobst It was a follow up of that PR : https://github.com/google/closure-compiler/pull/1790#issuecomment-218824187


@Dominator008 yes, with externs I did see that : 
> I think by default, the Compiler treats unannotated functions as returning ?, not undefined

I check some others externs and I saw some `@return {undefined}` so I though - at least for externs - that it was the correct way to annotate.

But the documentation ( https://developers.google.com/closure/compiler/docs/js-for-compiler#tag-return ) says : 
> If there is no return value, do not use a @return tag. 
But as the compiler doens't threat equally the absence of @return tag and `@return {undefined}`, I think the documentation is wrong.

IMHO, the current compilator behaviour ( treats unannotated functions as returning `?` ) is the more logic (because it's match the behaviour of unannotated arguments), but It's just an opinion.

Either way, something need to be fixed, either the compilator or the documentation :-)

Dominator008:
@tdelmas I think the documentation needs to be fixed. Can you file a separate issue for that? Thanks!

tdelmas:
@mprobst https://github.com/google/closure-compiler/issues/1795 is closed: the updated documentation says "If a function that is not in externs has no return value, you can omit the @return tag, and the compiler will assume that the function returns undefined."
So this PR is valid, no?

mprobst:
Sorry for dropping the ball on this, I meant to answer.

Indeed, the difference is that this function is in externs, so this is valid.

Issue #1374, opened 2016-01-07
The GWT debugger options are confusing


supersteves:
There are many checkboxes and they don't seem to correlate obviously to

* command-line options ```java -jar compiler.jar --help```
* CompilerOptions.java methods/fields (which I personally use directly) https://github.com/google/closure-compiler/wiki/Using-the-Compiler's-Java-API
* service API https://developers.google.com/closure/compiler/docs/api-ref
* additional webservice options https://github.com/google/closure-compiler/wiki/Additional-Webservice-Options

All this is rather confusing. To make it slightly easier when trying to create a test case in the debugger, I suggest updating the CompilationParam enum to have additional properties which describe how they map to at least some of the above options, and setting a tooltip in DebuggerGwtMain.

I might do this in a spare moment if you indicate it will be accepted as a PR.

dimvar:
We use the debugger to debug individual passes. If you want to repro something using simple or advanced optimizations, why can't you use http://closure-compiler.appspot.com?

supersteves:
I use the debugger to:

- find out what types are being inferred
- try to reproduce isolated test cases for specific weird errors or crashes so I can submit them; often I need to specify externs

... also I only discovered that other page yesterday!

I also don't know how to map the ~50 customisations we make (to CompilerOptions.java) to this page, e.g. specifying language in/out? If these are command-line options, I'm only familiar with the Java API.

All I'm suggesting is adding a tooltip to explain what Java API (and probably command-line option) each checkbox correlates to. And doing it myself. But only if it won't get rejected.

MatrixFrog:
The debugger UX could definitely use some love. I don't see a reason to reject it, but I kinda feel like I'm still digging myself out from a pile of email after the holidays (which is why I haven't responded on several existing PRs), so it might take some time to get merged.

PR #1802, opened 2016-05-20
Tooltips on debugger param checkboxes

supersteves:
Adding tooltip to debugger compilation params that clarifies what Java API methods/options they correlate to, where unclear.

Part resolves https://github.com/google/closure-compiler/issues/1374

I'd like also to know which checkboxes correlate to ADVANCED vs SIMPLE, but this is a start.

---
This change is [Reviewable](https://reviewable.io/reviews/google/closure-compiler/1802)

### DIFF
--- a/src/com/google/javascript/jscomp/debugger/CompilationParam.java
+++ b/src/com/google/javascript/jscomp/debugger/CompilationParam.java
@@ -43,6 +43,10 @@ enum CompilationParam {
         }
       }
     }
+    @Override
+    String getJavaInfo() {
+    	return "Sets all registered DiagnosticGroups to CheckLevel.WARNING";
+    }
   },
 
   /**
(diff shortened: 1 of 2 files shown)


MatrixFrog:
Nice. It might also be helpful to show the command-line equivalent, but since there isn't one for most of these (they're either part of simple or advanced mode) this is a nice start. We haven't done a lot with the debugger UI lately so it's nice to see some activity. I'll try to bring this in soon.

supersteves:
What I'd most like is buttons I could press which reset all checkboxes to SIMPLE or ADVANCED equivalents. Or an easy way to figure out what my active and finely tuned CompilerOptions config is closest to in the debugger.

Sometimes most of the problem is figuring out what I need to check in order to reproduce a weird bug. Other times I can easily do it by ticking items in order (but the ~30s page load time is a bit painful and seems to be getting worse). 

It's hard for me to develop anything much with the debugger because it's excluded from the Eclipse classpaths and so I'm coding blind without figuring out what GWT dependencies I need to add (I think the missing parts might be closed source).

supersteves:
Thanks!

PR #1808, opened 2016-05-21
Refined HTMLOutputElement

phistuck:
This change is [Reviewable](https://reviewable.io/reviews/google/closure-compiler/1808)

### DIFF
--- a/externs/browser/html5.js
+++ b/externs/browser/html5.js
@@ -3590,7 +3590,7 @@ HTMLOutputElement.prototype.name;
  * @const
  * @type {!string}
  */
-HTMLOutputElement.prototype.type = "output";
+HTMLOutputElement.prototype.type;
 
 /**
  * @type {string}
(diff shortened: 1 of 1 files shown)


externs/browser/html5.js
```diff
@@ -3570,9 +3570,64 @@ HTMLDataListElement.prototype.options;
...
+/**
+ * @type {string}
+ */
+HTMLOutputElement.prototype.name;
+
+/**
+ * @const
+ * @type {!string}
+ */
+HTMLOutputElement.prototype.type = "output";
```
Dominator008: Remove this assignment?

phistuck: This was intentional and I saw similar stuff (a function with an actual function body that returns a string, for example). This is a constant string in a read only attribute. Why is this problematic?

ChadKillingsworth: In general, externs should not have implementations. If they do - it's a mistake. It leads people to believe that the code actually executes.

phistuck: Done.

PR #1809, opened 2016-05-21
externs: useCapture parameter of EventTarget's addEventListener() and removeEventListener() methods should be optional

joshtynjala:
Documentation for these methods from MDN says its optional:

https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener
https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener

---

This change is [Reviewable](https://reviewable.io/reviews/google/closure-compiler/1809)

### DIFF
--- a/externs/browser/w3c_event.js
+++ b/externs/browser/w3c_event.js
@@ -32,19 +32,19 @@ function EventTarget() {}
 /**
  * @param {string} type
  * @param {EventListener|function(!Event):(boolean|undefined)} listener
- * @param {boolean} useCapture
+ * @param {boolean=} opt_useCapture
  * @return {undefined}
  */
-EventTarget.prototype.addEventListener = function(type, listener, useCapture)
+EventTarget.prototype.addEventListener = function(type, listener, opt_useCapture)
     {};
 
 /**
  * @param {string} type
  * @param {EventListener|function(!Event):(boolean|undefined)} listener
- * @param {boolean} useCapture
+ * @param {boolean=} opt_useCapture
  * @return {undefined}
  */
-EventTarget.prototype.removeEventListener = function(type, listener, useCapture)
+EventTarget.prototype.removeEventListener = function(type, listener, opt_useCapture)
     {};
 
 /**


Dominator008:
In the MDN documentation, there's this note:

> Note: useCapture became optional only in more recent versions of the major browsers; for example, it > was not optional before Firefox 6. You should provide this parameter for broadest compatibility.

So I guess the current version is better for compatibility with more browsers.

ChadKillingsworth:
If someone wants to try this out in IE7/8 and make sure it works, then this could probably be removed. IIRC FireFox was the only browser that was picky about having all 3 arguments present, and that restriction was removed many versions back.

joshtynjala:
> If someone wants to try this out in IE7/8 and make sure it works, then this could probably be removed.

According to the same MDN page, basic support for addEventListener() wasn't even added until IE9, and the useCapture argument was optional from the start. It says that before IE9, attachEvent() was used instead.

joshtynjala:
@googlebot I signed the CLA

Dominator008:
@Chadkillingsworth If Google is testing on very old Firefox browsers internally it would have to wait...

ChadKillingsworth:
@Dominator008 But this just makes the argument optional right? It's still part of the signature.

Dominator008:
@Chadkillingsworth Well people writing new code may omit the argument and find out their tests fail on, say, Firefox 3

Dominator008:
@Matrixfrog Are old Firefox versions (<6) still being tested internally? If not, we are safe to merge this.

MatrixFrog:
I can't say for sure there are zero projects that still care about Firefox <6 but it's probably okay to take this. Will give it a few days to see if anyone else objects.

dimvar:
We'll try to do an internal release early next week. Let's wait until after that.

concavelenz:
Tyler was working through updating these signatures.

MatrixFrog:
It looks like this PR is obsolete since this is now marked optional at https://github.com/google/closure-compiler/blob/50de7abb88f8f00c4e37b23da0515b840e9ed00c/externs/browser/w3c_event.js#L36 ?

MatrixFrog:
Oh cool when I commented it auto-unassigned me which is all I really wanted anyway ;)

PR #1829, opened 2016-06-06
Inline aliases inside of IIFEs

ChadKillingsworth:
Inline any constant alias within a scope that is executed exactly once. Enables aliases inlining in IIFEs created by module rewriting (`require.ensure` calls).

---
This change is [Reviewable](https://reviewable.io/reviews/google/closure-compiler/1829)

### DIFF
--- a/src/com/google/javascript/jscomp/InlineAliases.java
+++ b/src/com/google/javascript/jscomp/InlineAliases.java
@@ -57,18 +57,38 @@ final class InlineAliases implements CompilerPass {
     NodeTraversal.traverseEs6(compiler, root, new AliasesInliner());
   }
 
+  /**
+   * Check to see if the scope is global or a nested IIFE
+   */
+  static boolean isScopeGlobalOrIIFE(Scope scope) {
+    if (scope == null) {
+      return false;
+    }
+    while (scope != null && scope.isLocal()) {
+      if (scope.getRootNode().isFunction() && !scope.getRootNode().getParent().isCall()) {
+        return false;
+      }
+      if (!NodeUtil.isExecutedExactlyOnce(scope.getRootNode().getParent())) {
+        return false;
+      }
+
+      scope = scope.getParent();
+    }
+    return scope.isGlobal();
+  }
+
   private class AliasesCollector extends AbstractPostOrderCallback {
     @Override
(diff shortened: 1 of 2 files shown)


blickly:
What exactly is the code pattern that needs to be typechecked? I've been hoping to get rid of inlineAliases by doing file-local rewriting of modules instead.

ChadKillingsworth:
Yeah I'm not happy with this either. Here's the code pattern:

```js
(function() {
  const Foo = module$foo;
  new Foo();
})();
```

This code pattern can occur when:

```js
require.ensure(['./foo'], function(require) {
  const Foo = require('./foo');
  new Foo();
});
```

blickly:
Hmm, I think that the require module rewriting should inline the imports to end up with:
```javascript
(function() {
  new module$foo();
})();
```
rather than
```javascript
(function() {
  const Foo = module$foo;
  new Foo();
})();
```

That's at least how goog.module rewriting works.

ChadKillingsworth:
That makes much more sense. I'll work towards that then. It'll need to be changed for both ES6 and commonjs modules.

ChadKillingsworth:
@blickly I found a more difficult case:

Foo module:
```js
module.exports = {
  Foo
};
```

```js
(function() {
  const Foo = require('./foo').Foo;
  new Foo();
})()
```

It's this level of unwinding that makes ES6 and CommonJS modules more difficult. However, perhaps we could have an InlineModuleImports pass right after module rewriting that ONLY affected modules? Then we could run it in all scopes as well.

blickly:
We've recently disallowed this syntax for goog.module in favor of `const {Foo} = require('./foo');` which more closely resembles the ES6 import syntax.

But either way, yes, it's a little more work to inline a name reference with a qualified name. Would reusing the `safeSetMaybeQualifiedString` help?
https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/ClosureRewriteModule.java#L1320-1333

blickly:
And if you prefer to make a separate pass, that seems reasonable too. The important thing from my perspective is that the rewriting is local on a per-file level rather than global like inlineAliases.  (For reference, the goog.module rewriting is a single CompilerPass but does two passes over the AST)

PR #1838, opened 2016-06-11
Add optional param contentType to Blob.slice method

vobruba-martin:
This change is [Reviewable](https://reviewable.io/reviews/google/closure-compiler/1838)

### DIFF
--- a/externs/browser/fileapi.js
+++ b/externs/browser/fileapi.js
@@ -50,30 +50,33 @@ Blob.prototype.type;
  * @see http://www.w3.org/TR/FileAPI/#dfn-slice
  * @param {number} start
  * @param {number} length
+ * @param {string=} opt_contentType
  * @return {!Blob}
  * @nosideeffects
  */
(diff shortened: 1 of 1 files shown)


externs/browser/fileapi.js
```diff
@@ -50,30 +50,33 @@ Blob.prototype.type;
  * @see http://www.w3.org/TR/FileAPI/#dfn-slice
  * @param {number} start
  * @param {number} length
+ * @param {string=} contentType
```
MatrixFrog: Since it's optional it should be called `opt_contentType`

MatrixFrog:
Can you squash this into a single commit? That makes it a lot easier for us to import the change. I just added this as an instruction on https://github.com/google/closure-compiler/wiki/Contributors

MatrixFrog:
lgtm other than that

vobruba-martin:
Squashed. Thank you

PR #1871, opened 2016-07-01
Revert "Rewrite internal references to classes so that we don't crash…

redforks:
… on things like"

This reverts commit b78e9d690bca36ba0b78ac1330980142a613b8bd.

This commit cause error on Template es6 classes, such as:

/**
 * @constructor
 * @template T
 */
var A = class A {
  constructor() {}

  /**
   * @return {T}  // error here:  ERROR - Bad type annotation. Unknown type T
   */
  foo() {}
}

---
This change is [Reviewable](https://reviewable.io/reviews/google/closure-compiler/1871)

### DIFF
--- a/src/com/google/javascript/jscomp/Es6ExtractClasses.java
+++ b/src/com/google/javascript/jscomp/Es6ExtractClasses.java
@@ -16,17 +16,12 @@
 
 package com.google.javascript.jscomp;
 
-import static com.google.javascript.jscomp.Es6ToEs3Converter.CANNOT_CONVERT;
-
-import com.google.common.base.Preconditions;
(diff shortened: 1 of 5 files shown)


dimvar:
Tyler, AFAIU, breaking this double naming use case is expected? Can we give an easier to understand error message rather than this one?

MatrixFrog:
I don't think we can roll this back because there are some other changes in the works that depend on it. Yes, we should improve the error messages in cases like this. But if you rewrite it as

class A {...

or 

var A = class {...

I believe everything should work.

MatrixFrog:
Actually we should just copy over the JSDoc, for `@template` issues like this but also for things like `@unrestricted` or `@extends` etc. Working on that. Thanks.

PR #1891, opened 2016-07-14
Use the path for the maven-built jar now that we are recommending using maven to build above

MatrixFrog:


MatrixFrog:
(Did this as a PR instead of internally to make sure that I can 'git push' to here from my machine)

Dominator008:
LGTM

PR #1892, opened 2016-07-15
Warn if @abstract and @private appear on the same JSDocInfo

Dominator008:
"private abstract" methods don't make much sense so warn if one is
found.


dimvar:
LGTM. We'll merge this internally.

PR #1896, opened 2016-07-15
contrib/nodejs/globals.js: added some missing properties and methods on the global process object

joshtynjala:
Added a few things that were missing as documented here: https://nodejs.org/api/process.html

### DIFF
--- a/contrib/nodejs/globals.js
+++ b/contrib/nodejs/globals.js
@@ -104,10 +104,114 @@ function clearImmediate(id) {}
 function Process() {}
 
 /**
- * @param {Function} fn
- * @see http://nodejs.org/api/process.html#process_process_nexttick_callback
+ * @return {void}
+ * @see http://nodejs.org/api/process.html#process_process_abort
  */
-Process.prototype.nextTick = function (fn) {};
+Process.prototype.abort = function () {};
+
+/**
+ * @type {string}
+ * @see http://nodejs.org/api/process.html#process_process_arch
+ */
+Process.prototype.arch;
+
+/**
+ * @type {Array<string>}
+ * @see http://nodejs.org/api/process.html#process_process_argv
+ */
+Process.prototype.argv;
+
+/**
+ * @param {string} directory
+ * @see http://nodejs.org/api/process.html#process_process_chdir_directory
+ */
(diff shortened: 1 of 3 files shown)


contrib/nodejs/globals.js
```diff
@@ -104,10 +104,112 @@ function clearImmediate(id) {}
 function Process() {}
 
 /**
+ * @see http://nodejs.org/api/process.html#process_process_abort
+ */
+Process.prototype.abort = function () {};
```
MatrixFrog: If I remember correctly, this just tells the compiler that abort is a function, but the compiler won't know its param or return types. You need to add `@return {void}` to get the compiler to know that.

contrib/nodejs/globals.js
```diff
@@ -104,10 +104,112 @@ function clearImmediate(id) {}
...
+/**
+ * @return {string}
+ * @see http://nodejs.org/api/process.html#process_process_cwd
+ */
+Process.prototype.cwd = function () {};
+
+/**
+ * @see http://nodejs.org/api/process.html#process_process_disconnect
+ */
+Process.prototype.disconnect = function () {};
```
MatrixFrog: same here

contrib/nodejs/globals.js
joshtynjala: Added. Thanks for the tip!

contrib/nodejs/globals.js
```diff
@@ -104,10 +104,114 @@ function clearImmediate(id) {}
...
+
+/**
+ * @param {number} pid
+ * @param {string|number} signal
+ * @see http://nodejs.org/api/process.html#process_process_kill_pid_signal
+ */
+Process.prototype.kill = function (pid, signal) {};
+
+/**
  * @param {Function} fn
```
MatrixFrog: Should probably be {!Function} as I suspect passing null wouldn't make much sense.

MatrixFrog:
lgtm

MatrixFrog:
Oh and if you can squash this into a single commit, that would be great because it makes it easier to import the change.

Dominator008:
Copied in #2034.

PR #1928, opened 2016-08-02
Add Boolean.prototype.valueOf

vobruba-martin:


Dominator008:
Looks good, thanks!

externs/es3.js
```diff
@@ -935,6 +935,14 @@ Boolean.prototype.toSource = function() {};
 Boolean.prototype.toString = function() {};
 
 /**
+ * @return {boolean}
+ * @nosideeffects
+ * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean/valueOf
+ * @override
```
Dominator008: Not sure if this `@override` will cause a problem. Looks like `Boolean` does not extend from anything.

vobruba-martin: I know but `@override` is present in other Boolean methods so I've just copied it. Do you want to me to remove it? If yes, from all of them?

MatrixFrog: It looks like Boolean actually extends Object, because the JSTypeRegistry sets it up to be so, even though it doesn't say it explicitly in the externs file. This also reflects reality: http://jsconsole.com/?(new%20Boolean)%20instanceof%20Object

So I think the `@override` is correct.

PR #1930, opened 2016-08-02
Add missing states for XMLHttpRequest externs

zarov:
As seen here https://www.w3.org/TR/XMLHttpRequest/#states, some states are missing from the W3C XML extern.

---
This change is [Reviewable](https://reviewable.io/reviews/google/closure-compiler/1930)

### DIFF
--- a/externs/browser/w3c_xml.js
+++ b/externs/browser/w3c_xml.js
@@ -400,6 +400,36 @@ XMLHttpRequest.prototype.onreadystatechange;
 XMLHttpRequest.prototype.onerror;
 
 /**
+ * @type {number}
+ * @see https://www.w3.org/TR/XMLHttpRequest/#states
+ */
+XMLHttpRequest.UNSENT;
+
+/**
+ * @type {number}
+ * @see https://www.w3.org/TR/XMLHttpRequest/#states
+ */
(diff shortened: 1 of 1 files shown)


externs/browser/w3c_xml.js
```diff
@@ -400,6 +400,36 @@ XMLHttpRequest.prototype.onreadystatechange;
 XMLHttpRequest.prototype.onerror;
 
 /**
+ * @type {number}
+ * @see https://www.w3.org/TR/XMLHttpRequest/#states
+ */
+XMLHttpRequest.UNSENT = 0;
```
Dominator008: Assignments in externs are no-ops, so please remove them and only keep the type declaration.

zarov: No problem, done. But could you explain why some properties are assigned, like in this same file https://github.com/google/closure-compiler/blob/master/externs/browser/w3c_xml.js#L209 ? Is it bad behavior from before more strict writings of externs ?

Dominator008: We sometimes use assignments for aliasing constructors, but that's not a common use case.

MatrixFrog:
lgtm

PR #1938, opened 2016-08-05
Added missing methods for TypedArrays

vobruba-martin:
1) I've updated TypedArray's definitions according to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray). The diff is [here](https://github.com/google/closure-compiler/commit/52431d6f24faba0bd1573f95c197996e677de6b3)

2) I've simplified all Typed Array's definitions which leads to reducing number of lines from ~3400 to ~1400:

- `ITypedArray` is renamed to `TypedArray` because it's no longer treated as interface.
- `TypedArray` now extends `ArrayBufferView` and implements `IArrayLike<number>`.
- All TypedArrays' prototype methods are dropped because they all extend `TypedArray`

If you don't like it or ```ITypedArray``` is needed somewhere I will revert this change. The diff is [here](https://github.com/google/closure-compiler/commit/93ce55c34a647dca45fc6cd5d722009f4be7bdd9)

### DIFF
--- a/externs/es6.js
+++ b/externs/es6.js
@@ -330,41 +330,242 @@ var BufferSource;
 
 
 /**
- * An artificial interface to describe methods available on all TypedArray
- * objects so that they can be operated on in type safe but generic way.
- * @record
- * @extends {IArrayLike<number>}
+ * @constructor
+ * @implements {IArrayLike<number>}
+ * @extends {ArrayBufferView}
  */
-function ITypedArray() {}
-
-/** @type {number} */
-ITypedArray.prototype.length;
+function TypedArray() {};
 
 /** @const {number} */
-ITypedArray.prototype.BYTES_PER_ELEMENT;
+TypedArray.prototype.BYTES_PER_ELEMENT;
 
 /**
- * @param {ArrayBufferView|Array<number>} array
- * @param {number=} opt_offset
- * @return {undefined}
- */
-ITypedArray.prototype.set = function(array, opt_offset) {};
-
-/**
- * @param {number} begin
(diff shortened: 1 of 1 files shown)


MatrixFrog:
lgtm, will check if anyone is depending on the name ITypedArray or something.

vobruba-martin:
I will squash all the commits if everything is ok. Just let me know.

MatrixFrog:
Ah yes, please squash them together. Our importer still struggles with a PR that is multiple commits :(

MatrixFrog:
There seems to be a merge conflict, can you rebase?

vobruba-martin:
Rebased.

externs/es6.js
```diff
@@ -330,41 +330,228 @@ var BufferSource;
...
+ * @return {!TypedArray}
+ * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/sort
+ */
+TypedArray.prototype.sort = function(opt_compareFunction) {};
+
+/**
+ * @param {number} begin
  * @param {number=} opt_end
- * @return {ITypedArray}
+ * @return {!TypedArray}
```
MatrixFrog: To get Closure library to compile, I had to change this to

```
 * @return {THIS}
 * @this {THIS}
 * @template THIS
```

I assume the same would be correct for all methods that return a TypedArray?

vobruba-martin:
@MatrixFrog I've updated all methods which return TypedArray to return THIS template.

PR #1959, opened 2016-08-17
Update EventTarget according to the latest spec

vobruba-martin:
`EventTarget` is currently defined according to [DOM Level 2 spec](https://www.w3.org/TR/2000/REC-DOM-Level-2-Events-20001113/events.html#Events-EventTarget) and this is causing some problems:

1) You cannot pass options object when creating event listeners which is needed if you want to create passive or one time event listeners.

``` javascript
window.addEventListener('click', function() {}, { capture:false, passive:true, once:true });
```

```
JSC_TYPE_MISMATCH: actual parameter 3 of Window.prototype.addEventListener does not match formal parameter
found   : {
  capture: boolean,
  once: boolean,
  passive: boolean
}
required: (boolean|undefined) at line 1 character 48
window.addEventListener('click', function() {}, { capture:false, passive:true...
                                                ^
```

2) All 3 arguments are required on interfaces which extend `EventTarget` but do not override its methods. For example `WorkerGlobalScope`:

``` javascript
/** @type {!WorkerGlobalScope} */
var x;
x.addEventListener('click', function() {});
```

```
JSC_WRONG_ARGUMENT_COUNT: Function EventTarget.addEventListener: called with 2 argument(s). Function requires at least 3 argument(s) and no more than 3 argument(s). at line 3 character 0
x.addEventListener('click', function() {});
^
```

This PR updates `EventTarget` according to [the latest spec](https://dom.spec.whatwg.org/#interface-eventtarget).

### DIFF
--- a/contrib/externs/w3c_eventsource.js
+++ b/contrib/externs/w3c_eventsource.js
@@ -29,26 +29,15 @@
  */
 function EventSource(url, opt_eventSourceInitDict) {}
 
-/**
- * @param {boolean=} opt_useCapture
- * @override
- * @return {undefined}
- */
-EventSource.prototype.addEventListener = function(
-    type, listener, opt_useCapture) {};
+/** @override */
+EventSource.prototype.addEventListener = function(type, listener, opt_options)
+    {};
 
-/**
- * @param {boolean=} opt_useCapture
- * @override
- * @return {undefined}
- */
-EventSource.prototype.removeEventListener = function(
-    type, listener, opt_useCapture) {};
+/** @override */
+EventSource.prototype.removeEventListener = function(type, listener, opt_options)
+    {};
 
-/**
- * @override
- * @return {boolean}
- */
+/** @override */
 EventSource.prototype.dispatchEvent = function(evt) {};
 
 /**
--- a/externs/browser/fileapi.js
+++ b/externs/browser/fileapi.js
@@ -483,26 +483,15 @@ FileError.prototype.code;
  */
 function FileReader() {}
 
-/**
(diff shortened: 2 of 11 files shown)


aghassemi:
@MatrixFrog @Dominator008 What's the likelihood of this PR getting accepted? We are running into the same issue where we like to use the new Event API and considering we can't override builtin externs in our custom extern with a different signature, we are kind of stuck.

MatrixFrog:
Sounds good based on a quick glance. I can probably review it this week if someone else doesn't get to it first.

MatrixFrog:
This turns out to be harder to land than I thought. Every subclass of EventTarget has to be updated if it has the `@param` annotations written out instead of just `@override`. You've fixed a bunch of them in this PR but there are at least a few others within Google's codebase. Hopefully I can get them all updated in the next few days, and get this submitted.

vobruba-martin:
Thank you!

MatrixFrog:
Getting close.

vobruba-martin:
Rebased because of this commit 3b23f1e2029b30001925ae4b5c327165f1b07d91

@MatrixFrog Do you have any ETA on this?

vobruba-martin:
Rebased

MatrixFrog:
Waiting on a response from an internal team, so unfortunately I do not.

vobruba-martin:
@MatrixFrog Any update? :-)

MatrixFrog:
I'm going to really really try to get this in by next week.

MatrixFrog:
This is landed internally. However, don't get excited yet; we might need to roll it back for some reason.

PR #1963, opened 2016-08-18
Fix ResponseInit in Fetch API

vobruba-martin:
As you can see [here](https://fetch.spec.whatwg.org/#responseinit) all ResponseInit properties are optional


vobruba-martin:
I will drop this PR because I'm going to update whole Fetch API extern

PR #1966, opened 2016-08-19
Update w3c_audio externs to last published version

zarov:
See last published version from  08 December 2015 https://www.w3.org/TR/webaudio/, while these externs weren't updated since 2012.

---

This change is [Reviewable](https://reviewable.io/reviews/google/closure-compiler/1966)

### DIFF
--- a/externs/browser/w3c_audio.js
+++ b/externs/browser/w3c_audio.js
@@ -17,7 +17,7 @@
 /**
  * @fileoverview Definitions for the API related to audio.
  * Definitions for the Web Audio API.
- * This file is based on the W3C Working Draft 15 March 2012.
+ * This file is based on the W3C Working Draft 08 December 2015.
  * @see http://www.w3.org/TR/webaudio/
  *
  * @externs
@@ -42,7 +42,7 @@ BaseAudioContext.prototype.listener;
 
 /**
  * @type {string}
- * @see https://developer.mozilla.org/en-US/docs/Web/API/AudioContext/state
+ * See https://www.w3.org/TR/webaudio/#BaseAudioContext for valid values
  */
 BaseAudioContext.prototype.state;
 
@@ -56,10 +56,10 @@ BaseAudioContext.prototype.createBuffer =
     function(numberOfChannels, length, sampleRate) {};
 
 /**
- * @param {ArrayBuffer} audioData
- * @param {function(AudioBuffer)=} successCallback
+ * @param {!ArrayBuffer} audioData
+ * @param {function(!AudioBuffer)=} successCallback
  * @param {function(?)=} errorCallback
- * @return {!Promise<AudioBuffer>}
+ * @return {!Promise<!AudioBuffer>}
  */
 BaseAudioContext.prototype.decodeAudioData =
     function(audioData, successCallback, errorCallback) {};
@@ -70,34 +70,8 @@ BaseAudioContext.prototype.decodeAudioData =
 BaseAudioContext.prototype.createBufferSource = function() {};
 
 /**
- * @param {HTMLMediaElement} mediaElement
- * @return {!MediaElementAudioSourceNode}
- */
-BaseAudioContext.prototype.createMediaElementSource = function(mediaElement) {};
-
-/**
- * @param {MediaStream} mediaStream
- * @return {!MediaStreamAudioSourceNode}
- */
-BaseAudioContext.prototype.createMediaStreamSource = function(mediaStream) {};
-
-/**
- * @return {!MediaStreamAudioDestinationNode}
- */
-BaseAudioContext.prototype.createMediaStreamDestination = function() {};
-
-/**
- * To be deprecated. Use createScriptProcessor instead.
- * @param {number} bufferSize
- * @param {number} numberOfInputs
- * @param {number} numberOfOuputs
- * @return {!ScriptProcessorNode}
- */
-BaseAudioContext.prototype.createJavaScriptNode = function(bufferSize,
-    numberOfInputs, numberOfOuputs) {};
-
-/**
- * @param {number} bufferSize
+ * @deprecated Use createAudioWorker instead
+ * @param {number=} bufferSize
  * @param {number=} numberOfInputChannels_opt
  * @param {number=} numberOfOutputChannels_opt
  * @return {!ScriptProcessorNode}
@@ -111,22 +85,9 @@ BaseAudioContext.prototype.createScriptProcessor = function(bufferSize,
 BaseAudioContext.prototype.createAnalyser = function() {};
 
 /**
- * @deprecated Use createGain instead.
- * @return {!GainNode}
- */
-BaseAudioContext.prototype.createGainNode = function() {};
-
-/**
  * @return {!GainNode}
  */
 BaseAudioContext.prototype.createGain = function() {};
-
-/**
- * To be deprecated. Use createDelay instead.
- * @param {number=} maxDelayTime
- * @return {!DelayNode}
- */
-BaseAudioContext.prototype.createDelayNode = function(maxDelayTime) {};
 
 /**
  * @param {number=} maxDelayTime
@@ -140,19 +101,13 @@ BaseAudioContext.prototype.createDelay = function(maxDelayTime) {};
 BaseAudioContext.prototype.createBiquadFilter = function() {};
 
 /**
- * @param {!IArrayLike<number>} feedforward
- * @param {!IArrayLike<number>} feedback
- * @return {!IIRFilterNode}
- */
-BaseAudioContext.prototype.createIIRFilter = function(feedforward, feedback) {};
-
-/**
  * @return {!WaveShaperNode}
  */
 BaseAudioContext.prototype.createWaveShaper = function() {};
 
 /**
- * @return {!AudioPannerNode}
+ * @deprecated Use BaseAudioContext#createSpatialPanner or BaseAudioContext#createStereoPanner
+ * @return {!PannerNode}
  */
 BaseAudioContext.prototype.createPanner = function() {};
 
@@ -168,13 +123,13 @@ BaseAudioContext.prototype.createConvolver = function() {};
 
 /**
  * @param {number=} numberOfOutputs
- * @return {!AudioChannelSplitter}
+ * @return {!ChannelSplitterNode}
  */
 BaseAudioContext.prototype.createChannelSplitter = function(numberOfOutputs) {};
 
 /**
  * @param {number=} numberOfInputs
- * @return {!AudioChannelMerger}
+ * @return {!ChannelMergerNode}
  */
 BaseAudioContext.prototype.createChannelMerger = function(numberOfInputs) {};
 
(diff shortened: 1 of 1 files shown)


zarov:
Hi, any news on this ?

brad4d:
@zarov, there appear to be some problems with this externs file.

```
javascript/externs/w3c_audio.js:238: ERROR - property suspend already defined on superclass BaseAudioContext; use @override to override it
OfflineAudioContext.prototype.suspend = function() {};
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

javascript/externs/w3c_audio.js:243: ERROR - property resume already defined on superclass BaseAudioContext; use @override to override it
OfflineAudioContext.prototype.resume = function() {};
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

In this case, I think these 2 definitions can just be dropped. They're redundant anyway.
The compiler bails out at this point, so there may be more.

zarov:
Yes indeed, it seems that I duplicated several values like this. Do you have a test to run for checking the others duplicates ?

brad4d:
@zarov since this is one of the standard externs files, any advanced mode compilation should have caught this problem.
@ChadKillingsworth suggests this:

```shell
echo "alert('hi')" | java -jar compiler.jar -O=ADVANCED -W=VERBOSE
```

brad4d:
@zarov let me know when you have an update & have squashed it to a single commit.
I'll re-do the import and restart testing at that point.

brad4d:
It appears we also have at least one internal project that uses the constants defined on
`AudioBufferSourceNode.prototype` such as `PLAYING_STATE`, so it's not safe to remove them yet.
Could you instead mark these as `@deprecated`?

zarov:
Just updated, I got no warning/error on my side. I saw that I was a little too much picky when removing deprecated properties (but not marked as deprecated in the specification, only not present at all). Do you think it would be better if I move those properties as `@deprecated` for compatibility on your side, instead of deleting them ?

externs/browser/w3c_audio.js
```diff
@@ -26,33 +26,64 @@
...
-/** @type {!AudioListener} */
-AudioContext.prototype.listener;
+/**
+ * @type {!AudioListener}
+ * @readonly
+ */
+BaseAudioContext.prototype.listener;
 
 /**
+ * @readonly
```
brad4d: `@readonly` is not an annotation that closure-compiler understands.
https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/parsing/Annotation.java

Please drop it, since including it here misleads the reader into thinking it means something.

externs/browser/w3c_audio.js
```diff
@@ -232,39 +251,49 @@ function AudioNode() {}
...
+/**
+ * @type {number}
+ * @readonly
+ */
 AudioNode.prototype.numberOfInputs;
 
-/** @type {number} */
+/**
+ * @type {number}
+ * @readonly
```
brad4d: I think `@const` may be what you want here.

brad4d:
@zarov 
Please do mark things `@deprecated` instead of removing them wherever possible.
Thanks.

zarov:
@brad4d ok, just redid a complete pass on the file. I re-added deleted properties/methods as deprecated, change all `@readonly` to `@const`, so now it should be good. I got no error when testing on my side.

brad4d:
@zarov 
Our PR import tool is choking on this PR.
We believe this is because your branch is based on a very old commit from google/closure-compiler master.
Could you please rebase onto the current head of our master branch?
Thanks.

(It choked the first time I tried to import it, too. I manually copied then as a workaround, but that isn't as practical as we go back and forth on changes.)

zarov:
Oh sorry about that, thought I add rebase it ! Just did it, should be good now.

externs/browser/w3c_audio.js
```diff
@@ -26,245 +26,305 @@
...
  * @constructor
  */
-function AudioContext() {}
+function BaseAudioContext() {};
 
-/** @type {!AudioDestinationNode} */
-AudioContext.prototype.destination;
+/**
+ * @type {number}
+ * @const
```
brad4d: The compiler automatically recognizes a type given after `@const`, so this can just be.

```javascript
/** @const {number} */
```

externs/browser/w3c_audio.js
```diff
@@ -26,245 +26,305 @@
...
+ */
+BaseAudioContext.prototype.currentTime;
```
brad4d: I appreciate that you want to define things in a logical order and alphabetize these properties.
That's a good idea to make finding them easier for the reader.
Unfortunately doing it now makes it harder to review the functional changes.

If you like, you could do a separate PR that just reorders things, then rebase this PR onto that.
That should be a super easy PR to approve.

If you prefer not to do that, please reorder these in a way that minimizes the differences to review.
Thanks.

externs/browser/w3c_audio.js
```diff
@@ -292,56 +352,63 @@ AudioDestinationNode.prototype.maxChannelCount;
...
  */
-AudioParam.prototype.linearRampToValueAtTime = function(value, endTime) {};
+AudioParam.prototype.exponentialRampToValueAtTime = function(value, endTime) {};
 
 /**
  * @param {number} value
  * @param {number} endTime
- * @return {AudioParam}
+ * @return {!AudioParam}
+ * @throws {TypeError} if endTime is negative or not a finite number
```
brad4d: To avoid ambiguity our internal style guide requires that type annotations always specify explicitly whether a non-primitive type is nullable or not. (Primitive types are non-nullable by default.)

Should this be `{!TypeError}`?

externs/browser/w3c_audio.js
```diff
@@ -401,357 +478,829 @@ DelayNode.prototype.delayTime;
...
 AudioBuffer.prototype.duration;
 
-/** @type {number} */
-AudioBuffer.prototype.numberOfChannels;
-
 /**
- * @param {number} channel
- * @return {Float32Array}
+ * @deprecated
+ * @type {AudioGain}
```
brad4d: Should this be `{!AudioGain}` or `{?AudioGain}`?

brad4d:
@zarov 
Thanks a lot for rebasing. That fixed the import.
I've requested a few more changes, but I'll go ahead and see if what you have now passes testing.

zarov:
@brad4d no problem at all, thanks for taking all this time for review. I reorder it and squash the commit of this PR. I checked all the nullable/non-nullable properties according to the specification, it should be good now.

externs/browser/w3c_audio.js
```diff
@@ -26,33 +26,39 @@
...
+BaseAudioContext.prototype.destination;
 
-/** @type {number} */
-AudioContext.prototype.sampleRate;
+/** @const {number} */
+BaseAudioContext.prototype.sampleRate;
 
-/** @type {number} */
-AudioContext.prototype.currentTime;
+/** @const {number} */
```
brad4d: I may have misled you a bit. I've just realized it could be a problem to mark something `@const` when its value could change, you just aren't allowed to assign to it. I've confirmed with @dimvar that although the compiler doesn't currently have any optimizations that would break code annotated that way, it might in the future.

To sum up, we don't have a good way to annotate something as `@readonly`. If its value truly won't ever change, you can use `@const`, but if it could change you just have to use `@type` to declare its type and leave it at that.

I've filed #2306 for this.

externs/browser/w3c_audio.js
```diff
@@ -61,160 +67,189 @@ AudioContext.prototype.createBuffer =
...
+BaseAudioContext.prototype.suspend = function() {};
 
 /**
  * @return {!Promise<void>}
- * @see https://developer.mozilla.org/en-US/docs/Web/API/AudioContext/close
  */
-AudioContext.prototype.close = function() {};
+BaseAudioContext.prototype.close = function() {};
+
+/** @type {EventTarget} */
```
brad4d: I think this should be `/** @type {?EventTarget} */`

externs/browser/w3c_audio.js
```diff
@@ -61,160 +67,189 @@ AudioContext.prototype.createBuffer =
...
+BaseAudioContext.prototype.onstatechange;
+
+/**
+ * @param  {string} scriptURL
+ * @return {!Promise<AudioWorker>}
```
brad4d: `{!Promise<!AudioWorker>}`

externs/browser/w3c_audio.js
```diff
@@ -61,160 +67,189 @@ AudioContext.prototype.createBuffer =
...
+ */
+BaseAudioContext.prototype.createAudioWorker = function(scriptURL) {};
+
+/**
+ * @param  {Array<number>} feedforward
```
brad4d: `{!Array<number>}` or maybe `{!Array<number>=}`

brad4d:
@zarov 
Thanks a lot. That made it much easier to review.
Sent you some more comments.

externs/browser/w3c_audio.js
zarov: Well thought ! Reverting this change then.

zarov:
@brad4d just updated, should be even easier, lines changed have been reduced.

externs/browser/w3c_audio.js
```diff
@@ -485,156 +539,323 @@ AudioBufferSourceNode.prototype.loopStart;
...
+ */
+function AudioWorkerGlobalScope() {};
+
+/** @type {!EventTarget} */
+AudioWorkerGlobalScope.prototype.onaudioprocess;
+
+/** @type {!EventTarget} */
+AudioWorkerGlobalScope.prototype.onnodecreate;
+
+/** @type {Array<AudioWorkerParamDescriptor>} */
```
brad4d: Maybe `{!Array<!Audio...`?

externs/browser/w3c_audio.js
```diff
@@ -992,18 +1437,16 @@ function webkitAudioPannerNode() {}
  * @constructor
  * @extends {HTMLAudioElement}
  */
-function Audio(src) {}
+function Audio(src) {};
```
brad4d: Here and elsewhere function declarations shouldn't have terminating `;`. You only need those when you're assigning a function to a variable or property.

externs/browser/w3c_audio.js
zarov: Yep, missed it ! Well seen !

externs/browser/w3c_audio.js
zarov: Ok, just update with removing all non useful `;`

brad4d:
Testing a fresh import...

externs/browser/w3c_audio.js
```diff
@@ -919,40 +1344,54 @@ OscillatorNode.prototype.frequency;
 /** @type {!AudioParam} */
 OscillatorNode.prototype.detune;
 
-/** @type {function(number)} */
-OscillatorNode.prototype.start;
+/**
+ * @type {string}
+ * See https://www.w3.org/TR/webaudio/#the-oscillatornode-interface for valid values
+ */
+OscillatorNode.prototype.type;
```
brad4d: Duplicate definition causes compiler error.

brad4d:
Looks like there's at least one duplicate definition.
I added a comment on the file.

zarov:
@brad4d fixed, hoping I didn't missed anything again !

brad4d:
testing a fresh import...

brad4d:
Alas! IIRFilterNode is also duplicated.

zarov:
Yes, that's because of another commit yesterday... rebasing the file from the current state ! 
See https://github.com/google/closure-compiler/commit/fc1caf7bfa49ec4c888411daf3c17e9b524d443a

zarov:
Pushed the change back !

brad4d:
Importing and testing...

brad4d:
\o/ Tests passed.
Doing internal code review.

brad4d:
Had to rollback internal submit

```
REDACTED_FILE_NAME.js:24: ERROR - variable AudioBufferSourceNode.prototype.detune redefined with type (Object|null), original definition at javascript/externs/w3c_audio.js:546 with type AudioParam
AudioBufferSourceNode.prototype.detune;
```

brad4d:
More errors:

```
REDACTED/Sound.js:289: ERROR - actual parameter 1 of AudioNode.prototype.connect does not match formal parameter
found   : (DynamicsCompressorNode|null)
required: (AudioNode|AudioParam)
  gainNode.connect(this.compressor_);
                   ^^^^^^^^^^^^^^^^

REDACTED/Sound.js:316: ERROR - actual parameter 1 of AudioNode.prototype.connect does not match formal parameter
found   : (BiquadFilterNode|null)
required: (AudioNode|AudioParam)
    gainNode.connect(this.highFilter_);
                     ^^^^^^^^^^^^^^^^
```

brad4d:
The "redefined" error I posted first is because an internal project just defined something in their own externs that should be in ours.
I'll fix that internally.
The other errors look like a backward compatibility problem to me.

brad4d:
@zarov 
You can hold off on making further changes for now.

I'm going to see what it takes to update the project's code to match your update.

zarov:
@brad4d no problem ! As you said, the two other errors might need to be fixed internally, as non-primitive type should be non-nullable, as it is the problem here.

Let me know if I need to apply any other change. Thanks for the whole review ;)

brad4d:
Submitted an internal fix for the 'detune redefined' error.
I'm going to take a look at the errors in `Sound.js` next.

brad4d:
FYI, I'm doing internal changes to just make the first argument to `AudioNode.prototype.connect` non-nullable & fix whatever that breaks.

brad4d:
I've submitted fixes needed for 2 pieces of internal code, and I expect a change that makes the first argument to AudioNode.prototype.connect non-nullable to be submitted internally today.

zarov:
Thanks for doing all that work 😃

brad4d:
@zarov please rebase to resolve the conflict with my commit.
Thanks

zarov:
@brad4d done, basically just reverted ebaf022a94a81544fe4533a0541d8142acf0e0e6

brad4d:
Importing rebase and looking for the next fix to make...

brad4d:
I've had to update a few references from `AudioChannelSplitter` or `AudioChannelMerger` to `ChannelSplitterNode` or `ChannelMergerNode`.

I'm rerunning tests with those changes and hope to submit this PR internally tomorrow.

brad4d:
Discovered a couple more files referring to `AudioChannel*` instead of `Channel*Node`.
I've also discovered some cases that still use `RealtimeAnalyserNode`.
So this is going to take a bit longer.

brad4d:
Submitted changes to remove uses of `RealtimeAnalyserNode`.
Last round of testing found a few more errors.

brad4d:
@zarov 
I believe I've discovered a bug in your commit.
Several `onsomething` properties are defined as `{!EventTarget}`.
I believe they should all be `{!EventListener}`.

zarov:
@brad4d 
Well in the spec, they are all specified as `EventHandler`. I thought the closest type was `EventTarget`, but I could have been misguided. If you confirm the referring type is `EventListener` indeed, I'm changing it !

brad4d:
@zarov 
An `EventTarget` is an object on which events happen, not the code that does the handling.
I was mistaken about using `EventListener`, though.
Looking at other `.onsomething` properties in `html5.js` I see that we use '`{?function(!Event)}`',
so you should do the same here.

zarov:
@brad4d that sounds good to me, just pushed changes back.

zarov:
@brad4d I see new conflicts coming from nowhere, did I miss something or do I need to rebase and fix the conflicts on my side ?

brad4d:
@zarov 
An internal user made a partial version of your change independently of me.
I tried rebasing and pushing back to your branch, but I don't think you've granted me permission to do that.
Please rebase.

zarov:
@brad4d ok no problem, I added you in case it happens again ! Just rebase it back

brad4d:
@zarov 
I've imported your latest changes and am testing them internally.
Unfortunately, I already know there are going to be a few more bits of internal code I need to fix.

brad4d:
test-all-the-things failed, too many things affected.
hmm, this worked before.
trying again.

brad4d:
I now have 2 classes of failures:

**nullability problems**
Solution here is to fix the client code to avoid passing things the compiler thinks might be null.
3 files.

**Splitter/Merger issues**

previously: AudioNode -> AudioChannelSplitter
now: AudioNode -> ChannelSplitterNode -> AudioChannelSplitter

AudioChannelSplitter is now deprecated and all APIs are defined to take and 
return ChannelSplitterNode.

Same situation exists for ChannelMergerNode and AudioChannelMerger
4 files

fix:
1. Partial extern change: Add the new types to externs. Make methods and fields
accept the newer, base type, but have them return the older type for now.
2. Fix the client code to define everything in terms of the new type.
3. Partial extern change: update return values to use the new type.
4. maybe remove the deprecated type? Is it even really usable after step 3?

brad4d:
Submitted an internal change with almost everything from this PR.
There are TODOs for the bits that require internal fixes first.

brad4d:
I have 3 internal changes up for review to fix the nullability issues and splitter/merger type changes.
When they get submitted, I'll do a change that fixes the externs, and we'll be done.

zarov:
Yes wonderful, very happy to be finally here :)

brad4d:
Last internal fixes were submitted on Friday.
I'm submitting the fix for the TODOs today to close this PR.

brad4d:
A change fixing the TODOs was submitted this morning.
It should get pushed to github tomorrow and automatically close this PR.

zarov:
Thank you @brad4d for all this work !

PR #1967, opened 2016-08-19
Fetch API update based on Living Standard — Last Updated 17 August 2016

vobruba-martin:
See https://fetch.spec.whatwg.org/

### DIFF
--- a/externs/browser/fetchapi.js
+++ b/externs/browser/fetchapi.js
@@ -18,21 +18,43 @@
  *
  * This api is still in development and not yet stable. Use at your
  * own risk.
+ * 
+ * Based on Living Standard — Last Updated 17 August 2016
  *
  * @see https://fetch.spec.whatwg.org/
  * @externs
  */
 
+
 /**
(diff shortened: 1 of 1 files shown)


Dominator008:
Closed by https://github.com/google/closure-compiler/commit/f711e0c0e2f2e2cc0b69a91ab7618397e914d5ce

I changed `@inheritDoc`s to `@override`s because the former is deprecated and more confusing. I had to rename `RequestType` to `FetchRequestType` because the former conflicts with a global variable name from a project.

vobruba-martin:
@Dominator008 Thank you! The `@inheritDoc` deprecation should be mentioned [here](https://developers.google.com/closure/compiler/docs/js-for-compiler). Or is there any other page which is being actively used?

BTW is there any reason why to keep enums like `RequestType` in externs?

Dominator008:
@vobruba-martin No it's not explicitly documented, more of a best practice thing.

There's no reason to keep using those enums in externs. They should just be strings with a set of accepted values. I plan to migrate them to “string literal type” after I implement that.

PR #1968, opened 2016-08-19
Streams API updated based on Living Standard — Last Updated 5 August 2016

vobruba-martin:
See https://streams.spec.whatwg.org/

### DIFF
--- a/externs/browser/streamsapi.js
+++ b/externs/browser/streamsapi.js
@@ -14,42 +14,21 @@
  * limitations under the License.
  */
 /**
- * @fileoverview Definitions from the Streams API as of 22 June 2015.
- * @see https://streams.spec.whatwg.org/commit-snapshots/b2131ef6460c446817767d7575698f9fcfa554a3/
+ * @fileoverview Streams API definitions
+ * 
+ * Based on Living Standard — Last Updated 5 August 2016
+ * https://streams.spec.whatwg.org/commit-snapshots/34ecaadbcce8df9943d7a2cdb7fca4dc25914df4/
+ * 
+ * @see https://streams.spec.whatwg.org/
  * @externs
  */
 
-/**
- * @typedef {{ value: (*|undefined), done: boolean }}
- */
+
+/** @typedef {{ value:*, done:boolean }} */
 var IteratorResult;
 
-/**
- * @param {{ highWaterMark: number }} config
- * @constructor
- */
-function ByteLengthQueuingStrategy(config) {}
 
 /**
- * @param {*} chunk
- * @return {number}
- */
-ByteLengthQueuingStrategy.prototype.size = function(chunk) {};
-
-/**
- * @param {{ highWaterMark: number }} config
- * @constructor
- */
-function CountQueuingStrategy(config) {}
-
-/**
- * @param {*} chunk
- * @return {number}
- */
-CountQueuingStrategy.prototype.size = function(chunk) {};
-
-/**
- * The value of {@code highWaterMark} in a custom strategy must be non-negative.
  * @typedef {!CountQueuingStrategy|!ByteLengthQueuingStrategy|{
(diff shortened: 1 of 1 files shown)


vobruba-martin:
Forced Travis rebuild...

externs/browser/streamsapi.js
```diff
@@ -14,294 +14,446 @@
...
 
 /**
- * @see https://streams.spec.whatwg.org/#reader-class
- * @param {!ReadableStream} stream
- * @constructor
+ * @param {*} err
+ * @return {undefined}
+ * @see https://streams.spec.whatwg.org/#rs-default-controller-error
  */
-function ReadableStreamReader(stream) {}
```
MatrixFrog: There are a couple of projects referencing this type by name so we need to change them to `ReadableStreamDefaultReader` which is why this is taking a little bit of time to submit internally. Sorry for the delay.

vobruba-martin: Thank you for the info! Any ETA on this?

Dominator008: Can't promise, but should be similar to how long it took me to merge the Fetch API externs change, which was a few days :)

externs/browser/streamsapi.js
```diff
@@ -14,294 +14,446 @@
...
+PipeOptions.prototype.preventCancel;
+
+
+/**
+ * @record
+ */
+function ReadableStreamSource() {};
+
+/** @type {(undefined|function((!ReadableByteStreamController|!ReadableStreamDefaultController)):(!Promise<*>|undefined))}*/
+ReadableStreamSource.prototype.start;
```
vobruba-martin: I would like to change this and `ReadableStreamSource.prototype.pull;` to 
```
{(undefined
|function(!ReadableByteStreamController):(!Promise<*>|undefined)
|function(!ReadableStreamDefaultController):(!Promise<*>|undefined))}
```
because currently I have to deal with `!ReadableByteStreamController|!ReadableStreamDefaultController` type inside these methods.

What do you think @MatrixFrog ? Will it be OK if I update this PR? Or would be better to leave it for another PR after this one is merged?

MatrixFrog: Okay, I'll update it in the internal copy of this change. You can update the PR too if you want but it isn't necessary.

vobruba-martin: Thanks! I merge this PR already when building CC so I would like to have it in sync.

vobruba-martin: @MatrixFrog loot at #1992 . I guess we'll have to revert the last change until this gets resolved?

MatrixFrog: I don't think so. It just means there will be some warnings if people build with the new type inference. If I get all the other blockers cleared before the NTI issue is resolved, I'll probably just add a suppression or something.

MatrixFrog: I was wrong. We have tests that check to make sure the externs are warning-free under NTI, so it looks like we need to go back to `function(!ReadableByteStreamController|!ReadableStreamDefaultController)` (which, yes, means you may have to cast the argument, inside your start function)

externs/browser/streamsapi.js
```diff
@@ -14,294 +14,446 @@
  * limitations under the License.
  */
 /**
- * @fileoverview Definitions from the Streams API as of 22 June 2015.
- * @see https://streams.spec.whatwg.org/commit-snapshots/b2131ef6460c446817767d7575698f9fcfa554a3/
+ * @fileoverview Streams API definitions
+ * 
+ * Based on Living Standard — Last Updated 5 August 2016
+ * https://streams.spec.whatwg.org/commit-snapshots/34ecaadbcce8df9943d7a2cdb7fca4dc25914df4/
+ *
```
Dominator008: lkl

PR #1981, opened 2016-08-27
Fixed a broken reference link

phistuck:
ReferrerPolicy was probably recently moved to the W3C repository.

By the way, I do not see it exposed on Window in Chrome. Why is this in the externs? Does Firefox expose it? Is it specified to be exposed at all?

### DIFF
--- a/externs/browser/fetchapi.js
+++ b/externs/browser/fetchapi.js
@@ -28,7 +28,7 @@
 
 /**
  * @enum {string}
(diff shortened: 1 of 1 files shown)


Dominator008:
Thanks, will merge this internally and export it.

You are right, it does not exist on window and should not be a global enum in the externs. Once we have a notion of "string literal type" these enums should be gone.

PR #1971, opened 2016-08-22
Allow applying input source maps to the generated source map.

mprobst:
This allows users to easily map from generated source that's an
input to the compiler back into their original source files.

E.g. when compiling TypeScript, users would have:
    input.ts -(a)-> input.js -(b)-> compiled.js

Where (a) is the TypeScript compilation operation, and (b) is Closure
Compiler. With this option set, the source map generated in (a) is
applied to the operation (b), with the final source map mapping back
into `input.ts` from locations in `compiled.js`.

This also introduces an interface `SourceFileMapping` to avoid
increasing the coupling and responsibility of the Compiler object.

PR #1982, opened 2016-08-27
Parse //# sourceMappingURL= comments that contain inline source maps.

mprobst:
Standalone source maps can already be specified using --source_map_input. This adds support for inline source maps using data URLs with base64 encoding. Inline source maps make it easier to preserve the source map / file association across build steps and file transfers.

### DIFF
--- a/src/com/google/JsComp.gwt.xml
+++ b/src/com/google/JsComp.gwt.xml
@@ -16,6 +16,7 @@
 <module rename-to="jscomp">
   <inherits name="com.google.common.base.Base"/>
   <inherits name="com.google.common.collect.Collect"/>
+  <inherits name="com.google.common.io.Io"/>
   <inherits name="com.google.gwt.json.JSON"/>
   <source path="debugging/sourcemap"/>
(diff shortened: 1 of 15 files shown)


mprobst:
This is based on the changes in #1971.

ChadKillingsworth:
I was kind of waiting on #1971 to get merged - makes the diff easier to read.

Of particular note is what behavior is expected if an input sourcemap is provided AND an inline sourcemap for the same file? The gulp sourcemap plugin would never do this, but we should at least know the behavior.

mprobst:
@ChadKillingsworth source maps in files override any from parameters. I think that makes sense - the closer to the source code itself you are, the more "binding".

Also, source map comments later in the file override earlier ones, assuming there are multiple. Also working as intended for me - we don't support arbitrary depth mapping, so taking the last one is the closest reasonable thing we can do (at least we'll correctly map one level).

ChadKillingsworth:
@mprobst sounds good to me.

mprobst:
@ChadKillingsworth PTAL, this is now rebased on the submitted `apply_input_source_maps` branch. I've also introduced an option that allows disabling the inline source map parsing, in case somebody runs into issues with it (e.g. memory usage).

Issue #1826, opened 2016-06-02
Transpile uses of Promise to ES3 & ES5


brad4d:
Adding transpilation support for ES Promise will allow anyone using closure compiler to use standardized ES Promises. No need to use goog.Promise or some other library's implementation. It will also us to also transpile the soon-to-be-standard async functions (issue #1391)

brad4d:
I've just submitted the fix for this within Google.
It should get pushed to github by some time tomorrow.

Dominator008:
Closed by 22bb540d182d8ca9266e36dd952b9e0333b1f4d4.

PR #1984, opened 2016-08-28
Promises/A+ compilance test script and instructions

brad4d:
Now that polyfilling of Promises (#1826) is partially implemented, we can run the Promise/A+ conformance tests against them.


brad4d:
Not sure why this was still open.
I guess I failed to label the internal submission of these files correctly.

PR #1986, opened 2016-08-29
Added QUnit

elipoultorak:


elipoultorak:
I signed it!

MatrixFrog:
It's usually a good idea to put `/** @externs */` at the top of an externs file. Other than that, I don't know QUnit but since this is generated I assume it's reasonably correct.

ChadKillingsworth:
All of those functions will be assumed to be:

```js
/**
 * @param {...*} args
 * @param return {*}
 */
function fnc(args) {}
```

We have historically only accepted externs into the official repo that have type information.

MatrixFrog:
+1, let's not merge this. However if you want to put it somewhere else we would be happy to link to it from our wiki.

PR #1972, opened 2016-08-22
Update Service Worker events

vobruba-martin:
1) `InstallEvent` was dropped because it has been replaced by `ExtendableEvent`.
MDN says:

> As of May 2015, the install event is an instance of ExtendableEvent rather than a child of it.

I've tested it in the latest Chrome and it is using the `ExtendableEvent`.

2) `ServiceWorkerGlobalScope.prototype.onmessage` handler is fixed so a passed event may call the `ExtendableEvent.prototype.waitUntil` method.

3) Fixed `PushEvent` constructor.

4) Added `onpush` and `onpushsubscriptionchange` from [Push API spec](https://w3c.github.io/push-api/).

PR #1994, opened 2016-09-02
Definitions of the Web Background Synchronization API

vobruba-martin:
See https://wicg.github.io/BackgroundSync/spec/

[In Chrome since version 49](https://www.chromestatus.com/feature/6170807885627392)


vobruba-martin:
Ups, I forgot #1972 should be merged first...

PR #2002, opened 2016-09-07
Add NUMBER_FORMATS configuration definition in angular local service

sbrunner:


sbrunner:
I signed it! :-)

contrib/externs/angular-1.5.js
```diff
@@ -1795,6 +1795,20 @@ angular.$locale = function() {};
...
 angular.$locale.prototype.id;
 
+/**
+ * @typedef {{
+ *     CURRENCY_SYM: (string),
+ *     GROUP_SEP: (string),
+ *     DECIMAL_SEP: (string)
+ * }}
+ */
+angular.$locale.NUMBER_FORMATS;
```
Dominator008: I think this should be a `@record` instead. We will eventually migrate other `@typedef {{...}}`s in this file to be `@record`s.

sbrunner: Thanks for your review, I just update it :-)

Dominator008: Sorry I wasn't looking too closely. `@record` doesn't make sense here. Since this is an actual property available from Angular, what you had was almost right. I should have told you to just change `@typedef` to `@type`.

sbrunner: It I put `@type` I get this error (in my project):
```
ERR! compile ...: ERROR - Bad type annotation. Unknown type angular.NUMBER_FORMATS
ERR! compile  * @type {angular.NUMBER_FORMATS}
ERR! compile           ^
ERR! compile 
ERR! compile 
ERR! compile 1
ERR! compile  error(s), 
ERR! compile 0
ERR! compile  warning(s)
ERR! compile 
ERR! Process exited with non-zero status, see log for more detail: 1 
```
:-(
When I see the rest of the file I think that the @typedef is the best...

Dominator008:
Actually I'm curious why these need to be added. According to @mprobst's comment in #1497, these are supposed to be implementation details and not added to the public externs. Could you explain your use case?

Dominator008:
Per the comment by @mprobst in #1497, these should not be included in the externs so I'm closing this PR for now.

sbrunner:
It's because the number format of angular is essentially build to format currency, than we write our own number format filter...
It's OK for me to keep it probate by default :-)
for the curious: https://github.com/camptocamp/ngeo/blob/4d8b9d4527ab7cce09b0162f23b1ec5bb3e06acf/src/services/filters.js#L53-L113

Issue #1399, opened 2016-01-14
generate extern files


IorPerry:
Hello,
I was secure to find this issue already written, but with search I don't found it.
My request is adds a output_info to generate files to add as extern in compilation of another file .

I see there is: http://www.dotnetwise.com/code/externs/ but I think it should be better add it on compiler.
As plus I use the post api to compile local files and should be great if I can generate externs files also.

I am pretty sure it was already discussed, but I don't find the old discussion...
Regards,
Antonino Perricone

Ps. I attached my simple builder in vb.net, maybe someone is curious.
[mybuild.vb.txt](https://github.com/google/closure-compiler/files/90118/mybuild.vb.txt)

Dominator008:
There's an ExternExportsPass which creates an externs file containing all exported symbols and properties, but I don't think there's a command line flag for that (yet).

IorPerry:
I see it was created almost 6 years ago, it looks abandoned...

MatrixFrog:
There is a flag for it in the Google-internal version of the compiler, and it is used by a variety of projects.

joeyparrish:
Is anyone planning to add this flag to the open-source compiler?  If not, I volunteer as tribute.

Dominator008:
@joeyparrish I don't think anyone's actively working on adding this flag, so welcome!

PR #2019, opened 2016-09-19
Add --create_externs_from_exports flag

joeyparrish:
Adds the --create_externs_from_exports flag, which allows the creation of externs from exported methods.

Also fixes a minor bug in AbstractCommandLineRunner that prevented the use of that flag with relative paths containing no separators.

Finally, changes the way generated externs are sorted so that they respect dependency order.  Alphabetical sorting is still used, but only within a script.  This is important for the generated externs of a large project to be used directly without editing.

Issue #1399


joeyparrish:
Added a couple more commits to fix the order of the generated externs.  PTAL.

Dominator008:
Thanks Joey. Let's review this internally :)

Issue #1953, opened 2016-08-15
Compiler breaks working ES6 code when compiling a class which extends Date()


beingmrkenny:
I've written a class which extends the Date() object. When I use this class directly, before compilation, it works as expected.

This is a simplified example, which works in Chrome currently:

    <script type="text/javascript">

        class Dative extends Date {

            constructor (dateData) {
                super();
                super.setTime(Date.parse(dateData));
            }

            addMilliseconds (ms) {
                super.setTime(super.getTime() + ms);
            }

            addDays (days) {
                var self = this;

                var h = super.getHours(),
                    i = super.getMinutes(),
                    s = super.getSeconds(),
                    u = super.getMilliseconds();

                days = days || 0;

                while (days !== 0) {

                    if (days > 0) {
                        super.setHours(23, 59, 59, 999);
                        self.addMilliseconds(1);
                        days--;
                    } else {
                        super.setHours(0, 0, 0, 0);
                        self.addMilliseconds(-1);
                        days++;
                    }

                }

                super.setHours(h, i, s, u);
            }

        }

        var testDate = new Dative('24 december 2016');
        console.log(testDate.getHours());
        testDate.addDays(1)
        console.log(testDate);
        testDate.addMilliseconds(2 * 60 * 60 * 1000);
        console.log(testDate);
    </script>

The compiled version of the code breaks, returning
"Uncaught TypeError: Method Date.prototype.setTime called on incompatible receiver [object Object]" — on the line `Date.prototype.setTime.call(this, Date.parse(a))`.

    <script type="text/javascript">

        'use strict';
        var $jscomp = {
                scope: {},
                inherits: function(a, b) {
                    function d() {}
                    d.prototype = b.prototype;
                    a.prototype = new d;
                    a.prototype.constructor = a;
                    for (var c in b)
                        if (Object.defineProperties) {
                            var e = Object.getOwnPropertyDescriptor(b, c);
                            e && Object.defineProperty(a, c, e)
                        } else a[c] = b[c]
                }
            },
            Dative = function(a) {
                Date.call(this);
                Date.prototype.setTime.call(this, Date.parse(a))
            };
        $jscomp.inherits(Dative, Date);
        Dative.UTC = Date.UTC;
        Dative.parse = Date.parse;
        Dative.now = Date.now;
        Dative.prototype.addMilliseconds = function(a) {
            Date.prototype.setTime.call(this, Date.prototype.getTime.call(this) + a)
        };
[...]

PR #2024, opened 2016-09-21
Use the return value of super() calls in ES6 constructors

### LINKED ISSUE #1669: Return value of super() calls not used for `this`
If a constructor returns a value other than this, then in subclasses super() calls to that constructor should use the result as this in the subclass constructor.

This is specified in 12.3.5.1, step 10 of SuperCall: http://www.ecma-international.org/ecma-262/6.0/index.html#sec-super-keyword

Getting this behavior correct is going to be very important for supporting Custom Elements, which takes advantage of this to initialize browser-allocated elements with user-written constructors. See https://w3c.github.io/webcomponents/spec/custom/#htmlelement-constructor

Note, Babel 6 implements this correctly.

The folowing code should not fail the assertion, and doesn't in native ES2015:

``` javascript
class Foo {
  constructor() {
    return {
      x: 1,
    };
  }
}

class Bar extends Foo {

  constructor() {
    super();
    this.y = 2;
  }
}

let o = new Bar();
console.assert(o.x === 1 && y === 2);
```

The compiled version does fail the assertion:

``` javascript
// Input 0
var Foo = function() {
  return {x:1};
};
var Bar = function() {
  Foo.call(this);
  this.y = 2;
};
$jscomp.inherits(Bar, Foo);
var o = new Bar;
console.assert(o.x === 1 && o.y === 2);
```

The output for Bar should be:

``` javascript
var Bar = function() {
  var _this = Foo.call(this);
  _this.y = 2;
  return _this;
};
```

ChadKillingsworth:
Changes transpilation of ES6 super calls in constructors so that if the call returns a value, that value is used instead of `this`. Because you cannot assign to the `this` keyword, the transpilation creates the alias `$jscomp$super$this` which is then used instead of the `this` keyword throughout the constructor.

In order to simplify type inference, the original name of the alias is set to `this`. OTI has been updated to recognize the alias. I'll need a little help with NTI (or it can be done internally if desired).

This is a continuation of the work @MatrixFrog did MatrixFrog@b7948b8541f

Fixes #1669.

### DIFF
--- a/src/com/google/javascript/jscomp/Es6ConvertSuper.java
+++ b/src/com/google/javascript/jscomp/Es6ConvertSuper.java
@@ -30,6 +30,12 @@ import com.google.javascript.rhino.Token;
  * {@link Es6ToEs3Converter} pass.
  */
 public final class Es6ConvertSuper implements NodeTraversal.Callback, HotSwapCompilerPass {
+  static final DiagnosticType NO_SUPERTYPE = DiagnosticType.error(
+      "JSC_NO_SUPERTYPE",
+      "The super keyword may only appear in classes with an extends clause.");
+
+  private static final String SUPER_THIS = "$jscomp$super$this";
+
   private final AbstractCompiler compiler;
 
   public Es6ConvertSuper(AbstractCompiler compiler) {
@@ -45,6 +51,7 @@ public final class Es6ConvertSuper implements NodeTraversal.Callback, HotSwapCom
           member = member.getNext()) {
         if (member.isMemberFunctionDef() && member.getString().equals("constructor")) {
           hasConstructor = true;
+          break;
         }
       }
       if (!hasConstructor) {
@@ -75,11 +82,12 @@ public final class Es6ConvertSuper implements NodeTraversal.Callback, HotSwapCom
       }
       Node body = IR.block();
       if (!classNode.isFromExterns()) {
-        Node exprResult = IR.exprResult(IR.call(
+        Node baseCall = IR.call(
             IR.getprop(superClass.cloneTree(), IR.string("apply")),
             IR.thisNode(),
-            IR.name("arguments")));
-        body.addChildToFront(exprResult);
+            IR.name("arguments"));
+
+        body.addChildToFront(IR.returnNode(IR.or(baseCall, IR.thisNode())));
       }
       Node constructor = IR.function(
           IR.name(""),
@@ -97,9 +105,13 @@ public final class Es6ConvertSuper implements NodeTraversal.Callback, HotSwapCom
     classMembers.addChildToFront(memberDef);
   }
 
-  private void visitSuper(Node node, Node parent) {
+  /**
+   * @return Whether this is a 'super' node that we know how to handle.
+   */
+  private boolean validateSuper(Node superNode, Node parent) {
+    Preconditions.checkArgument(superNode.isSuper());
     Node enclosingCall = parent;
-    Node potentialCallee = node;
+    Node potentialCallee = superNode;
     if (!parent.isCall()) {
       enclosingCall = parent.getParent();
       potentialCallee = parent;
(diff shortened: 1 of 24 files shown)


dimvar:
Since you changed a file in the typechecker, please add a test for the changed behavior. Either in `TypeCheck` with enabled transpilation, or in `IntegrationTest`.

ChadKillingsworth:
@dimvar done. I double checked that the test fails without the TypeInference changes.

ChadKillingsworth:
Fixed now in a much more elegant fashion.

src/com/google/javascript/rhino/Node.java
```diff
@@ -3045,6 +3045,10 @@ public boolean isTemplateLitSub(){
   }
 
   public boolean isThis() {
```
MatrixFrog: I think we want the is* methods to just check the token type and nothing else.

src/com/google/javascript/jscomp/Es6ConvertSuper.java
```diff
@@ -148,18 +184,56 @@ private void visitSuper(Node node, Node parent) {
...
     baseCall.useSourceInfoIfMissingFromForTree(enclosingCall);
-    enclosingCall.getParent().replaceChild(enclosingCall, baseCall);
+    nodeToReplace.getParent().replaceChild(nodeToReplace, baseCall);
     compiler.reportCodeChange();
   }
 
+  private static class UpdateThisReferences implements NodeTraversal.Callback {
+    @Override
+    public void visit(NodeTraversal t, Node n, Node parent) {
+      if (n.isThis()) {
```
MatrixFrog: So then this would need to be `n.isThis() || "this".equals(n.getOriginalName())`. Are there any other `isThis` calls that need to change?

MatrixFrog:
I'll probably add a runtime test or two when I bring this in. LGTM overall though.

src/com/google/javascript/rhino/Node.java
ChadKillingsworth: Then let me make a `NodeUtils.isThisOrAlias(Node n)` type method and look through the code base for any reference to `n.isThis()` type calls and switch it where needed.

ChadKillingsworth:
@MatrixFrog Changes done, commits squashed.

I tried to look at each `n.isThis()` call and determine whether it should consider aliases or not. I didn't just wholesale search/replace everything.

The PR just got a lot bigger though. I ran it on my project and it compiled with no warnings and the unit tests (runtime tests in a browser) ran with no issues.

MatrixFrog:
Sorry for the delay. We're failing on this unusual case:

```
class SubFoo extends Object {
  constructor() {
    super();
    assertTrue(this instanceof SubFoo);
  }
}
```

which comes from j2cl in some cases, and transpiles to

```
var SubFoo = function() {
  var $jscomp$super$this = Object.call(this) || this;
  assertTrue($jscomp$super$this instanceof SubFoo);
  return $jscomp$super$this;
};
$jscomp.inherits(SubFoo, Object);
```

`Object.call(this)` returns an `Object` which is not an instance of `SubFoo` (Object is one of those functions that can be called either as a constructor or as a normal function.) We either need to special case this and emit slightly different code for this case, or get the j2cl team to tweak their output.

brad4d:
FWIW, I've just confirmed that the non-transpiled form has the same
behavior. There's no help for it. If super() returns the wrong type of
object, that's just what you get. There's no ES6 magic that tries to "fix"
it so instanceof will work.

ChadKillingsworth:
I expected there to be at least one curveball on this one. It's too significant of a change.

@brad4d is correct. The Babel transpiled version has the same problem:

```js
function _classCallCheck(instance, Constructor) {
  if (!(instance instanceof Constructor)) {
    throw new TypeError("Cannot call a class as a function");
  }
}
function _possibleConstructorReturn(self, call) {
  if (!self) {
    throw new ReferenceError("this hasn't been initialised - super() hasn't been called");
  }
  return call && (typeof call === "object" || typeof call === "function") ? call : self;
}
function _inherits(subClass, superClass) {
  if (typeof superClass !== "function" && superClass !== null) {
    throw new TypeError("Super expression must either be null or a function, not " + typeof superClass);
  }
  subClass.prototype = Object.create(superClass && superClass.prototype, {constructor:{value:subClass, enumerable:false, writable:true, configurable:true}});
  if (superClass) {
    Object.setPrototypeOf ? Object.setPrototypeOf(subClass, superClass) : subClass.__proto__ = superClass;
  }
}
var SubFoo = function(_Object) {
  _inherits(SubFoo, _Object);
  function SubFoo() {
    _classCallCheck(this, SubFoo);
    var _this = _possibleConstructorReturn(this, (SubFoo.__proto__ || Object.getPrototypeOf(SubFoo)).call(this));
    if(!(_this instanceof SubFoo)) throw new Error("_this not an instanceof SubFoo");
    return _this;
  }
  return SubFoo;
}(Object);
```

There are other objects where transpiled code doesn't work. Most of the native objects can only be extended with true ES6 syntax. ` class Foo extends Promise` for example.

I don't mind special casing this. Do you want me to special case it for all native types? Or just `Object`?

ChadKillingsworth:
#1953 highlighted similar problems.

ChadKillingsworth:
Another thought - what if we just add a check and re-assign it?

```js
// The super constructor returned something weird here...
if (!($jscomp$super$this instanceof SubFoo)) {
  $jscomp$super$this = this;
}
```

brad4d:
@ChadKillingsworth I think that could work.
BTW, does this pass run after type checking?
If not, won't rewriting all of the cases of 'this' with '$jscomp$super$this' mess up the member variable type definitions?

ChadKillingsworth:
@brad4d This runs WAY before type checking. The majority of the work in this PR was so make sure OTI could understand the alias properly. NTI wasn't covered.

@MatrixFrog I'll push up changes in the morning for you to try.

brad4d:
@ChadKillingsworth replacing 'this' is going to mess up type checking by hiding the member variable names.

I propose a different way of doing this.

```
var SubFoo = function(a, b, c) {
  // original code was 'super(a);'
  if (!$jscomp.wasConstructorCalled(this, Foo)) {
    return $jscomp.callSuperAndSimulateNew(this, Foo, [a], SubFoo, arguments);
  }
  this.b = b;
  this.c = c;
};
```

Implementation of the two methods called above is
```
$jscomp.wasConstructorCalled = function(obj, constructor) {
  return obj[getWasConstructorCalledSymbol(constructor)];
};

// assume this code is inside a closure
var WAS_CONSTRUCTOR_CALLED = new Symbol();

function getWasConstructorCalledSymbol(constructor) {
  if (!(WAS_CONSTRUCTOR_CALLED in constructor)) {
    constructor[WAS_CONSTRUCTOR_CALLED] = new Symbol();
  }
  return constructor[WAS_CONSTRUCTOR_CALLED];
}

$jscomp.callSuperAndSimulateNew = function(
    obj, superConstructor, superArgs, constructor, constructorArgs) {
  return simulateNew(simulateNew(obj, superConstructor, superArgs), constructor, constructorArgs);
}

function simulateNew(obj, constructor, args) {
  return constructor.apply(obj, args) || obj;
}
```

MatrixFrog:
@brad4d my internal copy of this change already has a few additional changes (new runtime tests mainly) so do you want to patch that and add your proposed change?

ChadKillingsworth:
The type checking wasn't a hard solve for OTI at least: https://github.com/google/closure-compiler/pull/2024/files#diff-1146e576b1cd6029ea1bd7fa9a19b891R4774

But switching to that method would probably work as well. I have no preference.

brad4d:
@MatrixFrog SGTM
Thanks

brad4d:
@MatrixFrog, sorry for the delay.
I somehow misread "do you want to patch" as "do you want **me** to patch" and thought you were offering to do it.
I'll work on this today.

ChadKillingsworth:
@brad4d Any update on this?

brad4d:
I ended up making significant changes to avoid mucking with the type checker.
I've just sent the results for review.

brad4d:
Example rewrite:
Original code
```javascript
class B extends A {
  constructor(a, b) {
    super(a);
    this.b = b;
  }
}
```
After super rewrite step (but before transpilation of the class itself):
```javascript
class B extends A {
  constructor(a, b) {
    // If this is a recursive call,
    // fetch the stashed result of calling the super constructor.
    // Otherwise, call the super constructor and store the new value for `this`.
    var $jscomp$super$this =
        this['$jscomp$super$this'] || A.call(this, a) || this;
    if ($jscomp$super$this != this) {
      // We've just executed the super constructor, and it changed the `this` value.
      $jscomp$super$this['$jscomp$super$this'] = $jscomp$super$this;
      return B.apply($jscomp$super$this, arguments) || $jscomp$super$this;
    }
    // Either super didn't change `this` or we're in the recursive call.
    delete this['$jscomp$super$this']; // clean up
    // all original code unchanged after this point
    this.b = b;
  }
}
```
Note: I actually have to cast to the unknown type before fetching `['$jscomp$super$this']` to keep the compiler happy.

brad4d:
Discussed this further with @concavelenz and @blickly.
We've agreed that the better solution is to keep the simpler method of replacing `this` with a variable, but move the super transpilation pass itself after type checking.

Moving transpilation passes after type checking will solve a lot of problems and is something we want to do in general. The super rewrite pass seems like a good first candidate.
I'm starting a change right now to see what problems doing that exposes.

ChadKillingsworth:
I'm cool with that - as long as it doesn't massively hold up this change.

What I had was working with OTI, and can probably be made to work with NTI with just a little tweaking. I'd rather do that in the short term if moving the super transpilation ends up being complicated.

brad4d:
I'm trying both ways to see which is likely to get me there sooner.

brad4d:
We're approaching this in a different way, so this PR should really be dropped.
See update on #1669

PR #2035, opened 2016-09-26
Indexed Database API updated based on W3C Recommendation 08 January 2015

vobruba-martin:
## [W3C Recommendation 08 January 2015](https://www.w3.org/TR/2015/REC-IndexedDB-20150108/)

1) [IDBDatabaseException](https://developer.mozilla.org/en-US/docs/Web/API/IDBDatabaseException), [IDBRequest](https://developer.mozilla.org/en-US/docs/Web/API/IDBRequest), [IDBCursor](https://developer.mozilla.org/en-US/docs/Web/API/IDBCursor) and [IDBTransaction](https://developer.mozilla.org/en-US/docs/Web/API/IDBTransaction) constants have been removed

2) Added missing nullability operators

3) New enums: `IDBTransactionMode`, `IDBRequestReadyState` and `IDBCursorDirection`

4) New records: `IDBObjectStoreParameters`, `IDBIndexParameters` and `IDBVersionChangeEventInit`

5) Updated objects: `IDBKeyRange`, `IDBVersionChangeEvent`, `IDBDatabase`, `IDBObjectStore`, `IDBIndex` and `IDBTransaction`

6) Added `IDBFactory.prototype.cmp`

**BTW do you still require squashing all the commits?**

### DIFF
--- a/externs/browser/w3c_indexeddb.js
+++ b/externs/browser/w3c_indexeddb.js
@@ -23,6 +23,7 @@
  *
  * @externs
  * @author guido.tapia@picnet.com.au (Guido Tapia)
+ * @author vobruba.martin@gmail.com (Martin Vobruba)
  */
 
 /** @type {!IDBFactory|undefined} */
@@ -39,6 +40,61 @@ Window.prototype.msIndexedDB;
 
 /** @type {!IDBFactory|undefined} */
 Window.prototype.indexedDB;
+
+
+/**
+ * Possible values: 'readonly', 'readwrite', 'versionchange'
+ * 
+ * @typedef {string}
+ * @see https://www.w3.org/TR/IndexedDB/#idl-def-IDBTransactionMode
+ */
+var IDBTransactionMode;
+
+
+/**
+ * Possible values: 'pending', 'done'
+ * 
+ * @typedef {string}
+ * @see https://www.w3.org/TR/IndexedDB/#idl-def-IDBRequestReadyState
+ */
+var IDBRequestReadyState;
+
+
+/**
+ * Possible values: 'next', 'nextunique', 'prev', 'prevunique'
+ * 
+ * @typedef {string}
+ * @see https://www.w3.org/TR/IndexedDB/#idl-def-IDBCursorDirection
+ */
+var IDBCursorDirection;
+
+
+/**
+ * @record
+ * @see https://www.w3.org/TR/IndexedDB/#idl-def-IDBIndexParameters
+ */
+function IDBIndexParameters() {};
+
+/** @type {(undefined|boolean)} */
+IDBIndexParameters.prototype.unique;
+
+/** @type {(undefined|boolean)} */
+IDBIndexParameters.prototype.multiEntry;
+
+
+/**
+ * @record
+ * @extends {EventInit}
+ * @see https://www.w3.org/TR/IndexedDB/#idl-def-IDBVersionChangeEventInit
+ */
+function IDBVersionChangeEventInit() {};
+
+/** @type {(undefined|number)} */
+IDBVersionChangeEventInit.prototype.oldVersion;
+
+/** @type {(undefined|number|null)} */
+IDBVersionChangeEventInit.prototype.newVersion;
+
 
 
 /**
@@ -74,6 +130,14 @@ IDBFactory.prototype.open = function(name, opt_version) {};
 IDBFactory.prototype.deleteDatabase = function(name) {};
 
 /**
+ * @param {*} first
+ * @param {*} second
+ * @return {number}
+ */
+IDBFactory.prototype.cmp = function(first, second) {};
+
+
+/**
  * @constructor
  * @implements {EventTarget}
  * @see http://www.w3.org/TR/IndexedDB/#idl-def-IDBRequest
@@ -91,6 +155,7 @@ IDBRequest.prototype.removeEventListener = function(
 /** @override */
 IDBRequest.prototype.dispatchEvent = function(evt) {};
 
+
 /**
  * @constructor
  * @extends {IDBRequest}
@@ -99,31 +164,7 @@ IDBRequest.prototype.dispatchEvent = function(evt) {};
 function webkitIDBRequest() {}
 
 /**
- * @type {number}
- * @const
- */
-IDBRequest.LOADING;
-
-/**
- * @type {number}
- * @const
- */
-webkitIDBRequest.LOADING;
-
-/**
- * @type {number}
- * @const
- */
-IDBRequest.DONE;
-
-/**
- * @type {number}
- * @const
- */
-webkitIDBRequest.DONE;
-
-/**
- * @type {number}
+ * @type {!IDBRequestReadyState}
  */
 IDBRequest.prototype.readyState; // readonly
 
@@ -150,11 +191,12 @@ IDBRequest.prototype.errorCode;  // readonly
 /** @type {!DOMError} */
 IDBRequest.prototype.error; // readonly
 
-/** @type {Object} */
+/** @type {?Object} */
 IDBRequest.prototype.source; // readonly
 
-/** @type {IDBTransaction} */
+/** @type {!IDBTransaction} */
 IDBRequest.prototype.transaction; // readonly
+
 
 /**
  * @constructor
@@ -173,6 +215,7 @@ IDBOpenDBRequest.prototype.onblocked = function(e) {};
  */
 IDBOpenDBRequest.prototype.onupgradeneeded = function(e) {};
 
+
 /**
  * @constructor
  * @implements {EventTarget}
@@ -187,26 +230,20 @@ function IDBDatabase() {}
 IDBDatabase.prototype.name;
 
 /**
- * @type {string}
- * @const
- */
-IDBDatabase.prototype.description;
-
-/**
- * @type {string}
+ * @type {number}
  * @const
  */
 IDBDatabase.prototype.version;
 
 /**
- * @type {DOMStringList}
+ * @type {!DOMStringList}
  * @const
  */
 IDBDatabase.prototype.objectStoreNames;
 
 /**
  * @param {string} name The name of the object store.
- * @param {?Object|?IDBObjectStoreParameters=} opt_parameters Parameters to be passed
+ * @param {!IDBObjectStoreParameters=} opt_parameters Parameters to be passed
  *     creating the object store.
  * @return {!IDBObjectStore} The created/open object store.
  */
@@ -220,15 +257,8 @@ IDBDatabase.prototype.createObjectStore =
 IDBDatabase.prototype.deleteObjectStore = function(name) {};
 
 /**
- * @param {string} version The new version of the database.
- * @return {!IDBRequest} The IDBRequest object.
- */
-IDBDatabase.prototype.setVersion = function(version) {};
-
-/**
- * @param {string|Array<string>} storeNames The stores to open in this
- *     transaction.
- * @param {(number|string)=} mode The mode for opening the object stores.
+ * @param {(string|!Array<string>)} storeNames The stores to open in this transaction.
+ * @param {!IDBTransactionMode=} mode The mode for opening the object stores.
  * @return {!IDBTransaction} The IDBRequest object.
  */
 IDBDatabase.prototype.transaction = function(storeNames, mode) {};
@@ -240,19 +270,24 @@ IDBDatabase.prototype.transaction = function(storeNames, mode) {};
 IDBDatabase.prototype.close = function() {};
 
 /**
- * @type {Function}
+ * @type {?function(!Event)}
  */
-IDBDatabase.prototype.onabort = function() {};
+IDBDatabase.prototype.onabort;
 
 /**
- * @type {Function}
+ * @type {?function(!Event)}
  */
-IDBDatabase.prototype.onerror = function() {};
+IDBDatabase.prototype.onclose;
 
 /**
- * @type {Function}
+ * @type {?function(!Event)}
  */
-IDBDatabase.prototype.onversionchange = function() {};
+IDBDatabase.prototype.onerror;
+
+/**
+ * @type {?function(!IDBVersionChangeEvent)}
+ */
+IDBDatabase.prototype.onversionchange;
 
 /** @override */
 IDBDatabase.prototype.addEventListener = function(type, listener, opt_options) {
@@ -265,6 +300,7 @@ IDBDatabase.prototype.removeEventListener = function(
 /** @override */
 IDBDatabase.prototype.dispatchEvent = function(evt) {};
 
+
 /**
  * Typedef for valid key types according to the w3 specification. Note that this
  * is slightly wider than what is actually allowed, as all Array elements must
@@ -273,6 +309,7 @@ IDBDatabase.prototype.dispatchEvent = function(evt) {};
  * @typedef {number|string|!Date|!Array<?>}
  */
 var IDBKeyType;
+
 
 /**
  * @constructor
@@ -286,16 +323,16 @@ function IDBObjectStore() {}
 IDBObjectStore.prototype.name;
 
 /**
- * @type {string}
+ * @type {*}
  */
 IDBObjectStore.prototype.keyPath;
 
 /**
- * @type {DOMStringList}
+ * @type {!DOMStringList}
  */
 IDBObjectStore.prototype.indexNames;
 
-/** @type {IDBTransaction} */
+/** @type {!IDBTransaction} */
 IDBObjectStore.prototype.transaction;
 
 /** @type {boolean} */
@@ -303,26 +340,26 @@ IDBObjectStore.prototype.autoIncrement;
 
 /**
  * @param {*} value The value to put into the object store.
- * @param {IDBKeyType=} key The key of this value.
+ * @param {!IDBKeyType=} key The key of this value.
  * @return {!IDBRequest} The IDBRequest object.
  */
 IDBObjectStore.prototype.put = function(value, key) {};
 
 /**
  * @param {*} value The value to add into the object store.
- * @param {IDBKeyType=} key The key of this value.
+ * @param {!IDBKeyType=} key The key of this value.
  * @return {!IDBRequest} The IDBRequest object.
  */
 IDBObjectStore.prototype.add = function(value, key) {};
 
 /**
- * @param {IDBKeyType} key The key of this value.
+ * @param {!IDBKeyType} key The key of this value.
  * @return {!IDBRequest} The IDBRequest object.
  */
 IDBObjectStore.prototype.delete = function(key) {};
 
 /**
- * @param {IDBKeyType|!IDBKeyRange} key The key of the document to retrieve.
+ * @param {!IDBKeyType|!IDBKeyRange} key The key of the document to retrieve.
  * @return {!IDBRequest} The IDBRequest object.
  */
 IDBObjectStore.prototype.get = function(key) {};
@@ -333,8 +370,9 @@ IDBObjectStore.prototype.get = function(key) {};
 IDBObjectStore.prototype.clear = function() {};
 
 /**
- * @param {IDBKeyRange=} range The range of the cursor.
- * @param {string=} direction The direction of cursor enumeration.
+ * @param {?IDBKeyRange=} range The range of the cursor.
+ *     Nullable because IE <11 has problems with undefined.
+ * @param {!IDBCursorDirection=} direction The direction of cursor enumeration.
  * @return {!IDBRequest} The IDBRequest object.
  */
 IDBObjectStore.prototype.openCursor = function(range, direction) {};
@@ -342,7 +380,7 @@ IDBObjectStore.prototype.openCursor = function(range, direction) {};
 /**
  * @param {string} name The name of the index.
  * @param {string|!Array<string>} keyPath The path to the index key.
- * @param {Object=} opt_paramters Optional parameters
+ * @param {!IDBIndexParameters=} opt_paramters Optional parameters
  *     for the created index.
  * @return {!IDBIndex} The IDBIndex object.
  */
@@ -361,11 +399,12 @@ IDBObjectStore.prototype.index = function(name) {};
 IDBObjectStore.prototype.deleteIndex = function(indexName) {};
 
 /**
- * @param {(IDBKeyType|IDBKeyRange)=} key The key of this value.
+ * @param {(!IDBKeyType|IDBKeyRange)=} key The key of this value.
  * @return {!IDBRequest} The IDBRequest object.
  * @see http://www.w3.org/TR/IndexedDB/#widl-IDBObjectStore-count
  */
 IDBObjectStore.prototype.count = function(key) {};
+
 
 /**
  * @constructor
@@ -386,7 +425,7 @@ IDBIndex.prototype.name;
 IDBIndex.prototype.objectStore;
 
 /**
- * @type {string}
+ * @type {*}
  * @const
  */
 IDBIndex.prototype.keyPath;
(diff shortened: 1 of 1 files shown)


vobruba-martin:
Rebased after 2fe56b9066cf2d565911cf9ffe6c798294900bc6. Any thought if this PR gets in?

vobruba-martin:
Rebased after c5ac564523b1fe978484afa52b776ffd52654672

brad4d:
Hi @vobruba-martin.
Sorry for the delay getting to this.

We do still need the branch to be squashed to a single commit.

I see that one file seems to be changed by just the addition of a blank line.
Could you remove that?

vobruba-martin:
Squashed. For a review you can see all the commits here https://github.com/vobruba-martin/closure-compiler/commits/indexeddb-update-old .

vobruba-martin:
@brad4d I've found a better link for the squashed commits: https://github.com/google/closure-compiler/compare/master...vobruba-martin:indexeddb-update-old

brad4d:
PR imported and testing underway.
FYI, the change includes TAB characters which aren't allowed in our externs files, so I removed them.

externs/browser/w3c_indexeddb.js
```diff
@@ -572,35 +471,52 @@ IDBIndex.prototype.keyPath;
...
+/**
+ * @type {boolean}
+ * @const
+ */
 IDBIndex.prototype.unique;
 
 /**
- * @param {IDBKeyRange=} range The range of the cursor.
- * @param {(number|string)=} direction The direction of cursor enumeration.
+ * TODO(vobruba-martin): Change direction param to !IDBCursorDirection once CC understands string literal types.
```
brad4d: Please make your to-dos just "TODO: ..."
Unfortunately non-googler names in TODOs break internal tools.

externs/browser/w3c_indexeddb.js
```diff
@@ -572,35 +471,52 @@ IDBIndex.prototype.keyPath;
...
+ * 
+ * @param {(!IDBKeyType|!IDBKeyRange)=} range The range of the cursor.
+ * @param {(!IDBCursorDirection|string)=} direction The direction of cursor enumeration.
```
brad4d: Sadly it isn't safe to disallow nulls for these arguments.
This definitely breaks some of our existing code that uses openCursor and openKeyCursor

externs/browser/w3c_indexeddb.js
```diff
@@ -406,19 +295,24 @@ IDBDatabase.prototype.transaction = function(storeNames, mode) {};
...
 /**
- * @type {Function}
+ * @type {?function(!Event)}
  */
-IDBDatabase.prototype.onerror = function() {};
+IDBDatabase.prototype.onerror;
 
 /**
- * @type {Function}
+ * @type {?function(!IDBVersionChangeEvent)}
```
brad4d: need to also allow webkitIDBVersionChangeEvent

vobruba-martin:
@brad4d Do you want to allow `webkitIDBVersionChangeEvent` also in [IDBOpenDBRequest](https://github.com/vobruba-martin/closure-compiler/blob/ed98833f51ce9a4e25dfc2c66771f1fed2f7b0a1/externs/browser/w3c_indexeddb.js#L397)?

brad4d:
@vobruba-martin 
It's best to maintain backward compatibility where you can, so yes that would probably be a good idea.

brad4d:
Importing and testing latest update to see what other problems turn up...

vobruba-martin:
@brad4d Ok, I've added this commit https://github.com/google/closure-compiler/commit/112d06f6f19f0148c6de4069e6486e90df4cc262

brad4d:
@vobruba-martin 
Could you clean up your commit description.
My PR import tool is choking on it, I think, and it needs to be more human-readable anyway.

vobruba-martin:
Updated because of 2ba0fc9cdc9a95c290ad2133765b2f71f6f4e4a5

See https://github.com/google/closure-compiler/commit/c3fd08849998384a0492e9241efaeaec8ca5b46a

externs/browser/w3c_indexeddb.js
```diff
@@ -316,12 +379,13 @@ IDBRequest.prototype.errorCode;  // readonly
 /** @type {!DOMError} */
 IDBRequest.prototype.error; // readonly
 
-/** @type {Object} */
+/** @type {?Object} */
 IDBRequest.prototype.source; // readonly
 
-/** @type {IDBTransaction} */
+/** @type {!IDBTransaction} */
```
brad4d: Not backward compatible, but probably not a problem.
I'm willing to take it as long as none of our internal stuff breaks.

externs/browser/w3c_indexeddb.js
```diff
@@ -353,26 +418,20 @@ function IDBDatabase() {}
 IDBDatabase.prototype.name;
 
 /**
- * @type {string}
- * @const
- */
-IDBDatabase.prototype.description;
```
brad4d: Not backward compatible.
Could you just mark this `@deprecated`?

externs/browser/w3c_indexeddb.js
```diff
@@ -353,26 +418,20 @@ function IDBDatabase() {}
...
- * @type {DOMStringList}
+ * @type {!DOMStringList}
  * @const
  */
 IDBDatabase.prototype.objectStoreNames;
 
 /**
  * @param {string} name The name of the object store.
- * @param {Object=} opt_parameters Parameters to be passed
+ * @param {!IDBObjectStoreParameters=} opt_parameters Parameters to be passed
```
brad4d: Not backward compatible.
How about `{?Object|!IDBObjectStoreParameters=}`?

externs/browser/w3c_indexeddb.js
```diff
@@ -386,15 +445,10 @@ IDBDatabase.prototype.createObjectStore =
...
- * @param {string} version The new version of the database.
- * @return {!IDBRequest} The IDBRequest object.
- */
-IDBDatabase.prototype.setVersion = function(version) {};
-
-/**
- * @param {string|Array<string>} storeNames The stores to open in this
- *     transaction.
- * @param {(number|string)=} mode The mode for opening the object stores.
+ * TODO: Change mode param to !IDBTransactionMode once CC understands string literal types.
```
brad4d: Please don't refer to closure-compiler as "CC" or "GCC". Those abbreviations have a long history of referring to the C language compiler and its GNU implementation.

externs/browser/w3c_indexeddb.js
```diff
@@ -386,15 +445,10 @@ IDBDatabase.prototype.createObjectStore =
 IDBDatabase.prototype.deleteObjectStore = function(name) {};
 
 /**
- * @param {string} version The new version of the database.
- * @return {!IDBRequest} The IDBRequest object.
- */
-IDBDatabase.prototype.setVersion = function(version) {};
```
brad4d: `@deprecated`

externs/browser/w3c_indexeddb.js
```diff
@@ -386,15 +445,10 @@ IDBDatabase.prototype.createObjectStore =
...
-
-/**
- * @param {string|Array<string>} storeNames The stores to open in this
- *     transaction.
- * @param {(number|string)=} mode The mode for opening the object stores.
+ * TODO: Change mode param to !IDBTransactionMode once CC understands string literal types.
+ * 
+ * @param {(string|!Array<string>)} storeNames The stores to open in this transaction.
+ * @param {(!IDBTransactionMode|string)=} mode The mode for opening the object stores.
```
brad4d: Why is `number` dropped?

externs/browser/w3c_indexeddb.js
```diff
@@ -406,19 +460,24 @@ IDBDatabase.prototype.transaction = function(storeNames, mode) {};
 IDBDatabase.prototype.close = function() {};
 
 /**
- * @type {Function}
+ * @type {?function(!Event)}
```
brad4d: It'll likely break existing code if we tighten these types.

brad4d:
@concavelenz 
This PR is a good example of the conflict between keeping the compiler happy as API standards change without breaking existing code and doing good type checking for those APIs.

@vobruba-martin 
I've marked a bunch of non-backward-compatible changes in your latest commit.

This file is part of our "standard externs". Their primary goal is to make sure the compiler doesn't choke on uses of standard APIs. We want to make sure that changes to these don't break existing code without a really good reason.

A higher goal is to actually do good type checking of the uses of those APIs, but then we start having problems with backwards compatibility. I've been discussing with @concavelenz and others the possibility of defining optional, strictly-typed, API version-specific externs files for use by those who want better type checking, but we don't have any agreement yet. For now, I think we need to make sure changes to standard externs are always backwards compatible.

ChadKillingsworth:
@brad4d @concavelenz This could easily be handled with the environments flag.

vobruba-martin:
Well, if you want to keep the changes always backward compatible do have any deprecation process for changing property types? And which browsers versions you want to stay compatible for?

I did dome research because of changes you want to keep BW compatible and tested browsers in versions with initial support for IDB without prefixes:

Google Chrome v23
Firefox v16
Internet Explorer v10

```javascript
/** @type {!IDBRequest} */
var request = indexedDB.open('test', 1);
/** @type {!IDBDatabase} */
var database = request.result;

// always string in both cases
// no reason for keeping "number" type
typeof request.readyState; 
typeof database.description;

// function in Chrome, undefined in others
// should be marked as @deprecated
typeof database.setVersion;
```

BTW I would love some strict mode option!!! As well as service worker mode, nodejs mode, etc...

externs/browser/w3c_indexeddb.js
vobruba-martin: `number` here was last specced in [Working Draft 2011](https://www.w3.org/TR/2011/WD-IndexedDB-20111206/) (6 versions before final recomendation). I don't believe we can meet a browser requiring `number` as a second param.

externs/browser/w3c_indexeddb.js
vobruba-martin: Do you have any example of what is broken by this change?

externs/browser/w3c_indexeddb.js
vobruba-martin: `?Object` was last specced in [Working Draft 2011](https://www.w3.org/TR/2011/WD-IndexedDB-20110419/) (7 versions before final recomendation). Again, I don't believe we can meet a browser with this.

vobruba-martin:
Another question: Is it really necessary to stay backward compatible with **Draft versions** of the specification when final recommendation was relased 2 years ago?

ChadKillingsworth:
> Another question: Is it really necessary to stay backward compatible with Draft versions of the specification when final recommendation was relased 2 years ago?

Yes. Changes to the compiler are not permitted to break existing projects at Google. Some of those projects support very old browsers.

brad4d:
Based on visual review there seem to still be some backwards incompatibilities.
I've just done a fresh import and I'm running tests to see if any of them break Google projects.

vobruba-martin:
Here is a list of what is unresolved from my point of view:

1) Changed type of `IDBRequest.prototype.readyState` and the second parameter of `IDBDatabase.prototype.createObjectStore`. These changes are not backward compatible but these changes were made before IDB was finally implemented in browsers.

2) Tighten types of `IDBDatabase.prototype.on*` handlers from `Function` to `?function(!Event)`.

Please let me know if it breaks something.

Thanks!

vobruba-martin:
Forced Travis rebuild

brad4d:
I've started testing the latest update.

FYI, the previous one caused these errors due to lack of backward compatibility.
```
redacted.js:188: ERROR - assignment to property onversionchange of IDBDatabase
found   : function (webkitIDBVersionChangeEvent): undefined
required: (function (IDBVersionChangeEvent): ?|null)
  db.onversionchange = goog.bind(this.handleVersionChange_, this);
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

redacted.js:82: ERROR - actual parameter 2 of IDBIndex.prototype.openCursor does not match formal parameter
found   : (number|string|undefined)
required: (string|undefined)
    request = this.index_.openCursor(opt_range, opt_direction);
                                                ^^^^^^^^^^^^^

redacted.js:112: ERROR - actual parameter 2 of IDBIndex.prototype.openKeyCursor does not match formal parameter
found   : (number|string|undefined)
required: (string|undefined)
    request = this.index_.openKeyCursor(opt_range, opt_direction);
                                                   ^^^^^^^^^^^^^
redacted.js:130: ERROR - actual parameter 1 of IDBObjectStore.prototype.openCursor does not match formal parameter
found   : (IDBKeyRange|null|undefined)
required: (IDBKeyRange|undefined)
    request = this.objectStore_.openCursor(opt_range, opt_direction);
                                           ^^^^^^^^^

redacted.js:130: ERROR - actual parameter 2 of IDBObjectStore.prototype.openCursor does not match formal parameter
found   : (number|string|undefined)
required: (string|undefined)
    request = this.objectStore_.openCursor(opt_range, opt_direction);
                                                      ^^^^^^^^^^^^^

redacted_file_name.js:132: ERROR - actual parameter 1 of IDBObjectStore.prototype.openCursor does not match formal parameter
found   : (IDBKeyRange|null|undefined)
required: (IDBKeyRange|undefined)
    request = this.objectStore_.openCursor(opt_range);
                                           ^^^^^^^^^
```

vobruba-martin:
@brad4d Please let me know after you finish the latest testing. I will fix all errors in one commit.

brad4d:
@vobruba-martin 
Here are the errors I got this time.

From Closure library
https://github.com/google/closure-library/blob/master/closure/goog/db/indexed.js

```
javascript/closure/db/index.js:65: ERROR - inconsistent return type
found   : *
required: string
  return this.index_.keyPath;
         ^^^^^^^^^^^^^^^^^^^
```

https://github.com/google/closure-library/blob/master/closure/goog/db/indexeddb.js

```
javascript/closure/db/indexeddb.js:145: ERROR - inconsistent return type
found   : number
required: string
  return this.db_.version;
         ^^^^^^^^^^^^^^^^

javascript/closure/db/indexeddb.js:175: ERROR - actual parameter 2 of IDBDatabase.prototype.createObjectStore does not match formal parameter
found   : (Object|null|undefined)
required: (IDBObjectStoreParameters|undefined)
        this.db_.createObjectStore(name, opt_params));
                                         ^^^^^^^^^^
```

From Lovefield
https://github.com/google/closure-library/blob/master/closure/goog/db/indexeddb.js
```
lovefield/lib/backstore/indexed_db.js:314: ERROR - actual parameter 1 of IDBObjectStore.prototype.openCursor does not match formal parameter
found   : null
required: (IDBKeyRange|undefined)
        req = tx.objectStore(tableName).openCursor(null, 'prev');
                                                   ^^^^
```

From internal Google code

```
REDACTED/docsdatabase.js:188: ERROR - assignment to property onversionchange of IDBDatabase
found   : function (webkitIDBVersionChangeEvent): undefined
required: (function (IDBVersionChangeEvent): ?|null)
  db.onversionchange = goog.bind(this.handleVersionChange_, this);
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

REDACTED/index.js:82: ERROR - actual parameter 2 of IDBIndex.prototype.openCursor does not match formal parameter
found   : (number|string|undefined)
required: (string|undefined)
    request = this.index_.openCursor(opt_range, opt_direction);
                                                ^^^^^^^^^^^^^

REDACTED/index.js:112: ERROR - actual parameter 2 of IDBIndex.prototype.openKeyCursor does not match formal parameter
found   : (number|string|undefined)
required: (string|undefined)
    request = this.index_.openKeyCursor(opt_range, opt_direction);
                                                   ^^^^^^^^^^^^^

REDACTED/objectstore.js:130: ERROR - actual parameter 1 of IDBObjectStore.prototype.openCursor does not match formal parameter
found   : (IDBKeyRange|null|undefined)
required: (IDBKeyRange|undefined)
    request = this.objectStore_.openCursor(opt_range, opt_direction);
                                           ^^^^^^^^^

REDACTED/objectstore.js:130: ERROR - actual parameter 2 of IDBObjectStore.prototype.openCursor does not match formal parameter
found   : (number|string|undefined)
required: (string|undefined)
    request = this.objectStore_.openCursor(opt_range, opt_direction);
                                                      ^^^^^^^^^^^^^

REDACTED/objectstore.js:132: ERROR - actual parameter 1 of IDBObjectStore.prototype.openCursor does not match formal parameter
found   : (IDBKeyRange|null|undefined)
required: (IDBKeyRange|undefined)
    request = this.objectStore_.openCursor(opt_range);
                                           ^^^^^^^^^
```

I'd like to reiterate that you're welcome to do a separate pull request with a strict version of this extern.
If you do, I'd like to propose we create a hierarchy that looks like this for it.

`contrib/externs/browser/w3c_indexeddb/VERSION_NUMBER.js`

@concavelenz what do you think?

vobruba-martin:
I have all changes recorded here https://github.com/google/closure-compiler/compare/master...vobruba-martin:indexeddb-update-old so no problem ;)

brad4d:
Discussed with @concavelenz and it seems I misunderstood our intended policy.
I'm going to start from where we are and try to fix the projects that are broken by this PR to match the newer externs.

Assuming I do that and get this PR through. You may wish to re-submit your full-blown change, and I'll start on further updates to make submitting it possible.

brad4d:
@vobruba-martin 
Actually, on the first case I tried to fix I realized I'd probably be better off if you just restored the fully-typed version for me to work against. Would you mind doing that?

Here's how I think it'll probably go from there.
1. Repeat until nothing is broken by this PR:
    1. I'll find something that breaks & fix it. I expect that will involve me applying a subset of your updates and simultaneously fixing internal code.
    2. You'll then need to rebase, since I'll have modified your file.
2. Submit the remaining changes

I sincerely apologize for how much of a pain this is.

vobruba-martin:
@brad4d 
This is great news! I look forward to work on this with you. This much less pain then previous fully backward compatible approach was :-) .

So I've reverted to the initial state of this PR and `cherry-pick`ed a few more commits. We can track all the changes [here](https://github.com/google/closure-compiler/compare/master...vobruba-martin:indexeddb-update-new).

Let me know if you need change something. Thanks!

brad4d:
@vobruba-martin 
OK, first problem.

It appears the code in closure library relies on the `IDBDatabaseException` class, which is no more in your version of the externs.

Perhaps we could just leave `IDBDatabaseException` in as deprecated, but I'm wondering if the closure code is really even working as it should, because the spec seems to have changed drastically with regard to exception handling. The current spec appears to prefer using DOM exceptions with custom types, 
https://www.w3.org/TR/IndexedDB/#exceptions

but the closure-library code expects to be able to dig into some existing exception types to pull out numeric error codes it then uses internally.
https://github.com/google/closure-library/blob/master/closure/goog/db/error.js

Fortunately you can actually look at this code yourself, and I'm hoping you understand the API in question well enough to help with fixing it. 

The good news is that outside of closure library no other code in Google seems to care about these exceptions.

brad4d:
Actually, I see an update was made in 2013 that is probably keeping the closure-library code working fine.
It looks like I can just tweak the parameter type here to expect a DOMError to get past this first hurdle.
https://github.com/google/closure-library/blob/master/closure/goog/db/error.js#L357

vobruba-martin:
@brad4d From what I see changing it to `!DOMError|!DOMException` is sufficient.

brad4d:
The submission of the closure library fix described above just went through internally.
Let's see if I can do an internal change to just drop the obsolete exception types now...

brad4d:
@vobruba-martin 
Please rebase to resolve the conflict with my commit removing `IDBDatabaseException` and related code.
https://github.com/google/closure-compiler/commit/ecb07fd743fe15848ee180871f4c12e48851929d

vobruba-martin:
Rebased and I've added one change https://github.com/google/closure-compiler/commit/2ebcd3ab057648ae8d69aae3499edcb2b596a5c4 :)

brad4d:
importing rebase & looking for next fix to make...

brad4d:
I'm changing the `openCursor` and `openKeyCursor` methods to accept `{!IDBKeyRange=}` instead of `{IDBKeyRange=}` parameters, and making the necessary internal code changes to prevent that breaking things.

brad4d:
@vobruba-martin 
Change of plans.
Test failures show that IE 11 and earlier throw an exception when `undefined` is passed to explicitly as the range for `openCursor`, but they're fine when you pass `null`.
So, please just change your code to use `{?IDBKeyRange=}` for range arguments so `null` will be explicitly allowed.

brad4d:
@vobruba-martin both

brad4d:
@vobruba-martin 
I'm working on adding the `IDBCursorDirection` type and updating relevant parameters.
It may hit some code that actually uses the `number` type here.
Do you know what the old numerical values were and how to map them to the correct values?

brad4d:
@vobruba-martin 
I'm afraid you need to drop the definition of `IDBCursorDirection` and instead just document what the valid string values are.

The trouble is that the compiler by design does not output definitions from externs files, and `IDBCursorDirection` doesn't exist in the browser environment. This means the compiler will end up outputting code containing e.g. `IDBCursorDirection.NEXT` that will fail at runtime because the value is undefined.

This is a problem for other APIs also, such as the fetch API.

vobruba-martin:
@brad4d 

`range` arguments were updated.

> Do you know what the old numerical values were and how to map them to the correct values?

```
    const unsigned short NEXT = 0;
    const unsigned short NEXT_NO_DUPLICATE = 1;
    const unsigned short PREV = 2;
    const unsigned short PREV_NO_DUPLICATE = 3;
```
They were last defined in this version https://www.w3.org/TR/2011/WD-IndexedDB-20111206/

> I'm afraid you need to drop the definition of IDBCursorDirection and instead just document what the valid string values are.

I thought that support for these enums will be implemented one time. See this comment https://github.com/google/closure-compiler/pull/1967#issuecomment-242992602

So should I remove `IDBTransactionMode`, `IDBRequestReadyState` and `IDBCursorDirection` enums?

Dominator008:
@vobruba-martin Yeah we probably won't add string literal types to OTI.

Instead of enums, I think you can keep a `@typedef {string}` for those types for now.

vobruba-martin:
Ok, summary of changes: b94c64724a3c977d4b4401163c2523f656d99787 and a6368f06b31d74da3e42a2ad767fc2b5f53265f4

brad4d:
@vobruba-martin 
Thanks for your updates.

I'll take a look.

brad4d:
FYI, I'm submitting an internal change that just changes the cursor direction type to `{string}` and fixes the tiny bit of google code that would be broken by that otherwise.

A rebase will be needed once that change gets pushed to github.

brad4d:
Turns out I could do the rebase myself.
I'm getting a fresh import now.

vobruba-martin:
@brad4d I would like to add changes for IndexedDB 2.0 which will be available in Chrome 58. What do you think? Or is it for another PR?

https://www.chromestatus.com/features/5812621622116352
https://hacks.mozilla.org/2016/10/whats-new-in-indexeddb-2-0/

brad4d:
@vobruba-martin 
Better to do a separate PR for further updates.
Even better would be to do multiple PRs adding one feature change at a time.
This PR is hard because it changes so many things at once.

The last test run produced 5 errors.
I'm working out how to address them.

externs/browser/w3c_indexeddb.js
brad4d: Nevermind. This is OK. It seems to be the way we generally type things the spec says should be `EventHandler`.

brad4d:
2 errors require minor tweaks to internal code.

3 others cause breakages in closure-library and are going to take more work to resolve

### IDBObjectStoreParameters
Requiring an `{?IDBObjectStoreParameters=}` argument for `createObjectStore` breaks the closure library.
See `db/indexeddb.js:175` in the google/closure-library project.
I recommend:
1. Create a separate pull request that only adds the definition of `IDBObjectStoreParameters` and makes the type passed to `createObjectStore` be `{!Object|!IDBObjectStoreParameters=}`. That should be easy to land.
2. Add a pull request for google/closure-library that updates their code to expect and pass this kind of object.
3. final pull request here again to drop `Object` from the `createObjectStore` parameter type.

### `IDBDatabase.prototype.version` changed from `string` to `number`
See `db/indexeddb.js:145` in closure-library
I suggest:
1. Create a PR in google/closure-library to just fix the `getVersion` method to return `Number(this.db_.version)`. Perhaps include a TODO saying `Number` can be removed when the externs are updated.
2. Create a PR here to just change the type of `version` from string to number.

### keyPath fields changed from `string` to `*`
See `db/index.js:65` in google/closure-library
I recommend.
1. Create a PR in google/closure-library to loosen types there.
2. Create a PR here to change the `keyPath` types.

brad4d:
@vobruba-martin 
Looking over my last update, I've realized it might not be clear that I'm asking you to create the PRs I listed. Sorry, about that.

I think we'll make better progress with more transparency to the open source community that way.

vobruba-martin:
Ok, I will do it.

vobruba-martin:
@brad4d PRs created but they don't pass Travis check. Do you know what is wrong?

brad4d:
@vobruba-martin 
sorry I missed your question, but it looks like you fixed the problem.
I'm working on getting the IDBObjectStoreParameters change in now.

vobruba-martin:
@brad4d Changes to google/closure-library were made. Is it necessary to create separate PRs for `createObjectStore`, `IDBDatabase.prototype.version` and `keyPath` here?

brad4d:
@vobruba-martin 
You may not need to do separate PRs.
Github still indicates your PR is good.
I'll try a fresh import and running it through tests to see if all the issues have now been addressed.

brad4d:
Woohoo! All the tests passed.
I hope to get this submitted today.

vobruba-martin:
Great news! Now we can start working on IDB 2.0 :-)

brad4d:
Drat! I spoke too soon.
Somehow a failure hid from my tests, but came out to play after I submitted, forcing a rollback.

I'm trying to fix the offending code so I can roll forward again without changes.

PR #2036, opened 2016-09-26
Fix recommended Maven command

vobruba-martin:
Currently recommended command changed in 87d979732554e7f695b02a648609ac304e037ae6 generates no jar files under the `target` directory


Dominator008:
Thanks!

PR #2041, opened 2016-09-27
Devirtualize duplicate methods

nicks:
The background on this PR is that we have a lot of autogenerated protobufs with the same method over and over again (e.g., `MyProtobuf.prototype.getIndex = function () { return this.getFieldValue(this, 'getIndex'); }`)

Closure Compiler doesn't do a great job optimizing these because it can't figure out how to coalesce the definitions. I'm hoping this will improve the situation.


dimvar:
Thanks Nick, this looks very useful! Can you add a unit test where the multiple definitions also use `this`? Like the proto definition you mention in the description. (And then squash the commits so we can bring the PR in.)

nicks:
updated!

dimvar:
LGTM. I'll bring this in and submit it.

nicks:
wooo thanks!

PR #2053, opened 2016-10-03
CSSStyleDeclaration implements IArrayLike<string>

vobruba-martin:
`CSSStyleDeclaration` is a collection of CSS property names. As an example you can try this on any webpage:

```javascript
window.getComputedStyle(document.body, null)
```

Or see this [MDN's example](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration) where `styleObj` is a `CSSStyleDeclaration` object:

```javascript
var styleObj= document.styleSheets[0].cssRules[0].style;
alert(styleObj.cssText);
for (var i = styleObj.length-1; i >= 0; i--) {
   var nameString = styleObj[i];
   styleObj.removeProperty(nameString);
}
alert(styleObj.cssText);
```

### DIFF
--- a/externs/browser/w3c_css.js
+++ b/externs/browser/w3c_css.js
@@ -416,6 +416,7 @@ function CSSUnknownRule() {}
 /**
  * @constructor
  * @extends {CSSProperties}
+ * @implements {IArrayLike<string>}
  * @see http://www.w3.org/TR/DOM-Level-2-Style/css.html#CSS-CSSStyleDeclaration
  */
 function CSSStyleDeclaration() {}


Dominator008:
Looks like it should be typed as `IObject<(string|number), (boolean|number|string|undefined)>` instead? I tried
```js
window.getComputedStyle(document.body, null);
```
on this particular page and got strings in keys and numbers in values.

vobruba-martin:
My intention was to allow passing this object to functions like `goog.array.every`. How about this?

Dominator008:
Since `IArrayLike<string>` extends `IObject<number, string>`, I don't think that it's compatible with `IObject<(string|number), (boolean|number|string|undefined)>}`. Currently some of our other externs have the same problem :(

But note that maybe `CSSStyleDeclaration` is not array-like. The ES6 version `Array.prototype.every` doesn't exist on its prototype.

vobruba-martin:
You're right. These types are not compatible :-(

Bounded generics wouldn't resolve this problem. `IArrayLike` has to have a `length` property. For example `{0:0}` cannot be an `IArrayLike` type.

And what to do with [PluginArray](https://github.com/google/closure-compiler/blob/master/externs/browser/html5.js#L4135) or [MimeTypeArray](https://github.com/google/closure-compiler/blob/master/externs/browser/html5.js#L4164)?

Dominator008:
I think the most precise parameter type for `goog.array.every` would be `(IObject<K super number, V> & IHasLength)|string`, where `IHasLength` would be a `@record` that specifies the `length` property of type `number`, but that requires both bounded generics and intersection type which are both unsupported at the moment.

So I've decided to go with the other route of restricting the index type to be `number` on these DOM collections and change existing string indexing to the equivalent `namedItem()`, `getNamedItem()`, or in the `CSSStyleDeclaration` case, `item()` methods. This seems cleaner IMO, and is in line with what TypeScript does: https://github.com/Microsoft/TypeScript/blob/2e7c83abc833534181d97c534bfac865cd93b1f0/lib/lib.d.ts#L5811.

So I think typing it as `IArrayLike<string>` makes sense. The only number value I saw was actually the `length` property so that doesn't count.

Dominator008:
Actually, `getPropertyValue()` is not supported on IE8 so we'd have to wait on that :(

vobruba-martin:
Ok, I've changed it back to just `IArrayLike<string>`. I guess dropping IE8 support is a longterm job so this PR can be closed? Anyway thank you!

concavelenz:
When we implemented IArrayLike/IObject we made sure that:

@implements IObject<number|string, T>
@implements IArrayLike<T>

The only question here is what type should be used for the value.  We don't need to revisit how we implemented IArrayLike/IObject

PR #2063, opened 2016-10-05
Make GeolocationPositionOptions a @record

chrismiddleton:
As a @constructor, user code which calls Geolocation methods generates a warning. Since there is no user-callable constructor, this needs to be a @record in order to allow an Object with the given properties to be treated as valid.

Additionally, the three properties are supposed to be optional, so their types now allow undefined. (Should they allow null too?)
From the spec (https://dev.w3.org/geo/api/spec-source.html#position_options_interface):

> In ECMAScript, the enableHighAccuracy, timeout and maximumAge properties are all optional: when creating a PositionOptions object, the developer may specify any of these properties.

### DIFF
--- a/externs/browser/w3c_geolocation.js
+++ b/externs/browser/w3c_geolocation.js
@@ -58,13 +58,20 @@ Geolocation.prototype.clearWatch = function(watchId) {};
  * @see http://www.w3.org/TR/geolocation-API/#coordinates
  */
 function GeolocationCoordinates() {}
-/** @type {number} */ GeolocationCoordinates.prototype.latitude;
-/** @type {number} */ GeolocationCoordinates.prototype.longitude;
-/** @type {number} */ GeolocationCoordinates.prototype.accuracy;
-/** @type {number} */ GeolocationCoordinates.prototype.altitude;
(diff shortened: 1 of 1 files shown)


chrismiddleton:
I signed the CLA.

externs/browser/w3c_geolocation.js
```diff
@@ -78,14 +78,14 @@ GeolocationPosition.prototype.coords;
...
+ * @record
  * @see http://www.w3.org/TR/geolocation-API/#position-options
  */
 function GeolocationPositionOptions() {}
-/** @type {boolean} */
+/** @type {boolean|undefined} */
 GeolocationPositionOptions.prototype.enableHighAccuracy;
-/** @type {number} */ GeolocationPositionOptions.prototype.maximumAge;
-/** @type {number} */ GeolocationPositionOptions.prototype.timeout;
+/** @type {number|undefined} */ GeolocationPositionOptions.prototype.maximumAge;
```
MatrixFrog: nit: Can you add a newline after the jsdoc comment like we have for enableHighAccuracy? Makes it easier to quickly scan through the file visually and see all the properties.

MatrixFrog:
lgtm

Issue #2083, opened 2016-10-14
Type Inference Fails to Find Type Information When assigned after the Var


ChadKillingsworth:
Fails on both OTI and NTI:

```js
var foo;
foo = class {

  bar() {
    console.log(this.baz); // type of this is unknown
  }

  get baz() {
    return 'baz';
  }
}
```

http://closure-compiler-debugger.appspot.com/#input0%3Dvar%2520foo%253B%250Afoo%2520%253D%2520class%2520%257B%250A%250A%2520%2520bar()%2520%257B%250A%2520%2520%2520%2520console.log(this.baz)%253B%250A%2520%2520%257D%250A%250A%2520%2520get%2520baz()%2520%257B%250A%2520%2520%2520%2520return%2520'baz'%253B%250A%2520%2520%257D%250A%257D%26input1%26conformanceConfig%26externs%26refasterjs-template%26includeDefaultExterns%3D1%26CHECK_SYMBOLS%3D1%26CHECK_TYPES%3D1%26LANG_IN_IS_ES6%3D1%26PRESERVE_TYPE_ANNOTATIONS%3D1%26PRETTY_PRINT%3D1%26TRANSPILE%3D1

This type of pattern is occurring during module rewriting. The pattern is specifically required because of hoisting:

```js
module.exports = Foo;
/** @constructor */
function Foo() {}
```

My current work has this rewritten to the pattern listed above.

dimvar:
Can we change module rewriting to not produce this pattern? It's more error prone to expand what we consider a "definition" in the rest of the compiler passes.

ChadKillingsworth:
I'm finding it pretty difficult to avoid:

```js
(function() {
  module.exports = class Foo{};
})()
```

Rewrites to:

```js
goog.provide('module$foo$');
var module$foo$;
(function() {
  module$foo$ = class {};
})()
```

dimvar:
[...]

PR #2085, opened 2016-10-15
Complete Rewrite of the ProcessCommonJsModules Pass

ChadKillingsworth:
The commonjs module rewriting was a very old pass and had many incremental changes. In addition, it suffered from some of the same problems as the old `goog.module` rewriting in that it relied on the early inling pass to handle all of the aliases created.

This is an almost complete rewrite. It had the following goals:

 * Avoids creating aliases for imports and exports thus removing the dependency on the early inlining pass.
 * Only traverses the AST twice.

I've tested it on my medium sized project that uses commonjs modules extensively. With this change, the number of warnings created with NTI is drastically reduced.

There's some other ancillary benefits: namely the PolymerPass requires behaviors to be global symbols so it now works with CommonJS modules.

### DIFF
--- a/src/com/google/javascript/jscomp/ProcessCommonJSModules.java
+++ b/src/com/google/javascript/jscomp/ProcessCommonJSModules.java
@@ -16,9 +16,9 @@
 package com.google.javascript.jscomp;
 
 import com.google.common.base.Preconditions;
-import com.google.common.collect.HashMultiset;
-import com.google.common.collect.Iterables;
-import com.google.common.collect.Multiset;
+import com.google.common.collect.ImmutableCollection;
+import com.google.common.collect.ImmutableCollection.Builder;
+import com.google.common.collect.ImmutableList;
 import com.google.javascript.jscomp.NodeTraversal.AbstractPostOrderCallback;
 import com.google.javascript.jscomp.NodeTraversal.AbstractPreOrderCallback;
 import com.google.javascript.jscomp.deps.ModuleLoader;
@@ -28,7 +28,9 @@ import com.google.javascript.rhino.JSDocInfo;
 import com.google.javascript.rhino.JSDocInfoBuilder;
 import com.google.javascript.rhino.Node;
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 /**
  * Rewrites a CommonJS module http://wiki.commonjs.org/wiki/Modules/1.1.1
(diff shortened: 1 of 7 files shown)


ChadKillingsworth:
This avoids the pattern highlighted in #2083 as much as possible. There are still cases where that pattern can be generated (as highlighted in the issue), but they aren't prevalent in any real world code I've encountered.

src/com/google/javascript/jscomp/ProcessCommonJSModules.java
```diff
@@ -338,371 +287,790 @@ private void visitRequireCall(NodeTraversal t, Node require, Node parent) {
...
+   * If a file is a CommonJS module, rewrite export statements.
+   * Typically exports create an alias - the rewriting tries to
+   * avoid such aliases.
+   */
+  private class RewriteModule extends
+      AbstractPostOrderCallback {
+    private final boolean allowFullRewrite;
+    private final ImmutableCollection<Node> exports;
+    private List<Node> imports =  new ArrayList<>();
+    private List<Node> rewrittenClassExpressions = new ArrayList<>();
```
blickly: Looks like these members can all be final

src/com/google/javascript/jscomp/ProcessCommonJSModules.java
```diff
@@ -338,371 +287,790 @@ private void visitRequireCall(NodeTraversal t, Node require, Node parent) {
...
+            initModule.setJSDocInfo(builder.build());
+          }
+        }
+        initModule.useSourceInfoIfMissingFromForTree(this.script);
+
+        Node refChild = this.script.getFirstChild();
+        while (refChild.getNext() != null & refChild.getNext().isExprResult() &&
+            refChild.getNext().getFirstChild().isCall() &&
+            ("goog.require".equals(refChild.getNext().getFirstChild().getFirstChild().getQualifiedName()) ||
+             "goog.provide".equals(refChild.getNext().getFirstChild().getFirstChild().getQualifiedName()))) {
```
blickly: I think these could be simplified with .matchesQualifiedName

If you don't mind the slightly strange name, there's also getFirstFirstChild.

src/com/google/javascript/jscomp/ProcessCommonJSModules.java
```diff
@@ -78,246 +88,185 @@ public ProcessCommonJSModules(Compiler compiler, boolean reportDependencies) {
 
   @Override
   public void process(Node externs, Node root) {
```
blickly: I was a bit confused at first, just because the root node isn't the usual jsRoot (the parent of all input files). Can we add a Preconditions check that root must be a SCRIPT node?

src/com/google/javascript/jscomp/ProcessCommonJSModules.java
```diff
@@ -78,246 +88,185 @@ public ProcessCommonJSModules(Compiler compiler, boolean reportDependencies) {
...
-        if (commonjsFinder.isFound()) {
-          visitCommonJSIfStatement(n);
-        } else {
-          FindDefineAmdStatements amdFinder = new FindDefineAmdStatements();
-          NodeTraversal.traverseEs6(compiler, condition, amdFinder);
-
-          if (amdFinder.isFound()) {
-            visitAMDIfStatement(n);
+      if (n.isGetProp()) {
+        String qualifiedName = n.getQualifiedName();
```
blickly: I think these might be slightly simpler as n.matchesQualifiedName

ChadKillingsworth:
@blickly done.

ChadKillingsworth:
I think the travis failures may be wrong - the tests all pass locally for me.

blickly:
Yeah, I think I've figured out the breakage and have a fix out. (And yes, it's unrelated to this PR)

PR #2088, opened 2016-10-16
Add command line option to include sources content to the source map

### LINKED ISSUE #1890: Support sourcesContent for sourcemaps
It is often handy to have all of the sources for an application in the source mapping file for ease of manipulation, upload, e.t.c. It would be great if the closure-compiler supported `sourcesContent` in source maps. Apologies if it does already, I couldn't find any docs or source on it.

mk0x9:
Should fix #1890. Not sure if everything is right (works for me at least).

### DIFF
--- a/src/com/google/javascript/jscomp/CommandLineRunner.java
+++ b/src/com/google/javascript/jscomp/CommandLineRunner.java
@@ -232,9 +232,10 @@ public class CommandLineRunner extends
         + "or property_renaming_report")
     private boolean createNameMapFiles = false;
 
-    @Option(name = "--source_map_include_sources_content",
+    @Option(name = "--source_map_include_content",
         handler = BooleanOptionHandler.class,
-        usage = "Includes sources content into source map")
+        usage = "Includes sources content into source map. Greatly increases "
+        + "the size of source maps but offers greater portability")
     private boolean sourceMapIncludeSourcesContent = false;
 
     @Option(name = "--property_renaming_report",
@@ -768,9 +769,9 @@ public class CommandLineRunner extends
                     "output_manifest",
                     "output_module_dependencies",
                     "property_renaming_report",
+                    "source_map_include_content",
                     "source_map_location_mapping",
-                    "variable_renaming_report",
-                    "source_map_include_sources_content"))
+                    "variable_renaming_report"))
             .putAll(
                 "Miscellaneous",
                 ImmutableList.of(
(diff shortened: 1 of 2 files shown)


ChadKillingsworth:
I think we can shorten that flag name: `--source_map_include_content`

mk0x9:
I would like to know, what @shicks would say.

ChadKillingsworth:
As a Google employee, @shicks uses a different command line runner and doesn't have to type that flag name.

src/com/google/javascript/jscomp/CommandLineRunner.java
```diff
@@ -232,6 +232,11 @@
         + "or property_renaming_report")
     private boolean createNameMapFiles = false;
 
+    @Option(name = "--source_map_include_sources_content",
+        handler = BooleanOptionHandler.class,
+        usage = "Includes sources content into source map")
```
ChadKillingsworth: Should also add that this greatly increases the size of the source map but offers greater portability.

ChadKillingsworth:
If you'll change the flag name, we can probably get this merged.

mk0x9:
@ChadKillingsworth, is it ok to rename only command line option? Should I rename variable too to reflect name changing?

ChadKillingsworth:
@mk0x9 It's fine to just rename the option. We have quite a few that aren't exact matches.

shicks:
Sorry, I was on leave last month and just now saw this.

For flags, I'd say the shorter the better, and no worries about renaming all the usages in the Java code.  We have a number of internal-only references to the option that would make landing such a change more difficult.

NeosinneR:
Guys, any progress on merging this PR? Would help debugging a lot.

ChadKillingsworth:
@MatrixFrog I've added commits to the original PR - but I can't squash the commits without losing the original author's attribution. Can you perhaps import the original commits and add the changes from 8923a64 during import? Don't worry about my contribution.

MatrixFrog:
Okay, that should work.

codecrackerx9:
I am a little confused by the resolution of this. Was this command line argument successfully committed to master and targeted for a release? When I look at the master branch, it seems like this argument is still missing.

I think this is a great improvement, I too would love to see it part of the release!

ChadKillingsworth:
The flag was renamed during import: https://github.com/google/closure-compiler/blob/592a877e21c8ff2b3d60444da9f2cf5d833adf45/src/com/google/javascript/jscomp/CommandLineRunner.java#L250-L254

It is part of the 20170218 release.

Issue #1773, opened 2016-05-05
Use with node_modules and commonjs requires


supersteves:
I'm exploring using Closure Compiler with node_modules and have run into a couple of blocking issues:
- Moocher dropping. All CJS modules are being considered moochers because they don't "provide" anything. I need moocher dropping in my codebase to avoid pulling in tons of unused files for compilation.
- node_module path resolution. I'm happy to write my paths explicitly like `require("react-bootstrap/lib/Action.js")` but where the module I'm requiring itself requires e.g. "react" rather than "react/main.js", it's an error. For this I'd like to supply my own loader which implements node_modules rules and package json awareness.

As it stands I think I can only realistically use CJS requires (or ES6 imports) when importing my own code rather than complex pre-existing libraries.

ChadKillingsworth:
> Moocher dropping. All CJS modules are being considered moochers because they don't "provide" anything. I need moocher dropping in my codebase to avoid pulling in tons of unused files for compilation.

Not sure why you see this - this definitely isn't occurring in my code bases. Make sure you use the `--process_common_js_modules` flag.

> As it stands I think I can only realistically use CJS requires (or ES6 imports) when importing my own code rather than complex pre-existing libraries.

This is currently true.

ChadKillingsworth:
This is fixed by #2130 - use the `--module_resolution=NODE` flag.

supersteves:
Great! Thanks Chad.

PR #2094, opened 2016-10-22
Implement the node module resolution algorithm for CommonJS modules

ChadKillingsworth:
Feature referenced by multiple issues and PRs, but especially by #1773.

Adds support for CommonJS module rewriting to handle named (non-path based) module lookups. This implements [Node Module Resolution](https://nodejs.org/api/modules.html#modules_all_together) in the compiler natively.

One note: the compiler does not discover source files but depends on all necessary files to be passed in as source. I have work started for the Java npm version which takes an entry point then discovers and passes all required files.

As it's required by the algorithm, this also adds support for JSON files to be passed to the compiler as source. These files are rewritten to module exports. This rewriting is dependent on the file name ending with ".json". That's the same restrictions node has.

I only enabled this support for CommonJS modules, but it will definitely be wanted for ES6 modules as well.

### DIFF
--- a/src/com/google/javascript/jscomp/Compiler.java
+++ b/src/com/google/javascript/jscomp/Compiler.java
@@ -29,6 +29,8 @@ import com.google.debugging.sourcemap.proto.Mapping.OriginalMapping;
 import com.google.javascript.jscomp.CompilerOptions.DevMode;
 import com.google.javascript.jscomp.ReferenceCollectingCallback.ReferenceCollection;
 import com.google.javascript.jscomp.TypeValidator.TypeMismatch;
+import com.google.javascript.jscomp.deps.DependencyInfo;
+import com.google.javascript.jscomp.deps.JsFileParser;
 import com.google.javascript.jscomp.deps.ModuleLoader;
 import com.google.javascript.jscomp.deps.SortedDependencies.MissingProvideException;
 import com.google.javascript.jscomp.parsing.Config;
@@ -1466,6 +1468,10 @@ public class Compiler extends AbstractCompiler implements ErrorHandler, SourceFi
 
         this.moduleLoader = new ModuleLoader(this, options.moduleRoots, inputs);
 
+        if (options.processCommonJSModules) {
+          this.moduleLoader.setPackageJsonMainEntries(processJsonInputs(inputs));
+        }
+
         if (options.lowerFromEs6()) {
           processEs6Modules();
         }
@@ -1675,6 +1681,34 @@ public class Compiler extends AbstractCompiler implements ErrorHandler, SourceFi
   private void repartitionInputs() {
     fillEmptyModules(modules);
     rebuildInputsFromModules();
+  }
+
+  /**
+   * Transforms JSON files to a module export that closure compiler can
+   * process and keeps track of any "main" entries in package.json files.
+   */
+  Map<String, String> processJsonInputs(List<CompilerInput> inputsToProcess) {
+    RewriteJsonToModule rewriteJson = new RewriteJsonToModule(this);
+    for (CompilerInput input : inputsToProcess) {
+      if (!input.getSourceFile().getOriginalPath().endsWith(".json")) {
+        continue;
+      }
+
+      input.setCompiler(this);
+
+      try {
+        // JSON objects need wrapped in parens to parse properly
+        input.getSourceFile().setCode("(" + input.getSourceFile().getCode() + ")");
+      } catch (IOException e) {
+        this.getErrorManager().report(CheckLevel.ERROR,
+            JSError.make(AbstractCompiler.READ_ERROR, input.getSourceFile().getOriginalPath()));
+        continue;
+      }
+
+      Node root = input.getAstRoot(this);
+      rewriteJson.process(null, root);
+    }
+    return rewriteJson.getPackageJsonMainEntries();
   }
 
   void processEs6Modules() {
--- a/src/com/google/javascript/jscomp/RewriteJsonToModule.java
+++ b/src/com/google/javascript/jscomp/RewriteJsonToModule.java
@@ -1,0 +1,150 @@
+/*
+ * Copyright 2016 The Closure Compiler Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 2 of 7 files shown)


ChadKillingsworth:
Example usage: https://github.com/ChadKillingsworth/closure-compiler-npm/blob/1c97ac06c428e13a5384256101a3d66feae0ee3e/test/gulp.js#L296-L315

ChadKillingsworth:
cc @samthor This will require changes in the JS version package as well:

 * https://github.com/ChadKillingsworth/closure-compiler-npm/blob/a163db18684cf88e93d81ce129438a2b7ddd1872/lib/gulp/index.js#L74
 * https://github.com/ChadKillingsworth/closure-compiler-npm/blob/a163db18684cf88e93d81ce129438a2b7ddd1872/lib/gulp/deps-to-vinyl.js#L68-L101

brad4d:
@ChadKillingsworth, can you use `goog.module` instead of `goog.provide` when rewriting JSON files as modules?

ChadKillingsworth:
Wouldn't it be rewritten twice that way?

brad4d:
@MatrixFrog tells me it doesn't really matter whether `goog.module` or `goog.provide` is used here. Continuing review...

brad4d:
This change is causing failures for some internal tests.
I don't yet know why.

brad4d:
Looks like a red herring. The tests were already broken.

src/com/google/javascript/jscomp/Compiler.java
```diff
@@ -1677,6 +1683,34 @@ private void repartitionInputs() {
...
+    for (CompilerInput input : inputsToProcess) {
+      if (!input.getSourceFile().getOriginalPath().endsWith(".json")) {
+        continue;
+      }
+
+      input.setCompiler(this);
+      try {
+        // JSON objects need wrapped in parens to parse properly
+        input.getSourceFile().setCode("(" + input.getSourceFile().getCode() + ")");
+      } catch (Exception e) {
```
brad4d: What exceptions do we expect here exactly?
Why is it OK to just ignore them?

src/com/google/javascript/jscomp/Compiler.java
```diff
@@ -1677,6 +1683,34 @@ private void repartitionInputs() {
...
+        continue;
+      }
+
+      Node root = input.getAstRoot(this);
+      if (root == null) {
+        continue;
```
brad4d: Are there valid cases where `root == null`? What are they?

src/com/google/javascript/jscomp/RewriteJsonToModule.java
```diff
@@ -0,0 +1,150 @@
...
+    n.removeFirstChild();
+
+    String moduleName = t.getInput().getPath().toModuleName();
+
+    n.addChildToFront(IR.var(IR.name(moduleName).useSourceInfoFrom(jsonObject),
+        jsonObject).useSourceInfoFrom(jsonObject));
+
+    n.addChildToFront(
+        IR.exprResult(
+            IR.call(IR.getprop(IR.name("goog"), IR.string("provide")), IR.string(moduleName)))
```
brad4d: This is for use with Node.js, right?
Shouldn't we be rewriting this to look like a non-JSON Node.js module then, rather than a goog.provide()?
Perhaps Node.js modules get rewritten to look like goog.provide() anyway?
Even if that's so, isn't it safer to rewrite this as a Node.js module & allow the later pass to rewrite it again as a goog.provide(). Otherwise, we're duplicating the logic here & risking inconsistent behavior.

src/com/google/javascript/jscomp/deps/ModuleLoader.java
```diff
@@ -125,21 +149,89 @@ public String toModuleName() {
...
+    }
+
+    private String resolveCommonJsModuleFromRegistry(String requireName) {
+      for (Map.Entry<String, ImmutableSet<String>> nodeModulesFolder : nodeModulesRegistry.entrySet()) {
+        if (!this.path.startsWith(nodeModulesFolder.getKey())) {
+          continue;
+        }
+
+        // Load as a file
+        String fullModulePath = nodeModulesFolder.getKey() + "node_modules/" + requireName;
```
brad4d: Shouldn't MODULE_SLASH be used here?

src/com/google/javascript/jscomp/deps/ModuleLoader.java
```diff
@@ -243,6 +335,67 @@ private static String normalize(String path, Iterable<String> moduleRootPaths) {
...
+    });
+
+    // For each modulePath, find all the node_modules folders
+    // There might be more than one:
+    //    /foo/node_modules/bar/node_modules/baz/foo_bar_baz.js
+    // Should add:
+    //   /foo/ -> bar/node_modules/baz/foo_bar_baz.js
+    //   /foo/node_modules/bar/ -> baz/foo_bar_baz.js
+    for (String modulePath : modulePaths) {
+      String[] nodeModulesDirs = modulePath.split("/node_modules/");
```
brad4d: Use MODULE_SLASH here & several other places below?

src/com/google/javascript/jscomp/RewriteJsonToModule.java
brad4d: Discussed this yesterday & agreed to leave it as-is.

src/com/google/javascript/jscomp/Compiler.java
ChadKillingsworth: I added the same Error as handling as other parse points use.

src/com/google/javascript/jscomp/Compiler.java
ChadKillingsworth: My thought was for invalid JSON. Since we wrap the JSON in parens, I think it will always have at least a root node.

ChadKillingsworth:
@brad4d Review items should now be addressed.

brad4d:
@ChadKillingsworth What's the resolution for the literal slashes in strings? I don't see a response to those questions.

brad4d:
@ChadKillingsworth nevermind about the '/' question. I see now.
The registry always uses '/' & you only need to use `MODULE_SLASH` for actual file paths.

MatrixFrog:
If we used the wrong slash somewhere I'm sure one of our valiant Windows users will let us know quite soon, so I'm not too worried about it :) (Does Travis run tests on Windows machines? Or can we make it do so if it doesn't already?)

ChadKillingsworth:
Travis doesn't run tests on windows.:( I think for now we're safe with just the "/" character. If someone complains - I'll be happy to fix it.

brad4d:
FYI, internal review appears to be complete.
I'm just waiting on test runs to complete successfully before submitting.

brad4d:
@MatrixFrog should I be concerned about the Travis failure on this PR?

ChadKillingsworth:
I can fix travis magically.

brad4d:
One of the suggested code improvements from our internal tools actually broke the code. :(
I had to undo the change and restart the global presubmit this morning.

brad4d:
Global presubmit looks good, so I'm submitting the change.

Issue #2113, opened 2016-10-28
Only use a default if output wrapper is null


sirajrauff:
Currently using

``` xml
<jscomp
   compilationLevel="simple"
   debug="false"
   output="js/test.min.js"
   forceRecompile="true"
   sourceMapFormat="V3"
   sourceMapOutputFile="js/test.js.map">
   <sources dir="js">
     <file name="test.js"/>
   </sources>
</jscomp>
```

results in a minified JS file with `//# sourceMappingURL=test.js.map` appended. Once "outputWrapper" option is used (`outputWrapper="%output%${line.separator}//# mapping_file_dir/test.map.js"`), this becomes

``` javascript
//# mapping_file_dir/test.map.js
//# sourceMappingURL=test.js.map
```

This breaks the functionality that one expects when using the jar from the command line (where there is no default at all), and does not allow the user to specify a specific sourceMappingURL.

This can be an issue in cases where the mapping file is not in the same directory, or the js file is loaded dynamically, for example using jQuery's `$.getScript` where relative sourceMappingURLs will not work.

PR #2114, opened 2016-10-28
Only append default sourceMappingURL if no output wrapper is specified

sirajrauff:
See #2113.

### DIFF
--- a/src/com/google/javascript/jscomp/ant/CompileTask.java
+++ b/src/com/google/javascript/jscomp/ant/CompileTask.java
@@ -414,8 +414,6 @@ public final class CompileTask
 
         if (result.sourceMap != null) {
           flushSourceMap(result.sourceMap);
-          source.append(System.getProperty("line.separator"));
-          source.append("//# sourceMappingURL=" + sourceMapOutputFile.getName());
         }
         writeResult(source.toString());
       } else {


sirajrauff:
Note that this change does not eliminate the default sourceMapURL from being specified, which is still divergent from the behaviour when run from the command line. Without information on why this was the original behaviour, I opted to simply change it so that the default was in in lieu of a given output wrapper argument.

ChadKillingsworth:
> Without information on why this was the original behaviour, I opted to simply change it so that the default was in in lieu of a given output wrapper argument.

I'm not sure anyone even knows that now. The ant plugin is completely community maintained.

src/com/google/javascript/jscomp/ant/CompileTask.java
```diff
@@ -410,12 +410,13 @@ public void execute() {
             throw new BuildException("Invalid output_wrapper specified. " +
                 "Missing '" + CommandLineRunner.OUTPUT_MARKER + "'.");
           }
+        } else {
```
ChadKillingsworth: Do you need to check here that `result.sourceMap != null`?

sirajrauff:
With your previous comment in mind, I removed the default output wrapper. This reduces complexity of having to check for the sourceMap or a user-given output wrapper before inserting a default, and brings this in line with the command-line behaviour.

ChadKillingsworth:
Toggling for Travis

brad4d:
Sorry about the delay on this.
@sirajrauff if you'll squash your branch down to 1 commit, I'll import this pull request.

sirajrauff:
@brad4d, I've rebased of current master and squashed, let me know if there's anything else.

brad4d:
@sirajrauff thanks
I've imported this and I'm confirming it doesn't break anything for us.

brad4d:
No problems found by standard tests or test for example projects.
Sent for internal code review.

PR #2117, opened 2016-10-30
Fix Math.hypot polyfill to call with zero/one params

teppeis:
[A compat-table test](https://kangax.github.io/compat-table/es6/#test-Math_methods_Math.hypot) fails due to this issue.
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/hypot

### DIFF
--- a/externs/es6.js
+++ b/externs/es6.js
@@ -152,7 +152,7 @@ Math.cbrt = function(value) {};
  * @nosideeffects
  * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/hypot
  */
-Math.hypot = function(value1, var_args) {};
+Math.hypot = function(var_args) {};
 
 /**
  * @param {number} value1
--- a/externs/es6.js
+++ b/externs/es6.js
@@ -147,7 +147,6 @@ Math.sign = function(value) {};
 Math.cbrt = function(value) {};
 
 /**
- * @param {number} value1
  * @param {...number} var_args
  * @return {number}
  * @nosideeffects
--- a/src/com/google/javascript/jscomp/js/es6/math/hypot.js
+++ b/src/com/google/javascript/jscomp/js/es6/math/hypot.js
@@ -24,15 +24,19 @@ $jscomp.polyfill('Math.hypot', function(orig) {
    *
    * <p>Polyfills the static function Math.hypot().
    *
-   * @param {number} x Any number, or value that can be coerced to a number.
-   * @param {number} y Any number, or value that can be coerced to a number.
-   * @param {...*} var_args More numbers.
+   * @param {...number} var_args Any number, or value that can be coerced to a number.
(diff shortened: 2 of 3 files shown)


MatrixFrog:
lgtm

MatrixFrog:
Internal review was more involved than I expected. Going to back-burner this for now.

concavelenz:
As I mentioned in the internal review, I don't think we want to change the type signature but it should be compatible with the actual implementation and I suggest we use V8's.  Which looks like:

```js
function MathHypot(x, y) { // Function length is 2.
    var length = arguments.length;
    var args = new Array(length);
    var max = 0;
    for (var i = 0; i < length; i++) {
      var n = arguments[i];
      if (typeof n !== "number") n = Number(n);
      if (n === Infinity || n === -Infinity) return Infinity;
      n = Math.abs(n);
      if (n > max) max = n;
      args[i] = n;
    }

    // Kahan summation to avoid rounding errors.
    // Normalize the numbers to the largest one to avoid overflow.
    if (max === 0) max = 1;
    var sum = 0;
    var compensation = 0;
    for (var i = 0; i < length; i++) {
      var n = args[i] / max;
      var summand = n * n - compensation;
      var preliminary = sum + summand;
      compensation = (preliminary - sum) - summand;
      sum = preliminary;
    }
    return Math.sqrt(sum) * max;
  };
```

concavelenz:
I changed my mind, this function isn't used enough for it to valuable to have a tighter signature.

brad4d:
@MatrixFrog are you proceeding with this?
@concavelenz is there any reason not to proceed. I think you cancelled out your previous objection.

teppeis:
I do not know if I should fix this PR ...
I will fix If you need.

MatrixFrog:
Probably not.

teppeis:
@MatrixFrog @brad4d @concavelenz any updates?

brad4d:
Created internal Google issue b/121320889

@concavelenz could you determine whether we should accept or reject this PR?

externs/es6.js
```diff
@@ -147,7 +147,6 @@ Math.sign = function(value) {};
 Math.cbrt = function(value) {};
 
 /**
- * @param {number} value1
  * @param {...number} var_args
```
brad4d: You need to delete the `value1` argument from the parameter list, too.

brad4d:
@concavelenz has confirmed to me that he is OK with accepting this PR.
There's one mistake I see in it for which I've added a comment.

externs/es6.js
teppeis: @brad4d Nice catch! I've fixed.

brad4d:
I've imported this CL and sent it for internal testing and review.

brad4d:
I submitted the internal version of this change a few minutes ago.
Our next daily push to GitHub should include it and close this PR.

teppeis:
@brad4d thanks!

PR #2121, opened 2016-11-01
Aggressively inline IIFEs even when assumeClosuresOnlyCaptureReferences is false

ChadKillingsworth:
Take 2 at fixing #2100. Follow up to discussion at #2103.

Tweak the function inlining heuristic such that it is much more aggressive about inlining IIFEs.

### DIFF
--- a/src/com/google/javascript/jscomp/InlineFunctions.java
+++ b/src/com/google/javascript/jscomp/InlineFunctions.java
@@ -325,10 +325,11 @@ class InlineFunctions implements CompilerPass {
           fs.setHasInnerFunctions(true);
           // If there are inner functions, we can inline into global scope
           // if there are no local vars or named functions.
+          // IIFEs can be aggressively inlined.
           // TODO(johnlenz): this can be improved by looking at the possible
           // values for locals.  If there are simple values, or constants
           // we could still inline.
-          if (!assumeMinimumCapture && hasLocalNames(fnNode)) {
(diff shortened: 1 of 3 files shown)


Dominator008:
I think we tried something similar but it was rolled-back: https://github.com/ChadKillingsworth/closure-compiler/commit/0660acdb4691cade499563c663a2c83243d0c4b2.

ChadKillingsworth:
Hmm your right. I had thought 0660acd had done it somewhat differently, but apparently not. I'll keep checking.

ChadKillingsworth:
Can someone lookup the failure cases that led to the rollback?

dimvar:
> Can someone lookup the failure cases that led to the rollback?

Unfortunately, I didn't record what the exact failure was, just the generic rollback message that it was an IIFE with local variables.

ChadKillingsworth:
I'm closing this. I looked extensively through the InlineFunctions pass. While the changes from the new ConvertIIFEArgsToVars pass could be added into InlineFunctions, it would simply serve to make a complicated process more so. There isn't any real benefit from combining the two.

It may prove that ConvertIIFEArgsToVars is unsafe for exactly the same reasons that the aggressive inlining of IIFEs, but it's worth a try.

Ideally, ConvertIIFEArgsToVars would be a peephole optimization, but it needs to run only at the beginning of the optimizations loop and Peephole optimizations run much more frequently..

PR #2126, opened 2016-11-04
Remove IIFE wrappers from CommonJS modules that have UMD patterns

ChadKillingsworth:
In these cases, the pattern from #2083 was produced by a rewrite. I encountered this in real world code: https://github.com/krisk/Fuse/blob/master/src/fuse.js

This simply removes the unnecessary IIFE wrapper.


ChadKillingsworth:
I'm reworking this.

PR #2130, opened 2016-11-05
Support the node module resolution algorithm for ES6 modules.

### LINKED ISSUE #1897: ES6 imports of paths with '.js' extension does not work
The compiler seems to want to add the extension.  I'm not super-clear on the spec, but my understanding is that an extension is allowed and if it's present, none will be added.

ChadKillingsworth:
Adds a `--module_resolution` flag to allow users to opt-out of loading from node_modules registries. However the path-based algorithm still follows the node module loading spec with regard to file extension priority and utilizing package.json file main enteries for directories.

Adds JSON file support if any module processing is enabled.

Note that the JSParser attempts to resolve ES6 module paths and issue loading errors. This seems strange because similar functionality exists in the ProcessEs6Modules pass (which seems more appropriate). I'm not sure why this logic exists in the parser, but I didn't want to move that until I understood the implications.

Fixes #1897 - any file extension is now allowed for either CommonJS or ES6 modules

### DIFF
--- a/src/com/google/javascript/jscomp/CommandLineRunner.java
+++ b/src/com/google/javascript/jscomp/CommandLineRunner.java
@@ -704,10 +704,23 @@ public class CommandLineRunner extends
         usage = "Rewrite ES6 library calls to use polyfills provided by the compiler's runtime.")
     private boolean rewritePolyfills = true;
 
-    @Option(name = "--print_source_after_each_pass",
-        hidden = true,
-        usage = "Whether to iteratively print resulting JS source per pass.")
-        private boolean printSourceAfterEachPass = false;
+    @Option(
+      name = "--print_source_after_each_pass",
+      hidden = true,
+      usage = "Whether to iteratively print resulting JS source per pass."
+    )
+    private boolean printSourceAfterEachPass = false;
+
+    @Option(
+      name = "--module_resolution",
+      hidden = false,
+      usage =
+          "Specifies how the compiler locates modules. BROWSER requires all module imports "
+              + "to begin with a '.' or '/' and have a file extension. NODE uses the node module "
+              + "rules. LEGACY prepends a '/' to any import not already beginning with a "
+              + "'.' or '/'."
+    )
+    private ModuleLoader.ResolutionMode moduleResolutionMode = ModuleLoader.ResolutionMode.NODE;
 
     @Argument
     private List<String> arguments = new ArrayList<>();
@@ -781,7 +794,10 @@ public class CommandLineRunner extends
             .putAll(
                 "JS Modules",
                 ImmutableList.of(
-                    "js_module_root", "process_common_js_modules", "transform_amd_modules"))
+                    "js_module_root",
+                    "module_resolution",
+                    "process_common_js_modules",
+                    "transform_amd_modules"))
             .putAll(
                 "Library and Framework Specific",
                 ImmutableList.of(
@@ -1507,7 +1523,8 @@ public class CommandLineRunner extends
           .setTracerMode(flags.tracerMode)
           .setInstrumentationTemplateFile(flags.instrumentationFile)
           .setNewTypeInference(flags.useNewTypeInference)
-          .setJsonStreamMode(flags.jsonStreamMode);
+          .setJsonStreamMode(flags.jsonStreamMode)
+          .setModuleResolutionMode(flags.moduleResolutionMode);
     }
     errorStream = null;
   }
@@ -1682,6 +1699,7 @@ public class CommandLineRunner extends
     options.setPrintSourceAfterEachPass(flags.printSourceAfterEachPass);
     options.setStrictModeInput(flags.strictModeInput);
     options.setEmitUseStrict(flags.emitUseStrict);
+    options.setModuleResolutionMode(flags.moduleResolutionMode);
 
     return options;
   }
--- a/src/com/google/javascript/jscomp/deps/ModuleLoader.java
+++ b/src/com/google/javascript/jscomp/deps/ModuleLoader.java
@@ -55,9 +55,8 @@ public final class ModuleLoader {
   /** The default module root, the current directory. */
   public static final String DEFAULT_FILENAME_PREFIX = "." + MODULE_SLASH;
 
-  public static final DiagnosticType LOAD_WARNING = DiagnosticType.warning(
-      "JSC_ES6_MODULE_LOAD_WARNING",
-      "Failed to load module \"{0}\"");
+  public static final DiagnosticType LOAD_WARNING =
+      DiagnosticType.warning("JSC_JS_MODULE_LOAD_WARNING", "Failed to load module \"{0}\"");
 
   @Nullable private final ErrorHandler errorHandler;
 
@@ -75,6 +74,8 @@ public final class ModuleLoader {
   /** Used to canonicalize paths before resolution. */
   private final PathResolver pathResolver;
 
+  private final ResolutionMode resolutionMode;
+
   /**
    * Creates an instance of the module loader which can be used to locate ES6 and CommonJS modules.
    *
@@ -84,7 +85,8 @@ public final class ModuleLoader {
       @Nullable ErrorHandler errorHandler,
       Iterable<String> moduleRoots,
       Iterable<? extends DependencyInfo> inputs,
-      PathResolver pathResolver) {
+      PathResolver pathResolver,
+      ResolutionMode resolutionMode) {
     checkNotNull(moduleRoots);
     checkNotNull(inputs);
     checkNotNull(pathResolver);
@@ -99,11 +101,31 @@ public final class ModuleLoader {
     this.packageJsonMainEntries = ImmutableMap.of();
 
     this.nodeModulesRegistry = buildRegistry(this.modulePaths);
+
+    this.resolutionMode = resolutionMode;
   }
 
-  public ModuleLoader(@Nullable ErrorHandler errorHandler,
-      Iterable<String> moduleRoots, Iterable<? extends DependencyInfo> inputs) {
-    this(errorHandler, moduleRoots, inputs, PathResolver.RELATIVE);
+  public ModuleLoader(
+      @Nullable ErrorHandler errorHandler,
+      Iterable<String> moduleRoots,
+      Iterable<? extends DependencyInfo> inputs,
+      ResolutionMode resolutionMode) {
+    this(errorHandler, moduleRoots, inputs, PathResolver.RELATIVE, resolutionMode);
+  }
+
+  public ModuleLoader(
+      @Nullable ErrorHandler errorHandler,
+      Iterable<String> moduleRoots,
+      Iterable<? extends DependencyInfo> inputs,
+      PathResolver pathResolver) {
+    this(errorHandler, moduleRoots, inputs, pathResolver, ResolutionMode.LEGACY);
+  }
+
+  public ModuleLoader(
+      @Nullable ErrorHandler errorHandler,
+      Iterable<String> moduleRoots,
+      Iterable<? extends DependencyInfo> inputs) {
+    this(errorHandler, moduleRoots, inputs, PathResolver.RELATIVE, ResolutionMode.LEGACY);
   }
 
   public Map<String, String> getPackageJsonMainEntries() {
@@ -120,6 +142,7 @@ public final class ModuleLoader {
    */
   public class ModulePath {
     private final String path;
+    private final String[] fileExtensionsToSearch = {"", ".js", ".json"};
 
     private ModulePath(String path) {
       this.path = path;
@@ -149,32 +172,97 @@ public final class ModuleLoader {
     }
 
     /**
-     * Find a CommonJS module {@code requireName}. See
+     * Find a JS module {@code requireName}. See
      * https://nodejs.org/api/modules.html#modules_all_together
      *
      * @return The normalized module URI, or {@code null} if not found.
      */
-    public ModulePath resolveCommonJsModule(String requireName) {
-      String loadAddress;
+    @Nullable
+    public ModulePath resolveJsModule(String moduleAddress) {
+      return resolveJsModule(moduleAddress, null, -1, -1);
+    }
+
+    /**
+     * Find a JS module {@code requireName}. See
+     * https://nodejs.org/api/modules.html#modules_all_together
+     *
+     * @return The normalized module URI, or {@code null} if not found.
+     */
+    @Nullable
+    public ModulePath resolveJsModule(
+        String moduleAddress, String sourcename, int lineno, int colno) {
+      String loadAddress = null;
 
       // * the immediate name require'd
-      if (isAbsoluteIdentifier(requireName) || isRelativeIdentifier(requireName)) {
-        loadAddress = resolveCommonJsModuleFileOrDirectory(requireName);
-      } else {
-        loadAddress = resolveCommonJsModuleFromRegistry(requireName);
+      switch (resolutionMode) {
+        case LEGACY:
+          loadAddress = resolveJsModuleLegacy(moduleAddress, sourcename, lineno, colno);
+          break;
+
+        case BROWSER:
+          if (isAbsoluteIdentifier(moduleAddress) || isRelativeIdentifier(moduleAddress)) {
+            loadAddress = resolveJsModuleFile(moduleAddress);
+          }
+          break;
+
+        case NODE:
+          if (isAbsoluteIdentifier(moduleAddress) || isRelativeIdentifier(moduleAddress)) {
+            loadAddress = resolveJsModuleFileOrDirectory(moduleAddress);
+          } else {
+            loadAddress = resolveJsModuleFromRegistry(moduleAddress);
+          }
+          break;
       }
+
       if (loadAddress != null) {
         return new ModulePath(loadAddress);
       }
+
+      if (errorHandler != null) {
+        errorHandler.report(
+            CheckLevel.WARNING,
+            JSError.make(sourcename, lineno, colno, LOAD_WARNING, moduleAddress));
+      }
+
       return null;
     }
 
-    private String resolveCommonJsModuleFile(String requireName) {
-      String[] extensions = {"", ".js", ".json"};
+    /**
+     * Find a module using the old LEGACY method.
+     * @return The normalized module path.
+     */
+    private String resolveJsModuleLegacy(String moduleName, String sourcename, int lineno, int colno) {
+      // Allow module names with or without the ".js" extension.
+      if (!moduleName.endsWith(".js")) {
+        moduleName += ".js";
+      }
 
-      // Load as a file
-      for (int i = 0; i < extensions.length; i++) {
-        String loadAddress = locate(requireName + extensions[i]);
+      String path = ModuleNames.escapePath(moduleName);
+      if (isRelativeIdentifier(moduleName)) {
+        String ourPath = this.path;
+        int lastIndex = ourPath.lastIndexOf('/');
+        path = ModuleNames.canonicalizePath(ourPath.substring(0, lastIndex + 1) + path);
+      }
+
+      String resolved = normalize(path, moduleRootPaths);
+      if (!modulePaths.contains(resolved) && errorHandler != null) {
+        errorHandler.report(CheckLevel.WARNING, JSError.make(sourcename, lineno, colno, LOAD_WARNING, moduleName));
+      }
+      return resolved;
+    }
+
+    @Nullable
+    private String resolveJsModuleFile(String moduleAddress) {
+      // Edge is the only browser supporting modules currently and requires
+      // a file extension. This may be loosened as more browsers support
+      // ES2015 modules natively.
+      if (resolutionMode == ResolutionMode.BROWSER) {
+        return locate(moduleAddress);
+      }
+
+      // Load node module as a file
+      for (int i = 0; i < fileExtensionsToSearch.length; i++) {
+        String loadAddress = locate(moduleAddress + fileExtensionsToSearch[i]);
         if (loadAddress != null) {
           return loadAddress;
         }
@@ -183,26 +271,28 @@ public final class ModuleLoader {
       return null;
     }
 
-    private String resolveCommonJsModuleFileOrDirectory(String requireName) {
-      String loadAddress = resolveCommonJsModuleFile(requireName);
+    @Nullable
+    private String resolveJsModuleFileOrDirectory(String moduleAddress) {
+      String loadAddress = resolveJsModuleFile(moduleAddress);
       if (loadAddress == null) {
-        loadAddress = resolveCommonJsModuleDirectory(requireName);
+        loadAddress = resolveJsModuleDirectory(moduleAddress);
       }
       return loadAddress;
     }
 
-    private String resolveCommonJsModuleDirectory(String requireName) {
+    @Nullable
+    private String resolveJsModuleDirectory(String moduleAddress) {
       String[] extensions = {
         MODULE_SLASH + "package.json", MODULE_SLASH + "index.js", MODULE_SLASH + "index.json"
       };
 
       // Load as a file
       for (int i = 0; i < extensions.length; i++) {
-        String loadAddress = locate(requireName + extensions[i]);
+        String loadAddress = locate(moduleAddress + extensions[i]);
         if (loadAddress != null) {
           if (i == 0) {
             if (packageJsonMainEntries.containsKey(loadAddress)) {
-              return resolveCommonJsModuleFile(packageJsonMainEntries.get(loadAddress));
+              return resolveJsModuleFile(packageJsonMainEntries.get(loadAddress));
             }
           } else {
             return loadAddress;
@@ -213,7 +303,8 @@ public final class ModuleLoader {
       return null;
     }
 
-    private String resolveCommonJsModuleFromRegistry(String requireName) {
+    @Nullable
+    private String resolveJsModuleFromRegistry(String moduleAddress) {
       for (Map.Entry<String, ImmutableSet<String>> nodeModulesFolder :
           nodeModulesRegistry.entrySet()) {
         if (!this.path.startsWith(nodeModulesFolder.getKey())) {
@@ -221,14 +312,14 @@ public final class ModuleLoader {
         }
 
         // Load as a file
-        String fullModulePath = nodeModulesFolder.getKey() + "node_modules/" + requireName;
-        String loadAddress = resolveCommonJsModuleFile(fullModulePath);
+        String fullModulePath = nodeModulesFolder.getKey() + "node_modules/" + moduleAddress;
+        String loadAddress = resolveJsModuleFile(fullModulePath);
         if (loadAddress != null) {
           return loadAddress;
         }
 
         // Load as a directory
-        loadAddress = resolveCommonJsModuleDirectory(fullModulePath);
+        loadAddress = resolveJsModuleDirectory(fullModulePath);
         if (loadAddress != null) {
           return loadAddress;
         }
@@ -238,46 +329,42 @@ public final class ModuleLoader {
     }
 
     /**
-     * Find an ES6 module {@code moduleName} relative to {@code context}.
-     * @return The normalized module URI, or {@code null} if not found.
+     * Locates the module with the given name, but returns null if there is no JS file in the
+     * expected location.
      */
-    public ModulePath resolveEs6Module(String moduleName) {
-      // Allow module names with or without the ".js" extension.
-      if (!moduleName.endsWith(".js")) {
-        moduleName += ".js";
-      }
-      String resolved = locateNoCheck(moduleName);
-      if (!modulePaths.contains(resolved) && errorHandler != null) {
-        errorHandler.report(CheckLevel.WARNING, JSError.make(LOAD_WARNING, moduleName));
-      }
-      return new ModulePath(resolved);
-    }
-
-    /**
-     * Locates the module with the given name, but returns successfully even if
-     * there is no JS file corresponding to the returned URI.
-     */
-    private String locateNoCheck(String name) {
+    @Nullable
+    private String locate(String name) {
       String path = ModuleNames.escapePath(name);
       if (isRelativeIdentifier(name)) {
         String ourPath = this.path;
         int lastIndex = ourPath.lastIndexOf('/');
         path = ModuleNames.canonicalizePath(ourPath.substring(0, lastIndex + 1) + path);
       }
-      return normalize(path, moduleRootPaths);
-    }
-
-    /**
-     * Locates the module with the given name, but returns null if there is no JS
-     * file in the expected location.
-     */
-    @Nullable
-    private String locate(String name) {
-      String path = locateNoCheck(name);
+      path = normalize(path, moduleRootPaths);
       if (modulePaths.contains(path)) {
         return path;
       }
       return null;
+    }
+
+    /**
+     * Treats the module address as a path and returns the name of that module. Does not verify that
+     * there is actually a JS file at the provided URI.
+     *
+     * <p>Primarily used for per-file ES6 module transpilation
+     */
+    public ModulePath resolveModuleAsPath(String moduleAddress) {
+      if (!moduleAddress.endsWith(".js")) {
+        moduleAddress += ".js";
+      }
+      String path = ModuleNames.escapePath(moduleAddress);
+      if (isRelativeIdentifier(moduleAddress)) {
+        String ourPath = this.path;
+        int lastIndex = ourPath.lastIndexOf('/');
+        path = ModuleNames.canonicalizePath(ourPath.substring(0, lastIndex + 1) + path);
+      }
+      path = normalize(path, moduleRootPaths);
+      return new ModulePath(path);
     }
   }
 
@@ -370,6 +457,14 @@ public final class ModuleLoader {
     for (String modulePath : modulePaths) {
       String[] nodeModulesDirs = modulePath.split("/node_modules/");
       String parentPath = "";
+
+      if (nodeModulesDirs.length > 0 && nodeModulesDirs[0].startsWith("node_modules/")) {
+        if (!registry.containsKey(parentPath)) {
+          registry.put(parentPath, new HashSet<String>());
+        }
+        registry.get(parentPath).add("node_modules/");
+      }
+
       for (int i = 0; i < nodeModulesDirs.length - 1; i++) {
         if (i + 1 < nodeModulesDirs.length) {
           parentPath += nodeModulesDirs[i] + "/";
@@ -433,5 +528,16 @@ public final class ModuleLoader {
 
   /** A trivial module loader with no roots. */
   public static final ModuleLoader EMPTY =
-      new ModuleLoader(null, ImmutableList.<String>of(), ImmutableList.<DependencyInfo>of());
+      new ModuleLoader(
+          null,
+          ImmutableList.<String>of(),
+          ImmutableList.<DependencyInfo>of(),
+          ResolutionMode.LEGACY);
+
+  /** An enum used to specify what algorithm to use to locate non path-based modules */
+  public enum ResolutionMode {
+    BROWSER,
+    LEGACY,
+    NODE
+  }
 }
--- a/src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
+++ b/src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
@@ -2525,6 +2525,12 @@ public abstract class AbstractCommandLineRunner<A extends Compiler,
       return this;
     }
 
+    public CommandLineConfig setModuleResolutionMode(ModuleLoader.ResolutionMode mode) {
+      this.moduleResolutionMode = mode;
+      return this;
+    }
+
+    private ModuleLoader.ResolutionMode moduleResolutionMode = ModuleLoader.ResolutionMode.LEGACY;
   }
 
   /**
--- a/src/com/google/javascript/jscomp/Compiler.java
+++ b/src/com/google/javascript/jscomp/Compiler.java
@@ -1481,9 +1481,15 @@ public class Compiler extends AbstractCompiler implements ErrorHandler, SourceFi
           || options.transformAMDToCJSModules
           || options.processCommonJSModules) {
 
-        this.moduleLoader = new ModuleLoader(this, options.moduleRoots, inputs);
+        this.moduleLoader =
+            new ModuleLoader(
+                this,
+                options.moduleRoots,
+                inputs,
+                ModuleLoader.PathResolver.RELATIVE,
+                options.moduleResolutionMode);
 
-        if (options.processCommonJSModules) {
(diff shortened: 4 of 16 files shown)


brad4d:
Sorry for the delay.
I'm importing and reviewing this now.

ChadKillingsworth:
After mulling this over a bit, I think it might be better to restrict the PATH_ONLY mode to loading files only and exclude the directory loading logic. I don't think directory loading logic will every be supported by browsers.

brad4d:
@ChadKillingsworth will that be a significant change to what you have so far?
I've been making changes to my internal copy to get tests to pass. Should I just wait for your update?

ChadKillingsworth:
It should not be a significant update.

Github has this new feature though - you can push any changes you make back to git@github.com:ChadKillingsworth/closure-compiler.git -> node-modules-es6 branch. I can add my changes on top of that.

brad4d:
That's good to know. Thanks.
However I don't know right off how to easily get my changes back into git,
so don't wait for me.
I'll keep trying to find out though.

ChadKillingsworth:
I pushed up a separate commit with the changes.

A lot of the tests had modules without a path identifier (didn't start with a `.` or `/` character). This definitely didn't work in external code (see #2193), but I'm unsure of the impact of this at Google. Edge is the only browser that supports ES6 modules currently. Edge not only requires import strings to begin with a `.` or `/` character, but also requires a file extension. However the specs seem pretty silent on this front.

We should probably discuss the impact of module path changes at the meeting on Monday.

brad4d:
@ChadKillingsworth could you go ahead and squash this into one commit?
Our import tool chokes otherwise, and it isn't clear how to work around it.
I'll just need to re-apply my previous changes on a new import.

ChadKillingsworth:
@brad4d done

brad4d:
Thanks, re-importing and testing now.

brad4d:
The runtime_tests/module_tests are failing with this imported.
I'm now running `mvn` in my git repo including this change to see if it fails there, too.

brad4d:
No failures from running `mvn`, but I'm not sure it's actually running the tests in runtime_tests.

ChadKillingsworth:
How can I run the runtime_tests? I always just use the mvn test command.

brad4d:
@ChadKillingsworth 
That is an excellent question.
It seems that we haven't developed any way for external folks to conveniently run these, which is obviously a problem. I'll file an issue for that.

In the meantime the problem is occurring with absolute_path_test.js.
The failure I'm seeing right now happens at compile time using these options.
I've modified them so I *think* they will work if you have closure-library checked out as a sibling directory to closure-compiler.

```
--jscomp_off=checkTypes
--jscomp_off=undefinedVars
--export_test_functions=true
--language_in=ECMASCRIPT6_STRICT
--language_out=ES3
--property_renaming=OFF
--strict
--variable_renaming=OFF
--jscomp_error=accessControls
--jscomp_error=ambiguousFunctionDecl
--jscomp_error=checkDebuggerStatement
--jscomp_error=checkEventfulObjectDisposal
--jscomp_error=checkRegExp
--jscomp_error=checkTypes
--jscomp_error=checkVars
--jscomp_error=const
--jscomp_error=constantProperty
--jscomp_error=deprecated
--jscomp_error=duplicate
--jscomp_error=duplicateMessage
--jscomp_error=externsValidation
--jscomp_error=extraRequire
--jscomp_error=fileoverviewTags
--jscomp_error=globalThis
--jscomp_error=internetExplorerChecks
--jscomp_error=invalidCasts
--jscomp_error=missingProperties
--jscomp_error=missingRequire
--jscomp_error=nonStandardJsDocs
--jscomp_error=strictModuleDepCheck
--jscomp_error=undefinedNames
--jscomp_error=undefinedVars
--jscomp_error=unknownDefines
--jscomp_error=uselessCode
--jscomp_error=visibility
--collapse_properties
--export_test_functions
--js test/com/google/javascript/jscomp/runtime_tests/module_tests/module_test_resources/classProtectedMethod.js
--js test/com/google/javascript/jscomp/runtime_tests/module_tests/module_test_resources/decl_leg_name_a.js
--js test/com/google/javascript/jscomp/runtime_tests/module_tests/module_test_resources/exportClass.js
--js test/com/google/javascript/jscomp/runtime_tests/module_tests/module_test_resources/exportDefault.js
--js test/com/google/javascript/jscomp/runtime_tests/module_tests/module_test_resources/exportDefaultClass.js
--js test/com/google/javascript/jscomp/runtime_tests/module_tests/module_test_resources/exportDefaultOldStyleClass.js
--js test/com/google/javascript/jscomp/runtime_tests/module_tests/module_test_resources/exportFrom.js
--js test/com/google/javascript/jscomp/runtime_tests/module_tests/module_test_resources/exportSpec.js
--js test/com/google/javascript/jscomp/runtime_tests/module_tests/module_test_resources/fq_jsdoc_a.js
--js test/com/google/javascript/jscomp/runtime_tests/module_tests/module_test_resources/goog_module.js
--js test/com/google/javascript/jscomp/runtime_tests/module_tests/module_test_resources/importAndExport.js
--js test/com/google/javascript/jscomp/runtime_tests/module_tests/module_test_resources/moduleOrderA.js
--js test/com/google/javascript/jscomp/runtime_tests/module_tests/module_test_resources/moduleOrderB.js
--js test/com/google/javascript/jscomp/runtime_tests/module_tests/module_test_resources/moduleOrderC.js
--js test/com/google/javascript/jscomp/runtime_tests/module_tests/module_test_resources/moduleOrderRecorder.js
--js test/com/google/javascript/jscomp/runtime_tests/module_tests/module_test_resources/module_destructuring.js
--js test/com/google/javascript/jscomp/runtime_tests/module_tests/module_test_resources/module_es3_a.js
--js test/com/google/javascript/jscomp/runtime_tests/module_tests/module_test_resources/module_es6_a.js
--js test/com/google/javascript/jscomp/runtime_tests/module_tests/module_test_resources/namespace_export.js
--js test/com/google/javascript/jscomp/runtime_tests/module_tests/module_test_resources/nesting.Outer.Inner.js
--js test/com/google/javascript/jscomp/runtime_tests/module_tests/module_test_resources/nesting.Outer.js
--js test/com/google/javascript/jscomp/runtime_tests/module_tests/module_test_resources/referenceOtherModule.js
--js test/com/google/javascript/jscomp/runtime_tests/module_tests/module_test_resources/simpleExport.js
--js test/com/google/javascript/jscomp/runtime_tests/module_tests/module_test_resources/typedef.js
--js ../closure-library/closure/goog/base.js
--js ../closure-library/closure/goog/deps.js
--js ../closure-library/closure/goog/testing/stacktrace.js
--js ../closure-library/closure/goog/testing/asserts.js
--js ../closure-library/closure/goog/debug/error.js
--js ../closure-library/closure/goog/dom/nodetype.js
--js ../closure-library/closure/goog/string/string.js
--js ../closure-library/closure/goog/asserts/asserts.js
--js ../closure-library/closure/goog/array/array.js
--js ../closure-library/closure/goog/dom/tagname.js
--js ../closure-library/closure/goog/object/object.js
--js ../closure-library/closure/goog/dom/tags.js
--js ../closure-library/closure/goog/i18n/bidi.js
--js ../closure-library/closure/goog/labs/useragent/util.js
--js ../closure-library/closure/goog/labs/useragent/browser.js
--js ../closure-library/closure/goog/string/typedstring.js
--js ../closure-library/closure/goog/string/const.js
--js ../closure-library/closure/goog/html/safestyle.js
--js ../closure-library/closure/goog/html/safestylesheet.js
--js ../closure-library/closure/goog/fs/url.js
--js ../closure-library/closure/goog/html/safeurl.js
--js ../closure-library/closure/goog/html/trustedresourceurl.js
--js ../closure-library/closure/goog/html/safehtml.js
--js ../closure-library/closure/goog/html/safescript.js
--js ../closure-library/closure/goog/html/uncheckedconversions.js
--js ../closure-library/closure/goog/structs/collection.js
--js ../closure-library/closure/goog/functions/functions.js
--js ../closure-library/closure/goog/math/math.js
--js ../closure-library/closure/goog/iter/iter.js
--js ../closure-library/closure/goog/structs/map.js
--js ../closure-library/closure/goog/structs/structs.js
--js ../closure-library/closure/goog/structs/set.js
--js ../closure-library/closure/goog/labs/useragent/engine.js
--js ../closure-library/closure/goog/labs/useragent/platform.js
--js ../closure-library/closure/goog/reflect/reflect.js
--js ../closure-library/closure/goog/useragent/useragent.js
--js ../closure-library/closure/goog/debug/debug.js
--js ../closure-library/closure/goog/debug/logrecord.js
--js ../closure-library/closure/goog/debug/logbuffer.js
--js ../closure-library/closure/goog/debug/logger.js
--js ../closure-library/closure/goog/debug/relativetimeprovider.js
--js ../closure-library/closure/goog/debug/formatter.js
--js ../closure-library/closure/goog/debug/console.js
--js ../closure-library/closure/goog/math/coordinate.js
--js ../closure-library/closure/goog/math/size.js
--js ../closure-library/closure/goog/dom/browserfeature.js
--js ../closure-library/closure/goog/dom/safe.js
--js ../closure-library/closure/goog/dom/dom.js
--js ../closure-library/closure/goog/async/freelist.js
--js ../closure-library/closure/goog/debug/entrypointregistry.js
--js ../closure-library/closure/goog/async/nexttick.js
--js ../closure-library/closure/goog/async/workqueue.js
--js ../closure-library/closure/goog/async/run.js
--js ../closure-library/closure/goog/promise/resolver.js
--js ../closure-library/closure/goog/promise/thenable.js
--js ../closure-library/closure/goog/promise/promise.js
--js ../closure-library/closure/goog/testing/testcase.js
--js ../closure-library/closure/goog/testing/testrunner.js
--js ../closure-library/closure/goog/testing/jsunit.js
--js ../closure-library/closure/goog/disposable/idisposable.js
--js ../closure-library/closure/goog/disposable/disposable.js
--js ../closure-library/closure/goog/events/browserfeature.js
--js ../closure-library/closure/goog/events/eventid.js
--js ../closure-library/closure/goog/events/event.js
--js ../closure-library/closure/goog/events/eventtype.js
--js ../closure-library/closure/goog/events/browserevent.js
--js ../closure-library/closure/goog/events/listenable.js
--js ../closure-library/closure/goog/events/listener.js
--js ../closure-library/closure/goog/events/listenermap.js
--js ../closure-library/closure/goog/events/events.js
--js ../closure-library/closure/goog/events/eventtarget.js
--js ../closure-library/closure/goog/events/keycodes.js
--js ../closure-library/closure/goog/dom/vendor.js
--js ../closure-library/closure/goog/html/legacyconversions.js
--js ../closure-library/closure/goog/math/box.js
--js ../closure-library/closure/goog/math/irect.js
--js ../closure-library/closure/goog/math/rect.js
--js ../closure-library/closure/goog/style/style.js
--js ../closure-library/closure/goog/testing/events/events.js
--js ../closure-library/closure/goog/testing/objectpropertystring.js
--js ../closure-library/closure/goog/testing/propertyreplacer.js
--js ../closure-library/closure/goog/testing/mockclock.js
--js ../closure-library/closure/goog/testing/mockinterface.js
--js ../closure-library/closure/goog/testing/mockmatchers.js
--js ../closure-library/closure/goog/testing/mock.js
--js ../closure-library/closure/goog/testing/loosemock.js
--js ../closure-library/closure/goog/testing/strictmock.js
--js ../closure-library/closure/goog/testing/functionmock.js
--js ../closure-library/closure/goog/testing/mockcontrol.js
--js ../closure-library/closure/goog/labs/testing/environment.js
--js ../closure-library/closure/goog/testing/testsuite.js
--js ../closure-library/closure/goog/events/eventhandler.js
--js test/com/google/javascript/jscomp/runtime_tests/module_tests/absolute_path_test.js
--externs externs/google.js
--externs externs/es3.js
--externs externs/es5.js
--externs externs/es6.js
--externs externs/es6_collections.js
--externs externs/w3c_event.js
--externs externs/w3c_event3.js
--externs externs/w3c_device_sensor_event.js
--externs externs/gecko_event.js
--externs externs/ie_event.js
--externs externs/webkit_event.js
--externs externs/w3c_pointer_events.js
--externs externs/w3c_gamepad.js
--externs externs/w3c_dom1.js
--externs externs/w3c_dom2.js
--externs externs/w3c_dom3.js
--externs externs/w3c_dom4.js
--externs externs/html5.js
--externs externs/gecko_dom.js
--externs externs/ie_dom.js
--externs externs/webkit_dom.js
--externs externs/w3c_css.js
--externs externs/gecko_css.js
--externs externs/ie_css.js
--externs externs/webkit_css.js
--externs externs/chrome.js
--externs externs/deprecated.js
--externs externs/fileapi.js
--externs externs/flash.js
--externs externs/gecko_xml.js
--externs externs/page_visibility.js
--externs externs/ie_vml.js
--externs externs/intersection_observer.js
--externs externs/iphone.js
--externs externs/mediakeys.js
--externs externs/mediasource.js
--externs externs/url.js
--externs externs/v8.js
--externs externs/webstorage.js
--externs externs/whatwg_encoding.js
--externs externs/w3c_anim_timing.js
--externs externs/w3c_audio.js
--externs externs/w3c_batterystatus.js
--externs externs/w3c_clipboardevent.js
--externs externs/w3c_css3d.js
--externs externs/w3c_elementtraversal.js
--externs externs/w3c_eventsource.js
--externs externs/w3c_geolocation.js
--externs externs/w3c_indexeddb.js
--externs externs/w3c_midi.js
--externs externs/w3c_navigation_timing.js
--externs externs/w3c_range.js
--externs externs/w3c_rtc.js
--externs externs/w3c_screen_orientation.js
--externs externs/w3c_selectors.js
--externs externs/w3c_speech.js
--externs externs/w3c_touch_event.js
--externs externs/w3c_webcrypto.js
--externs externs/w3c_xml.js
--externs externs/window.js
--externs externs/webkit_notifications.js
--externs externs/webkit_usercontent.js
--externs externs/webgl.js
--externs externs/w3c_serviceworker.js
--externs externs/w3c_permissions.js
--externs externs/intl.js
--externs externs/fetchapi.js
--externs externs/streamsapi.js
```

ChadKillingsworth:
I noticed. Trying to create a custom build with those options.

ChadKillingsworth:
@brad4d I've gotten some of the tests to compile and run - not even close to all the requirements are available externally. Can you list which test files have failures? I'll restrict my work to looking just at those.

brad4d:
My test run bailed after the file I already mentioned above.
I'm running again with an option to make it keep going for other files.
Stay tuned.

ChadKillingsworth:
@brad4d That test isn't in the open source repo: 
`test/com/google/javascript/jscomp/runtime_tests/module_tests/absolute_path_test.js`

If you can send it to me via email, I'll see what's going on.

brad4d:
Well, that's interesting. `absolute_path_test.js` is explicitly excluded.

MatrixFrog:
Right because the absolute path is different within Google than it is here. Until external people can run the tests, I don't think it's worth it to make an external version of that test.

brad4d:
OK, well it's only excluded because it contains a hard-coded path that won't work in the git repo.
It's tiny, so here's its full text with the full path modified to match the git repo.

```javascript
import s from 'test/com/google/javascript/jscomp/runtime_tests/module_tests/module_test_resources/exportDefault';

function testDefault() {
  assertEquals('this is the default export', s);
}
```

ChadKillingsworth:
So I can see right away the problem: the path doesn't start with either a '.' or '/'. This is by design. We can change that behavior for PATH_ONLY, but it seems like this is a case we should fix. However if it's a big fix, we can do it in stages.

Is there by chance a `--js_module_root` flag used on that test?

MatrixFrog:
I think internally Blaze automatically passes a default --js_module_root flag, yes.

ChadKillingsworth:
How do we want to normalize this then? Outside of Google other module loaders let you specify roots, but still require the '/' or '.' character preceding.

MatrixFrog:
Let's switch to requiring the `/` then I guess. Might take a lot of work to migrate code to that internally, but it should be pretty easy to automate.

brad4d:
I did notice the missing '/'. I tried adding it, but the test still failed in what looked like the same way.
`--jsmodule_root` is not used. If it were, it would have appeared in my list of options above.

ChadKillingsworth:
Ok I'll run it locally and see what I can find out.

ChadKillingsworth:
Hmm the old version just allows this.  The other option is to add a `LEGACY` mode and set it default internal to Google.

brad4d:
@ChadKillingsworth 
I'm not clear on what you mean by "old version" here.

ChadKillingsworth:
Sorry - by "old version" I meant the current HEAD builds.

Here's why this is really difficult: the `--js_module_root` flag does something very different than pretty much anyone would expect and doesn't help solve this problem.

Given these flags: `--js_module_root=/Users/chad/git/closure-compiler --js entry.js --js module.js`

You would expect this to work (but it doesn't):
```js
import foo from '/module';
```

Instead, the `--js_module_root` flag actually does this:
**Original:**
```js
import foo from '/Users/chad/git/closure-compiler/module';
```

**Rewritten:**
```js
goog.provide('module$entry');
goog.require('module$module');
```

So all the js_module_root is seems to be good for is to shorten the rewritten module symbol. I can fix that in a separate PR, but I have no idea what kind of impact that would have at Google.

brad4d:
@jart 
Looks like the only code within Google that uses the `--js_module_root` flag is your bazel rules code.
Perhaps you can help us puzzle this out?

brad4d:
@MatrixFrog pointed out I overlooked some code that used `--js_module_root` from within Java code.

@ChadKillingsworth 
It does seem safest to have a `LEGACY` mode preserving existing behavior. How much of a problem is that?

ChadKillingsworth:
It's not hard. I'll get fixes pushed.

I'm going to create a separate PR for fixing the `--js_module_root` flag.

brad4d:
@ChadKillingsworth what does "fixing" mean in this case?

ChadKillingsworth:
Make it act like most people expect. Namely so that absolute imports like so:

```js
import foo from '/module';
```

Are resolved against any provided module roots.

jart:
> A lot of the tests had modules without a path identifier (didn't start with a . or / character). This definitely didn't work in external code

Why not? If you `import foo from 'bar'` then if a `bar.js` flag is passed to the compiler, it will work.

> see #2139

What he was doing works for me. I tested it on the command line to be sure.

> jart@ said in https://github.com/google/closure-compiler/issues/2139#issuecomment-261120739
> If you remove the --js_module_root ./ flag then what you did should work. That flag is meant to strip prefixes off the filename paths passed to the compiler via flags. I don't think it actually takes the file system into consideration. So because the ./ was being stripped, it started treating your imports as relative to the current directory, rather than being relative to the source file basename.

If you want to learn more about the problem that `--js_module_root` was designed to solve then read [find_roots() in def.bzl (Closure Rules)](https://github.com/bazelbuild/rules_closure/blob/master/closure/private/defs.bzl#L121). In Bazel, all source files are treated as relative to the root of the repository. Closure Compiler's current directory when it runs is the root of that gigantic repository, and all the paths it gets passed as flags are relative paths to that directory. However sometimes Bazel creates weird generated directories like bazel-out/genfiles/ which become a root in and of themselves that needs to be stripped off when the JSCompiler runs.

> Until external people can run the tests [...]

@MatrixFrog Maybe we should set this up repository up with Bazel Closure Rules (once we're done migrating Closure Rules from PhantomJS to [rules_webtesting](https://github.com/bazelbuild/rules_webtesting).)

src/com/google/javascript/jscomp/CommandLineRunner.java
```diff
@@ -682,7 +682,12 @@
     @Option(name = "--print_source_after_each_pass",
         hidden = true,
         usage = "Whether to iteratively print resulting JS source per pass.")
-        private boolean printSourceAfterEachPass = false;
+    private boolean printSourceAfterEachPass = false;
+
+    @Option(name = "--module_resolution",
+        hidden = false,
+        usage = "Specifies how the compiler locates modules. Defaults to NODE.")
```
ChadKillingsworth: The current compiler behavior isn't standard in any way. It doesn't match any other module loading system. The expectation of users (especially new users) is that it works like Node because to date the biggest use case is bundling.

ChadKillingsworth:
> Why not? If you import foo from 'bar' then if a bar.js flag is passed to the compiler, it will work.

Only if 'bar.js' is in the compilation root folder. Once you move to a subfolder, it breaks. This is very counter-intuitive and doesn't follow URL or folder rules.

> However sometimes Bazel creates weird generated directories like bazel-out/genfiles/ which become a root in and of themselves that needs to be stripped off when the JSCompiler runs.

What I'm trying to figure out is if it can be made to work in reverse (without a huge migration). Namely, if you provide a list of folders which are considered roots and absolute module paths are anchored to one of those. This is what most module loaders outside of Google do. So for your example:

`--js_module_root=/bazel/compilation/root --js_module_root=/bazel/compilation/root/bazel-out/genfiles --js src/foo.js --js bazel-out/genfiles/bar.js`

And in foo.js:
```js
import s from '/bar'
```

jart:
> What I'm trying to figure out is if it can be made to work in reverse (without a huge migration). Namely, if you provide a list of folders which are considered roots and absolute module paths are anchored to one of those.

But then the filename arguments passed to the compiler wouldn't be filenames anymore; they would be module names. Then the build is no longer hermetic because the full set of inputs isn't being specified.

I'm not sure if I understand this change. Does the `--module_resolution` flag only apply if `--process_common_js_modules` is also specified? Or does it apply to *all* ES6 code regardless of whether or not CommonJS is enabled? If so, why would we want ES6 code to follow non-standard Node conventions?

jart:
Also would it be possible for you to write a shell script wrapper around the Closure Compiler that inverts the `--js_module_root` flag? You could just do the module resolution in the shell script to get the real filename, and then recalculate the `--js` and `--js_module_root` flags to make Closure understand it.

ChadKillingsworth:
> Does the --module_resolution flag only apply if --process_common_js_modules is also specified? Or does it apply to all ES6 code regardless of whether or not CommonJS is enabled? If so, why would we want ES6 code to follow non-standard Node conventions?

Module resolution will apply to both CommonJS and ES6 modules.

Node runs v8 or Chakra - ES6 modules are coming natively. In addition, ES6 modules are already used for node code via transpilation.

Browser libraries also use Node node dependencies and it is extremely common for their build systems to import (via ES6 or CommonJS) dependencies from the `node_modules` folder.

Both of these use cases are very common in live code today.

As for the `--js_module_root` flag - I'm going to leave it as is in this PR. I'll make a new PR just for changes there where we can either satisfy everyone's use cases or create a new flag.

jart:
Ok if the behavior is being changed for all ES6 code, then help me understand what exactly is being changed, as someone who is largely ignorant of Node and CommonJS. For example, will this cause Closure Compiler to start searching the filesystem for files that weren't specified via flags?

ChadKillingsworth:
> For example, will this cause Closure Compiler to start searching the filesystem for files that weren't specified via flags?

No. You still are responsible for providing the complete source set of files.

Node module resolution solves many of the same problems that `goog.module` does inside Google: namely that you don't have to know the exact filesystem path of dependencies in code. Instead, any module path that doesn't begin with a `.` or a `/` is looked for from the `node_modules` folder. There's quite a few intricacies involved, but that folder is managed in such a way so that if 2 packages have conflicting dependencies, the import will return the correct version for each.

To provide the correct source set there are a couple of options:

 1. Pass in all the JS and JSON files in the "node_modules" folder as source and let the compiler dependency management drop and sort them.
 2. Run a pre-build step that discovers dependencies from an entry point and passes in the exact set. This option is the recommended one.

I'm working on tooling for the NPM deployment to facilitate the 2nd option. This will probably lead into an easy way to configure and support the code-splitting functionality of the compiler as well: https://twitter.com/chadhikes/status/799266788290613249

jart:
How does it know where the `node_modules` directory is? Does it only look in the same directory as the module that's importing it? If so, wouldn't that give modules quadratic disk space complexity? If so, that kind of blows my mind.

ChadKillingsworth:
The algorithm walks up the directory tree for a "node_modules" folder. It uses the first one it finds.

In practice the folder is almost always at the root of your project. Something like:

```
Project Root
├─┬ src
│ └── entry.js
└─┬ node_modules
  ├─┬ dep1  
  │ └─┬ node_modules
  │   └── sub-dep@1.24.0
  └─┬ dep2
    └─┬ node_modules
      └── sub-dep@2.3.0
```

kuus:
sorry dismiss my code review please, I clicked by mistake

brad4d:
@ChadKillingsworth 
I believe I'm waiting for you to make further modifications here regarding how paths are handled starting with '/', '.', or something else. Is that right?

ChadKillingsworth:
@brad4d Ready for review again. I opted not to refactor the deps package mainly due to the concerns in #2175.

There are now 3 resolution modes - all of which effect both CommonJS and ES6 modules:

 * `LEGACY` keeps the same behavior the compiler has used historically.
    + Modules which do not begin with a "." or "/" character are assumed to be relative to the compilation root.
    + ".js" file extensions are added if the import omits them.
 * `NODE` uses the node module resolution algorithm
    + Modules which do not begin with a "." or "/" character are looked up from the appropriate `node_modules` folder. Includes the ability to require directories and JSON files.
    + Exact match, then ".js", then ".json" file extensions are searched.
  * `BROWSER` mimics the behavior of MS Edge
    + Modules must begin with a "." or "/" character
    + Modules must include the file extension

jart:
Why call the old way LEGACY? Why not PATH? I like PATH. I think it makes sense. I think we should try to get browsers to implement support for PATH, where dotless paths are considered relative to the `/` root directory on the web server. That way it will be easy to export a Blaze or Bazel raw sources project to a web server.

jart:
Also, let's be real, do we really want IE and Node to be the ones setting the standards?

jart:
Also take into consideration that the algorithmic complexity of PATH is better than NODE.

ChadKillingsworth:
@jart The whatwg work to support ES6 modules in browsers has been long and contentious. I doubt any effort to change what is currently specced would get traction. The `BROWSER` mode is designed to track whatever browsers ship so as more browsers add support that one should be changed to match. I'm watching intently to see what Chrome does, but so far it's not an option.

On the Node side - they HAVE set the standards and have participated in TC39. Node modules easily have the largest implementation in the wild. While inside Google this might not make sense, it is what the vast majority of external developers are familiar with.

As for LEGACY, that algorithm was the best guess of the team at the time. But outside of Google the exact behavior is not documented (I only figured it out by reading code) and doesn't match anything else. If several feel strongly about changing the name, then we can. However, to call it `PATH` I would argue that he module paths should look like `~/foo/bar` to make the intent clear. For URIs, paths which don't begin with "." or "/" are relative to the current folder/file - not the root.

brad4d:
I'm testing this now.

brad4d:
The absolute_path_test.js test is still failing.

brad4d:
I suspect this is because the default resolution mode in CompilerOptions.java is being set to PATH_ONLY instead of LEGACY.

brad4d:
Nope, it is LEGACY. I was looking at the old change...

ChadKillingsworth:
Ok good - because I don't even know how it would compile that way.

I'm going to run the test locally.

ChadKillingsworth:
The test compiles and runs locally for me. Do you get a compile time warning like this?

```
test/com/google/javascript/jscomp/runtime_tests/module_tests/absolute_path_test.js:1: WARNING - Failed to load module "test/com/google/javascript/jscomp/runtime_tests/module_tests/module_test_resources/exportDefault"
import s from 'test/com/google/javascript/jscomp/runtime_tests/module_tests/module_test_resources/exportDefault';
^

0 error(s), 1 warning(s), 96.9% typed
```

Because that's what I get if the module resolution isn't `LEGACY`.

ChadKillingsworth:
Here's the flags I used:

```
{
  output_wrapper: '(function(){%output%}).call(this)',
  compilation_level: 'SIMPLE',
  language_in: 'ECMASCRIPT_NEXT',
  language_out: 'ECMASCRIPT5',
  jscomp_off: '*',
  jscomp_error: [
    'accessControls',
    'ambiguousFunctionDecl',
    'checkDebuggerStatement',
    'checkEventfulObjectDisposal',
    'checkRegExp',
    'checkTypes',
    'checkVars',
    'const',
    'constantProperty',
    'deprecated',
    'duplicate',
    'duplicateMessage',
    'externsValidation',
    'extraRequire',
    'fileoverviewTags',
    'globalThis',
    'internetExplorerChecks',
    'invalidCasts',
    'missingProperties',
    'missingRequire',
    'nonStandardJsDocs',
    'strictModuleDepCheck',
    'undefinedNames',
    'undefinedVars',
    'unknownDefines',
    'uselessCode',
    'visibility'
  ],
  module_resolution: 'NODE',
  js_module_root: '/Users/chad/git/closure-compiler'
}
```

brad4d:
I get this
```
"module$$third_party$java_src$jscomp$javatests$com$google$javascript$jscomp$runtime_tests$module_tests$module_test_resources$exportDefault" is required but never provided.
```

It's actually failing when running MakeJsDeps.

ChadKillingsworth:
Which I don't believe is open source ...

jart:
> The whatwg work to support ES6 modules in browsers has been long and contentious. I doubt any effort to change what is currently specced would get traction. 

If the standards committees have failed to reach a consensus, then maybe that means the standard is going to evolve organically. Therefore what is decided upon by Google, the Closure Compiler team, and its contributors—is going to have an impact on that process by influencing the opinions of others. Thank you for all the work you've done, and for being a part of that conversation.

> The BROWSER mode is designed to track whatever browsers ship so as more browsers add support that one should be changed to match. I'm watching intently to see what Chrome does, but so far it's not an option.

If the definition of BROWSER is going to change over time, and other browsers haven't decided yet, then maybe it should be called EDGE?

> As for LEGACY, that algorithm was the best guess of the team at the time.

What you call LEGACY was designed to meet the requirements of a monolithic source code repository containing petabytes of data. It's a common practice at large corporations to force all sources to share a common root, because it's the only way to make certain things tractable, like for example, compiling millions of C++ files.

I understand that not everyone does things this way. But please take into consideration that PATH has a use case and the only reason why we would consider calling it LEGACY would be if we were considering migrating our repository to NODE, which is not something we're considering.

> But outside of Google the exact behavior is not documented (I only figured it out by reading code) and doesn't match anything else. 

You're right. We should do a better job explaining this. But also remember that NODE is defined [one page algorithm](https://nodejs.org/api/modules.html#modules_all_together) and PATH can be defined by one sentence: "it's the same as the paths you pass to the compiler command."

ChadKillingsworth:
@jart `BROWSER` is what I expect most Google products to migrate to. The change from `LEGACY` should be one that refactoring tools can handle. The following should be equivalent:

`LEGACY` with `--js_module_root=.`
```
import s from 'path/to/file'
```

`BROWSER` with `--js_module_root=.`
```
import s from '/path/to/file.js'
```

There should be no loss in functionality from what `LEGACY` currently has - just a different way to express it.

jart:
Now I think I have a better understanding of what you mean. Refactoring our entire internal codebase so a `/` is added to the beginning of all ES6 imports is possible and not something I'm opposed to.

brad4d:
I'm trying to make an open source version of an internal test that this change causes to fail.
Along the way I've found a couple of other bugs I've also needed to fix.

ToDo:
1.  Submit latest fix - DepsGenerator.java shouldn't be ignoring non-existent files to avoid throwing an IO error. Currently in review.
2.  Submit DepsGeneratorTest.java including a test that demonstrates the failure this PR causes.

Things I'm uncertain about:
1. Is it sufficient to create DepsGeneratorTest.java, or do I need to do other stuff to make sure it gets run for the open-source project?

ChadKillingsworth:
> Is it sufficient to create DepsGeneratorTest.java, or do I need to do other stuff to make sure it gets run for the open-source project?

It should be sufficient for the test to exist @brad4d

probins:
> Edge is the only browser that supports ES6 modules currently. Edge not only requires import strings to begin with a . or / character, but also requires a file extension. However the specs seem pretty silent on this front.

Edge is implementing the [module script spec](https://html.spec.whatwg.org/#integration-with-the-javascript-module-system), which loads from urls, not paths. If Edge is requiring a file extension, that's not compliant with the current spec (which may well be because the spec has changed since they wrote the code :-)). Firefox has also implemented this, but atm only for 'chrome documents' https://bugzilla.mozilla.org/show_bug.cgi?id=1240072

cramforce:
@ChadKillingsworth Was just wondering whether you have an idea when you'll be able to continue working on this? Looks amazing!

ChadKillingsworth:
@cramforce hopefully next week. There is an internal test at Google breaking. @brad4d is working on an open source equivalent test so I can fix it. Progress was slowed by holiday breaks.

brad4d:
@ChadKillingsworth 
Just before I left for vacation I added `DepsGeneratorTest.java` to the repo.
This should be the place to put a test that reproduces the internal test failure.

brad4d:
I think I'm now seeing some other test failures likely due to the repo moving forward.
I'm going to do a fresh import of this PR to work from...

brad4d:
@ChadKillingsworth I think you made a small change that seems to have fixed the problem test in the latest import I did. Yay!
The other failures turned out to be just things I needed to tweak or timeouts.
I'm doing a "test all the things" run, but I'm feeling pretty optimistic about getting this submitted now.

brad4d:
The only failing tests now are 2 belonging to @jart in closure rules.

Errors look like this:

```
HINT: Closure will compile faster if you run: echo "build --strategy=Closure=worker" >>~/.bazelrc
third_party/bazel_rules/rules_closure/closure/testing/test/arithmetic_es6module_test.js:18: ERROR - Failed to load module "third_party/bazel_rules/rules_closure/closure/testing/test/arithmetic_es6module"
import { add } from 'third_party/bazel_rules/rules_closure/closure/testing/test/arithmetic_es6module';
^
  ProTip: "JSC_JS_MODULE_LOAD_WARNING" can be added to the `suppress` attribute of:
  //third_party/bazel_rules/rules_closure/closure/testing/test:arithmetic_es6module_test_lib
```

src/com/google/javascript/jscomp/deps/ModuleLoader.java
```diff
@@ -149,32 +157,80 @@ public String toModuleName() {
...
+      String loadAddress = null;
 
       // * the immediate name require'd
-      if (isAbsoluteIdentifier(requireName) || isRelativeIdentifier(requireName)) {
-        loadAddress = resolveCommonJsModuleFileOrDirectory(requireName);
-      } else {
-        loadAddress = resolveCommonJsModuleFromRegistry(requireName);
+      switch (resolutionMode) {
+        case LEGACY:
+          loadAddress = resolveJsModuleFile(moduleAddress);
```
brad4d: @ChadKillingsworth, I'd expect LEGACY to preserve the behavior as it was prior to this change, but this is not.
I think I should change this to contain something like the if/else statement above.

ChadKillingsworth: LEGACY mode should never resolve a directory. I thought this code path was identical to the old behavior, but it's possible that it's a change.  Is there a specific case that this is breaking?

brad4d: @ChadKillingsworth 
The only breakages I know of are the 2 related to closure rules I mentioned above.
I'm not sure of exactly why they fail,
but the LEGACY code path is definitely not defining the same behavior as before this PR,
so I believe the answer lies somewhere in that difference.

Since you said you updated it, I'll grab a fresh copy of this PR.
I'm going to modify it a bit before submission to ensure that the LEGACY code path preserves current behavior.

ChadKillingsworth: @brad4d So I missed the last set of comments about the closure-rules. Github page caching just didn't even show them to me. I think one minor tweak ought to fix the Closure-rules caching. It's nice to see this getting close.

ChadKillingsworth: I can even add a test for that case ...

brad4d: Shiny!
I'd really like to see the PR show a clear preservation of the previous behavior for the LEGACY code path, though, rather than just a fix for this one case.

Could you do something like this?

1. Define private `ModuleLoader.legacyResolveJsModule()` (& related methods) that completely preserves current behavior in a way that is clear in the diff.
2. Call that from the `LEGACY` case in the switch statement.

Thanks.

ChadKillingsworth: Yeah that's no problem at all. I'll do it this weekend.

ChadKillingsworth:
@brad4d Changes made.

The legacy function always returns a module path - even if it doesn't exist. This is how the current functionality works. Both of the other modes return null in this case.

src/com/google/javascript/jscomp/deps/ModuleLoader.java
```diff
@@ -149,32 +172,97 @@ public String toModuleName() {
...
+    @Nullable
+    public ModulePath resolveJsModule(
+        String moduleAddress, String sourcename, int lineno, int colno) {
+      String loadAddress = null;
 
       // * the immediate name require'd
-      if (isAbsoluteIdentifier(requireName) || isRelativeIdentifier(requireName)) {
-        loadAddress = resolveCommonJsModuleFileOrDirectory(requireName);
-      } else {
-        loadAddress = resolveCommonJsModuleFromRegistry(requireName);
```
brad4d: @ChadKillingsworth 
I expected to see this if-else in resolveJsModuleLegacy(), probably calling different subroutines, though (e.g. resolveLegacyJsModuleFileOrDirectory())

I believe you are trying to simplify things by eliminating cases that shouldn't apply in LEGACY mode,
but this still makes the change more risky.
I'd rather such simplification be done in a separate change.

brad4d:
Latest version imported & essential internal tweaks applied.
I've resolved the conflict github noticed in `CommandLineRunner.java` in my internal copy.
Internal testing in progress.

brad4d:
This has been submitted internally & will appear in github on our daily export.

probins:
@ChadKillingsworth I see you've updated the JS Modules wiki page. Node and Browser resolution mode both say 'This is the only mode available for releases after Feb 2017`. I assume that this should read 'This mode is only available for releases after Dec 2016' (or possibly 'from Feb 2017', assuming there'll be a release containing this change in Feb)?

ChadKillingsworth:
@probins You should simply read that as the code has yet to be released. Utilizing the behavior requires building the compiler from HEAD. I'll be updating the page again one a release containing the code has occurred.

This PR has such a high level of interest, I wanted the documentation out there for people to use it.

supersteves:
I think @probins main point was that I think you copy pasted wrong.
"This is the only mode available for releases after Feb 2017"
should be
"This mode is only available for releases after Feb 2017"
for the 2 new modes.

ChadKillingsworth:
Ahh good catch - fixed.

jart:
Note for anyone reading in the future, TensorBoard learned the hard way that ES6 modules are grimly fiendish. https://github.com/tensorflow/tensorboard/pull/747

PR #2133, opened 2016-11-08
Don't convert IIFE params to variables when the arguments contain "this" or "arguments" references

ChadKillingsworth:
Followup to #2103


dimvar:
This pass was rolled back internally yesterday. I'm assigning John to this PR, because he had hesitations about including this pass (because he feels the right place for this change is function inlining), and I don't have a strong preference about whether we include it or not.

ChadKillingsworth:
Closing this - we're going to try rolling forward 0660acd

PR #2137, opened 2016-11-09
Fixing angular's $q.Promise.finally externs

Krun:
Externs for .finally() is missing the second argument for notifyCallback


Krun:
I signed it!

PR #2140, opened 2016-11-10
Make the second argument to Polymer's toggleAttribute() optional.

jrobinson01:
The second argument is optional:
https://www.polymer-project.org/1.0/docs/api/Polymer.Base#method-toggleAttribute


jrobinson01:
I signed it!

Issue #2119, opened 2016-11-01
Polymer Behaviors Not Compatible with Type Based Optimizations


ChadKillingsworth:
I've yet to figure out how to create a simple repo, but the type based optimizations can break Polymer behaviors. The behavior literal keys are renamed in a conflicting manner:

```js
/** @polymerBehavior */
var fooBehavior = {
  foo: function() { alert('foo'); },
};

/** @polymerBehavior */
var barBehavior = {
  bar: function() { alert('bar'); },
};

(function() {
  Polymer({
    is: 'foo-bar',
    behaviors: [fooBehavior, barBehavior]
  });
})();
```

Becomes something like:

```js
/** @polymerBehavior */
var a = {
  b: function() { alert('foo'); },
};

/** @polymerBehavior */
var c = {
  b: function() { alert('bar'); },
};

(function() {
  Polymer({
    is: 'foo-bar',
    behaviors: [a, c]
  });
})();
```

I believe the way to fix this is to change the PolymerPass processing of behaviors such that a synthetic interface is created for each behavior. The behavior object literal will then be type cast to that interface and any Polymer elements using the behavior will be marked as implementing it.

I'm guessing this hasn't come up at Google because the Polymer-Renamer is already incompatible with type based optimizations. However https://github.com/banno/polymer-rename does support them.

dimvar:
AFAIU, you're saying that the root of the problem is that the two object literals appear unrelated to the type-based optimizations, but are combined in `behaviors: [a, c]`, so their properties should have distinct names.

[...]

PR #2143, opened 2016-11-11
Ambiguate Properties Bug Fix

ChadKillingsworth:
Came from investigating #2119 - it's very tricky to reproduce.

Namely in certain cases, the keys from unrelated object literals were ambiguated to the same name, however both were used as mixins. Since the object lits themselves don't have a type (their types weren't inferred from an assignment), this shouldn't have occurred.

Tracing through the ambiguate properties pass, I discovered that for GETPROP nodes, we use the type of the left portion of the getprop (example: `Foo.prototype.bar` uses the type of `Foo.prototype`), where as for object literals we use the type from the **first key** (`{foo: 4, bar: '1'}` both use the type of `foo` which is a number) - which just seems wrong.

This warrants some close scrutiny however. Git blame says that this code was in the initial open source commit.

### DIFF
--- a/src/com/google/javascript/jscomp/AmbiguateProperties.java
+++ b/src/com/google/javascript/jscomp/AmbiguateProperties.java
@@ -35,7 +35,6 @@ import com.google.javascript.rhino.jstype.JSType;
 import com.google.javascript.rhino.jstype.JSTypeNative;
 import com.google.javascript.rhino.jstype.JSTypeRegistry;
 import com.google.javascript.rhino.jstype.ObjectType;
-
 import java.util.ArrayList;
 import java.util.BitSet;
 import java.util.Comparator;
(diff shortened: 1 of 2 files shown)


concavelenz:
I looked back on the history of this.  At the time the pass was originally written object literals children were structured as a list of "key, value, key, value, ..." so the call to get first child returned "null" and the getJsType return UNKNOWN so the bug was harmless however we later restructured object literals as "key, key, key" with the value under it and this bug was introduced.

Your change looks good but I would like to see more test cases. It does worry me just because this code has been this way a long time but, if we have trouble landing this,  the alternative would be to use "unknown" as the type to land a fix and go from there.

ChadKillingsworth:
Tests added (or rather copied then updated to use object literals). I converted these tests to use `LINE_JOINER` while I was there.

concavelenz:
LGTM

PR #2145, opened 2016-11-12
Treat any type annotation with a `/` character as a module reference

ChadKillingsworth:
Goes along (but does not require) #2094  and #2130.

When updating type node annotations during module rewriting, treat any type name with a `/` as a module identifier. Needed for named modules loaded from a registry.

Example `/** @type {!page/index.Context} */`

### DIFF
--- a/src/com/google/javascript/jscomp/ProcessCommonJSModules.java
+++ b/src/com/google/javascript/jscomp/ProcessCommonJSModules.java
@@ -1101,7 +1101,7 @@ public final class ProcessCommonJSModules implements CompilerPass {
       if (typeNode.isString()) {
         String name = typeNode.getString();
         // Type nodes can be module paths.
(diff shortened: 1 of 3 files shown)


brad4d:
I'm importing & testing this PR now.

brad4d:
@ChadKillingsworth 
All tests passed and LGTM, but why no test cases? Are they included in your related PRs?

ChadKillingsworth:
@brad4d Tests would be dependent on the other PRs. We can sit on this until then if you wish. Or I can add tests as a separate pr after everything lands.

Mainly I needed this functionality on my code base.

brad4d:
@ChadKillingsworth 
I'm fine with submitting it now.
I'm just waiting for an LGTM from @MatrixFrog

brad4d:
This has been submitted internally.

PR #2146, opened 2016-11-13
Remove IIFE wrappers from CommonJS modules that have UMD patterns

ChadKillingsworth:
A rework of #2126.

Frequently modules using a UMD pattern are wrapped in an IIFE to support inclusion as source instead of only a module. [Example](https://github.com/krisk/Fuse/blob/master/src/fuse.js#L784-L800)

To effectively rewrite the module, we need to inline the IIFE before rewriting. The IIFE should be in a module, and should be the only statement in the script.


ChadKillingsworth:
Conflicts resolved - tests passing.

PR #2147, opened 2016-11-13
Fix Travis

ChadKillingsworth:
I'm using this PR to try some tweaks for Travis. I think it's running out of memory.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,11 +3,17 @@ jdk:
   - openjdk7
   - oraclejdk8
 
-script:
-  - mvn javadoc:javadoc
-  - mvn test
+cache:
+  directories:
+  - $HOME/.m2
 
-sudo: false
+install: true
+
+script:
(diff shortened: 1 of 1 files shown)


ChadKillingsworth:
I tried everything I could think of, and lots of tweaks from articles. The GWT compiler is getting killed by the docker host for using too many resources. The only solution I could find was to switch off the container-based build infrastructure at Travis.

Longer build times are far superior to random failures.

Dominator008:
Hmm. It looks like something GWT and Java 7 specific. The Java 8 build's been fine.

ChadKillingsworth:
True - which prompted me to try Oracle's JDK instead of Open JDK. I'll toggle the build a couple more times to see if that's true.

ChadKillingsworth:
Nope. Still flaky.

ChadKillingsworth:
Squashed commits to make this easier

PR #2172, opened 2016-11-25
Fix asssorted compiler internal errors

rgpower:
I am hoping you might consider this pull request, which proposes fixes for various internal compiler errors I noticed when I switched on _newTypeInference_ when using a project that includes [react](https://cdnjs.cloudflare.com/ajax/libs/react/0.14.8/react-with-addons.js) and [react-dom](https://cdnjs.cloudflare.com/ajax/libs/react/0.14.8/react-dom.js) sources.  

I did up a demo github project [/rgpower/nti-fixes-demo](https://github.com/rgpower/nti-fixes-demo) that you can err... check out to see the fixes in action and provides a README which describes the motivations for each change.

Thanks for your time, and for the closure-compiler.

### DIFF
--- a/src/com/google/javascript/jscomp/NewTypeInference.java
+++ b/src/com/google/javascript/jscomp/NewTypeInference.java
@@ -3213,7 +3213,9 @@ final class NewTypeInference implements CompilerPass {
           // retain Namespace instances after GTI instead of turning them into
           // ObjectTypes, and then update those with the summaries and stop
           // specializing here.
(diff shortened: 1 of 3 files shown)


dimvar:
Hi. Thank you for your contribution. This PR is not in a form we can accept. Can you submit separate PRs for each crash you are fixing, and add an associated unit test? See commit 3d325b9d1220fac5eede7aef8e49c3e9248825b7 for an example. Thanks.

PR #2173, opened 2016-11-28
Allow trailing comma in Array Pattern

alexeykuzmin:
It's allowed by ECMAScript 2015 spec.
See "13.3.3 Destructuring Binding Patterns".

The check was added in b0e9123ac763defb149359a9a99f8645b4b07977.
Are there any reasons why it should be there?


blickly:
Hmm, perhaps the final spec differs from the in-progress state back then?
Anyway, thanks for the fix!

alexeykuzmin:
You are welcome )

PR #2190, opened 2016-12-09
Patches needed by splittable.

cramforce:


src/com/google/javascript/jscomp/CompilationLevel.java
```diff
@@ -163,7 +163,7 @@ private static void applyFullCompilationOptions(CompilerOptions options) {
     options.setRemoveAbstractMethods(true);
     options.setRemoveSuperMethods(true);
     options.setReserveRawExports(true);
-    options.setRenamingPolicy(VariableRenamingPolicy.ALL, PropertyRenamingPolicy.ALL_UNQUOTED);
+    options.setRenamingPolicy(VariableRenamingPolicy.ALL, PropertyRenamingPolicy.OFF);
```
ChadKillingsworth: Was this intended to be included? Turning off property renaming isn't something we can just do - it's one of the biggest advantages of `ADVANCED` mode.

src/com/google/javascript/jscomp/ProcessCommonJSModules.java
```diff
@@ -666,7 +666,7 @@ private void visitExport(NodeTraversal t, Node export) {
       }
 
       if (root.getParent().isAssign()
-          && (root.getNext().isName() || root.getNext().isGetProp())
+          && (root.getNext() != null && (root.getNext().isName() || root.getNext().isGetProp()))
```
ChadKillingsworth: What case was this solving for? I'd like to add a test.

src/com/google/javascript/jscomp/CompilationLevel.java
cramforce: I know it must have been discussed many times to add a flag for this. It breaks React's plugin system. I didn't have time yet to take a closer look as to what the exact cause was. It is likely easy to fix on the React side.

cramforce:
I, uhm, didn't actually mean to make this a pull request yet. Well, here we go :)

src/com/google/javascript/jscomp/CompilationLevel.java
cramforce: Ah, just realized I made a PR out of this already. It cannot be this simple, of course. Just wanted to document my local changes for now.

src/com/google/javascript/jscomp/ProcessCommonJSModules.java
cramforce: It failed on this line https://github.com/shama/bel/blob/master/index.js#L1 but I had no idea why.

You will need `splittable` to compile the package for now. I implemented a subset of the `browser` key in `package.json` as an overlay. Is this something you had on your radar?

src/com/google/javascript/jscomp/CompilationLevel.java
ChadKillingsworth: @concavelenz Wondered if that was the case. No worries.

We'd be very surprised if turning off property renaming was safe in this case - because property removal can still happen. So can other breaking optimizations. In general, if it's not safe to rename, then most of the other property optimizations will have similar problems.

src/com/google/javascript/jscomp/ProcessCommonJSModules.java
ChadKillingsworth: > I implemented a subset of the browser key in package.json as an overlay. Is this something you had on your radar?

No it wasn't. You referring to this "spec"? https://github.com/defunctzombie/package-browser-field-spec

cramforce: Yes

ChadKillingsworth:
I'm going to close this since it wasn't intentional in the first place. We can reopen individual changes if needed.

Issue #1811, opened 2016-05-23
Ability to define restricted record types


supersteves:
I'd like to be able to define a record type such as 

``` javascript
{ 
  color: (string|undefined), 
  size: (number|undefined)
}
```

Currently this is somewhat meaningless because it will match against any object since it will allow unknown properties. So, while I want it to match against any of these:

```
{
  color: "#800"
}
{
  color: "#800",
  size: 5
}
{
  size: 5
}
```

... I don't want it to match these, which it does currently:

```
{
  color: "#800",
  size_typo: 5 
}
```

(Background: this is for use defining React props, to allow templated externs for React that allow the createElement params to be validated).

See also https://github.com/google/closure-compiler/issues/1804#issuecomment-221020830

[...]

Issue #1804, opened 2016-05-21
Ability to define loose types (NTI)


supersteves:
I want to be able to define some React props and validate them in the compiler. Some or all of the props are optional. I want the compiler to check them, permit missing props, but complain if unrecognised props are present.

[...]

Issue #1696, opened 2016-03-29
Allow marking @record types as @final to disallow undeclared fields


shicks:
Use case: "keyword" arguments to functions. It would be wonderful to get an error for a typo-ed optional argument key.

```js
/** @record @final */ const ExtraArgs = function() {};
/** @type {!string|undefined} */ ExtraArgs.prototype.foo;
```

Then `{}` and `{foo: 'bar'}` would both match, but `{fop: 'bar'}` would fail to match because the interface is sealed and `fop` isn't a known field.

vobruba-martin:
I would love this change!

concavelenz:
/** @record @struct */ this is intended to disallow accessing properties off the record (or interface) that are not declared.

vobruba-martin:
But it doesn't cover this scenario:

```javascript
/**
 * @record
 * @struct
 */
function MyOpts() {};

/** @type {(undefined|string)} */
MyOpts.prototype.foo;


/**
 * @param {!MyOpts} opts
 */
function doSomething(opts) {
	console.log(opts);
}


doSomething({fop:'will this throw a warning?'}); // no warning

```

supersteves:
Probable dup: https://github.com/google/closure-compiler/issues/1811

PR #2198, opened 2016-12-14
Supporting `@struct` annotation for `@typedef` in NTI

supersteves:
Generating an NTI warning if there are additional unknown properties.

This is a first attempt to solve a requirement of mine and other users - the ability to defined a restricted typedef such that unknown properties generate warnings. This is needed for at least 2 real-world cases of mine:

1. Using React with Closure Compiler, where the createElement takes a props object which we type in externs using a typedef. 
2. Using object literals to rapidly build new instances of types without using classes and the syntactic overhead of designing, instantiating and configuring them.

In both cases a typo in the properties will go undetected without this PR. Here's a React example, showing this PR's approach of annotationg the `@typedef` with `@struct`:
``` javascript
/**
 * @struct @typedef {{
 *      opt: (number|undefined),
 *      mand: string
 * }}
 */
var PropTypeDef;

/** @extends {React.Component<PropTypeDef>} */
class TestPropTypes extends React.Component {
    constructor(props) {
        super(props);
    }
}

// JSX conversion result of <TestPropTypes opt={5} mand="a" bad={true} />:
React.createElement(TestPropTypes, { 
    opt: 5, // can be omitted, but if present, the type must match (existing typecheck)
    mand: "a", // mandatory property, cannot be omitted (existing typecheck)
    bad: true // illegal property, e.g. typo (new typecheck)
});
```
with externs (simplified):
``` javascript
/**
 * @constructor
 * @template P
 * @param {P} props
 */

/** @constructor */
React.Element = function() {};

/**
 * @template PP
 * @param {!function(new:React.Component<PP>, PP)} tag
 * @param {?PP} props
 * @return {React.Element}
 */
React.createElement = function(tag, props) {};
```

In https://github.com/google/closure-compiler/issues/1804#issuecomment-221020830, @dimvar mentioned:

>  It's a grey area in the semantics and we haven't quite decided what we want to do there.

Unfortunately I could not wait, and have gone ahead with this change in a fork. I wanted to share it for others to see even if if is rejected. If you do think this approach (perhaps with some feedback) will be accepted, I'll rebase against the latest tip (this fork is against a month-old tip).

This PR addresses:
* https://github.com/google/closure-compiler/issues/1811
* https://github.com/google/closure-compiler/issues/1696

Also see 
* https://github.com/google/closure-compiler/issues/1804
* https://github.com/google/closure-compiler/issues/1150

### DIFF
--- a/src/com/google/javascript/jscomp/GlobalTypeInfo.java
+++ b/src/com/google/javascript/jscomp/GlobalTypeInfo.java
@@ -1006,7 +1006,7 @@ class GlobalTypeInfo implements CompilerPass, TypeIRegistry {
         return;
       }
       JSDocInfo jsdoc = NodeUtil.getBestJSDocInfo(qnameNode);
-      Typedef td = Typedef.make(jsdoc.getTypedefType());
+      Typedef td = Typedef.make(jsdoc.getTypedefType(), jsdoc.makesStructs());
       currentScope.addTypedef(qnameNode, td);
     }
 
--- a/src/com/google/javascript/jscomp/newtypes/JSTypeCreatorFromJSDoc.java
+++ b/src/com/google/javascript/jscomp/newtypes/JSTypeCreatorFromJSDoc.java
@@ -116,6 +116,11 @@ public final class JSTypeCreatorFromJSDoc {
         "JSC_NTI_CIRCULAR_TYPEDEF_ENUM",
         "Circular typedefs/enums are not allowed");
 
+  public static final DiagnosticType COMPLEX_TYPEDEF_STRUCT =
+    DiagnosticType.warning(
+        "JSC_NTI_COMPLEX_TYPEDEF_STRUCT",
+        "Typedefs with struct are only allowed for simple record types");
+
   public static final DiagnosticType ENUM_WITH_TYPEVARS =
     DiagnosticType.warning(
         "JSC_NTI_ENUM_WITH_TYPEVARS",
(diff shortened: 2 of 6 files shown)


dimvar:
Haven't looked at the diff yet, but it's unlikely that I'll merge this, sorry. The problem with these objects that expect a fixed number of properties is: if you expect a PropTypeDef and you pass some nominal type Foo, the typechecker needs to look up the inheritance chain of Foo to see if there are extra properties. I don't have time at the moment to think carefully about this, so I can't accept the PR.

supersteves:
OK. I probably got that wrong because I only checked own class props. If I changed it to use some guarded call to getAllPropsOfClass(), it should satisfy that. But I won't if you're not going to have time to consider the ramifications. If you do get time for this issue in future, there's some units you can copy, at least.

supersteves:
@dimvar I should add that in the 24 hours of developing in our large codebase since using our forked version (CC with this PR), it's proven itself hugely useful particularly when locking down externs. E.g. a ton of libraries accept 'options' objects in their function args, and this allows you to catch typos. And it works. Deliberately introduce an object literal typo anywhere deep in our codebase and instantly NTI finds it. This increases the sense of security when developing in JS.

If I were to solve the nominal type inherited props issue, would you look at the PR? Or is there anything else of concern? Clearly I'd prefer not to have a long-running fork for something that a custom pass couldn't solve, and would like to contribute this back.

dimvar:
@supersteves TypeScript has the concept of ["fresh" object literals](https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#3115-excess-properties), where basically they check that object literals don't have excess properties, and they don't check it for other objects. This helps them catch typos and also avoids the inheritance-chain issue I mentioned earlier.

Would that behavior work for you? The difference with what you are proposing is that the behavior will be the default for all object literals, you won't need to add `@struct` to the places where you want this checked.

supersteves:
@dimvar Interesting - this seems to be the same - but I can't switch to TypeScript! It's often object literals but there might be cases where you've used Object.assign() and need to tell CC the result type. Also we frequently use typedef to define lightweight interfaces and it's specifically those that need this most. My fork attempts to make this only apply to typedefs, in fact.

dimvar:
I'm not suggesting that you switch to TypeScript :P just that we add a check like this to NTI. I'm considering adding it but I'm not sure yet of all the implications.

supersteves:
Ah I see, a new check which you can opt into, which assumes that all object literals are strict / like my `@struct`. Would this work for cases like the `@typedef` and associated use at the top of this PR?

PR #2200, opened 2016-12-15
Expose the RenamePrefix compiler option as a command line argument.

christiang88:
This allows for consumers of the closure compiler to set the renamePrefix compiler option from the command line interface.


MatrixFrog:
lgtm

ChadKillingsworth:
This doesn't document the flag. We'll need to add it to the dictionaries below if you want it to show up for users when they type `java -jar compiler.jar --help`

MatrixFrog:
I'm okay with that I think. This is a fairly obscure feature.

ChadKillingsworth:
Fine by me - easy to change later

MatrixFrog:
We decided on `--rename_variable_prefix`

PR #2213, opened 2016-12-23
getComputedStyle returns a non-null value

vobruba-martin:
See:
https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle
https://www.w3.org/TR/DOM-Level-2-Style/css.html#CSS-OverrideAndComputed
https://drafts.csswg.org/cssom/#extensions-to-the-window-interface

### DIFF
--- a/externs/browser/gecko_dom.js
+++ b/externs/browser/gecko_dom.js
@@ -896,7 +896,8 @@ BoxObject.prototype.width;
 /**
  * @param {Element} element
  * @param {?string=} pseudoElt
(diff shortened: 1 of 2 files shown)


Dominator008:
Looks like it returns null on Firefox in some cases: https://bugzilla.mozilla.org/show_bug.cgi?id=548397

vobruba-martin:
@Dominator008 Thanks! I've added the link into externs if you want to merge it like this. Otherwise I think we can close this PR.

PR #2220, opened 2017-01-03
Add angular.$q.prototype.resolve definition to 1.5

papadopa:
Per https://code.angularjs.org/1.5.9/docs/api/ng/service/$q, resolve is an alias of when so I just copied the definition here. For some reason, those typedefs do not include several optional parameters but I'll keep this change light just in case there's good reasons to have it like that.


Dominator008:
@papadopa Looks like this requires fixing too many things internally so I'm not sure if it can land :( Maybe you can try migrating to Angular 1.6?

dimvar:
Forgot to reply here. I pointed @papadopa to a related thread internally. Closing this.

PR #2226, opened 2017-01-09
Support symbol primitive

JoelEinbinder:
Most of this code was written with a poor understanding of what the surrounding code was doing. I'm sure that some parts are done incorrectly and that I missed some places where the symbol primitive needs to be taken into account. But all the tests pass and it works for my purposes.

### DIFF
--- a/externs/es3.js
+++ b/externs/es3.js
@@ -30,10 +30,6 @@
 // Some types require them to be part of their definition (such as Array).
 
 
-// TODO(johnlenz): symbol should be a primitive type.
-/** @typedef {?} */
-var symbol;
-
 /**
  * @param {string=} opt_description
  * @return {symbol}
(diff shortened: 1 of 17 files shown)


dimvar:
Hi Joel, apologies but we won't be able to take this PR. Core changes to the type system usually require changes to other google code as well, and need planning. We are not working on native ES6 type checking right now, and we can't deprioritize our other tasks for it. We will be working on native ES6 type checking later this year, and most likely we will only do it for the [new type checker](https://github.com/google/closure-compiler/wiki/Using-NTI-(new-type-inference)), not the old one.

For future reference, a PR is more likely to be merged if it is fixing a bug, because that's uncontroversial. If it is a design contribution, please talk to us first to decide on the design before moving to implementation.

PR #2227, opened 2017-01-09
Handle exceptions when a property can't be inherited

HitkoDev:
Prevent code from breaking on Safari with ```TypeError: Attempting to change attribute of unconfigurable property.```


HitkoDev:
I signed it!

shicks:
This seems reasonable. Do you have an example of a case where this shows up? What is it peculiar to Safari?

ChadKillingsworth:
Closing due to lack of response.

PR #2235, opened 2017-01-13
Better definition of webkitPersistentStorage and webkitTemporaryStorage

vobruba-martin:
I had these methods defined on my own in my project but after the commit 3f385f8f7c4205ba6084aec7000b59d72e19632c I see errors like this:
```
Property queryUsageAndQuota never defined on navigator.webkitPersistentStorage of type Object
```

As you can see `StorageQuota` is well defined in [fileapi.js](https://github.com/google/closure-compiler/blob/master/externs/browser/fileapi.js#L984). Maybe these methods should be moved there but I know you don't like moving things between externs files. :)

### DIFF
--- a/externs/browser/html5.js
+++ b/externs/browser/html5.js
@@ -4165,30 +4165,6 @@ Navigator.prototype.plugins;
  */
 Navigator.prototype.javaEnabled = function() {};
 
-/**
- * @type {Object}
- */
-Navigator.prototype.webkitPersistentStorage;
-/**
- * @param {number} newQuotaInBytes
- * @param {!Function=} opt_quotaCallback
- * @param {!Function=} opt_errorCallback
- *
- * @return {undefined}
- * @see https://developer.chrome.com/apps/offline_storage#method_overview
- */
-Navigator.prototype.webkitPersistentStorage.requestQuota = function(
-    newQuotaInBytes, opt_quotaCallback, opt_errorCallback) {};
-
-/**
- * @param {!Function=} opt_successCallback
- * @param {!Function=} opt_errorCallback
- *
- * @return {undefined}
- * @see https://developer.chrome.com/apps/offline_storage#method_overview
- */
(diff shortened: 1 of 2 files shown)


vobruba-martin:
Forced Travis rebuild...

externs/browser/html5.js
```diff
@@ -4166,29 +4166,17 @@ Navigator.prototype.plugins;
...
 /**
- * @param {!Function=} opt_successCallback
- * @param {!Function=} opt_errorCallback
- *
- * @return {undefined}
- * @see https://developer.chrome.com/apps/offline_storage#method_overview
+ * @type {!StorageQuota}
+ * @see https://developer.chrome.com/apps/offline_storage
  */
-Navigator.prototype.webkitPersistentStorage.queryUsageAndQuota = function(
```
Dominator008: Was this accidentally deleted?

Dominator008: Was this accidentally deleted?

vobruba-martin: No, `queryUsageAndQuota` is a member of `StorageQuota` defined in [fileapi.js](https://github.com/google/closure-compiler/blob/master/externs/browser/fileapi.js#L999)

Dominator008:
Since this change will make `html5.js` depend on `fileapi.js`, I think this list should change:
https://github.com/google/closure-compiler/blob/feb48c375ae3031ac7ff9cf9df14e333f55687a6/src/com/google/javascript/jscomp/DefaultExterns.java#L40

vobruba-martin:
That brings me to my original note. What if we move `Navigator.prototype.webkitPersistentStorage` and `Navigator.prototype.webkitTemporaryStorage` under the `fileapi.js` where I think these properties should belong?

Dominator008:
@vobruba-martin I think we can try that and see what breaks internally :) These are not in the HTML5 standard anyway.

Issue #2211, opened 2016-12-21
goog.scope creates variable masking issues


nicks:
Repro code:
```js
goog.scope(function() {
  var x = function RequestInfo() {}
})
```

Repro link:
http://closure-compiler-debugger.appspot.com/#input0%3Dgoog.scope(function()%2520%257B%250A%2520%2520var%2520x%2520%253D%2520function%2520RequestInfo()%2520%257B%257D%250A%257D)%26input1%26conformanceConfig%26externs%26refasterjs-template%26includeDefaultExterns%3D1%26CHECK_SYMBOLS%3D1%26CHECK_TYPES%3D1%26CLOSURE_PASS%3D1%26MISSING_PROPERTIES%3D1%26PRESERVE_TYPE_ANNOTATIONS%3D1%26PRETTY_PRINT%3D1%26TRANSPILE%3D1

Result:
The compiler outputs the following code:
```js
/** @const */ var $jscomp = {};
/** @const */ $jscomp.scope = {};
$jscomp.scope.x = function RequestInfo() {
};
```

and then emits the warning:
```
input0:2: WARNING - function RequestInfo masks variable (IE bug)
  var x = function RequestInfo() {}
```

I hope you appreciate the irony that the compiler is producing sketchy code that triggers an IE bug, and then complains about the code it produced.

Expected:
I would expect the compiler to namespace RequestInfo when it unboxed the goog.scope, like `$jscomp.RequestInfo` or something.

Alternatively, this is a pretty old IE bug and maybe the warning should just go away.

Dominator008:
The IE bug is probably still present on IE8, so I'm not sure if it can go away :(

brad4d:
I'm not sure why the message says "IE bug", but the warning seems legit to me.
`RequestInfo` is defined in the externs for Fetch API, so using that name does mask a variable.

https://github.com/google/closure-compiler/blob/55cf43ee31e80d89d7087af65b5542aa63987874/externs/browser/fetchapi.js#L155

Dominator008:
@brad4d JScript has some quirkiness with named functions hence the message "IE bug". In general, masking a global variable name with a function name shouldn't give a warning because the function name is scoped to the function.

nicks:
it shouldn't be hard to fix, i can send you a fix if that would make it easier.

Dominator008:
@nicks If `RequestInfo` is scoped to the function, why is it necessary to namespace it?

nicks:
[...]

PR #2238, opened 2017-01-16
Rename bleeding function expressions when lifting them out of a goog.…

nicks:
…scope block. Fixes issue #2211

### DIFF
--- a/src/com/google/javascript/jscomp/ScopedAliases.java
+++ b/src/com/google/javascript/jscomp/ScopedAliases.java
@@ -584,6 +584,17 @@ class ScopedAliases implements HotSwapCompilerPass {
       }
     }
 
+    private void renameBleedingFunctionName(Node fnName) {
+      MakeDeclaredNamesUnique.Renamer renamer =
+          new MakeDeclaredNamesUnique.WhitelistedRenamer(
+              new MakeDeclaredNamesUnique.ContextualRenamer(),
+              ImmutableSet.of(fnName.getString()));
+      renamer.addDeclaredName(fnName.getString(), false);
+
+      MakeDeclaredNamesUnique uniquifier = new MakeDeclaredNamesUnique(renamer);
+      NodeTraversal.traverseEs6(compiler, fnName.getParent().getParent(), uniquifier);
+    }
+
     private void validateScopeCall(NodeTraversal t, Node n, Node parent) {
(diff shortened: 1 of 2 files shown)


nicks:
oops, didn't link the issue properly: https://github.com/google/closure-compiler/issues/2211

src/com/google/javascript/jscomp/ScopedAliases.java
```diff
@@ -658,26 +658,36 @@ public void visit(NodeTraversal t, Node n, Node parent) {
...
+      // If this is a bleeding function expression, like
+      // var x = function y() { ... }
+      // then old versions of IE declare "y" in the current scope. We don't
+      // want the scope unboxing to add "y" to the global scope, so we
+      // need to rename it.
+      //
+      // The simplest solution is to add "y" to forbiddenLocals so it will get
+      // renamed.
+      if (NodeUtil.isBleedingFunctionName(n)) {
+        forbiddenLocals.add(n.getString());
```
Dominator008: Looks like this is going to introduce a lot of artificial "namespace shadow"s, which we currently don't handle very well. I've hit a failure case that looks like:
```js
/** @param {?} g */
function f(g) {}
use(function g() {})
```
Since `g` is a bleeding function name, it will be renamed to `g$jscomp$1` and the code will look like:
```js
/** @param {?} g */
function f(g$jscomp$1) {}
use(function g$jscomp$1() {})
```
which will fail typechecking because the parameter is only renamed in code, but not in the JsDoc.

nicks: long-term, it might make sense to change the renamer to also rename jsdoc params. For now, I just modified this to do a more narrow rename.

nicks:
comments addressed. Thanks for shepherding this change!

Dominator008:
@nicks Thanks. This has been merged internally and will be exported soon.

Dominator008:
Closed by https://github.com/google/closure-compiler/commit/0a83811448f1cb0d6313aceda26204b3c8df6698.

Dominator008:
Unfortunately it got rolled-back again due to a breakage. Will look into it :(

Dominator008:
Closed by https://github.com/google/closure-compiler/commit/bf13e3a23c952c16c14c11dabbf5c5a3db19332f.

PR #2263, opened 2017-01-27
Optional useCapture parameter

adam-rocska:
According to the [W3C Spec](http://www.w3.org/TR/dom/#eventtarget), [mdn](https://developer.mozilla.org/hu/docs/Web/API/EventTarget) and web browser practices, the useCapture parameters of `addEventListener` and `removeEventListener` are optional, and their default value is false. This is an attempt with respect to fix the given extern file.

```
[Exposed=(Window,Worker)]
interface EventTarget {
  void addEventListener(DOMString type, EventListener? callback, optional boolean capture = false);
  void removeEventListener(DOMString type, EventListener? callback, optional boolean capture = false);
  boolean dispatchEvent(Event event);
};

callback interface EventListener {
  void handleEvent(Event event);
};
```

### DIFF
--- a/externs/browser/w3c_event.js
+++ b/externs/browser/w3c_event.js
@@ -35,7 +35,7 @@ function EventTarget() {}
  *
  * @param {string} type
  * @param {EventListener|function(!Event):(boolean|undefined)} listener
- * @param {boolean} useCapture
+ * @param {boolean} [useCapture=false]
  * @return {undefined}
  */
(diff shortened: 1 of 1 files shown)


MatrixFrog:
This will conflict with another PR which is (slowly) making its way toward being submitted.

adam-rocska:
I signed it!

ChadKillingsworth:
Right this is a duplicate of #1809.

PR #2273, opened 2017-02-03
Add the timeout.cancel function from angular 1.5

sbrunner:
Get from https://github.com/google/closure-compiler/blob/master/contrib/externs/angular-1.5.js#L2575-L2589

### DIFF
--- a/contrib/externs/angular-1.6.js
+++ b/contrib/externs/angular-1.6.js
@@ -2562,6 +2562,21 @@ angular.$templateCache;
  */
 angular.$timeout;
 
+/**
+ * Augment the angular.$timeout type definition by reopening the type via an
+ * artificial angular.$timeout instance.
+ *
+ * This allows us to define methods on function objects which is something
+ * that can't be expressed via typical type annotations.
+ *
+ * @type {angular.$timeout}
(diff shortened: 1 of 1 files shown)


mprobst:
@dpurp did we leave out the reopening + `$timeout.cancel` on purpose?

dpurp:
Yes and no.  It was actually moved to [angular-1.6-function-properties.js](https://github.com/google/closure-compiler/blob/master/contrib/externs/angular-1.6-function-properties.js), since it didn't compile with strict checks due to an issue with Closure compiler.  It errored with properties-on-functions, wherein two functions have properties with the same name (`$timeout` and `$interval`).

The issue was fixed internally a few weeks ago, but I have been waiting on a public release of closure.  Since that happened last week, I *should* be merge `function-properties` back into the main file.  I'll look into that today.

sbrunner:
Thanks in advance @dpurp :-)

dpurp:
Exported to github in: https://github.com/google/closure-compiler/commit/3a99f4ddacee6173a2d00e2803612628a79a8407

sbrunner:
Thanks :-)

PR #2275, opened 2017-02-03
Prevent NO_TRANSPILE being set as input language

oxc:
The #setLanguage(LanguageMode) method has a check that NO_TRANSPILE is not set as input language, but it checked the instance variable instead of checking the method parameter.


blickly:
Nice catch, thanks!

PR #2276, opened 2017-02-03
add webkitFullscreenEnabled property

nederland074:
webkitFullscreenEnabled property was missing in the externs/browser/html5.js file


nederland074:
I signed it!

Issue #2087, opened 2016-10-15
PolymerPass and Polymer 2.0


ChadKillingsworth:
Polymer 2.0 changes for the compiler look to be straightforward. The one stumbling block I see is detecting whether a class extends `Polymer.Element`. Ideally the PolymerPass would run after type checking and therefore be able to simply know this. However since it also adds type information needed by type inference, it also needs to run beforehand.

Possible solutions:
1. Use a special annotation like `@PolymerElement`. This is the most simple solution - but it feels like we should be able to do better.
2. Use a heuristic to detect a Polymer class by non-type means. This could include checking for a static `is` getter or a `config` getter. Or maybe even just checking to see if the class is referenced by a `customElements.define` call.
3. Ask the rest of the jscomp team if they can think of a way for the PolymerPass to participate in the "middle" of type checking.

cc @justinfagnani, @samthor

ChadKillingsworth:
What's needed here is a way to extend the type inference system. Since that doesn't really exist, I'm planning on implementing a custom jsdoc annotation. However I hope that the annotation is only needed for classes which indirectly extend `Polymer.Element`.

jklein24:
Hey Chad, feel free to cc @MatrixFrog and myself on a PR when you get there. Your analysis seems pretty right-on here. I chatted briefly with a couple folks on the Polymer team about this exact problem when 2.0 was announced. Tyler may have some thoughts about how we can make this work.

ChadKillingsworth:
@jklein24 Will do. The proposed methods for mixins (behaviors) will also be challenging but I'm saving that for a separate issue as it's not Polymer specific.

PR #2277, opened 2017-02-03
Add basic support for Polymer2 elements

ChadKillingsworth:
Part one of addressing #2087 

This adds basic rewriting support for Polymer2 classes. This should support most `SIMPLE` mode uses, but does not yet support `ADVANCED` mode for Polymer2 elements.

Notes:

 * Synthetic externs are not used
 * Class names are not forced to be globally defined
 * Only supports classes which directly extend `Polymer.Element`. Other cases will be added in a separate PR.
 * Polymer2 elements do not currently require extern definitions for the compiler.

cc @jklein24, @MatrixFrog, @justinfagnani

### DIFF
--- a/src/com/google/javascript/jscomp/NodeUtil.java
+++ b/src/com/google/javascript/jscomp/NodeUtil.java
@@ -2213,9 +2213,31 @@ public final class NodeUtil {
     return null;
   }
 
-  /**
-   * @return The first computed property in the objlit whose key matches {@code key}.
-   */
+  /** @return The first getter in the class members that matches the key. */
+  @Nullable
+  static Node getFirstGetterMatchingKey(Node n, String keyName) {
+    Preconditions.checkState(n.isClassMembers() || n.isObjectLit());
+    for (Node keyNode : n.children()) {
+      if (keyNode.isGetterDef() && keyNode.getString().equals(keyName)) {
+        return keyNode;
+      }
+    }
+    return null;
+  }
+
+  /** @return The first setter in the class members that matches the key. */
+  @Nullable
+  static Node getFirstSetterMatchingKey(Node n, String keyName) {
+    Preconditions.checkState(n.isClassMembers() || n.isObjectLit());
+    for (Node keyNode : n.children()) {
+      if (keyNode.isSetterDef() && keyNode.getString().equals(keyName)) {
+        return keyNode;
+      }
+    }
+    return null;
+  }
+
+  /** @return The first computed property in the objlit whose key matches {@code key}. */
   @Nullable
   static Node getFirstComputedPropMatchingKey(Node objlit, Node key) {
     Preconditions.checkState(objlit.isObjectLit());
--- a/src/com/google/javascript/jscomp/PolymerClassDefinition.java
+++ b/src/com/google/javascript/jscomp/PolymerClassDefinition.java
@@ -33,6 +33,13 @@ import javax.annotation.Nullable;
  * class.
  */
 final class PolymerClassDefinition {
+  static enum DefinitionType {
+    ObjectLiteral,
+    ES6Class
+  }
+
+  /** The declaration style used for the Polymer definition */
+  final DefinitionType defType;
 
   /** The target node (LHS) for the Polymer element definition. */
   final Node target;
@@ -44,18 +51,19 @@ final class PolymerClassDefinition {
   final MemberDefinition constructor;
 
   /** The name of the native HTML element which this element extends. */
-  final String nativeBaseElement;
+  @Nullable final String nativeBaseElement;
 
   /** Properties declared in the Polymer "properties" block. */
   final List<MemberDefinition> props;
 
   /** Flattened list of behavior definitions used by this element. */
-  final ImmutableList<BehaviorDefinition> behaviors;
+  @Nullable final ImmutableList<BehaviorDefinition> behaviors;
 
   /** Language features that should be carried over to the extraction destination. */
-  final FeatureSet features;
+  @Nullable final FeatureSet features;
 
   PolymerClassDefinition(
+      DefinitionType defType,
       Node target,
       Node descriptor,
       JSDocInfo classInfo,
@@ -64,8 +72,9 @@ final class PolymerClassDefinition {
       List<MemberDefinition> props,
       ImmutableList<BehaviorDefinition> behaviors,
       FeatureSet features) {
+    this.defType = defType;
     this.target = target;
-    Preconditions.checkState(descriptor.isObjectLit());
+    Preconditions.checkState(descriptor == null || descriptor.isObjectLit());
     this.descriptor = descriptor;
     this.constructor = constructor;
     this.nativeBaseElement = nativeBaseElement;
@@ -134,7 +143,9 @@ final class PolymerClassDefinition {
       overwriteMembersIfPresent(allProperties, behavior.props);
     }
     overwriteMembersIfPresent(
-        allProperties, PolymerPassStaticUtils.extractProperties(descriptor, compiler));
+        allProperties,
+        PolymerPassStaticUtils.extractProperties(
+            descriptor, DefinitionType.ObjectLiteral, compiler));
 
     FeatureSet newFeatures = null;
     if (!behaviors.isEmpty()) {
@@ -145,6 +156,7 @@ final class PolymerClassDefinition {
     }
 
     return new PolymerClassDefinition(
+        DefinitionType.ObjectLiteral,
         target,
         descriptor,
         classInfo,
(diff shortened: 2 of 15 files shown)


ChadKillingsworth:
One note: read-only property setters are intentionally not supported as of yet. I'm still trying to come up with a valid strategy for them that supports renaming.

justinfagnani:
Awesome @ChadKillingsworth! @MatrixFrog and I are going to get together next week to go though how to run a patch through the Google internal tests, so we can probably start with this one.

brad4d:
@MatrixFrog, does the comment from @justinfagnani above indicate this PR should be assigned to you or him?

MatrixFrog:
Me, for now, and we'll work together on importing it on Tuesday.

test/com/google/javascript/jscomp/PolymerPassTest.java
```diff
@@ -184,6 +186,19 @@ public void testVarTarget() {
             "X = Polymer(/** @lends {X.prototype} */ {",
             "  is: 'x-element',",
             "});"));
+
+    disableTypeCheck();
```
justinfagnani: Add a comment that the type checker doesn't understand ES6 and to remove when it does

test/com/google/javascript/jscomp/PolymerPassTest.java
```diff
@@ -201,6 +216,19 @@ public void testLetTarget() {
             " */",
             "var X = function() {};",
             "X = Polymer(/** @lends {X.prototype} */ {is:'x-element'});"));
+
+    disableTypeCheck();
```
justinfagnani: here too

test/com/google/javascript/jscomp/PolymerPassTest.java
```diff
@@ -184,6 +186,19 @@ public void testVarTarget() {
             "X = Polymer(/** @lends {X.prototype} */ {",
             "  is: 'x-element',",
             "});"));
+
+    disableTypeCheck();
+    testEs6(
```
justinfagnani: use `testSameEs6`

test/com/google/javascript/jscomp/PolymerPassTest.java
```diff
@@ -201,6 +216,19 @@ public void testLetTarget() {
             " */",
             "var X = function() {};",
             "X = Polymer(/** @lends {X.prototype} */ {is:'x-element'});"));
+
+    disableTypeCheck();
+    testEs6(
```
justinfagnani: use `testSameEs6`

justinfagnani: and below...

src/com/google/javascript/jscomp/PolymerClassDefinition.java
```diff
@@ -33,6 +33,8 @@
  * class.
  */
 final class PolymerClassDefinition {
+  /** Either 1 or 2 - the Polymer version of the definition */
+  final int version;
```
ChadKillingsworth: Polymer 2 elements created with the `Polymer()` function are treated as Polymer 1 elements by the pass. They are only the Hybrid elements, correct? So behaviors for instance should be supported?

There is a significant difference in how the pass is going to treat the class based syntax elements. Namely I'm trying to make them fully work with renaming and obey the standard compiler rules regarding quoted properties. While this same behavior can be made to work with the 1.0 version elements, landing that change inside Google would probably be quite difficult.

I'm changing this to the enum - that makes sense. Just wanted you to be aware that the two are definitely not treated equivalently right now.

src/com/google/javascript/jscomp/NodeUtil.java
```diff
@@ -2212,9 +2212,31 @@ static Node getFirstPropMatchingKey(Node n, String keyName) {
     return null;
   }
 
-  /**
-   * @return The first computed property in the objlit whose key matches {@code key}.
-   */
+  /** @return The first getter in the class members that matches the key. */
+  @Nullable
+  static Node getFirstGetterMatchingKey(Node n, String keyName) {
+    Preconditions.checkState(n.isClassMembers());
```
ChadKillingsworth: There is - simple change.

src/com/google/javascript/jscomp/PolymerPass.java
```diff
@@ -89,15 +84,22 @@ public void hotSwapScript(Node scriptRoot, Node originalRoot) {
...
     if (isPolymerCall(node)) {
-      rewriteClassDefinition(node, parent, traversal);
+      if (polymerElementExterns != null) {
+        rewritePolymer1ClassDefinition(node, parent, traversal);
+      } else if (!warnedPolymer1ExternsMissing) {
+        compiler.report(JSError.make(polymerElementExterns, POLYMER_MISSING_EXTERNS));
+        warnedPolymer1ExternsMissing = true;
+      }
+    } else if (isPolymerClass(node)) {
+      rewritePolymer2ClassDefinition(node, parent, traversal);
```
ChadKillingsworth: Currently for the 2.0 (class based elements), the compiler doesn't need externs at this point. The 1.0 version of the pass heavily depended on them, but I'm trying to plan for the case where Polymer 2.0 can be used as either source or externs.

If the Polymer externs or source definitions are omitted, the standard compiler passes that come later will highlight the problem.

src/com/google/javascript/jscomp/PolymerPass.java
```diff
@@ -171,9 +183,16 @@ public static boolean isPolymerCall(Node value) {
     return value != null && value.isCall() && value.getFirstChild().matchesQualifiedName("Polymer");
   }
 
-  /**
-   * Any member of a Polymer element or Behavior. These can be functions, properties, etc.
-   */
+  /** @return Whether the class extends Polymer.Element */
+  @VisibleForTesting
+  public static boolean isPolymerClass(Node value) {
```
ChadKillingsworth: Done. Moving the `isPolymerCall` method immediately proceeding as well.

src/com/google/javascript/jscomp/PolymerPass.java
```diff
@@ -171,9 +183,16 @@ public static boolean isPolymerCall(Node value) {
...
+    return value != null
+        && value.isClass()
+        && !value.getSecondChild().isEmpty()
+        && value.getSecondChild().matchesQualifiedName("Polymer.Element");
```
ChadKillingsworth: Yes. I actually have the next large chunk of work done which includes this, but I am trying to limit the size of the PRs.

src/com/google/javascript/jscomp/PolymerClassDefinition.java
justinfagnani: Polymer()-created elements do not need to be hybrid - they could use Polymer 2.0 only features, or Web Components v1 features. They are "legacy" in that they have another layer of 1.x back-compat APIs added to them.

Ack on the difference between renaming ability, but also, there's a chance we'll introduce a new class-based syntax for 1.x for better forward compatibility with 2.x, in which case Polymer() vs class, and renaming vs. not might be completely orthogonal.

src/com/google/javascript/jscomp/PolymerPass.java
justinfagnani: 👍

src/com/google/javascript/jscomp/PolymerClassDefinition.java
justinfagnani: I wonder if we need a way to globally detect the Polymer version so the concerns are fully separated.

ChadKillingsworth: That would definitely make things clearer. We could change the flag to be `--polymer_pass=VERSION`.

justinfagnani: yeah, there can be only one version of Polymer in a program, so that makes things very clear

ChadKillingsworth: I can change the open source flags to indicate this, but you'll have to handle the internal version.

That change will take me a bit longer than the rest (which will be pushed in just a few minutes).

justinfagnani: @MatrixFrog is recommending a new flag name, like `--polymer_version=VERSION` and for a while `--polymer_pass` will be an alias for `--polymer_version=1`

ChadKillingsworth: That's because @MatrixFrog knows how difficult it would be to land that change :) It's a good plan - I'll go with that.

ChadKillingsworth:
I think all the comments have been addressed (except the flag change - I may defer that one). Per normal, commits have been squashed for importing.

ChadKillingsworth:
New flag issue created: #2288. I've got a couple of questions listed in there too.

ChadKillingsworth:
Updated to use the new `--polymerVersion` flag. Renaming rules are now based off the Polymer Version. Tests now run against both versions.

This also now adds synthetic externs for the read only properties in Polymer 2. After discussion, there is no way to coordinate the renaming of `element.readOnlyProp` and `element._setReadOnlyProp`.

ChadKillingsworth:
@justinfagnani Where are we on this?

ChadKillingsworth:
Updated to include the change from `static get config() {}` to `static get properties() {}`

ChadKillingsworth:
This PR has been superseded by #2338

PR #2279, opened 2017-02-03
Always inline IIFEs

ChadKillingsworth:
This is a rollforward of 0660acdb. I fully expect this to be rolled back again, but we don't know what cases triggered the original rollback.

IIFEs blocking optimizations are frequently coming up - and they are generated by Typescript.

 * #2100 
 * #2103 
 * http://stackoverflow.com/a/41986846/1211524

If someone can get me test cases that break, I'll update this optimization to address them.

### DIFF
--- a/src/com/google/javascript/jscomp/NodeUtil.java
+++ b/src/com/google/javascript/jscomp/NodeUtil.java
@@ -2362,6 +2362,14 @@ public final class NodeUtil {
         && n.getFirstChild().isCall();
   }
 
+  static boolean isIIFE(Node n) {
+    Node parent = n.getParent();
+    return n.isFunction()
+        && parent.isCall()
+        && parent.getFirstChild() == n
+        && parent.getParent().isExprResult();
+  }
+
   static boolean isVanillaFunction(Node n) {
     return n.isFunction() && !n.isArrowFunction();
   }
--- a/test/com/google/javascript/jscomp/InlineFunctionsTest.java
+++ b/test/com/google/javascript/jscomp/InlineFunctionsTest.java
@@ -802,7 +802,7 @@ public final class InlineFunctionsTest extends CompilerTestCase {
          "function foo(){return a}" +
          "(function(){var a=5;(function(){foo()})()})()",
          "var a=3;" +
-         "{var a$jscomp$inline_0=5;{a}}"
+         "{var a$jscomp$inline_1=5;{a}}"
          );
 
     assumeMinimumCapture = true;
@@ -899,7 +899,7 @@ public final class InlineFunctionsTest extends CompilerTestCase {
          "function foo(){return a}" +
          "(function(){var a=5;(function(){foo()})()})()",
          "var a=3;" +
-         "{var a$jscomp$inline_0=5;{a}}"
+         "{var a$jscomp$inline_1=5;{a}}"
          );
 
     assumeMinimumCapture = true;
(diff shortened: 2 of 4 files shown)


MatrixFrog:
I tested this out a while ago. If a Googler wants to look at the test failures and try to boil them down to minimal test cases they can ping me.

brad4d:
I can't give this much of my time, but I'm willing to import it and see what breaks.

ChadKillingsworth:
One thought would be to restrict the "always inline" directive to the global scope. I'm guessing that nested IIFEs are what are causing the problem.

MatrixFrog:
I already imported this and saw what breaks. I just didn't have time to investigate *why* things broke.

brad4d:
Import done & standard compiler tests pass.
Next level of testing initiated...

brad4d:
First attempt at doing wide testing failed due to a testing system problem related to having too many tests to run.
Trying again with a bit smaller set of tests.

brad4d:
@ChadKillingsworth 
There seem to be a lot of cosmetic changes in this PR.
Could you remove those? They make it hard to see the important functional changes.
Thanks.

ChadKillingsworth:
@brad4d I'm guessing that's from rolling forward a commit from 8 months ago. I'll re-do the changes from scratch.

ChadKillingsworth:
@brad4d Rebased and kept the test formatting changes to a minimum - which means I didn't run the Google formatting tool on this commit.

src/com/google/javascript/jscomp/NodeUtil.java
```diff
@@ -2362,6 +2362,14 @@ public static boolean isExprCall(Node n) {
         && n.getFirstChild().isCall();
   }
 
+  static boolean isIIFE(Node n) {
+    Node parent = n.getParent();
+    return n.isFunction()
+        && parent.isCall()
+        && parent.getFirstChild() == n
+        && parent.getParent().isExprResult();
```
brad4d: For my own information, why does it matter if the grandparent is an ExprResult?
Perhaps a comment giving a bad case we're trying to avoid would make this clearer.

test/com/google/javascript/jscomp/InlineFunctionsTest.java
```diff
@@ -1589,8 +1589,11 @@ public void testComplexFunctionWithFunctionDefinition2a() {
     assumeMinimumCapture = false;
 
     // Don't inline if local names might be captured.
```
brad4d: I don't think this comment is correct anymore.

test/com/google/javascript/jscomp/InlineFunctionsTest.java
```diff
@@ -2288,7 +2290,9 @@ public void testAnonymous2() {
   public void testAnonymous3() {
     // Introducing a new value into is tricky
```
brad4d: Would you be able to fix this comment so it makes sense?

brad4d:
@ChadKillingsworth 
Thanks for your updates.
My review comments are all cosmetic, so I'm going to try again to get some useful test results without waiting for another update from you.

src/com/google/javascript/jscomp/NodeUtil.java
ChadKillingsworth: I think it's avoiding this case:

```js
var foo = (function() {
  return some.dynamically.calculated.value;
})();
```

brad4d:
FYI, I am actually making slow progress on this.
I'm being extra careful since this change could introduce runtime failures rather than just compilation errors.

Currently I have 33 test targets that failed or were broken on my last attempt.
Likely many of these are flaky cases. I'm re-trying them to weed those out.

brad4d:
Sent for internal review.
All failing or broken tests either succeeded on rerun or were found to be already broken/failing without this change.

brad4d:
Submitting this internally.
Fingers crossed...

brad4d:
The change submitted successfully internally, and no one has screamed yet, so this PR should be closed by Monday's push to github.

ChadKillingsworth:
And now we wait ...

brad4d:
Had to roll this back temporarily. See #2408
I'll roll it forward after the fix for that lands.

Issue #1693, opened 2016-03-29
Extending ES6 class with static method leads to undefined function call


SunboX:
## sample code

```js
// ==ClosureCompiler==
// @compilation_level ADVANCED_OPTIMIZATIONS
// @output_file_name default.js
// @language ECMASCRIPT6_STRICT
// @language_out ES5_STRICT
// @formatting pretty_print
// ==/ClosureCompiler==

class BaseClass {
  static run() {
    console.log(this._isBaseClass() ? 'base class' : 'extended class');
  }

  static _isBaseClass() {
    return true;
  }
}

class ExtendedClass extends BaseClass {
  static _isBaseClass() {
    return false;
  }
}

class ClassImpl extends ExtendedClass {
  static doRun() {
    this.run();
  }
}

ClassImpl.doRun();
```

## demo link

[...]

PR #2286, opened 2017-02-07
Prevent renaming of Polymer 2 static properties

ChadKillingsworth:
Polymer 2 classes define static getters. When Polymer is used as an external library, these properties must not be renamed. Since the type system doesn't model static properties on a constructor currently, this pass adds `is` and `config` properties to `Function.prototype`. This will block renaming of any `is` or `config` property when defined on a function.

cc @justinfagnani

### DIFF
--- a/src/com/google/javascript/jscomp/DefaultPassConfig.java
+++ b/src/com/google/javascript/jscomp/DefaultPassConfig.java
@@ -435,6 +435,12 @@ public final class DefaultPassConfig extends PassConfig {
       checks.add(generateExports);
     }
 
+    if (options.polymerVersion != null
+        && options.polymerVersion > 1
+        && options.propertyRenaming != PropertyRenamingPolicy.OFF) {
+      checks.add(polymerProtectStaticProperties);
+    }
+
     checks.add(createEmptyPass("afterStandardChecks"));
 
     assertAllOneTimePasses(checks);
@@ -3068,6 +3074,15 @@ public final class DefaultPassConfig extends PassConfig {
         @Override
         protected FeatureSet featureSet() {
           return ES8;
+        }
+      };
+
+  /** Prevents renaming of static properties on Polymer Element classes */
+  private final PassFactory polymerProtectStaticProperties =
+      new PassFactory("polymerProtectStaticProperties", true) {
+        @Override
+        protected CompilerPass create(AbstractCompiler compiler) {
+          return new PolymerProtectStaticProperties(compiler);
         }
       };
 
--- a/src/com/google/javascript/jscomp/PolymerProtectStaticProperties.java
+++ b/src/com/google/javascript/jscomp/PolymerProtectStaticProperties.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright 2017 The Closure Compiler Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.javascript.jscomp;
+
+import com.google.javascript.rhino.IR;
+import com.google.javascript.rhino.Node;
+
+/**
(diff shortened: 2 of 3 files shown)


src/com/google/javascript/jscomp/DefaultPassConfig.java
```diff
@@ -429,6 +429,10 @@ private void addOldTypeCheckerPasses(
       checks.add(generateExports);
     }
 
+    if (options.polymerPass && options.propertyRenaming != PropertyRenamingPolicy.OFF) {
```
justinfagnani: does this need to be updated for the new flag names?

test/com/google/javascript/jscomp/PolymerProtectStaticPropertiesTest.java
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright 2008 The Closure Compiler Authors.
```
justinfagnani: 2017?

src/com/google/javascript/jscomp/DefaultPassConfig.java
ChadKillingsworth: Yes - just hadn't circled back since that landed.

ChadKillingsworth:
Changes made - commits squashed.

ChadKillingsworth:
Update to reflect the changes in Polymer/polymer/pull/4332

src/com/google/javascript/jscomp/PolymerProtectStaticProperties.java
```diff
@@ -0,0 +1,88 @@
...
+    compiler.reportCodeChange();
+  }
+
+  /** Lazily create a "new" externs root for undeclared variables. */
+  private Node getSynthesizedExternsRoot() {
+    return compiler.getSynthesizedExternsInput().getAstRoot(compiler);
+  }
+
+  /** Traverse the externs to find references to Polymer.Element. */
+  class FindPolymerExterns implements NodeTraversal.Callback {
```
justinfagnani: Why doesn't the main PolymerPass need to find externs?

src/com/google/javascript/jscomp/PolymerProtectStaticProperties.java
```diff
@@ -0,0 +1,88 @@
...
+    this.compiler = compiler;
+  }
+
+  @Override
+  public void process(Node externs, Node root) {
+    FindPolymerExterns findPolymerExterns = new FindPolymerExterns();
+    NodeTraversal.traverseEs6(compiler, externs, findPolymerExterns);
+
+    if (findPolymerExterns.hasPolymerElementExterns()) {
+      addExtern("is");
```
justinfagnani: We may add optional mixins that support `hostAttributes` and `listener` static properties. How should we handle those? The set of important statics is nearly as open as the set of instance methods now. Is there any way the _users_ of these properties can protect them or allow for renaming?

That seems to be what this is attempting: https://github.com/Polymer/polymer/blob/master/lib/mixins/element-mixin.html#L141

Is this not sufficient?

src/com/google/javascript/jscomp/PolymerProtectStaticProperties.java
ChadKillingsworth: Currently, only Polymer 1 version passes is making use of the externs directly. Checking for the extern being present is already done by other passes in the compiler - namely you'll have type annotations for unknown symbols if they are missing.

Also, the long term plan was that Polymer 2 would be included in the compilation - not an extern.

src/com/google/javascript/jscomp/PolymerProtectStaticProperties.java
ChadKillingsworth: No - it's not. With the class based style, the compiler does MUCH better at understanding the syntax and fully enables optimizations. I think the real answer is https://github.com/google/closure-compiler/issues/1693#issuecomment-303981793 but I don't know what it would take to land such a change at Google.

justinfagnani:
Hey @ChadKillingsworth can we get a rebase? :)

ChadKillingsworth:
@justinfagnani Done.

FYI This pass is a temporary workaround. I have a general plan with consensus of the compiler team on how to handle this in the general case for ES6 static properties and methods.

justinfagnani:
cc @azakus 

Would this PR be merged, even temporarily?

ChadKillingsworth:
Yes I would merge it temporarily. The changes needed for static properties might take just a bit for me to get worked out, then could take a while to get landed. After static properties are handled properly, we'll just remove this pass.

TL;DR - I have to ensure that the disambiguate and ambiguate properties passes are updated to account for static inheritance. Collapse Properties needs updated to not collapse static properties from ES6 classes.

azakus:
@ChadKillingsworth is this PR still necessary after #2338?

PR #2292, opened 2017-02-09
Convert the --polymer_pass flag from a boolean to an optional integer

### LINKED ISSUE #2288: Add a flag to specify the polymer version
From discussion: https://github.com/google/closure-compiler/pull/2277#discussion_r99972133

Add a `--polymer_pass_version=X` flag so that the Polymer pass can have different behaviors between Polymer 1.x and 2.x. Since both the `Polymer({})` style and `class FooElement extends Polymer.Element {}` will potentially be supported in both versions, the compiler cannot infer this on it's own.

The 1.x version disables renaming (using synthetic externs) of any declared property. 2.x is going to follow the ALL_UNQUOTED property renaming policy.

How are behaviors handled? Currently behaviors are supported only by the `Polymer({})` style call for hybrid elements when using the 2.x version of Polymer.

* Should behaviors be supported by any potential class syntax that might be added to Polymer 1?
* Are Polymer 2 elements using the class style declaration ever going to support behaviors (I'm really hoping this answer is "no")?

ChadKillingsworth:
Convert the `--polymer_pass` flag from a boolean to an optional integer. When used without a value (like a boolean flag), the flag defaults to "1".

Example: `--polymer_pass` is equivalent to `--polymer_pass=1`

Also switches the CompilerOption polymerPass field to polymerVersion.

This allows users to specify whether they want Polymer 1 or Polymer 2 behavior.

I'm not sure if something similar can be done with the internal command line runner or not.

Fixes #2288

### DIFF
--- a/src/com/google/javascript/jscomp/CommandLineRunner.java
+++ b/src/com/google/javascript/jscomp/CommandLineRunner.java
@@ -532,10 +532,13 @@ public class CommandLineRunner extends
         + "annotated with @ngInject")
     private boolean angularPass = false;
 
-    @Option(name = "--polymer_pass",
-        handler = BooleanOptionHandler.class,
-        usage = "Rewrite Polymer classes to be compiler-friendly.")
-    private boolean polymerPass = false;
+    @Option(
+      name = "--polymer_pass",
+      handler = OptIntegerOptionHandler.class,
+      usage =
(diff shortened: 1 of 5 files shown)


MatrixFrog:
I already have a fix internally. That's why I added the "work in progress" label.

ChadKillingsworth:
Sorry - I should have asked. I saw the label but didn't pick up on that's what it meant.

PR #2309, opened 2017-02-15
Remove <string> from Object.entries() return type

tjlav5:
Going through https://tc39.github.io/ecma262/#sec-object.entries, I don't see where the <string> type is coming from, or why it should be appended to the template type.


dimvar:
The return type of this method is an array of 2-element tuples. But we can't express tuple types in closure compiler, so the closest to that is `Array<(T|string)>`.

PR #2331, opened 2017-02-22
Allow suppressing of conformance violations

vobruba-martin:
Currently `@suppress {conformanceViolations}` is working but it raises
another warning: `Parse error. unknown @suppress parameter:
conformanceViolations`

### DIFF
--- a/src/com/google/javascript/jscomp/parsing/ParserConfig.properties
+++ b/src/com/google/javascript/jscomp/parsing/ParserConfig.properties
@@ -152,6 +152,7 @@ jsdoc.suppressions =\
     checkTypes,\
     checkVars,\
     closureDepMethodUsageChecks,\
+    conformanceViolations,\
     const,\
     constantProperty,\
     deprecated,\


dimvar:
In your conformance config, you can whitelist files where you don't want the conformance checks to apply. Why do you need the suppression?

vobruba-martin:
The file is quite large (3000+ lines) and I don't want to whitelist the whole file because of one line.

dimvar:
Makes sense. I'll bring this in.

vobruba-martin:
Thanks!

concavelenz:
As with previous discussions, conformance violations aren't intended to be suppressible.  If we want suppressible rules we need to add a "lint" or "warning" category.  I'm recommending a rollback here.

dimvar:
Hey Martin, turns out I was missing some details here. The conformance checks are supposed to be enforced project-wide, which is why we didn't allow use with `@suppress` in the past. To whitelist individual warnings without whitelisting the whole file, you can use a whitelist file with the option `--warnings_whitelist_file`. It's a bit more hassle to set up than `@suppress`, but it will do what you want.

vobruba-martin:
@dimvar Thanks for the explanation but I still don't understand why am I able to whitelist by file or even by file:line but I'm not able to whitelist a block of code. Whitelisting by file:line isn't really helpful in my case because the line is changing with almost every commit.

MatrixFrog:
dimvar@ added an explanatory comment yesterday but I'm having some trouble with MOE which is why you don't see it on github yet:

```
  // This diagnostic group is intentionally absent in ParserConfig.properties.
  // Conformance checks are supposed to be enforced project-wide, so we don't
  // allow suppressions on individual functions.
  // In the future, we may carve out a subset of the conformance checks that is
  // OK to suppress.
  // For now, the only way to suppress a check at a granularity smaller than
  // the file level is by using a whitelist file.
```

Conformance checks are often used to limit use of dangerous APIs like the `innerHTML` property of HTML elements. You want to make sure that no one is using it except for a few spots that have been very carefully reviewed to make sure there's no risk of XSS. Allowing `@suppress` means you can't really enforce that. On the other hand some conformance checks are less crucial from a security standpoint which is why we may carve out a suppressible subset in the future.

dimvar:
> Whitelisting by file:line isn't really helpful in my case because the line is changing with almost every commit.

Line numbers are ignored when comparing warnings against a whitelist.

PR #2332, opened 2017-02-22
Update properties of PipeOptions in Streams API

vobruba-martin:
See https://streams.spec.whatwg.org/#rs-pipe-to

### DIFF
--- a/externs/browser/streamsapi.js
+++ b/externs/browser/streamsapi.js
@@ -54,13 +54,13 @@ TransformStream.prototype.readable;
  */
 function PipeOptions() {};
 
-/** @type {boolean} */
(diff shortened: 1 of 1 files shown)


MatrixFrog:
What's the use case for setting these to undefined instead of false?

vobruba-martin:
Without this change you have to do this
```javascript
/** @type {!PipeOptions} */
var options = {
    preventClose: true,
    preventAbort: false,
    preventCancel: false
};
```
instead of just this
```javascript
/** @type {!PipeOptions} */
var options = {
    preventClose: true
};
```

MatrixFrog:
Cool. Thanks!

PR #2334, opened 2017-02-24
createImageData and getImageData always return ImageData

dsanders11:
My reading of the HTML5 spec is that these two methods will always return an `ImageData` if they didn't throw an exception, so `null` isn't a valid return type. Relevant parts of the spec quoted and linked below (emphasis mine).

> [When the createImageData() method is invoked with two numeric arguments sw and sh, **it must create an ImageData object**, with parameter pixelsPerRow set to the absolute magnitude of sw, and parameter rows set to the absolute magnitude of sh. Initialize the image data of the new ImageData object to transparent black. If both sw and sh are non-zero, then return the new ImageData object. If one or both of sw and sh are zero, then throw an "IndexSizeError" DOMException instead.](https://html.spec.whatwg.org/multipage/scripting.html#dom-context-2d-createimagedata)

> [The getImageData(sx, sy, sw, sh) method, when invoked, must, if either the sw or sh arguments are zero, throw an "IndexSizeError" DOMException; otherwise, if the CanvasRenderingContext2D's origin-clean flag is set to false, it must throw a "SecurityError" DOMException; otherwise, **it must create an ImageData object**, with parameter pixelsPerRow set to sw, and parameter rows set to sh.](https://html.spec.whatwg.org/multipage/scripting.html#dom-context-2d-getimagedata)

### DIFF
--- a/externs/browser/html5.js
+++ b/externs/browser/html5.js
@@ -478,7 +478,7 @@ CanvasRenderingContext2D.prototype.drawImage = function(
 /**
  * @param {number} sw
  * @param {number} sh
- * @return {ImageData}
+ * @return {!ImageData}
  * @nosideeffects
  */
 CanvasRenderingContext2D.prototype.createImageData = function(sw, sh) {};
@@ -488,7 +488,7 @@ CanvasRenderingContext2D.prototype.createImageData = function(sw, sh) {};
  * @param {number} sy
  * @param {number} sw
  * @param {number} sh
- * @return {ImageData}
+ * @return {!ImageData}
  * @throws {Error}
  */
 CanvasRenderingContext2D.prototype.getImageData = function(sx, sy, sw, sh) {};


dsanders11:
CLA signed to make @googlebot happy.

dsanders11:
For further color, here's an example of where it causes a type warning that doesn't really exist:

```
// ==ClosureCompiler==
// @compilation_level ADVANCED_OPTIMIZATIONS
// @output_file_name default.js
// ==/ClosureCompiler==

/**
 * @param {!ImageData} imgData required ImageData
 */
function requiresImageData(imgData) { }

var canvas = /** @type {HTMLCanvasElement} */ (document.createElement('canvas'));
var context = /** @type {CanvasRenderingContext2D} */ (canvas.getContext('2d'));
requiresImageData(context.getImageData(0, 0, 200, 200));
```

Warning:

```
JSC_TYPE_MISMATCH: actual parameter 1 of requiresImageData does not match formal parameter
found   : (ImageData|null)
required: ImageData at line 8 character 18
requiresImageData(context.getImageData(0, 0, 200, 200));
```

brad4d:
importing and testing..

brad4d:
I've found some internal code that breaks due to this change. I'll have to fix that before taking in this PR.

PR #2335, opened 2017-02-25
Vibration API (Second Edition)

vobruba-martin:
Based on "W3C Recommendation 18 October 2016"

See https://www.w3.org/TR/vibration/

### DIFF
--- a/externs/browser/w3c_vibration.js
+++ b/externs/browser/w3c_vibration.js
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2017 The Closure Compiler Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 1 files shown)


brad4d:
@concavelenz this is a new extern file.
Is being standardized by W3C sufficient to add it to standard externs? How do we decide?

vobruba-martin:
BTW i forgot this link http://caniuse.com/#feat=vibration . It may help you with the decision.

brad4d:
LGTM
I'm importing and submitting this.
If you want it added to standard externs, that'll require another PR.

vobruba-martin:
@brad4d Adding a file into externs/browser isn't enough?

brad4d:
@vobruba-martin 
I haven't looked up where it is yet, but according to @concavelenz there's an explicit list of the "standard" externs specified in the order they must be read by the compiler.

I asked him about it because I wondered how we could be certain that `Navigator` gets defined before this new extern defines `Navigator.prototype.vibrate`.

ChadKillingsworth:
It's here: https://github.com/google/closure-compiler/blob/0a0a8ec45ab0df1f5d35bd2881a9dea2b4f7cceb/src/com/google/javascript/jscomp/DefaultExterns.java#L36

ChadKillingsworth:
Only if you need to have a specific order in relation to one of the externs on that list does it need updated. Everything else is just added to the end of the list.

brad4d:
Ah, I see how it works now.
the `pom.xml` file for externs does grab every `.js` file from `externs/browser`.

Close to getting this submitted. Looks like there may be one internal project that I have to tweak, because they created their own version of this extern.

brad4d:
submitted internally

brad4d:
For some reason I don't see my submission in our latest pushes.
I think there must be something I need to do to get the file pushed back to github.

PR #2337, opened 2017-02-27
Add restrictions for property name generation for Angular and Polymer

### LINKED ISSUE #1371: $ prefix on renamed properties causes angular.equals to misbehave
When allowing the compiler to rename properties, it sometimes results in properties beginning with a $ character (e.g. "$a"). Unfortunately, this causes those properties to be ignored by angular.equals, which means that calling $watch(foo, bar, true) doesn't invoke its callback reliably.

Is there any chance an option can be provided for preventing such a prefix from being used during the renaming process?

ChadKillingsworth:
Both Angular and Polymer place restrictions on property names. Angular ignores properties beginning with '$' in equality checks. Polymer serializes DOM attributes to properties and thus requires names to match their restrictions (Only alpha-numeric chars and must begin with a lowercase alpha).

This restricts the property renaming passes to the set of characters permitted by the frameworks.

cc @mprobst, @justinfagnani 

Fixes #1371

### DIFF
--- a/src/com/google/javascript/jscomp/DefaultPassConfig.java
+++ b/src/com/google/javascript/jscomp/DefaultPassConfig.java
@@ -2559,17 +2559,19 @@ public final class DefaultPassConfig extends PassConfig {
   };
 
   /**
-   * Renames properties so that the two properties that never appear on
-   * the same object get the same name.
+   * Renames properties so that the two properties that never appear on the same object get the same
+   * name.
    */
   private final PassFactory ambiguateProperties =
       new PassFactory("ambiguateProperties", true) {
-    @Override
-    protected CompilerPass create(AbstractCompiler compiler) {
-      return new AmbiguateProperties(
-          compiler, options.anonymousFunctionNaming.getReservedCharacters());
-    }
-  };
+        @Override
+        protected CompilerPass create(AbstractCompiler compiler) {
+          return new AmbiguateProperties(
+              compiler,
+              options.getPropertyReservedNamingFirstChars(),
+              options.getPropertyReservedNamingNonFirstChars());
+        }
+      };
 
   /**
    * Mark the point at which the normalized AST assumptions no longer hold.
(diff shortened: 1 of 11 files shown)


src/com/google/javascript/jscomp/NameGenerator.java
```diff
@@ -17,27 +17,31 @@
...
 
 import java.util.Set;
-
 import javax.annotation.Nullable;
 
 /**
  * A class that generates unique JavaScript variable/property names.
  */
 interface NameGenerator {
+  void reset(Set<String> reservedNames, String prefix, @Nullable char[] reservedCharacters);
```
mprobst: nit: maybe add a `@see` tag pointing to docs below?

ChadKillingsworth: I just copied the full doc comment up and tweaked it.

src/com/google/javascript/jscomp/DefaultPassConfig.java
```diff
@@ -184,6 +185,38 @@ void maybeInitializeModuleRewriteState() {
     }
   }
 
+  private char[] getReservedNamingFirstChars(
```
dimvar: Looks good generally. I would move these two methods to CompilerOptions.

```
  char[] getReservedNamingFirstChars() {
    char[] reservedChars = null;
    if (this.polymerVersion != null && this.polymerVersion > 1) {
      reservedChars = PolymerPass.PROPERTY_RESERVED_FIRST_CHARS;
    } else if (this.angularPass) {
      reservedChars = AngularPass.PROPERTY_RESERVED_FIRST_CHARS;
    }
    if (this.anonymousFunctionNaming.getReservedCharacters() != null) {
      reservedChars =
          Chars.concat(this.anonymousFunctionNaming.getReservedCharacters(), reservedChars);
    }
    return reservedChars;
  }
```

Also, it'd be good if we could have an integration test that enables one of these passes and checks that property renaming uses appropriate names, but I'm not sure how to write one that is not overly long.

ChadKillingsworth: Done. I added an integration test for Angular. I can't add one for Polymer yet as it requires either #2277 or #2238 for the `polymerVersion` to be valid > 1.

If this lands first, I'll add an integration test to one of those PRs.

dimvar:
LGTM. I'll bring it in for testing.

PR #2343, opened 2017-03-01
Fix the interaction between absolute path imports and module roots

ChadKillingsworth:
The interaction between the `--js_module_root` flag and absolute imports `require('/foo/bar.js')` results in the module loader unable to locate the module. This applies equally to all 3 module resolution modes and both CommonJS and ES6 modules.

Fixing this without breaking existing functionality has proven challenging. I'm hopeful this PR actually accomplishes it.

Given a module root of `base/`, the following modules should properly resolve:

**base/mod1.js**
```js
module.exports = true;
```

**base/app.js**
```js
const mod1 = require('/mod1.js');
```

### DIFF
--- a/src/com/google/javascript/jscomp/deps/BrowserModuleResolver.java
+++ b/src/com/google/javascript/jscomp/deps/BrowserModuleResolver.java
@@ -42,17 +42,15 @@ public class BrowserModuleResolver extends ModuleResolver {
       String scriptAddress, String moduleAddress, String sourcename, int lineno, int colno) {
 
     if (ModuleLoader.isAmbiguousIdentifier(moduleAddress)) {
-      if (errorHandler != null) {
-        errorHandler.report(
-            CheckLevel.WARNING,
-            JSError.make(
-                sourcename,
-                lineno,
-                colno,
-                ModuleLoader.INVALID_MODULE_PATH,
-                moduleAddress,
-                "BROWSER"));
-      }
+      errorHandler.report(
+          CheckLevel.WARNING,
+          JSError.make(
+              sourcename,
+              lineno,
+              colno,
+              ModuleLoader.INVALID_MODULE_PATH,
+              moduleAddress,
+              "BROWSER"));
       return null;
     }
 
--- a/src/com/google/javascript/jscomp/deps/BrowserModuleResolver.java
+++ b/src/com/google/javascript/jscomp/deps/BrowserModuleResolver.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2017 The Closure Compiler Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.javascript.jscomp.deps;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+import com.google.javascript.jscomp.CheckLevel;
+import com.google.javascript.jscomp.ErrorHandler;
+import com.google.javascript.jscomp.JSError;
+import javax.annotation.Nullable;
+
+/**
+ * Resolution algorithm for Browsers.
+ *
+ * <p>Only unambiguous paths are supported. Paths must specify a file extension.
+ */
+public class BrowserModuleResolver extends ModuleResolver {
+
+  public BrowserModuleResolver(
+      ImmutableSet<String> modulePaths,
+      ImmutableList<String> moduleRootPaths,
+      ErrorHandler errorHandler) {
+    super(modulePaths, moduleRootPaths, errorHandler);
+  }
+
+  @Nullable
+  public String resolveJsModule(
+      String scriptAddress, String moduleAddress, String sourcename, int lineno, int colno) {
+
+    if (ModuleLoader.isAmbiguousIdentifier(moduleAddress)) {
+      if (errorHandler != null) {
+        errorHandler.report(
+            CheckLevel.WARNING,
+            JSError.make(
+                sourcename,
+                lineno,
+                colno,
+                ModuleLoader.INVALID_MODULE_PATH,
+                moduleAddress,
+                "BROWSER"));
+      }
+      return null;
+    }
+
+    String loadAddress = locate(scriptAddress, moduleAddress);
+    if (loadAddress == null && errorHandler != null) {
+      errorHandler.report(
+          CheckLevel.WARNING,
+          JSError.make(sourcename, lineno, colno, ModuleLoader.LOAD_WARNING, moduleAddress));
+    }
+    return loadAddress;
+  }
+}
--- a/src/com/google/javascript/jscomp/deps/BrowserModuleResolver.java
+++ b/src/com/google/javascript/jscomp/deps/BrowserModuleResolver.java
@@ -57,7 +57,7 @@ public class BrowserModuleResolver extends ModuleResolver {
     }
 
     String loadAddress = locate(scriptAddress, moduleAddress);
-    if (loadAddress == null && errorHandler != null) {
+    if (loadAddress == null) {
       errorHandler.report(
           CheckLevel.WARNING,
           JSError.make(sourcename, lineno, colno, ModuleLoader.LOAD_WARNING, moduleAddress));
--- a/src/com/google/javascript/jscomp/deps/ModuleLoader.java
+++ b/src/com/google/javascript/jscomp/deps/ModuleLoader.java
@@ -23,7 +23,7 @@ import com.google.common.base.Function;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.ImmutableSortedMap;
+import com.google.common.collect.ImmutableSortedSet;
 import com.google.common.collect.Iterables;
 import com.google.javascript.jscomp.CheckLevel;
 import com.google.javascript.jscomp.DiagnosticType;
@@ -34,8 +34,8 @@ import java.util.Comparator;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
-import java.util.SortedMap;
-import java.util.TreeMap;
+import java.util.SortedSet;
+import java.util.TreeSet;
 import javax.annotation.Nullable;
 
 /**
@@ -58,6 +58,12 @@ public final class ModuleLoader {
   public static final DiagnosticType LOAD_WARNING =
       DiagnosticType.warning("JSC_JS_MODULE_LOAD_WARNING", "Failed to load module \"{0}\"");
 
+  private static final String[] NODE_FILE_EXTENSIONS_TO_SEARCH = {"", ".js", ".json"};
+  private static final String[] EMPTY_FILE_EXTENSIONS_TO_SEARCH = {""};
+  private static final String[] NODE_FILES_TO_SEARCH = {
+    MODULE_SLASH + "package.json", MODULE_SLASH + "index.js", MODULE_SLASH + "index.json"
+  };
+
   @Nullable private final ErrorHandler errorHandler;
 
   /** Root URIs to match module roots against. */
@@ -66,7 +72,7 @@ public final class ModuleLoader {
   private final ImmutableSet<String> modulePaths;
 
   /** Named modules found in node_modules folders */
-  private final ImmutableSortedMap<String, ImmutableSet<String>> nodeModulesRegistry;
+  private final ImmutableSortedSet<String> nodeModulesFolders;
 
   /** Named modules found in node_modules folders */
   private ImmutableMap<String, String> packageJsonMainEntries;
@@ -100,7 +106,7 @@ public final class ModuleLoader {
 
     this.packageJsonMainEntries = ImmutableMap.of();
 
-    this.nodeModulesRegistry = buildRegistry(this.modulePaths);
+    this.nodeModulesFolders = buildNodeModulesFoldersRegistry(this.modulePaths);
 
     this.resolutionMode = resolutionMode;
   }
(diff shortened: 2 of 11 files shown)


src/com/google/javascript/jscomp/deps/ModuleLoader.java
```diff
@@ -339,10 +362,22 @@ private String locateNoCheck(String name) {
      */
     @Nullable
     private String locate(String name) {
-      String path = locateNoCheck(name);
-      if (modulePaths.contains(path)) {
-        return path;
+      InputPathData pathData = new InputPathData(canonicalizePath(name));
+      InputPathData modulePath = modulePaths.get(pathData.ensureLeadingSlash());
+      if (modulePath != null) {
```
brad4d: Please add comments to explain.
e.g. If we find the module by forcing the path to be absolute, why do we return the path unmodified?

src/com/google/javascript/jscomp/deps/ModuleLoader.java
```diff
@@ -388,51 +424,55 @@ public static boolean isPathIdentifier(String name) {
...
+      builder.add(new InputPathData(ModuleNames.escapePath(resolver.apply(root))));
     }
     return builder.build();
   }
 
-  private static ImmutableSet<String> resolvePaths(
-      Iterable<String> names, ImmutableList<String> roots) {
-    HashSet<String> resolved = new HashSet<>();
+  private static ImmutableMap<String, InputPathData> resolvePaths(
+      Iterable<String> names, ImmutableList<InputPathData> roots) {
```
brad4d: nit: Why not `Iterable<InputPathData> roots`?

src/com/google/javascript/jscomp/deps/ModuleLoader.java
```diff
@@ -388,51 +424,55 @@ public static boolean isPathIdentifier(String name) {
...
-      builder.add(ModuleNames.escapePath(resolver.apply(root)));
+      builder.add(new InputPathData(ModuleNames.escapePath(resolver.apply(root))));
     }
     return builder.build();
   }
 
-  private static ImmutableSet<String> resolvePaths(
-      Iterable<String> names, ImmutableList<String> roots) {
-    HashSet<String> resolved = new HashSet<>();
+  private static ImmutableMap<String, InputPathData> resolvePaths(
```
brad4d: Could you add a Javadoc comment explaining what the keys and values of the return value represent?

src/com/google/javascript/jscomp/deps/ModuleLoader.java
```diff
@@ -388,51 +424,55 @@ public static boolean isPathIdentifier(String name) {
...
+      Iterable<String> names, ImmutableList<InputPathData> roots) {
+    Map<String, InputPathData> resolved = new HashMap<>();
```
brad4d: nit: use ImmutableMap.builder()

src/com/google/javascript/jscomp/deps/ModuleLoader.java
```diff
@@ -388,51 +424,55 @@ public static boolean isPathIdentifier(String name) {
...
     // Not underneath any of the roots.
-    return path;
+    return inputPath;
   }
 
   /** Build the module registry from the set of module paths */
-  private static ImmutableSortedMap<String, ImmutableSet<String>> buildRegistry(
-      ImmutableSet<String> modulePaths) {
-    SortedMap<String, Set<String>> registry =
+  private static ImmutableSortedMap<String, ImmutableMap<String, InputPathData>> buildRegistry(
```
brad4d: Please expand the JavaDoc to define the meaning of the keys and values of both the input and output.

src/com/google/javascript/jscomp/deps/ModuleLoader.java
```diff
@@ -564,4 +607,30 @@ public String apply(DependencyInfo info) {
      */
     NODE
   }
+
+  private static class InputPathData {
```
brad4d: Need javadoc for this.

src/com/google/javascript/jscomp/deps/ModuleLoader.java
ChadKillingsworth: Mainly because of the great difficulty of landing anything that deals with the deps package. I did my absolute best to change the bare minimum of test cases to force the functionality to adhere to existing results.

My preferences is to return the path as absolute here if you're willing to try it.

src/com/google/javascript/jscomp/deps/ModuleLoader.java
```diff
@@ -564,4 +607,30 @@ public String apply(DependencyInfo info) {
...
+
+    InputPathData(String path) {
+      this.path = path;
+    }
+
+    boolean isRelativeToCompilationRoot() {
+      return !(isAbsoluteIdentifier(this.path) || isRelativeIdentifier(this.path));
+    }
+
+    String ensureLeadingSlash() {
```
ChadKillingsworth: The struggle I have is one of terminology. Within this file:

 * isAbsoluteIdentifier('/foo') => true
 * isAbsoluteIdentifier('./foo') => false
 * isAbsoluteIdentifier('foo') => false
 * isRelativeIdentifier('/foo') => false
 * isRelativeIdentifier('./foo') => true
 * isRelativeIdentifier('foo') => false

I don't know how to label paths that do not begin with either a '.' or '/' as they are neither relative or absolute.

If it's a module path `foo/bar.js` can either be relative to the compilation root (LEGACY mode), looked up from the node module registry (NODE mode) or invalid (BROWSER mode).

If it's an input path, `foo/bar.js` is relative to either the compilation or module roots.

It's all horribly confusing.

ChadKillingsworth:
@brad4d I took a hard run through and reworked this PR. I was able to drastically simplify things and hopefully the entire file is now much easier to read and follow.

src/com/google/javascript/jscomp/deps/ModuleLoader.java
```diff
@@ -232,6 +248,14 @@ public ModulePath resolveJsModule(
 
     @Nullable
     private String resolveJsModuleFile(String moduleAddress) {
+      String[] fileExtensionsToSearch;
```
brad4d: How about passing the file extensions to search into this method instead?
I'd prefer to keep checking of ResolutionMode to only the one switch statement, even if it means creating some custom-named functions for particular code paths.
Such functions make it easier to know the context when you're reading.

src/com/google/javascript/jscomp/deps/ModuleLoader.java
```diff
@@ -254,13 +278,9 @@ private String resolveJsModuleFileOrDirectory(String moduleAddress) {
...
     private String resolveJsModuleDirectory(String moduleAddress) {
-      String[] extensions = {
-        MODULE_SLASH + "package.json", MODULE_SLASH + "index.js", MODULE_SLASH + "index.json"
-      };
-
       // Load as a file
-      for (int i = 0; i < extensions.length; i++) {
-        String loadAddress = locate(moduleAddress + extensions[i]);
+      for (int i = 0; i < NODE_FILES_TO_SEARCH.length; i++) {
+        String loadAddress = locate(moduleAddress + NODE_FILES_TO_SEARCH[i]);
```
brad4d: If this method is actually specific to Node.js modules, could you rename it to indicate that?

src/com/google/javascript/jscomp/deps/ModuleLoader.java
```diff
@@ -277,23 +297,21 @@ private String resolveJsModuleDirectory(String moduleAddress) {
...
+        if (loadAddress == null) {
+          // Load as a directory
+          loadAddress = resolveJsModuleDirectory(fullModulePath);
         }
 
-        // Load as a directory
-        loadAddress = resolveJsModuleDirectory(fullModulePath);
         if (loadAddress != null) {
-          return loadAddress;
+          return removeAbsoluteIdentifierIndicator(loadAddress);
```
brad4d: Could you add a comment to say why you remove the absolute identifier indicator here?
Maybe in javadoc on the method.

src/com/google/javascript/jscomp/deps/ModuleLoader.java
```diff
@@ -311,26 +329,30 @@ private String resolveJsModuleLegacy(
       if (!moduleName.endsWith(".js")) {
         moduleName += ".js";
       }
-      String resolved = locateNoCheck(moduleName);
-      if (!modulePaths.contains(resolved) && errorHandler != null) {
-        errorHandler.report(
-            CheckLevel.WARNING, JSError.make(sourcename, lineno, colno, LOAD_WARNING, moduleName));
+      String resolved = resolveJsModuleFile(moduleName); //locateNoCheck(moduleName);
```
brad4d: should probably remove the `//locateNoCheck` comment.

src/com/google/javascript/jscomp/deps/ModuleLoader.java
ChadKillingsworth: Apparently there wasn't a good reason ...

ChadKillingsworth:
@brad4d Fixed.

src/com/google/javascript/jscomp/deps/ModuleLoader.java
```diff
@@ -388,38 +417,66 @@ public static boolean isPathIdentifier(String name) {
     return name.contains(MODULE_SLASH);
   }
 
+  /** Removes the leading slash from absolute identifiers */
+  public static String removeAbsoluteIdentifierIndicator(String name) {
```
brad4d: This isn't called anywhere.

src/com/google/javascript/jscomp/deps/ModuleLoader.java
```diff
@@ -388,38 +417,66 @@ public static boolean isPathIdentifier(String name) {
...
+    if (isAbsoluteIdentifier(name)) {
+      return name.substring(1);
+    }
+    return name;
+  }
+
+  /** Converts paths which don't start with either a '.' or '/' to absolute paths */
+  public static String toAbsoluteIdentifier(String name) {
```
brad4d: Why is this public?

src/com/google/javascript/jscomp/deps/ModuleLoader.java
```diff
@@ -388,38 +417,66 @@ public static boolean isPathIdentifier(String name) {
...
+    if (name.length() == 0 || isAbsoluteIdentifier(name) || isRelativeIdentifier(name)) {
+      return name;
```
brad4d: Wait, this can return a relative identifier?
That's just misleading.
I saw your post regarding troublesome terminology, though I cannot find it now.
I propose this.

- "absolute" means "starts with '/'"
- "relative" means "starts with './'"
- "ambiguous" means everything else.

How about we detect ambiguous paths ASAP and either report an error or explicitly force them to be relative or absolute at that point?

I'd rather see some duplicated logic that makes it clear what is happening than a "fix it if it is broken, otherwise leave it alone" method like this one.

brad4d:
@ChadKillingsworth 
Our "squash to a single commit" policy doesn't seem to work well with code reviews.
Since we're having some back and forth here, I suggest you not do the squashing until we've come to agreement.

src/com/google/javascript/jscomp/deps/ModuleLoader.java
ChadKillingsworth: > How about we detect ambiguous paths ASAP and either report an error or explicitly force them to be relative or absolute at that point?

I think ambiguous paths (as you labeled them) are used widely inside google. For instance, this path is ambiguous:

```js
import {foo} from 'path/to/bar'
```

I can force them to be absolute if you are willing to try landing it ...

ChadKillingsworth: FYI that path means two completely different things depending on the resolution mode. This has been a source of great confusion to all parties.

```js
import {foo} from 'path/to/bar'
```

In `LEGACY` mode, this path is absolute.
In `NODE` mode, this path is looked up from the registry.
In `BROWSER` mode, this path is invalid.

ChadKillingsworth:
@brad4d I reworked the resolution algorithms into independent classes.

brad4d:
Looking very good.
Thanks!
Importing and testing.

brad4d:
Had to fix some internal stuff to know about the new files.
All our project tests passed.
I'm testing against all internal projects now.

brad4d:
Got a couple-dozen failures when testing all the things.
I'm part way through confirming which are "real" breaks and which were actually already broken.

ChadKillingsworth:
I would actually expect a few project fixes to be required by this change - but it is definitely worth the trouble.

brad4d:
None of the breaks appear to be real.
Sent for internal review.

MatrixFrog:
Just to be clear, you don't have to push changes like formatting changes back to the Github PR for MOE to work.

src/com/google/javascript/jscomp/deps/BrowserModuleResolver.java
```diff
@@ -0,0 +1,67 @@
...
+      ErrorHandler errorHandler) {
+    super(modulePaths, moduleRootPaths, errorHandler);
+  }
+
+  @Nullable
+  public String resolveJsModule(
+      String scriptAddress, String moduleAddress, String sourcename, int lineno, int colno) {
+
+    if (ModuleLoader.isAmbiguousIdentifier(moduleAddress)) {
+      if (errorHandler != null) {
```
brad4d: Could we forbid null for errorHandler and just make the creator pass in a no-op Errorhandler if they don't care about errors?

src/com/google/javascript/jscomp/deps/ModuleResolver.java
```diff
@@ -0,0 +1,102 @@
...
+      ImmutableSet<String> modulePaths,
+      ImmutableList<String> moduleRootPaths,
+      ErrorHandler errorHandler) {
+    this.modulePaths = modulePaths;
+    this.moduleRootPaths = moduleRootPaths;
+    this.errorHandler = errorHandler;
+  }
+
+  Map<String, String> getPackageJsonMainEntries() {
+    return null;
```
brad4d: Could we return an empty map instead of null so this method doesn't have to be `@Nullable`?

brad4d:
@ChadKillingsworth 
I've pushed some tweaks from my internal review with @MatrixFrog onto your branch.

I've also passed along 2 requests for changes in a github code review.

Regarding testing we are looking great.
All of the global presubmit tests passed.
As long as we make no further behavior changes on this PR request, we should be able to submit with just our standard project tests, which complete very quickly.

brad4d:
@MatrixFrog 
Yeah, I know I didn't have to push the changes back, but I'm trying to make the review process more transparent, and I've worked out a way to do it that's pretty easy and will get easier as I tweak it further.

ChadKillingsworth:
@brad4d So the 64million dollar question... Do I need to squash my commits when I push next?

brad4d:
@ChadKillingsworth 
Nope, no need for you to squash at all.
I'll effectively squash it on my end.

In case you're interested, I'm basically working like this.

```sh
cd $my_github_clone
git pull $chads_pr_branch
git format-patch -stdout >patchfile
patch-munge <patchfile >patchfile.munge # secret sauce to fix paths, etc.
cd $my_internal_repo_clone
git am patchfile.munge
```

Then the reverse to go back out to you.
This has no effect on what happens when I submit internally and the result is pushed to our github repo.

ChadKillingsworth:
Awesome. I made the requested changes. Should be good-to-go.

brad4d:
I've imported your latest change and hit "the submit button".
With luck this will be in tomorrow's push to github.

PR #2351, opened 2017-03-06
Consistency update to README to make instructions clearer

sveningsonrobin:
After trying to build Closure Compiler and run it using Eclipse I noticed that some of the instructions were inconsistent, especially when it came to describing when a task is to be done inside of Eclipse or on the command line. Since the *Running* section started with "On the command line, at the root of this project", I figured that the same sentence should be used elsewhere when you are referring to running a command on the command line. The consistency update regarding the dots is less important 😉 

If I am misinterpreting your instructions, and there is some way of running commands inside of Eclipse that I am unaware of, please let me know.

PR #2353, opened 2017-03-06
README run instructions update

sveningsonrobin:
Running the Closure Compiler from inside of Eclipse is a bit of a struggle, thanks to the bug in eclipse [described here](http://stackoverflow.com/questions/4711098/passing-end-of-transmission-ctrl-d-character-in-eclipse-cdt-console). Since the README instructions contains information about how to build the project using Eclipse, I find it appropriate that these extra instructions on how to run the project using Eclipse are included in the README as well. This has been tested and verified by me and @lbercken, but I suggest you try it out as well if you have the time.

This PR goes well together with my two previous PRs (#2351 and #2352).

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,14 @@ java -jar compiler.jar --help
 More detailed information about running the Closure Compiler is available in the
 [documentation](http://code.google.com/closure/compiler/docs/gettingstarted_app.html).
 
+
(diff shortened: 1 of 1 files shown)


sveningsonrobin:
Squashed the commits. All three PRs now only contain a single commit each.

Issue #180, opened 2014-04-29
Conditional operator: when the condition and first expression are the same, OR could be used instead


blickly:
This issue was imported from Closure Compiler's previous home at http://closure-compiler.googlecode.com

The original discussion is archived at:
http://blickly.github.io/closure-compiler-issues/#1194

PR #2354, opened 2017-03-06
Conditional operator: when the condition and first expression are the same, OR could be used instead

lbercken:
This should (partially) fix https://github.com/google/closure-compiler/issues/180.

When using the ? operator the following can be transformed:

```x ? x : y```

to a shorter equivalent:

```x || y```.

I added a Unit test to verify that the change works. However the transformation only works for conditionals now, e.g. in an `if` statement. It does not work for example in an `assignment` statement yet.

### DIFF
--- a/test/com/google/javascript/jscomp/PeepholeMinimizeConditionsTest.java
+++ b/test/com/google/javascript/jscomp/PeepholeMinimizeConditionsTest.java
@@ -337,6 +337,12 @@ public final class PeepholeMinimizeConditionsTest extends CompilerTestCase {
   }
 
   public void testMinimizeHook() {
+    fold("x ? x : y",
+	 "x || y");
+    // These two cases cannot be folded, because of possible side effects
+    foldSame("x() ? x() : y()");
+    foldSame("x.y ? x.y : x.z");
+    
     fold("!x ? foo() : bar()",
          "x ? bar() : foo()");
     fold("while(!(x ? y : z)) foo();",
--- a/src/com/google/javascript/jscomp/PeepholeMinimizeConditions.java
+++ b/src/com/google/javascript/jscomp/PeepholeMinimizeConditions.java
@@ -1148,6 +1148,7 @@ class PeepholeMinimizeConditions
    *   x && true        --> x
    *   x ? false : true --> !x
    *   x ? true : y     --> x || y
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by Dominator008** [comment]
This is not a safe optimization in the general case. Folding `x() ? x() : y()` to `x() || y()` is incorrect if `x()` has side effects. Even folding `x.y ? x.y : x.z` is unsafe if `x.y` triggers a getter with side effects.

I think this should be restricted to `NAME` accesses.

**Comment by lbercken** [comment]
Right, I see what you mean, I'll try to restrict this to `NAME` accesses.

**Comment by lbercken** [comment]
I guess I can squash the commits later into a single commit, right?

File: test/com/google/javascript/jscomp/PeepholeMinimizeConditionsTest.java [hunk 1]
```diff
@@ -337,6 +337,8 @@ public void testNoSwapWithDanglingElse() {
   }
 
   public void testMinimizeHook() {
+    fold("x ? x : y",
+	 "x || y");
```
Dominator008: Can you add test cases to make sure `x() ? x() : y()` and `x.y ? x.y : x.z` are not folded?

Dominator008:
@lbercken Yes, please squash and rebase.

dimvar:
I'll bring this in and test it.

Dominator008:
@lbercken I've submitted this internally and it should be synced externally later today. I changed the condition to `!mayHaveSideEffects(condition) && !mayHaveSideEffects(trueNode) && condition.isEquivalentTo(trueNode)` so it folds function calls that are deemed side-effect free by the compiler, and also for GETPROPs, which are assumed to be side-effect free: https://github.com/google/closure-compiler/wiki/Compiler-Assumptions#details.

lbercken:
@Dominator008 Alright, thanks! Also thanks for your support and help. Didn't know about the `mayHaveSideEffects` method and the assumption detail page. By the way, we might have to change the comments (I added) saying:

`//   Only when x is NAME, hence x does not have side effects`
`//   x ? x : y        --> x || y`

accordingly.

Dominator008:
@lbercken I think the documentation in the unit test suffices :)

PR #2363, opened 2017-03-09
Add Object to isUnextendableNativeClass

dsanders11:
Spent a good amount of time spinning my wheels trying `class Foo extends Object { }` and not getting the desired results until I noticed other built-ins like `Array` have an error when trying to extend.

I don't see any reason `Object` shouldn't be included in the list of unextendable native classes, so I've added it to prevent anyone else from spinning their wheels.

### DIFF
--- a/src/com/google/javascript/jscomp/Es6ConvertSuperConstructorCalls.java
+++ b/src/com/google/javascript/jscomp/Es6ConvertSuperConstructorCalls.java
@@ -380,6 +380,7 @@ implements NodeTraversal.Callback, HotSwapCompilerPass {
       case "InternalError":
       case "Map":
       case "Number":
+      case "Object":
       case "Promise":
       case "Proxy":
       case "RegExp":


Dominator008:
@brad4d Wanna take a look?

brad4d:
@dsanders11
I'm pretty sure we have to handle code that explicitly extends `Object` to avoid breaking internal google projects at least, and possibly external ones as well.

Could you file an issue explaining the problems you encountered trying to do this yourself?
Perhaps there's something we need to fix there.

dsanders11:
@brad4d, I'm not sure it's worth a separate issue. Basically I just spun my wheels a bit trying to do:

```
/**
 * @dict
 */
class Foo extends Object { }

var bar = new Foo();
bar['bar'] = 42;
var foobar = Foo.assign(new Foo(), bar);
```

There's a warning from the compiler about `assign` not existing on `Foo` but it's not clear to the end user why that is, and the transpiled code from ES6 to ES5 doesn't work. It was only after trying a different built-in type (`Array`) that the compiler had an error about not being able to extend native classes which made me realize that was the problem with the code.

So perhaps a different warning if extending `Object`? I try to run the compiler with as many warnings enabled as possible but didn't get anything helpful other than the message about `assign` not being defined, but that was vague and led to more confusion than anything.

brad4d:
@dsanders11 
Trying to use `[]` on a class-defined object is something we explicitly disallow on purpose, because the point of a class is to have a clearly defined API. Allowing "random" access with `[]` violates that goal, as does using `assign`.

It would be good if we could explicitly warn or error when someone is trying to extend `Object` to just add special behavior on top of the standard `Object` "(almost) any property is OK" behavior. Unfortunately, we have to deal with automatically-generated code that regularly creates classes that extend `Object`.

@dimvar 
Would you consider it a bug that the type checker doesn't recognize static methods inherited from Object

brad4d:
Actually, it isn't a bug because all of the static methods on `Object` are non-enumerable, so the `$jscomp.inherits` method that handles copying them won't see them.

Here's a repro.
http://closure-compiler.appspot.com/home#code%3D%252F%252F%2520%253D%253DClosureCompiler%253D%253D%250A%252F%252F%2520%2540compilation_level%2520ADVANCED_OPTIMIZATIONS%250A%252F%252F%2520%2540output_file_name%2520default.js%250A%252F%252F%2520%2540formatting%2520pretty_print%250A%252F%252F%2520%253D%253D%252FClosureCompiler%253D%253D%250A%250A%252F%252F%2520ADD%2520YOUR%2520CODE%2520HERE%250Aclass%2520Foo%2520extends%2520Object%2520%257B%257D%250Aconsole.log(Foo.assign)%253B%250Aconsole.log(Foo.isSealed)%253B%250AObject.assign(new%2520Foo()%252C%2520%257Ba%253A'a'%257D)%253B%250A%250Aclass%2520Bar%2520%257B%250A%2520%2520static%2520staticFunc()%2520%257B%257D%250A%257D%250A%250Aclass%2520Baz%2520extends%2520Bar%2520%257B%257D%250A%250Aconsole.log(Bar.staticFunc)%253B%250Aconsole.log(Baz.staticFunc)%253B%250Aconsole.log(new%2520Foo().hasOwnProperty)%253B

ChadKillingsworth:
@brad4d By spec they should be copied though. Our inheritance model just doesn't support this. It's a problem for Polymer.

brad4d:
To sum up, someone who tries to explicitly extend `Object` probably isn't going to get what they expect, because it'll end up acting just like not extending anything at all when compiled, but this isn't true for uncompiled code.

```javascript
class Foo extends Object {}
console.log(Foo.assign); // native code function
class Bar {}
console.log(Bar.assign); // undefined
```

It would be good to turn this into a warning at least.
The automatically generated code I'm aware of that regularly extends `Object` is coming from the TypeScript compiler.

@dimvar 
Do you know of other cases than that one?
Do you think we could generate a warning here?

dsanders11:
> Trying to use [] on a class-defined object is something we explicitly disallow on purpose, because the point of a class is to have a clearly defined API. Allowing "random" access with [] violates that goal, as does using assign.

@brad4d, yes, that's why I added `@dict` to get around the `[]` warnings. I was experimenting with trying to find a way to get the compiler to derive type for individual key-value pairs in a heterogeneous collection and the code I was working in already used `Object.assign` for the collection, which is how I ended up where I ended up.

> It would be good to turn this into a warning at least.

👍

brad4d:
@ChadKillingsworth 
Could you elaborate on your statement "it's a problem for Polymer"?
Thanks.

ChadKillingsworth:
Polymer 2 depends on static ES6 properties to be present on a class. Because the compiler doesn't properly model this, we have to add nasty workarounds like #2286.

I wish we would just add support for static properties on a class in the type system.

brad4d:
@ChadKillingsworth 
Thanks for the explanation. Do you know if there's already an issue defined for the static properties typing problem?

I'm afraid we cannot accept this PR since it will break a large amount of known code.

PR #2400, opened 2017-04-03
Add missing jsonpCallbackParam option in angular externs

fredj:
This option was added in version 1.6, see:
  https://docs.angularjs.org/guide/migration#migrate1.5to1.6-ng-services-$http


mprobst:
Thanks for the fix @fredj !

MatrixFrog:
For future PRs, please LGTM but don't merge, and then we will merge it internally first then MOE push it back out to the master branch.

PR #2401, opened 2017-04-04
[WIP] Provide correct location info for getprop, getelem and call invocations

ChadKillingsworth:
Currently, location information for some expressions is in the wrong place:

**Before change**
```js
console.log('foo');
^
```
The location information for both the `.` and the `(` symbols is the start of the line.

**After change**
Correct location info
```js
console.log('foo');
       ^   ^
```

Because of this we can't provide suspicious code checks for missing semicolons between function calls and source mappings don't match other tools.

There are several broken tests. In particular, I have yet to determine why the ProcessCommonJS test failure is even related to this change.

### DIFF
--- a/src/com/google/javascript/jscomp/parsing/parser/Parser.java
+++ b/src/com/google/javascript/jscomp/parsing/parser/Parser.java
@@ -2824,7 +2824,7 @@ public class Parser {
 
     if (isStartOfAsyncArrowFunction(left)) {
       // re-evaluate as an async arrow function.
-      resetScanner(left);
+      resetScanner(start);
       return parseAsyncArrowFunction(expressionIn);
     }
     if (peek(TokenType.ARROW)) {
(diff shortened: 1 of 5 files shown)


src/com/google/javascript/jscomp/parsing/parser/Parser.java
```diff
@@ -2824,7 +2824,7 @@ private ParseTree parseAssignment(Expression expressionIn) {
 
     if (isStartOfAsyncArrowFunction(left)) {
       // re-evaluate as an async arrow function.
-      resetScanner(left);
+      resetScanner(start);
```
brad4d: The intention here is "reparse `left` as something else".
Resetting to an arbitrary position isn't something I think we should be doing.

ChadKillingsworth: I'll take another look, but `left` was no longer referring to `async` - it now referred to the beginning of the call `(`. That's why this was changed.

brad4d: Ah, I see.
Looks like I was incorrect in my mental model when I wrote `resetScanner()`.

In that case, please remove the version of `resetScanner()` that takes a `ParseTree`.
Thanks.

ChadKillingsworth: @brad4d Your mental model was completely correct - but this PR changes what `start` refers to.

brad4d: If my mental model were correct, then that would mean

`a + b` => AND node source location covers the entire expression, not just the location of the `+`

If that's correct, then the CALL node's source location really must include the entire expression and not point at the `(`.

brad4d:
@ChadKillingsworth, do you still want to make this change, or can we close this PR?

Issue #2425, opened 2017-04-11
extern definition for measureText should return non-null


myphysicslab:
In the extern [html5.js](https://github.com/google/closure-compiler/blob/master/externs/browser/html5.js#L550) the return type should be non-null.  Add an exclamation point like this:

```javascript
/**
 * @param {string} text
 * @return {!TextMetrics}
 * @nosideeffects
 */
CanvasRenderingContext2D.prototype.measureText = function(text) {};
```

The [spec for measureText](https://html.spec.whatwg.org/multipage/scripting.html#dom-context-2d-measuretext) makes no mention of possibly returning null.  

This change helps avoid needing code like this:

```javascript
var textMetrics = context.measureText(s);
if (!textMetrics) {
 throw new Error();
}
var textWidth = textMetrics.width;
```
instead you can write
```javascript
var textWidth = context.measureText(s).width;
```

myphysicslab:
Same situation for [createLinearGradient](https://html.spec.whatwg.org/multipage/scripting.html#dom-context-2d-createlineargradient) and probably several other calls on `CanvasRenderingContext2D` such as `createRadialGradient` and  `createPattern`.

brad4d:
@myphysicslab 
Fortunately, fixing return values to be non-nullable should be an easy change to land.
Care to do a pull request?

myphysicslab:
As discussed on #860 I've also changed some parameter types to be non-nullable which can break existing code. Shall I remove those for now?  I think it would be good to have those definitions sooner, before more code is written to the looser incorrect specification.

myphysicslab:
After thinking it over, I'll modify the pull request to only change return values to be non-nullable, not parameter values.  The question about non-nullable parameter values should be done on a separate issue and pull request.

myphysicslab:
Fixed by #2435

Issue #860, opened 2015-03-01
html5.js should use non-nullable types


myphysicslab:
Several types specified in html5.js should be indicated as non-nullable types.  For example, here is an error I get (with "new type inference" compiler option) which should not occur:

```
src/lab/graph/StandardAxes.js:296: WARNING - Attempt to use nullable type TextMetrics|null.
      var textWidth = context.measureText(s).width;
                      ^
```

where `context` is defined by:

```
@param {!CanvasRenderingContext2D} context 
```

The definition referenced in html5.js is http://www.w3.org/TR/2dcontext/#textmetrics.  That page says that "Interfaces are defined in terms of [Web IDL](http://heycam.github.io/webidl/)".  Looking at that page, in the [section about Nullable Types](http://heycam.github.io/webidl/#idl-nullable-type) we read:

> A nullable type is an IDL type constructed from an existing type (called the inner type), which just allows the additional value null to be a member of its set of values. Nullable types are represented in IDL by placing a U+003F QUESTION MARK ("?") character after an existing type. The inner type must not be any, another nullable type, or a union type that itself has includes a nullable type or has a dictionary type as one of its flattened member types.

The [w3.org definition of CanvasRenderingContext2D](http://www.w3.org/TR/2dcontext/#canvasrenderingcontext2d) shows that measureText returns a non-nullable type:

```
interface CanvasRenderingContext2D {
 // code omitted...
 TextMetrics measureText(DOMString text);
// code omitted...
};
interface TextMetrics {
  readonly attribute double width;
};
```

The same argument applies to other return types in CanvasRenderingContext2D such as CanvasGradient, CanvasPattern, and ImageData.

myphysicslab:
Another property to perhaps make non-nullable is `window.localStorage`

```
src/lab/util/Terminal.js:412: WARNING - Attempt to use nullable type Storage|null.
  window.localStorage.removeItem(this.pageKey());
  ^
```
This would be in the extern webstorage.js:
```javascript
/**
 * Window implements WindowLocalStorage
 *
 * @type {Storage}
 */
Window.prototype.localStorage;
```

myphysicslab:
Another property to consider for making non-nullable: `navigator`
```
src/lab/util/UtilityCore.js:228: WARNING - Attempt to use nullable type Navigator|null.
  return navigator.userAgent.match(/.*Chrome.*/) != null;
         ^
```
Or can `navigator` be null in older browsers?

dimvar:
[...]

PR #2435, opened 2017-04-13
update HTML5 externs to have non-null types

myphysicslab:
Update definitions in HTML5 externs to specify non-null types (adding ! exclamation mark) where appropriate.   Only return and property types have been changed. I've done some cursory internet research on these changes.

Add the ? nullable type for some return and property types in cases where the spec is clear about that.

This addresses issues #860 and #2425.

### DIFF
--- a/externs/browser/html5.js
+++ b/externs/browser/html5.js
@@ -257,7 +257,7 @@ Path2D.prototype.arc = function(
     x, y, radius, startAngle, endAngle, optAnticlockwise) {};
 
 /**
- * @param {Path2D} path
+ * @param {!Path2D} path
  * @return {undefined}
  */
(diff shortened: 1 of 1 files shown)


brad4d:
The PR looks good to me.
I've imported it and will do a 'test all the things' run on it.

brad4d:
Overlooked change of `setLineDash()` parameter to non-nullable.
It broke 4000+ tests. :)
Nothing else was a problem, though, so I'm trying to submit this internally with the `setLineDash()` parameter restored to nullable.

brad4d:
Submitted internally.
Should get pushed to github tomorrow.

myphysicslab:
Great, sorry I missed undoing that change to `setLineDash` parameter.

PR #2444, opened 2017-04-18
Add externs for hls.js

IvanRF:
Externs definition for the JavaScript library [hls.js](https://github.com/video-dev/hls.js)

### DIFF
--- a/contrib/externs/hls.js
+++ b/contrib/externs/hls.js
@@ -1,0 +1,785 @@
+/*
+ * Copyright 2017 The Closure Compiler Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


IvanRF:
I must add that I tested the player with a basic code like the one on the _Getting Started_ docs of `hls.js` and the player on the resulting minification works fine.
Before:
```
if (Hls.isSupported()) {
	var video = document.getElementById("video");
	var src = video.dataset.source;
	var hls = new Hls();
	hls.loadSource(src);
	hls.attachMedia(video);
	hls.on(Hls.Events.MANIFEST_PARSED, function() {
		video.play();
	});
}
```

After:
```
if(Hls.isSupported()){var a=document.getElementById("video"),b=a.dataset.source,c=new Hls;c.loadSource(b);c.attachMedia(a);c.on(Hls.Events.MANIFEST_PARSED,function(){a.play()});
```

dimvar:
Hi Ivan, we don't really host externs for all libraries, only for a few very popular ones.

MatrixFrog:
You can add it to https://github.com/google/closure-compiler/wiki/Externs-For-Common-Libraries if you like.

MatrixFrog:
(Er, I mean, host it somewhere else, and then link to it from that wiki page)

IvanRF:
@MatrixFrog I just edited the wiki page, thanks!

PR #2451, opened 2017-04-23
Adds the PolymerElement.prototype.create method to externs

lukeschaefer:
https://www.polymer-project.org/1.0/docs/api/Polymer.Base#method-create

Otherwise, compiling this.create() in a PolymerElement results in 'Property create never defined on WhateverElement', despite working at run-time

### DIFF
--- a/contrib/externs/polymer-1.0.js
+++ b/contrib/externs/polymer-1.0.js
@@ -646,6 +646,14 @@ PolymerElement.prototype.updateStyles = function(properties) {};
 PolymerElement.prototype.customStyle;
 
 /**
+ * Convenience method for creating an element and configuring it.
+ * @param {string} tagName HTML tag name
+ * @param {IObject<string, *>=} properties Object of properties to configure on the instance
+ * @return {!Element}
+ */
+PolymerElement.prototype.create = function(tagName, properties) {};
+
+/**
  * Returns the computed style value for the given property.
  * @param {string} property
  * @return {string} the computed value


contrib/externs/polymer-1.0.js
```diff
@@ -646,6 +646,14 @@ PolymerElement.prototype.updateStyles = function(properties) {};
 PolymerElement.prototype.customStyle;
 
 /**
+ * Convenience method for creating an element and configuring it.
+ * @param {string} tagName HTML tag name
+ * @param {Object=} properties Object of properties to configure on the instance
```
ChadKillingsworth: I think this should be `IObject<string, *>`.

contrib/externs/polymer-1.0.js
```diff
@@ -646,6 +646,14 @@ PolymerElement.prototype.updateStyles = function(properties) {};
...
+ * @return {?Element}
```
ChadKillingsworth: Is this really nullable?

contrib/externs/polymer-1.0.js
lukeschaefer: Agreed - updated.

contrib/externs/polymer-1.0.js
lukeschaefer: Nope - I thought it was if the element wasn't registered - but was wrong.

ChadKillingsworth:
Great! Can you squash these commits? Imports into the Google code base only work on a single commit.

lukeschaefer:
Sure thing - but also I believe GitHub can be configured to do squashing by itself:

https://help.github.com/articles/configuring-commit-squashing-for-pull-requests/

ChadKillingsworth:
GitHub can squash on merge. We need squash without merge.

lukeschaefer:
Ah I see - I had already squashed to one commit by the way!

lukeschaefer:
Awesome - thanks for the update!

PR #2453, opened 2017-04-24
Update the default language levels

ChadKillingsworth:
It's becoming increasingly common for modern apps to target no less than IE 11. This updates the command line runner to reflect that reality. Input language default is set to `ECMASCRIPT_NEXT` and the default output language is set to `ECMASCRIPT5`.

### DIFF
--- a/src/com/google/javascript/jscomp/CommandLineRunner.java
+++ b/src/com/google/javascript/jscomp/CommandLineRunner.java
@@ -584,9 +584,9 @@ public class CommandLineRunner extends
           "Sets what language spec that input sources conform. "
               + "Options: ECMASCRIPT3, ECMASCRIPT5, ECMASCRIPT5_STRICT, "
               + "ECMASCRIPT6_TYPED (experimental), ECMASCRIPT_2015, ECMASCRIPT_2016, "
-              + "ECMASCRIPT_NEXT"
+              + "ECMASCRIPT_2017, ECMASCRIPT_NEXT"
     )
-    private String languageIn = "ECMASCRIPT_NEXT";
+    private String languageIn = "ECMASCRIPT_2017";
 
     @Option(
       name = "--language_out",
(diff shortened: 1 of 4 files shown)


src/com/google/javascript/jscomp/CommandLineRunner.java
```diff
@@ -583,15 +583,20 @@
...
-    @Option(name = "--language_out",
-        usage = "Sets what language spec the output should conform to. "
-        + "Options: ECMASCRIPT3 (default), ECMASCRIPT5, ECMASCRIPT5_STRICT, "
-        + "ECMASCRIPT6_TYPED (experimental)")
-    private String languageOut = "ECMASCRIPT3";
+    @Option(
+      name = "--language_out",
+      usage =
+          "Sets what language spec the output should conform to. "
+              + "Options: ECMASCRIPT3, ECMASCRIPT5, ECMASCRIPT5_STRICT, "
```
dasa: Should it still include `(default)` like before?

ChadKillingsworth: No - args4j already outputs that. It was redundant.

src/com/google/javascript/jscomp/CommandLineRunner.java
```diff
@@ -583,15 +583,20 @@
       usage =
           "Sets what language spec that input sources conform. "
               + "Options: ECMASCRIPT3, ECMASCRIPT5, ECMASCRIPT5_STRICT, "
-              + "ECMASCRIPT6 (default), ECMASCRIPT6_STRICT, ECMASCRIPT6_TYPED (experimental)"
+              + "ECMASCRIPT6_TYPED (experimental), ECMASCRIPT_2015, ECMASCRIPT_2016, "
+              + "ECMASCRIPT_NEXT"
     )
-    private String languageIn = "ECMASCRIPT6";
+    private String languageIn = "ECMASCRIPT_NEXT";
```
brad4d: `ECMASCRIPT_NEXT` would include features that aren't part of the current standard yet, so we shouldn't make it the default.
This could be `ECMASCRIPT_2016`, or if you like you could introduce an alias `ECMASCRIPT_CURRENT` that we'll update as new versions of the standard are released.

ChadKillingsworth:
Since this is just setting defaults, and we are going to move pointers yearly anyway, I did not create an `ECMASCRIPT_CURRENT` flag. After thinking about it, I didn't see any value.

Instead I created an `ECMASCRIPT_2017` flag and set it as the default (the feature set was finalized in January). This coming January, we can create an `ECMASCRIPT_2018` flag and set it to the default.

brad4d:
Importing and testing...

brad4d:
passed all the standard tests.
I expect it to be submitted in the morning.

PR #2461, opened 2017-04-27
Add extern for `Document.prototype.readyState`.

bicknellr:


Dominator008:
I tried this one before: https://github.com/google/closure-compiler/pull/1168
It seems that `XMLDOMDocument.prototype.readyState` has type `number` and that conflicts with the `string` type of `Document.prototype.readyState`.

bicknellr:
Ah, thanks; I'll bring this up again over there.

PR #2469, opened 2017-05-01
Update type of angular.$resource to include options

jpevarnek:
$resource has supported options as a fourth parameter since v1.3.0, this
updates the types for resource to reflect that.

https://code.angularjs.org/1.3.0/docs/api/ngResource/service/$resource

### DIFF
--- a/contrib/externs/angular-1.3-resource.js
+++ b/contrib/externs/angular-1.3-resource.js
@@ -26,7 +26,7 @@
  * The $resource service is a factory for creating a resource class.
  *
  * @typedef {function(string, !Object=,
- *     !Object.<string, angular.ResourceAction>=):angular.Resource}
+ *     !Object.<string, angular.ResourceAction>=, Object=):angular.Resource}
  */
 angular.$resource;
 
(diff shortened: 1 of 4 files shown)


brad4d:
Created internal issue b/122075334

brad4d:
Imported and sent for internal testing.

brad4d:
Internal testing looks good.
Internal reviewer pointed out that the argument being added should indicate exactly what properties are expected.
I'm rerunning testing with it changed to look like this.

```javascript
/**
 * @typedef {{
 *     stripTrailingSlashes: (boolean|undefined),
 *     cancellable: (boolean|undefined),
 * }}
 */
angular.$ResourceOptions;

/**
 * The $resource service is a factory for creating a resource class.
 *
 * @typedef {function(
 *       string,
 *       !Object=,
 *       !Object.<string, angular.ResourceAction>=,
 *       !angular.$ResourceOptions=)
 *     :angular.Resource}
 */
angular.$resource;
```
See docs https://docs.angularjs.org/api/ngResource/service/$resource

PR #2472, opened 2017-05-02
Update angular-material typing to clarify type of controller

jpevarnek:
The type of the controller property for a few of the option objects was
listed as (Function|string|undefined), this updates that to reflect that
it can any injectable type rather than just function.  Injectable type
is pulled from existing angular typings.


brad4d:
Imported and running through tests.
Sent to @jelbourn for internal Google review.

brad4d:
Got internal approval for this, but having some trouble getting all the necessary tests run.
I expect them to run overnight tonight.

brad4d:
To be clear, trouble with tests is not due to the content of this PR.

brad4d:
Submitted internally this morning.
If all goes well, it'll get pushed out to github later today.

PR #2482, opened 2017-05-07
Clipboard API and events based on W3C Working Draft 13 December 2016

vobruba-martin:
`ClipboardEvent` introduced in Chrome 58: https://www.chromestatus.com/feature/4687833935642624
Specification: https://www.w3.org/TR/clipboard-apis/

There is also WIP on asynchronous access to this API which will be added later:
https://www.chromestatus.com/feature/5861289330999296
https://w3c.github.io/clipboard-apis/#async-clipboard-api

### DIFF
--- a/externs/browser/w3c_clipboard.js
+++ b/externs/browser/w3c_clipboard.js
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2017 The Closure Compiler Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


brad4d:
We seem to already have externs for this internally, and I cannot immediately see why they aren't being exported to github.
I'll see if I can get that fixed first.

brad4d:
I hit a snag initially because `w3c_clipboardevent.js` depends on another extern file, but
I think I have that sorted & will get it submitted tomorrow morning.

brad4d:
submitted internally
it'll be in tomorrow's push to github

brad4d:
@vobruba-martin 
We now have an open-sourced version of `w3c_clipboardevent.js`, but it's in `contrib/externs`.
Please rework this PR accordingly or file a new one.
Thanks.

vobruba-martin:
@brad4d Is there any reason why not move it into the standard externs folder?

brad4d:
@vobruba-martin 
We don't want to expand the list of default externs files, which includes everything under `externs/browser`.
Most projects don't need these features.
Projects that do should explicitly ask to use these externs files.

brad4d:
@vobruba-martin 
I believe I'm awaiting changes from you.
Shall we just close this PR?

vobruba-martin:
@brad4d 
Ups, I forgot about this PR. I will get back to it in next few days.

PR #2485, opened 2017-05-09
Move Canvas 2D Context externs closer to published spec

dsanders11:
Mainly just looking to add the `CanvasImageSource` typedef as it makes working with `drawImage` easier instead of repeating the type union in code constantly.

### DIFF
--- a/externs/browser/html5.js
+++ b/externs/browser/html5.js
@@ -94,6 +94,11 @@ HTMLCanvasElement.prototype.getContext = function(contextId, opt_args) {};
 HTMLCanvasElement.prototype.captureStream = function(opt_framerate) {};
 
 /**
+ * @typedef {HTMLImageElement|HTMLVideoElement|HTMLCanvasElement}
+ */
+var CanvasImageSource;
+
+/**
  * @interface
  * @see https://www.w3.org/TR/2dcontext/#canvaspathmethods
  */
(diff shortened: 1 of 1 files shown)


dsanders11:
Hmm, looks like `CanvasRenderingContext2D` can't implement `CanvasDrawingStyles` and inherit the properties. I tried to use `closureidl` on the IDL in the spec but it didn't seem to like it and that project is clearly no longer maintained.

Is there a preferred pattern for an IDL interface like `CanvasDrawingStyles` that has properties?

brad4d:
@dsanders11 

`CanvasRenderingContext2D` cannot inherit properties from an interface, because it is a class.
That's obfuscated a bit since this is an externs file, so there's no real implementation, which makes you think of it as an interface.

You'll need to put back the explicit definitions of the properties.
The duplication is ugly, but unavoidable here.
I *think* you can get away with dropping the doc comments for them though.
See the description of `@override` here https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler#override for why I say that.

dsanders11:
Latest build failure appears to be because `master` was already failing with that error.

brad4d:
@dsanders11 
yes, I'm sure the failures are due to an existing breakage.
sorry about that.

ChadKillingsworth:
Toggled open/close to rebuild with travis ...

brad4d:
imported & running tests

brad4d:
Looks like I was wrong about dropping the `@type` annotations from the implementation fields.
It causes errors in Google code because the compiler thinks the field type is unknown.

I've put back the `@type` annotations, and I'm re-running tests with that.

brad4d:
All internal tests passed.
Sent for internal review.

brad4d:
Submitted internally on Friday.
Should be in today's push to github.

PR #2495, opened 2017-05-16
Adds Polymer.ArraySplice.calculateSplices to externs.

nik3daz:


nik3daz:
I signed it!

MatrixFrog:
Hm the googlebot is not updating the label :(

PR #2496, opened 2017-05-16
Make deploy_snapshot.sh executable

MatrixFrog:
Hopefully this will actually fix https://github.com/google/closure-compiler/issues/1689 once and for all.


MatrixFrog:
I think we figured out the magic MOE button to push so nevermind.

PR #2527, opened 2017-06-11
Added externs for chrome.declarativeContent API and chrome.events

IvanRF:
Event methods and types: https://developer.chrome.com/extensions/events
Added `addRules`, `getRules`, `removeRules`, `Rule` type and `UrlFilter`
type.

Added `chrome.declarativeContent` API:
https://developer.chrome.com/extensions/declarativeContent


IvanRF:
@MatrixFrog did you have time to watch this? I'm already using this new externs to minify my Chrome extension

contrib/externs/chrome_extensions.js
```diff
@@ -11016,3 +11016,72 @@ chrome.downloads.DeterminingFilenameEvent.prototype.hasListeners =
...
+/** @type {string} */
+chrome.declarativeContent.SetIconInstanceType;
+
+
+/** @type {string} */
+chrome.declarativeContent.RequestContentScriptInstanceType;
+
+
+/**
+ * @typedef {!{
```
brad4d: Please remove the `!` to make this `@typedef {{` and do the same for other code you're adding or modifying, also removing the `?` from `{?{`.

I know there's already code in here that does this, but let's not add more.
As a matter of style we (Google) have decided that for readability's sake you should always specify nullability explicitly at the point where you **use** a typedef. Given that, setting an explicit default at the point where it is defined is just confusing noise.

contrib/externs/chrome_extensions.js
```diff
@@ -11016,3 +11016,72 @@ chrome.downloads.DeterminingFilenameEvent.prototype.hasListeners =
...
+ *   css: (!Array<string>|undefined),
+ *   isBookmarked: (boolean|undefined)
+ * }}
+ * @see https://developer.chrome.com/extensions/declarativeContent#type-PageStateMatcher
+ */
+chrome.declarativeContent.PageStateMatcher;
+
+
+/**
+ * @constructor
```
brad4d: Do you ever actually write code that says `new ShowPageAction()`, or do you just need to declare parameters and variables with this type? If you never say `new`, then please replace `@constructor` with `@record` here to make this an interface type.

Please ask yourself the same question for all the other new types you've added and update as appropriate.
Thanks.

contrib/externs/chrome_extensions.js
```diff
@@ -11016,3 +11016,72 @@ chrome.downloads.DeterminingFilenameEvent.prototype.hasListeners =
...
+ * @see https://developer.chrome.com/extensions/declarativeContent#type-RequestContentScript
+ */
+chrome.declarativeContent.RequestContentScript;
+
+
+/**
+ * @type {!ChromeEvent}
+ * @see https://developer.chrome.com/extensions/declarativeContent#event-onPageChanged
+ */
+chrome.declarativeContent.onPageChanged;
```
brad4d: please add a terminating newline

externs/browser/chrome.js
```diff
@@ -76,29 +76,170 @@ function ChromeEvent() {}
 /**
  * @param {!Function} callback
  * @return {undefined}
+ * @see https://developer.chrome.com/extensions/events#method-Event-addListener
```
brad4d: Thanks for adding all of these documentation references.
That's great.

externs/browser/chrome.js
```diff
@@ -76,29 +76,170 @@ function ChromeEvent() {}
...
+ * @return {boolean}
+ * @see https://developer.chrome.com/extensions/events#method-Event-hasListeners
+ */
 ChromeEvent.prototype.hasListeners = function() {};
 
-// TODO(tbreisacher): Add the addRules, getRules, and removeRules methods?
 
+/**
+ * @param {!Array<Rule>} rules
+ * @param {function(!Array<Rule>): void=} callback
```
brad4d: Are you allowed to have null rules in the array?
If not, please make this `!Array<!Rule>`.
If so, please make nullability explicit with `!Array<?Rule>`.
Same for all other similar cases.

contrib/externs/chrome_extensions.js
brad4d: I've had some internal Google discussion regarding this comment.
Found out that it's official chrome extensions externs style to use `@typedef {?{...` to indicate that the typedef includes `null`. However, `@typedef {!{...` is redundant and should definitely be avoided.

FYI, I think the policy probably will change to eliminate `{?{` also for the reasons stated above, but for now, you should leave them in for consistency's sake.

IvanRF: OK, so I should only fix the `@typedef {!{...` occurrences, right? do you want me to change older occurrences of this in these files?

brad4d: No need to change older occurrences, just your new ones.

brad4d:
@IvanRF I believe I'm waiting for you to push an update. Let me know if I am mistaken.

IvanRF:
@brad4d Oh, I was waiting a reply for the other two questions I made

brad4d:
@IvanRF I'm glad I asked.
I only see the one question about `{!{` in typedefs.

Maybe you have comments that are somehow in a "draft" mode or something?
Where should I be looking?

IvanRF:
Oh, that was weird, comments were in a "Pending" state, do you see them now?

contrib/externs/chrome_extensions.js
brad4d: You need to change `PageStateMatcher` from a typedef to a class, or the compiler won't let you call `new` on it.

e.g.
```javascript
/** @constructor */
chrome.declarativeContent.PageStateMatcher = function() {};
/** @type {!UrlFilter|undefined} */
chrome.declarativeContent.PageStateMatcher.prototype.pageUrl;
// etc.
```

externs/browser/chrome.js
brad4d: I think you should enforce it (`!Array<!Rule>`), since a null rule makes no sense.
If you don't enforce it, it's easier for bad code to slip in and it becomes harder to enforce it later.

brad4d:
Seen and responded.

contrib/externs/chrome_extensions.js
IvanRF: The compiler worked fine with `PageStateMatcher` as a `@typedef`, I will test it as a class

IvanRF:
@brad4d I have a doubt for https://developer.chrome.com/extensions/events#method-Event-getRules

The first parameter `ruleIdentifiers` is optional and the second one `callback` is not. The compiler does not like that and wants the optional parameters at the end, but that will break the function definition.

contrib/externs/chrome_extensions.js
IvanRF: I'm not sure about changing `PageStateMatcher`, check the docs at the top of the file: `Permissions` is a type too and it's defined as `@typedef`:

>   For #1, use a typedef so object literals and objects created via goog.object are acceptable, for example, the Permissions type defined at http://developer.chrome.com/extensions/permissions.html#type-Permissions should be:
```
   / **
     * at-typedef {?{
     *   permissions: (!Array<string>|undefined),
     *   origins: (!Array<string>|undefined)
     * }}
     * /
   chrome.permissions.Permissions;
```

IvanRF: I tested it as a `class`:
```
/**
 * @constructor
 * @see https://developer.chrome.com/extensions/declarativeContent#type-PageStateMatcher
 */
chrome.declarativeContent.PageStateMatcher = function() {};

/** @type {!UrlFilter|undefined} */
chrome.declarativeContent.PageStateMatcher.prototype.pageUrl;

/** @type {!Array<string>|undefined} */
chrome.declarativeContent.PageStateMatcher.prototype.css;

/** @type {boolean|undefined} */
chrome.declarativeContent.PageStateMatcher.prototype.isBookmarked;
```

and the compiler throws a Warning '... called with 1 argument(s). Function requires ar least 0 argument(s) and no more than 0 argument(s).'

brad4d: The compiler should complain if you do this:

```javascript
// in externs
/** @typedef {{foo: string}} */
var MyType;

// in code
new MyType()
```

It does complain with our new type-inference engine, but apparently not with the old one, which is still the default.

I believe what you really want to do here is this.

```javascript
/**
 * @typedef {{
 *   pageUrl: (!UrlFilter|undefined),
 *   css: (!Array<string>|undefined),
 *   isBookmarked: (boolean|undefined)
 * }}
 * @see https://developer.chrome.com/extensions/declarativeContent#type-PageStateMatcher
 */
chrome.declarativeContent.PageStateMatcherLiteral;

/**
 * @constructor
 * @param {?chrome.declarativeContent.PageStateMatcherLiteral=} literalValue
 */
chrome.declarativeContent.PageStateMatcher = function(literalValue) {};
/** @type {!UrlFilter|undefined} */
chrome.declarativeContent.PageStateMatcher.prototype.pageUrl;
// etc.

// if there is a function 
```

externs/browser/chrome.js
```diff
@@ -76,29 +76,170 @@ function ChromeEvent() {}
...
+/**
+ * @param {!Array<!Rule>} rules
+ * @param {function(!Array<!Rule>): void=} callback
+ * @see https://developer.chrome.com/extensions/events#method-Event-addRules
+ */
+ChromeEvent.prototype.addRules = function(rules, callback) {};
+
+
+/**
+ * @param {Array<string>} ruleIdentifiers
```
brad4d:
```
 * @param {!Array<string>|undefined}
```

brad4d:
@IvanRF added a comment to show how to handle an optional argument that isn't at the end.

brad4d:
@IvanRF BTW the broken travis build is not related to your update. Please ignore it.

contrib/externs/chrome_extensions.js
IvanRF: You are the expert, but that seems redundant.
The current `@typedef` definition for `PageStateMatcher` is working fine with the compiler on my project. In the code I make this call: 
`new chrome.declarativeContent.PageStateMatcher({ pageUrl: { hostEquals: 'host', pathEquals: 'path' } })`
and the compiler successfully maintains the class name and all the parameters' names.

IvanRF: shouldn't we keep the same structure as previous definitions?, like the case of `chrome.permissions.Permissions`

brad4d: As I understand it, the compiler should complain if you do any of these.

```javascript
/** @type {!Array<!chrome.declarativeContent.PageStateMatcher>} */
const rules = [
    new chrome.declarativeContent.PageStateMatcher(15),  // wrong argument type
    chrome.declarativeContent.PageStatematcher({  // missing 'new'
      pageUrl: { hostEquals: 'www.google.com', schemes: ['https'] },
      css: ["input[type='password']"]
    }),
    {  // bare literal where you needed to pass a constructed object.
      pageUrl: { hostEquals: 'www.google.com', schemes: ['https'] },
      css: ["input[type='password']"]
    } ];
```
If you define `PageStateMatcher` as a typedef the compiler won't catch any of these.
It won't catch the first 2 because the old type system basically treats the typedef'ed symbol as an unknown type when it is used in code instead of JSDoc, so it doesn't check it at all.
It won't catch the last one because the typedef actually says that the bare literal is a `PageStateMatcher`.

Does that clarify things at all?

IvanRF: Ok. My doubt is why we need to define the prototypes too? (duplicating the variables and their types). In your example, wouldn't `PageStateMatcherLiteral` be enough to define the class constructor parameters? these variables will never be accessed outside the `PageStateMatcher` class, they are only used as parameters for the class constructor.

IvanRF: @brad4d In the last fix I added the `PageStateMatcherLiteral` and changed `PageStateMatcher`. When I tested it, the compiler showed one warning on my code, and I had to add a `@type {!UrlFilter}` to a variable. So, I think that's a good sign.

IvanRF:
@brad4d let me know if there is something else to change in the pull request.

Related question: Two days ago I received an email from the Chrome Web Store team which confused me. As you mentioned before, you are able to have "some internal Google discussion" which could help since this project is involved.
I published an update to a Chrome extension in which the code was minified using Closure Compiler. The update entered a "pending review" state and mainly what the email said was:

> ...it currently violates the section of the Developer Program Policy relating to Malicious Products as its obfuscating a part of code. Per our policies, where possible, make as much of your code visible in the package as you can. If some of your app's logic is hidden and it appears to be suspicious, we may remove it.

Of course, I quickly uploaded the extension without minification and the revision was approved.

Now my question, when I decided to use Closure Compiler on the extension was because I saw that [Google Publisher Toolbar](https://chrome.google.com/webstore/detail/google-publisher-toolbar/omioeahgfecgfpfldejlnideemfidnkc) minifies the code. Also, the file in which I added some new code here was available in this project: `contrib/externs/chrome_extensions.js` and the only purpose is to use it on a Chrome extension. So I am confused. Every question at StackOverflow says it is ok, but it seems it is not. What do you think?

brad4d:
@IvanRF 
I've asked some people your question internally. Not sure if I'll get an answer, but I'll let you know if I do.

I'd guess the problem might be resolved by disabling variable and property renaming, or even falling back to `SIMPLE_OPTIMIZATIONS` instead of `ADVANCED_OPTIMIZATIONS`.

I'm importing your latest version for testing.

danbeam:
/cc @MatrixFrog

Chrome uses a .json description of this extension API to generate bindings:
https://cs.chromium.org/chromium/src/chrome/common/extensions/api/declarative_content.json

Using this python script, we can generate externs from this .json fairly easily:
https://cs.chromium.org/chromium/src/tools/json_schema_compiler/compiler.py

The results of running
```shell
tools/json_schema_compiler/compiler.py -g externs \
chrome/common/extensions/api/declarative_content.json > declarative_content_externs.js
```
Are: https://gist.github.com/danbeam/15836fea2e2c767318b6cbb004fbe5e4

It might also be worthwhile to update the internal file with these contents.

IvanRF:
@danbeam looking at https://developer.mozilla.org/en-US/docs/Web/API/ImageData the definition is the same as `ImageData`: https://github.com/google/closure-compiler/blob/ea509e4ba9601f273c606d824a7f37b8a600b7e4/externs/html5.js#L524 which is the type I used, no need to declare

brad4d:
@danbeam 
I believe the automatically generated externs are incorrect.
They are defining as `@typedef` symbols that you intend should be called as a constructor with a literal object matching the `@typedef`ed type.

Please see this comment and surrounding discussion for more explanation.
https://github.com/google/closure-compiler/pull/2527#discussion_r127330992

IvanRF:
@brad4d any news regarding this PR?

contrib/externs/chrome_extensions.js
```diff
@@ -11214,3 +11214,80 @@ chrome.downloads.DeterminingFilenameEvent.prototype.hasListeners =
...
+ */
+chrome.declarativeContent.PageStateMatcherLiteral;
+
+
+/**
+ * @constructor
+ * @param {?chrome.declarativeContent.PageStateMatcherLiteral=} literalValue
+ * @see https://developer.chrome.com/extensions/declarativeContent#type-PageStateMatcher
+ */
+chrome.declarativeContent.PageStateMatcher = function(literalValue) {};
```
brad4d: Do you expect to write code like this?

```javascript
let matcher = new chrome.declarativeContent.PageStateMatcher(pageStateLiteral);
...
someFunction(matcher.pageUrl); // reference to object property assigned from the literal
```

If so, you need to declare the properties on `PageStateMatcher`.

contrib/externs/chrome_extensions.js
```diff
@@ -11214,3 +11214,80 @@ chrome.downloads.DeterminingFilenameEvent.prototype.hasListeners =
...
+chrome.declarativeContent.ShowPageAction = function() {};
+
+
+/**
+ * @typedef {{
+ *   imageData: (!ImageData|!Object<number, !ImageData>|undefined)
+ * }}
+ * @see https://developer.chrome.com/extensions/declarativeContent#type-SetIcon
+ */
+chrome.declarativeContent.SetIcon;
```
brad4d: Do you expect to have code that does `new SetIcon()`?
It's not clear to me from the doc reference.

contrib/externs/chrome_extensions.js
```diff
@@ -11214,3 +11214,80 @@ chrome.downloads.DeterminingFilenameEvent.prototype.hasListeners =
...
+/**
+ * @typedef {{
+ *   css: (!Array<string>|undefined),
+ *   js: (!Array<string>|undefined),
+ *   allFrames: (boolean|undefined),
+ *   matchAboutBlank: (boolean|undefined)
+ * }}
+ * @see https://developer.chrome.com/extensions/declarativeContent#type-RequestContentScript
+ */
+chrome.declarativeContent.RequestContentScript;
```
brad4d: Do you expect to have code that does `new RequestContentScript()`?
It's not clear to me from the doc reference.

brad4d:
@IvanRF I thought I was waiting for @danbeam to correct his autogeneration of externs, and then you to use that autogeneration to update this PR.

Assuming you intend to just submit a manually created update here, I've added some questions to the code changes.

contrib/externs/chrome_extensions.js
IvanRF: I don't think so, `PageStateMatcher` is only used as a parameter for rules at `chrome.declarativeContent.onPageChanged`.
Check https://developer.chrome.com/extensions/declarativeContent#rules

BTW, you suggested those changes, I first wrote it as a property: https://github.com/google/closure-compiler/pull/2527/commits/c419db9aa7a3c1b6cd91bd9cc9bc0b6354c85ed3#diff-0dbd98fbf87a9a1918041aecf8035ad0R11052

contrib/externs/chrome_extensions.js
IvanRF: Never used it, but from https://developer.chrome.com/extensions/declarativeContent#rules:

> As a declarative API, this API lets you register rules on the onPageChanged event object which take an action (ShowPageAction and SetIcon)....

It could be deduced that it can be expected (since `new ShowPageAction()` is used).

contrib/externs/chrome_extensions.js
IvanRF: Yes, it used as an action on the `actions` parameter.
`var rule = {..., actions: [new chrome.declarativeContent.RequestContentScript(...) ]}`

IvanRF:
@brad4d thanks for the reply, let me know if new changes are needed.

contrib/externs/chrome_extensions.js
brad4d: I do remember asking you to change that. Thanks for doing so.

I think this is fine then.

contrib/externs/chrome_extensions.js
brad4d: If you need to call `new SetIcon()`, then `SetIcon` must be defined as `@constructor` rather than `@typedef`.

contrib/externs/chrome_extensions.js
brad4d: This needs to be defined as `@constructor` with separate declarations for each property then.

brad4d:
@IvanRF requests for changes added.

In general:
1. if you need to create an instance of something with `new`, you must define it as `@constructor`.
2. The `@record` form of declaring interface types is generally preferable to `@typedef`. It's more verbose, but it's also easier to document the individual properties and the compiler does a better job of type checking with `@record`.  e.g. I've recently been looking at a bug where properties defined with `@typedef` in externs are getting renamed when they really shouldn't. Also since you can extend `@record` types, it's easier to define several related types that way.

danbeam:
> @IvanRF I thought I was waiting for @danbeam to correct his autogeneration of externs, and then you to use that autogeneration to update this PR.

https://bugs.chromium.org/p/chromium/issues/detail?id=772603

IvanRF:
@brad4d I'm testing the new changes and I've found that the `externs.zip` file inside the latest release `closure-compiler-v20170910.jar` **does not include** `chrome.js`. Was that intentional or a mistake on making on the release?
The previous version 20170806 has it.

IvanRF:
@brad4d In the last commit I defined `SetIcon` and `RequestContentScript` as `@constructor`. I maintained the "literal" structure and the `@typedef` since `@record` was never used in that file.

Nevertheless, let me know if you think I should make the 3 `chrome.declarativeContent.*Literal` a `@record` and define each property as `chrome.declarativeContent.*Literal.prototype.*`.

contrib/externs/chrome.js
```diff
@@ -76,29 +76,170 @@ function ChromeEvent() {}
...
+ * @param {Array<string>=} ruleIdentifiers
+ * @param {function(): void=} callback
+ * @see https://developer.chrome.com/extensions/events#method-Event-removeRules
+ */
+ChromeEvent.prototype.removeRules = function(ruleIdentifiers, callback) {};
+
+
+/**
+ * @see https://developer.chrome.com/extensions/events#type-Rule
+ * @constructor
```
brad4d: To get the behavior you want, this needs to be `@record`

I'm looking at the example here:
https://developer.chrome.com/extensions/declarativeContent#type-Rule

It creates rules by just defining objects with the appropriately named properties.
That's what `@record` gets you. The compiler will consider any object with the right properties to have the type.

With `@constructor` the compiler expects to see `new Rule()` or a call to some function that returns a `Rule` in order to get one.

More info here
https://github.com/google/closure-compiler/wiki/Structural-Interfaces-in-Closure-Compiler
and here
https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler#interface-record

IvanRF: @brad4d oh, same goes to `UrlFilter` then, I will change that.

IvanRF: @brad4d done 👍

brad4d:
@IvanRF 
I think this looks pretty good.
I've imported it and started a global presubmit to make sure it doesn't break any Google code.

IvanRF:
@brad4d Excellent, thanks!

brad4d:
Internal testing looks good.
I've sent this to @MatrixFrog for internal review.

contrib/externs/chrome_extensions.js
```diff
@@ -11214,3 +11214,96 @@ chrome.downloads.DeterminingFilenameEvent.prototype.hasListeners =
...
+ *   imageData: (!ImageData|!Object<number, !ImageData>|undefined)
+ * }}
+ * @see https://developer.chrome.com/extensions/declarativeContent#type-SetIcon
+ */
+chrome.declarativeContent.SetIconLiteral;
+
+
+/**
+ * @constructor
+ * @param {?chrome.declarativeContent.SetIconLiteral=} literalValue
```
brad4d: I suspect that if you actually pass a `null` value as the argument for this constructor and the similar ones below you'll get an exception.

Could you please confirm that passing `null` doesn't break things or change the `?` here and below to `!`?

Thanks.

contrib/externs/chrome_extensions.js
```diff
@@ -11214,3 +11214,96 @@ chrome.downloads.DeterminingFilenameEvent.prototype.hasListeners =
...
+chrome.declarativeContent.SetIconInstanceType;
+
+
+/** @type {string} */
+chrome.declarativeContent.RequestContentScriptInstanceType;
+
+
+/**
+ * @typedef {{
+ *   pageUrl: (!UrlFilter|undefined),
```
brad4d: Internal reviewer pointed out that the docs do not say this field is a `UrlFilter`.
Instead, they explicitly list all of the fields.

Do you have any way of knowing whether the "real" requirement is for a `UrlFilter` or if the fields allowed may diverge in the future?

IvanRF: @brad4d It is actually mentioned in the [documentation](https://developer.chrome.com/extensions/declarativeContent#type-PageStateMatcher):

> (optional) pageUrl: Matches if the condition of the **UrlFilter** are fulfilled for the top-level URL of the page.

Other than that, I don't have access to the implementation of `PageStateMatcher` in order to check what is being used.

contrib/externs/chrome_extensions.js
```diff
@@ -11214,3 +11214,96 @@ chrome.downloads.DeterminingFilenameEvent.prototype.hasListeners =
...
+
+/**
+ * @const
+ * @see https://developer.chrome.com/extensions/declarativeContent
+ */
+chrome.declarativeContent = {};
+
+
+/** @type {string} */
+chrome.declarativeContent.PageStateMatcherInstanceType;
```
brad4d: AFAICT the docs don't actually show any use of any of these `*InstanceType` values.
How would you feel about dropping them?

contrib/externs/chrome_extensions.js
IvanRF: @brad4d I tested passing `null` to `PageStateMatcher` and no error was thrown.
Minified version result: `new chrome.declarativeContent.PageStateMatcher(null)`.

When passing `null` to `RequestContentScript` it throws:
`Unchecked runtime.lastError while running events.addRules: Missing parameter is required: css or js`.
So, docs say they are optional parameters but at least one is needed.

When passing `null` to `SetIcon` it throws:
`Error in event handler for runtime.onInstalled: TypeError: Cannot use 'in' operator to search for 'iconIndex' in null`

contrib/externs/chrome_extensions.js
IvanRF: I don't really know if they are used. They are listed in the docs, just like types for [chrome.downloads](https://developer.chrome.com/extensions/downloads#type-FilenameConflictAction).

IvanRF: @brad4d I just found this code: https://github.com/dart-gde/chrome.dart/blob/master/lib/gen/declarative_content.dart
Is that the official repository? It looks like.
If so, following the `chrome.downloads` example, these `*InstanceType` should be `@enum {string}` using the `const` names from there, like `DECLARATIVE_CONTENT_PAGE_STATE_MATCHER`.

brad4d: The difference is that in the chrome.downloads docs `FilenameConflictAction` actually appears as the type for an object property. I don't see anything similar for the declarativeContent instance type definitions.

Also, if we're going to define them, they need to be defined like [ViewType](https://github.com/google/closure-compiler/blob/master/contrib/externs/chrome_extensions.js#L2322).
That's especially weird since these have only a single allowed value and they're not used anywhere.

contrib/externs/chrome_extensions.js
IvanRF: @brad4d Here you have the answer 😃 
https://github.com/dart-gde/chrome.dart/blob/master/lib/gen/declarative_content.dart#L76

> `PageStateMatcher({UrlFilter pageUrl, List<String> css, bool isBookmarked})`

contrib/externs/chrome_extensions.js
IvanRF: I must add that when I defined all the types I used this file as reference:
https://chromium.googlesource.com/chromium/src/+/lkgr/chrome/common/extensions/api/declarative_content.json

contrib/externs/chrome_extensions.js
IvanRF: Also, I found this documentation that seems to be generated from that code:
https://www.dartdocs.org/documentation/chrome/0.9.33/chrome.declarativeContent/PageStateMatcher-class.html

contrib/externs/chrome_extensions.js
IvanRF: @brad4d I updated the `*InstanceType` to respect the code.

contrib/externs/chrome_extensions.js
IvanRF: @brad4d based on the errors thrown, should I change the `?` to `!`?

contrib/externs/chrome_extensions.js
brad4d: I don't think those references count since they are for the dart wrapper around the API rather than the API itself.

contrib/externs/chrome_extensions.js
brad4d: @IvanRF yes, please change `?` to `!` for methods where passing `null` causes an exception.
Thanks.

brad4d:
Internal review revealed that some refactoring is needed so we don't end up having to add the rules-related methods separately to every one of the event types.
I'll make that change separately, then this PR will need to build on top of it.

contrib/externs/chrome_extensions.js
IvanRF: @brad4d done

contrib/externs/chrome_extensions.js
IvanRF: @brad4d I thought that project was from Google too.
If you check the [docs](https://developer.chrome.com/extensions/declarativeContent#type-PageStateMatcher) for `pageUrl` it says:
> Matches if the condition of the **UrlFilter** are fulfilled for the top-level URL of the page.

IvanRF:
@brad4d OK, let me know when the refactoring is finished

brad4d:
@IvanRF 
I've just submitted a refactoring change affecting `chrome.js` and `chrome_extensions.js`.
It should get pushed out to github tomorrow.

The point of the refactoring is to remove lots of duplicated code and avoid having to write lots of duplicates of the new rules-related methods in this pull request.

Sadly, it turned out that there were more special cases to be dealt with.
I believe I need to do another refactoring before we'll be ready to add the rules methods.
See the TODO comments in `chrome_extensions.js` after tomorrow's push to github for more details.

brad4d:
I've discovered that I should have used `@interface` instead of `@record` in my refactoring.
A change is on the way to do that.

brad4d:
I've just sent another change for internal review.
It adds the rule methods in a way that ensures they are inherited by all of the event types defined in `chrome.js` and `chrome_extentions.js`.

Once that is submitted and pushed to github, @IvanRF should sync and resolve conflicts.
Then I'll import them to finish this PR.

brad4d:
@IvanRF 
My addition of the rule methods has been pushed to github.
Please rebase the branch for this PR.
Thanks.

IvanRF:
@brad4d Is `ChromeBaseEvent` the new version of `ChromeEvent`? I moved and added the functions `addRules`, `getRules` and `removeRules` to `ChromeBaseEvent`.

IvanRF:
@brad4d I merged the new changes.

brad4d:
I've imported the latest and am continuing the internal review.

contrib/externs/chrome_extensions.js
brad4d: I'm inclined to agree with you, so I've made the same point to the internal reviewer.

contrib/externs/chrome_extensions.js
```diff
@@ -10262,3 +10262,116 @@ chrome.downloads.DeterminingFilenameEvent = function() {};
...
+ * @see https://developer.chrome.com/extensions/declarativeContent
+ */
+chrome.declarativeContent = {};
+
+
+/**
+ * @enum {string}
+ * @see https://developer.chrome.com/extensions/declarativeContent#type-PageStateMatcherInstanceType
+ */
+chrome.declarativeContent.PageStateMatcherInstanceType = {
```
brad4d: Thanks for figuring out the right format for these `InstanceType` types, but I think we'd be better off dropping them.
Here's my reasoning.

1. Even though it is small having these definitions has a cost in machine parsing time and human maintenance.
2. AFAICT there is no API that requires these values or exposes them to the client of the API. They seem to be totally internal and of no use to the clients of the API.

Can you show me where I'm wrong about the second point?
If not, let's just dump these until the API changes in some way that makes them useful.
Thanks.

brad4d: FWIW, the internal reviewer is also asking to drop these.

contrib/externs/chrome_extensions.js
brad4d: Internal reviewer OK'd leaving this as-is.

contrib/externs/chrome.js
```diff
@@ -143,12 +143,99 @@ Rule.prototype.actions;
...
+
+/** @type {string|undefined} */
+UrlFilter.prototype.urlSuffix;
+
+
+/** @type {!Array<string>|undefined} */
+UrlFilter.prototype.schemes;
+
+
+/** @type {!Array<(number|Array<number>)>|undefined} */
```
brad4d: I believe the inner `Array` also needs to be `!Array`.

contrib/externs/chrome_extensions.js
```diff
@@ -10262,3 +10262,116 @@ chrome.downloads.DeterminingFilenameEvent = function() {};
...
+ * @enum {string}
+ * @see https://developer.chrome.com/extensions/declarativeContent#type-RequestContentScriptInstanceType
+ */
+chrome.declarativeContent.RequestContentScriptInstanceType = {
+  DECLARATIVE_CONTENT_REQUEST_CONTENT_SCRIPT: ''
+};
+
+
+/**
+ * @typedef {{
```
brad4d: See section C of the "Best Practices" documentation comment at the top of this file.
All typedefs need to be declared with `{!{`
Thanks.

contrib/externs/chrome_extensions.js
```diff
@@ -10262,3 +10262,116 @@ chrome.downloads.DeterminingFilenameEvent = function() {};
...
+/**
+ * @constructor
+ * @param {!chrome.declarativeContent.RequestContentScriptLiteral=} literalValue
+ * @see https://developer.chrome.com/extensions/declarativeContent#type-RequestContentScript
+ */
+chrome.declarativeContent.RequestContentScript = function(literalValue) {};
+
+
+/**
+ * @type {!ChromeEvent}
```
brad4d: Please use `ChromeVoidEvent`
We don't want to add any more uses of the loosely typed `ChromeEvent`.

IvanRF:
@brad4d all changes done 👍

brad4d:
Updates imported for internal review.
I believe this will get submitted internally today.

contrib/externs/chrome_extensions.js
```diff
@@ -10262,3 +10262,80 @@ chrome.downloads.DeterminingFilenameEvent = function() {};
...
+ *   isBookmarked: (boolean|undefined)
+ * }}
+ * @see https://developer.chrome.com/extensions/declarativeContent#type-PageStateMatcher
+ */
+chrome.declarativeContent.PageStateMatcherLiteral;
+
+
+/**
+ * @constructor
+ * @param {!chrome.declarativeContent.PageStateMatcherLiteral=} literalValue
```
brad4d: Is this parameter really optional?

brad4d: Same question for similar constructors below.

IvanRF: @brad4d I tested calling the constructor without parameters for `PageStateMatcher` and `RequestContentScript` and no error was thrown.

For `SetIcon` it throws (similar as passing null):
Uncaught TypeError: Cannot use 'in' operator to search for 'iconIndex' in undefined

All parameters are optional according to the documentation, but I can't say for sure if `literalValue` is optional.

brad4d: Thanks a lot for checking that out.
I'll submit this as-is.

brad4d: Actually, I will go ahead and make the arg required for `SetIcon`.
It may save some developers pain.

PR #2536, opened 2017-06-16
Ensure that dependency pruning doesn't remove modules which were force rewritten as ES6

### LINKED ISSUE #2535: Scripts without either import or export are dropped from compilation
The new ES6 module force rewriting is no longer adding `goog.provide` calls and dependency information - the script is thus dropped.

ChadKillingsworth:
Fixes #2535

Modules which do not import or export anything can only be recognized by the fact that they were imported by another module or script. Changes to the force rewriting were pre-parsing the file and marking it as a module body, but the dependency information was not included and then wasn't regenerated because it had already been parsed.


blickly:
Oops, it looks like github is missing a change that got committed yesterday that affects this code. Let me check if this bug is already fixed internally.

ChadKillingsworth:
Closing in favor of fixes by @blickly.

PR #2548, opened 2017-06-28
Adding missing $mdThemingProvider externs

Krun:
Adding missing registerStyles and enableBrowserColor functions to $mdThemingProvider externs.


brad4d:
imported and set for internal review

PR #2596, opened 2017-08-02
Revisit vars after splitting when processing CommonJS Modules

### LINKED ISSUE #2593: Combined var statement with more than two vars doesn't work
Related to #2450, #2579, 8fc82299c45b0a33f19834222acda4850fd77bc0

We just tested and debugged Closure-compiler master with ClojureScript with @anmonteiro and we found out that 8fc82299c45b0a33f19834222acda4850fd77bc0 breaks for example Lodash which worked previously. One of the problematic files is https://unpkg.com/lodash@4.17.4/_baseGetTag.js which has combined var statement with three vars. Only the two first vars are processed.

Here is a simple test case:

foo.js
```js
var first = require('./bar'),
    second = require('./bar'),
    third = require('./bar'),
    fourth = require('./bar'),
    fifth = require('./bar');

module.exports = {};
```
bar.js
```js
module.exports = {};
```

test and output
```
❯ java -jar target/closure-compiler-1.0-SNAPSHOT.jar bar.js foo.js --process_common_js_modules --module_resolution node
var module$bar={};
var module$foo={},
first$$module$foo=module$bar,
second$$module$foo=module$bar,
third=require("./bar"),
fourth=require("./bar"),
fifth=require("./bar");
```

ChadKillingsworth:
Fixes #2593 

Because rewriting commonjs modules is a post-order traversal, we need to specifically revisit vars after splitting multiple declarations apart.

### DIFF
--- a/src/com/google/javascript/jscomp/ProcessCommonJSModules.java
+++ b/src/com/google/javascript/jscomp/ProcessCommonJSModules.java
@@ -694,8 +694,28 @@ public final class ProcessCommonJSModules implements CompilerPass {
           }
           break;
 
+        case VAR:
+        case LET:
+        case CONST:
+          // Multiple declarations need split apart so that they can be refactored into
+          // property assignments or removed altogether.
+          if (n.hasMoreThanOneChild()) {
+            List<Node> vars = splitMultipleDeclarations(n);
+            t.reportCodeChange();
+            for (Node var : vars) {
+              visit(t, var.getFirstChild(), var);
+            }
+          }
+          break;
+
         case NAME:
           {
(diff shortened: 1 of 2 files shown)


anmonteiro:
Tested this patch and fixes my previously failing tests 👍

dimvar:
LGTM. I'll submit this internally.

dimvar:
FYI, this broke an internal project and was rolled back. I'll try to fix it and roll it forward tomorrow.

dimvar:
The fix was to avoid splitting a VAR inside a FOR header. Hope to roll this forward and push externally today.

Issue #2586, opened 2017-07-31
Some UMD wrappers are not detected


Deraen:
I've hit problems with a few libraries which use UMD wrappers which Closure doesn't detect:

Leaflet https://cdnjs.cloudflare.com/ajax/libs/leaflet/1.1.0/leaflet-src.js:
```js
(function (global, factory) {
	typeof exports === 'object' && typeof module !== 'undefined' ? factory(exports) :
	typeof define === 'function' && define.amd ? define(['exports'], factory) :
	(factory((global.L = global.L || {})));
}(this, (function (exports) {
   exports.foo = "bar";
})));
```

Bowser https://github.com/lancedikson/bowser/blob/master/src/bowser.js:
```js
!function (root, name, definition) {
  if (typeof module != 'undefined' && module.exports) module.exports = definition()
  else if (typeof define == 'function' && define.amd) define(name, definition)
  else root[name] = definition()
}(this, 'bowser', function () {
    return {foo: 'bar'};
});
```

ChadKillingsworth:
There are so many of these, it's very hard to properly detect. It'll probably be quite some time before I can look into this.

ChadKillingsworth:
Most of these variations are now properly rewritten.

PR #2597, opened 2017-08-02
Improved UMD wrapper support

Deraen:
WIP: #2586 

## Before:

Leaflet wrapper, UMD check is not removed:
```
Expected: goog.provide("module$other");goog.provide("module$yet_another");goog.provide("module$test");var module$test={foo:"bar"}
Result:   goog.provide("module$other");goog.provide("module$yet_another");goog.provide("module$test");var module$test={};(function(global,factory){typeof module$test==="object"&&typeof module!=="undefined"?factory(module$test):typeof define==="function"&&define.amd?define(["exports"],factory):factory(global.foobar=global.foobar||{})})(this,function(exports){exports.foo="bar"})
```

Bowser wrapper, IIFE is not removed:
```
Expected: goog.provide("module$other");goog.provide("module$yet_another");goog.provide("module$test");var module$test={foo:"bar"}
Result:   goog.provide("module$other");goog.provide("module$yet_another");goog.provide("module$test");var module$test={};!function(root,name,definition){var module$test=definition()}(this,"foobar",function(){return{foo:"bar"}})
```

## After

Leaflet wrapper

Problem with this one was use of ternary operator instead of `if` statements. I fixed this by adding a `isHook` check to `getOutermostIfAncestor`.

```
Expected: goog.provide("module$other");goog.provide("module$yet_another");goog.provide("module$test");var module$test={foo:"bar"}
Result:   goog.provide("module$other");goog.provide("module$yet_another");goog.provide("module$test");var module$test={};var factory$$module$test=function(exports){exports.foo="bar"};factory$$module$test(module$test)
```

Bowser wrapper

Problem with this one was that the IIFE was created using `!` instead of wrapping the expression in parentheses. I fixed this by adding check on `removeIIFEWrapper` to check if node is Not expression and select the first children instead. Probably this should also support other ways to invoke IIFEs? Like `+`, `-`, `~`, `void`. Not sure if best way to support this would be to check for these five tokens or something else? Some of these can also be repeated but no reasonable wrapped does that...

```
Expected: goog.provide("module$other");goog.provide("module$yet_another");goog.provide("module$test");var module$test={foo:"bar"}
Result:   goog.provide("module$other");goog.provide("module$yet_another");goog.provide("module$test");var definition$$module$test=function(){return{foo:"bar"}};var module$test=definition$$module$test()
```

## Next

I didn't yet try to run the code generated with these changes, but my understanding is that it won't work yet. I think the return value of function call in `module.exports` assignment should be inlined in case of Bowser, and the function body lifted and exported names renamed in case of Leaflet:

```
// leaflet
function foo(a) {
  a.foo = 'bar';
}
foo(exports);
// =>
var module$test = {};
module$test.a = 'bar';

// bowser
function foo() {
  return {foo: 'bar'};
}
module.exports = foo();
// =>
var module$test = {foo: 'bar'};
```

I will need some help to continue with this (ping @ChadKillingsworth). My understanding is that this replacement would need to be done after `removeIIFEWrapper` in `process`?

### DIFF
--- a/test/com/google/javascript/jscomp/ProcessCommonJSModulesTest.java
+++ b/test/com/google/javascript/jscomp/ProcessCommonJSModulesTest.java
@@ -643,6 +643,22 @@ public final class ProcessCommonJSModulesTest extends CompilerTestCase {
     testModules(
         "test.js",
         LINE_JOINER.join(
+            "!function(){",
+            "var foobar = {foo: 'bar'};",
+            "if (typeof module === 'object' && module.exports) {",
+            "  module.exports = foobar;",
+            "} else if (typeof define === 'function' && define.amd) {",
+            "  define([], function() {return foobar;});",
+            "} else {",
+            "  this.foobar = foobar;",
+            "}}()"),
+        LINE_JOINER.join(
+            "goog.provide('module$test');",
+            "var module$test = {foo: 'bar'};"));
+
+    testModules(
+        "test.js",
+        LINE_JOINER.join(
             ";;;(function(){",
             "var foobar = {foo: 'bar'};",
             "if (typeof module === 'object' && module.exports) {",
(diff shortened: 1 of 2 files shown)


ChadKillingsworth:
@tadeegan I've made an attempt at this. You might want to test out the commit I pushed while I clean up the code more.

tadeegan:
@ChadKillingsworth Did you mean to tag me? I don't think I'm involved with this.

ChadKillingsworth:
@tadeegan I did not - sorry about that.

I meant to tag @Deraen

test/com/google/javascript/jscomp/ProcessCommonJSModulesTest.java
```diff
@@ -945,6 +945,7 @@ public void testIssue2593() {
             "var fifth$$module$test=5;"));
   }
 
+<<<<<<< HEAD
```
Deraen: Conflict marker here

Deraen:
@ChadKillingsworth Great! After removing the problematic merge conflict marker I was able to build  package and test this.

The first problem I now see with Leaflet is that 

line 1796:
```
var webkit = userAgentContains('webkit');
```

line 1905:
```
function userAgentContains(str) { ... }
```

doesn't work because the function is rewritten to `var` so function hoisting doesn't work. But this is probably a separate problem? Unless functions are supposed to be moved but it doesn't work in this case.

I was able to workaround these hoisting problems by manually moving `userAgentContains` & others, after this it looks like Leaflet is correctly evaluated in browser. Found some problems with React-leaflet now however :)

ChadKillingsworth:
We hoist functions that are rewritten, so it could very well be a bug introduced by this change. Can you simplify the leaflet issue enough that it can be added to a test?

Deraen:
```
❯ cat leaflet2.js 
(function (global, factory) {
	typeof exports === 'object' && typeof module !== 'undefined' ? factory(exports) :
	typeof define === 'function' && define.amd ? define(['exports'], factory) :
	(factory((global.L = {})));
}(this, (function (exports) { 'use strict';
var webkit = userAgentContains('webkit');
function userAgentContains(str) {
	return navigator.userAgent.toLowerCase().indexOf(str) >= 0;
}
})));

~/Source/closure-compiler umd-wrappers*
❯ java -jar target/closure-compiler-1.0-SNAPSHOT.jar --js leaflet2.js  --process_common_js_modules
var module$leaflet2,
webkit$jscomp$inline_4=userAgentContains$jscomp$inline_5("webkit"),
userAgentContains$jscomp$inline_5=function(a){
  return 0<=navigator.userAgent.toLowerCase().indexOf(a)
};
```

Deraen:
Okay, so without the wrapper, function is not converted to var.

Deraen:
I opened new issue about one problem with React-leaflet, doesn't seem to me that it would be related to this but I can't be sure: https://github.com/google/closure-compiler/issues/2634

Another problem with Lodash-es6 seems like it might be related to changes to UMD wrapper tests here (could have been problem before also).

Lodash-es modules contain code that checks `exports`, and I think this triggers UMD wrapper code and disables ES6 import handling:

```
❯ cat anything.js 
export default "anything";

❯ cat lodashProblem.js 
import anything from './anything.js';

var haha = typeof exports == 'object' && exports && exports;

export default {};


❯ java -jar target/closure-compiler-1.0-SNAPSHOT.jar --js anything.js lodashProblem.js  --process_common_js_modules --module_resolution NODE -O WHITESPACE_ONLY

goog.provide("module$anything");
var $jscompDefaultExport$$module$anything="anything";
module$anything.default=$jscompDefaultExport$$module$anything;

goog.provide("module$lodashProblem");
var module$lodashProblem={};
import anything from"./anything.js";
var haha=typeof module$lodashProblem=="object"&&module$lodashProblem&&module$lodashProblem;
export default{};
```

ChadKillingsworth:
The hoisting issue was due to the FunctionToBlockMutator which is what I use to inline the functions. I've added a fix in #2636.

ChadKillingsworth:
@Deraen The function hoist fix landed and I've merged in master. Want to try again?

Deraen:
I think I tested this already with that fix already, and Leaflet worked. I'll check again tomorrow :)

ChadKillingsworth:
Great - then we just need to squash all these commits. Can you handle that? If I do it, your contribution would be lost.

Deraen:
Okay, done. Thank you very much for finishing this!

I tested this with Leaflet again, and it worked.

Deraen:
@ChadKillingsworth Is this OK now? I thought we could also split this to two PR's. I could change my part so it doesn't add broken tests.

ChadKillingsworth:
Yeah this LGTM. Just need a Google engineer to import it and review it.

PR #2598, opened 2017-08-03
Add browser and module field support to package.json processing

### LINKED ISSUE #2433: Add browser and module field support to package.json processing
Closure-compiler uses the standard node field `main` in a package.json to locate the entry point of a package. However, several other defacto standards have emerged including the `module` and `browser` fields. We need to honor those.

Also, it's become increasingly common for build tooling to add their own field and use that. Angular is using `es2015:main` and `esnext:main` fields. We should provide a flag to allow specifying a custom set and order of fields to search.

See the discussion in #2413

anmonteiro:
This PR fixes #2433 by introducing a new option, `packageJsonEntryNames`, which allows specifying a list of ordered entries which the compiler should look for when resolving Node modules. The default is the single entry `"main"`.

When compiling for the web, however, we would pass in a list of `"browser", "main"` which tells the compiler to look for a `"browser"` entry before checking the `"main"` entry.

This also allows people who want to consume packages that instead rely on `es2015:main` or `esnext:main` to include those too in the order of precedence they prefer.

### DIFF
--- a/src/com/google/javascript/jscomp/CompilerOptions.java
+++ b/src/com/google/javascript/jscomp/CompilerOptions.java
@@ -1169,6 +1169,9 @@ public class CompilerOptions implements Serializable {
   /** Which algorithm to use for locating ES6 and CommonJS modules */
   ModuleLoader.ResolutionMode moduleResolutionMode;
 
+  /** Which entries to look for in package.json files when processing modules */
+  List<String> packageJsonEntryNames;
+
   /**
    * Should the compiler print its configuration options to stderr when they are initialized?
    *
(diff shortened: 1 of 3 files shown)


anmonteiro:
I signed it!

ChadKillingsworth:
The PR itself looks good - a flag can be added later for CLI usage.

@anmonteiro Do you have any idea how prevalent the browser field advanced usage form is? https://github.com/defunctzombie/package-browser-field-spec#replace-specific-files---advanced

That's been my largest hangup on this to date.

anmonteiro:
Interesting. I didn't know about the advanced usage of the browser field. 

I have no idea how prevalent it is, but my intuition would be to solve that as needed.

This PR introduces basic support for it,  and we can build on it later.

ChadKillingsworth:
@dimvar LGTM. We'll continue to iterate on this in future PRs.

dimvar:
OK, I'll import and submit the PR internally.

PR #2600, opened 2017-08-04
Add `--package_json_entry_names` command line flag

anmonteiro:
This PR builds on #2598 and adds a command line flag that allows specifying the ordered entries which `RewriteJsonToModule` should look for when computing the main entry in `package.json` files.

refs #2433

### DIFF
--- a/src/com/google/javascript/jscomp/CommandLineRunner.java
+++ b/src/com/google/javascript/jscomp/CommandLineRunner.java
@@ -767,6 +767,13 @@ public class CommandLineRunner extends
     )
     private ModuleLoader.ResolutionMode moduleResolutionMode = ModuleLoader.ResolutionMode.BROWSER;
 
+    @Option(name = "--package_json_entry_names",
+        usage = "Ordered list of entries to look for in package.json files when processing "
+        + "modules with the NODE module resolution strategy (i.e. esnext:main,browser,main). "
+        + "Defaults to a list with the following entries: \"browser\", \"module\", \"main\"."
+    )
+    private String packageJsonEntryNames = null;
+
     @Argument
     private List<String> arguments = new ArrayList<>();
     private final CmdLineParser parser;
@@ -844,7 +851,8 @@ public class CommandLineRunner extends
                     "js_module_root",
                     "module_resolution",
                     "process_common_js_modules",
-                    "transform_amd_modules"))
+                    "transform_amd_modules",
+                    "package_json_entry_names"))
             .putAll(
                 "Library and Framework Specific",
                 ImmutableList.of(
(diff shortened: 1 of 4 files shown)


dimvar:
FYI, I won't have time to look at this today.

anmonteiro:
Thanks for setting expectations. No rush.

dimvar:
Also, you may be amused to know that your previous PR crashed the tool we use to bring PRs internally, because of the unicode character ó in your name :) So I had to edit the script to add the name without the accent.

Consider using the name without the accent in your git settings to avoid this in the future. (Up to you, we can work around this, just thought I'd let you know.)

anmonteiro:
that's just too funny 😃 

I can do that in my git settings, I should probably amend this commit then.

anmonteiro:
Just amended this commit to not include the ó character in the author line.

alexeagle:
+1 we need this for Angular as well, @jasonaden and I built locally and tested today.

anmonteiro:
@ChadKillingsworth fixed and squashed commits. Any other notes?

src/com/google/javascript/jscomp/CommandLineRunner.java
```diff
@@ -767,6 +767,13 @@
     )
     private ModuleLoader.ResolutionMode moduleResolutionMode = ModuleLoader.ResolutionMode.BROWSER;
 
+    @Option(name = "--package_json_entry_names",
+        usage = "Ordered list of entries to look for in package.json files when processing "
+        + "modules with the NODE module resolution strategy (i.e. esnext:main,browser,main). "
+        + "Defaults to a list with a single entry, \"main\"."
```
ChadKillingsworth: Need to update this comment.

anmonteiro: Nice catch. fixed.

src/com/google/javascript/jscomp/CompilerOptions.java
```diff
@@ -1173,12 +1173,6 @@ public void setWrapGoogModulesForWhitespaceOnly(boolean enable) {
   List<String> packageJsonEntryNames;
 
   /**
-   * Needed by {@link RewriteJsonToModule}, but defined here because RewriteJsonToModule is not
-   * part of the core build.
-   */
-  public static final String PACKAGE_JSON_MAIN = "main";
```
anmonteiro: this is not actually used anywhere else, since we generalized package.json fields.

ChadKillingsworth: Then I'd just remove it.

anmonteiro: It's removed. My comment is to justify the deletion.

PR #2604, opened 2017-08-05
Add support for browser field advanced usage to package.json processing

### LINKED ISSUE #2433: Add browser and module field support to package.json processing

anmonteiro:
this is the last milestone in adding full support for the browser field when processing `package.json` files.

Fixes #2433 

cc @ChadKillingsworth

### DIFF
--- a/src/com/google/javascript/jscomp/RewriteJsonToModule.java
+++ b/src/com/google/javascript/jscomp/RewriteJsonToModule.java
@@ -18,6 +18,7 @@ package com.google.javascript.jscomp;
 import static com.google.common.base.Preconditions.checkState;
 
 import com.google.common.collect.ImmutableMap;
+import com.google.javascript.jscomp.deps.NodeModuleResolver;
 import com.google.javascript.rhino.IR;
 import com.google.javascript.rhino.Node;
 import java.util.HashMap;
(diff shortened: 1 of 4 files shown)


src/com/google/javascript/jscomp/RewriteJsonToModule.java
```diff
@@ -36,6 +36,8 @@
     implements CompilerPass {
   public static final DiagnosticType JSON_UNEXPECTED_TOKEN =
       DiagnosticType.error("JSC_JSON_UNEXPECTED_TOKEN", "Unexpected JSON token");
+  public static final String JSC_BROWSER_BLACKLISTED_MARKER = "$jscomp$browser$blacklisted";
```
blickly: It's not easy for us to add a dependency from the deps package onto the jscomp package.  Is it possible to move the definition of JSC_BROWSER_BLACKLISTED_MARKER into the deps package?

anmonteiro: just did this in a separate commit for easier review. Let me know if it looks good and I'll squash them

blickly:
LGTM, thanks!

anmonteiro:
Awesome. Just rebased & squashed commits.

blickly:
Nice, importing this now.

PR #2633, opened 2017-08-30
Fixed maven not cleaning generated sources

havrikov:
Executing `mvn clean` will now properly delete  the `gen/target` directory.  
(I had problems with IntelliJ Idea due to old files duplicating this way.)


havrikov:
I signed it!

PR #2636, opened 2017-08-30
Hoist rewritten function expressions to the top of the block

ChadKillingsworth:
When using the FunctionToBlockMutator, any nested functions are rewritten to an expression. However they are not currently hoisted. Proper hoisting would be difficult in this case, but we can at least move them to the top of the defining block.

### DIFF
--- a/src/com/google/javascript/jscomp/FunctionToBlockMutator.java
+++ b/src/com/google/javascript/jscomp/FunctionToBlockMutator.java
@@ -21,11 +21,13 @@ import static com.google.common.base.Preconditions.checkState;
 import static com.google.common.base.Strings.isNullOrEmpty;
 import static com.google.javascript.jscomp.FunctionArgumentInjector.THIS_MARKER;
 
+import com.google.common.base.Predicate;
 import com.google.common.base.Supplier;
 import com.google.javascript.jscomp.MakeDeclaredNamesUnique.InlineRenamer;
 import com.google.javascript.rhino.IR;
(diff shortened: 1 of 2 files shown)


src/com/google/javascript/jscomp/FunctionToBlockMutator.java
```diff
@@ -21,11 +21,13 @@
 import static com.google.common.base.Strings.isNullOrEmpty;
 import static com.google.javascript.jscomp.FunctionArgumentInjector.THIS_MARKER;
 
+import com.google.common.base.Predicate;
```
blickly: Looks unused

src/com/google/javascript/jscomp/FunctionToBlockMutator.java
```diff
@@ -180,14 +182,31 @@ private static void rewriteFunctionDeclarations(Node n) {
...
     }
 
+    // Keep track of any rewritten functions and hoist them to the top
+    // of the block they are defined in. This isn't fully compliant hoisting
+    // but it does address a large set of use cases.
+    List<Node> functionsToHoist = new ArrayList<>();
     for (Node c = n.getFirstChild(), next; c != null; c = next) {
       next = c.getNext(); // We may rewrite "c"
-      rewriteFunctionDeclarations(c);
+      Node fnExpr = rewriteFunctionDeclarations(c);
```
blickly: Technically not an expression. I'd maybe call this fnVar or similar.

PR #2637, opened 2017-08-31
Restore ES6 module processing of input

thelgevold:
With this I am proposing a fix to a problem we have been having when integrating Angular builds with versions of Closure Compiler > 20170409.0.0. 

It appears that Input processing of ES6 modules was removed in the following commit: https://github.com/google/closure-compiler/commit/d24d175572ebf97b06e480f04c3f452bb6355afa

Putting parts of this back fixes the issue.

The test repo for this can be found here: https://github.com/angular/closure-demo

Some of the input files in the flag file are written in ES6 JavaScript (all the Angular bundles). 

With the current release of CC the ES6 input files are not resolved. The compilation fails with errors messages like: 
```
ERROR - required "module$$angular$core" namespace never provided
import * as import0 from '@angular/core';
```
cc: @alexeagle

### DIFF
--- a/src/com/google/javascript/jscomp/Compiler.java
+++ b/src/com/google/javascript/jscomp/Compiler.java
@@ -1785,6 +1785,11 @@ public class Compiler extends AbstractCompiler implements ErrorHandler, SourceFi
         if (!inputsToRewrite.isEmpty()) {
           forceToEs6Modules(inputsToRewrite.values());
         }
+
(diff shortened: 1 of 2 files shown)


ChadKillingsworth:
@blickly I thought this was fixed?

ChadKillingsworth:
The reason this was originally changed was that anytime ES6 or higher was specified as the input language, a full parse of every input file was required. We don't want to undo that fix.

I'm closing this in favor of #2641 - but it may take a bit to get that one landed.

PR #2641, opened 2017-09-04
Use a deterministic depth-first order when managing dependencies

### LINKED ISSUE #2247: ES6 module with no exports is not found with --js_module_root
Repro:

```
$ npm install google-closure-compiler@20161201.0.0
$ cat user.js
import 'rxjs/add_first';

$ cat stripme/rxjs/add_first.js
import { Observable } from './Observable';
import { first } from './operator_first';
Observable.prototype.first = (first);

$ cat stripme/rxjs/operator_first.js
export function first() {}

$ cat stripme/rxjs/Observable.js
export class Observable {}

$ java -jar node_modules/google-closure-compiler/compiler.jar user.js stripme/rxjs/*.js --entry_point=user.js --js_module_root=stripme
user.js:1: ERROR - required "module$rxjs$add_first" namespace never provided
import 'rxjs/add_first';
^^^^^^^^^^^^^^^^^^^^^^^^

1 error(s), 0 warning(s)
```

It could be worked around in these ways:

1) add a trivial export to stripme/rxjs/add_first:
```
$ cat >> stripme/rxjs/add_first.js
export var a;

$ java -jar node_modules/google-closure-compiler/compiler.jar user.js stripme/rxjs/*.js --entry_point=user.js --js_module_root=stripme
var module$rxjs$Observable={},Observable$$module$rxjs$Observable=function(){};module$rxjs$Observable.Observable=Observable$$module$rxjs$Observable;var module$rxjs$operator_first={};function first$$module$rxjs$operator_first(){}module$rxjs$operator_first.first=first$$module$rxjs$operator_first;var module$rxjs$add_first={};Observable$$module$rxjs$Observable.prototype.first=module$rxjs$operator_first.first;var a$$module$rxjs$add_first;module$rxjs$add_first.a=a$$module$rxjs$add_first;
```

2) move the rxjs directory out of stripme/ and remove the `--js_module_root` argument.

### LINKED ISSUE #2362: ES6 modules can be reordered incorrectly
I have a project which seems to make Closure Compiler reorder ES6 modules improperly, breaking functionality.

A small reproduction of this issue can be found at https://github.com/azakus/closure-reorder-import

In this reproduction, I use rollup and the npm package for closure on the same entry point. Rollup will produce the expected output, while closure seems to reorder the modules and produce an incorrect output

@ChadKillingsworth

### LINKED ISSUE #2535: Scripts without either import or export are dropped from compilation

### LINKED ISSUE #2559: Dependency scanning does not support import statements on more than one line
I just updated to 20170626, and I get this error while compiling:

> ERROR - required "...." namespace never provided

20170521 works just fine.

### LINKED ISSUE #2590: LOOSE mode includes ES6 modules that aren't imported by the entry point
I'm trying to convert our code from `goog.provide` modules to ES6 modules, but I've run into a difference in behavior that I didn't expect. When running in `LOOSE` dependency mode, the compiler includes all ES6 modules regardless of whether they're a dependency of the entry point. When I used `goog.provide`, these files were excluded from compilation. Based on https://github.com/google/closure-compiler/wiki/Managing-Dependencies#working-with-files-that-do-not-specify-dependencies, I'd expect the ES6 modules to work the way `goog.provide` modules do.

Running in `STRICT` mode eliminates files that aren't imported by the entry point, but it also eliminates non-module files (as expected). We have some non-module legacy files that we include in some of our builds, so I don't want to switch to `STRICT` if I can avoid it.

Can you make ES6 modules work like `goog.provide`, or at least clarify that section of the wiki?

Thanks!

ChadKillingsworth:
**Warning:** this is potentially a high-impact change.

Due to the history of ordering dependencies as scripts with `goog.provide` and `goog.require` calls, the dependency sorting could produce incorrect results when using ES6 or CommonJS modules. In addition, the regex based parsing of dependencies did not recognize destructuring imports that spanned more than one line - and these are quite common external to Google.

This change begins with entry points and fully parses each dependency as discovered. The order a dependency reference is encountered is preserved and the module sorting later performed now honors that ordering.

Benefits of this change:

 * Avoids a full parse on every input when CommonJS module processing is enabled.
 * ES6 and CommonJS modules no longer generate synthetic goog.require and goog.provide calls.
 * ES6 module transpilation no longer depends on Closure-Library primitives.
 * CommonJS module rewriting no longer depends on ES6 input

Side effects of this change:

 * Multi-threaded parsing will now only apply to externs.

Fixes #2590 
Fixes #2535 
Fixes #2559 
Fixes #2362
Fixes #2247

### DIFF
--- a/src/com/google/javascript/jscomp/Compiler.java
+++ b/src/com/google/javascript/jscomp/Compiler.java
@@ -64,7 +64,6 @@ import java.io.PrintStream;
 import java.io.Serializable;
 import java.util.AbstractSet;
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -1749,7 +1748,19 @@ public class Compiler extends AbstractCompiler implements ErrorHandler, SourceFi
                   options.moduleResolutionMode,
                   processJsonInputs(inputs));
         }
+      } else {
+        // Use an empty module loader if we're not actually dealing with modules.
+        this.moduleLoader = ModuleLoader.EMPTY;
+      }
 
+      if (options.getDependencyOptions().needsManagement()) {
+        findDependenciesFromEntryPoints(
+            options.getLanguageIn().toFeatureSet().has(Feature.MODULES),
+            options.processCommonJSModules,
+            options.transformAMDToCJSModules);
+      } else if (options.needsTranspilationFrom(FeatureSet.ES6_MODULES)
+          || options.transformAMDToCJSModules
+          || options.processCommonJSModules) {
         if (options.getLanguageIn().toFeatureSet().has(Feature.MODULES)) {
           parsePotentialModules(inputs);
         }
@@ -1782,12 +1793,12 @@ public class Compiler extends AbstractCompiler implements ErrorHandler, SourceFi
           }
         }
 
-        if (!inputsToRewrite.isEmpty()) {
-          forceToEs6Modules(inputsToRewrite.values());
+        for (CompilerInput input : inputsToRewrite.values()) {
+          forceInputToPathBasedModule(
+              input,
+              options.getLanguageIn().toFeatureSet().has(Feature.MODULES),
+              options.processCommonJSModules);
         }
-      } else {
-        // Use an empty module loader if we're not actually dealing with modules.
-        this.moduleLoader = ModuleLoader.EMPTY;
       }
 
       orderInputs();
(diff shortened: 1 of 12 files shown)


thelgevold:
Thanks @ChadKillingsworth 
I just pulled down this PR and confirmed that it fixes the original issue described in https://github.com/google/closure-compiler/pull/2637

ChadKillingsworth:
@blickly I haven't been able to create a repo case where base.js is ordered incorrectly. I did add a couple of test cases to prove dependency ordering is not dependent on the user provided input order.

The one oddity I found was https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/JSModuleGraph.java#L542-L545
While I can't figure out why it would cause a problem, the order of entry points matters. And always adding base.js as an entry point seems "odd".

ChadKillingsworth:
One other note: historical precedence adds moochers to the front of the entry point list and user defined entry points at the end. If moochers are traditionally tests, should this order be reversed?

blickly:
I'm seeing cases where this is incorrectly ordering base.js after other files.  For example, with the following setup.

```javascript
// bootstrap.js
goog.require('foo');
```

```javascript
// foo.js
goog.provide('foo');
goog.setTestOnly();
```

When running the compiler with sorting but no pruning with the files passed in the order:
`--js=bootstrap.js --js=closure/base.js --js=foo.js`
we were previously getting the (correct) ordering of
`base, foo, bootstrap`,
but after this PR, we're getting an ordering of
`foo, bootstrap, base`

blickly:
OK, looks resolved in the current version. Thanks Chad!
Now I'm just seeing issues that need to be fixed on our end...

alexeagle:
@blickly what's your priority for merging this? Angular can't release our version 5.0 right now because ES6 modules are broken. We are hoping to cut an RC next week and might need to escalate.
/cc @IgorMinar

MatrixFrog:
We're working on an issue with some internal tests that have done some unusual things.

alexeagle:
Any estimate of how many tests, and how broken they are? Do you need help to land the CL?

blickly:
Sure, never hurts to have more help. Thanks!
I've cc-ed you a task on the internal tracker bug.

Ramblurr:
Could [this behavior](https://stackoverflow.com/questions/46430494/seemingly-simple-import-module-case-not-compiling-with-namespace-never-provided) be cause by this bug?

ChadKillingsworth:
@Ramblurr Yes definitely

Ramblurr:
@ChadKillingsworth good to know. I've checked out this PR and while the original error message is gone, I've ran into another one. See the **edited** portion of the SO question I just added. Could it still be related to this PR?

lauraharker:
This PR was rolled back in https://github.com/google/closure-compiler/commit/22240a5551da2c426c1e5ddf9cca98dc5a5ff3c1 after it broke some internal projects. @MatrixFrog is working on a fix.

PR #2668, opened 2017-10-03
Hide and mark as deprecated the --transform_amd_modules flag

ChadKillingsworth:
I wish to delete the TransformAMDToCJSModule pass. As a first step, indicate that the open source flag is deprecated and remove it from the help display.


dimvar:
LGTM. I'll bring it in.

PR #2673, opened 2017-10-05
Fix arity of setInterval and setImmediate.

PaulCapron:
setInterval() and setImmediate(), like setTimeout(), take a fixed
number of mandatory arguments, and, if other arguments are present
after them, they are passed to the handler when it is called.

See https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#timers
and https://w3c.github.io/setImmediate/#si-setImmediate

The current externs definition of setTimeout is correct in this regard.
But this is not the case for the definitions of setInterval/setImmediate.
This commit fixes that.

### DIFF
--- a/externs/browser/window.js
+++ b/externs/browser/window.js
@@ -172,20 +172,22 @@ function prompt(message, opt_value) {}
 
 /**
  * @param {function()} callback
+ * @param {...*} var_args
  * @return {number}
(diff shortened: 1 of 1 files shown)


dimvar:
LGTM. I'll submit this internally.

PR #2674, opened 2017-10-08
Robust source map resolver (prevent exception when source map contains invalid entry)

gregmagolan:
Change prevents exception in SourceMapResolver when there is an invalid entry in the source map such as `"../../"` seen here https://github.com/angular/closure-demo/issues/25.

### DIFF
--- a/src/com/google/javascript/jscomp/SourceMapResolver.java
+++ b/src/com/google/javascript/jscomp/SourceMapResolver.java
@@ -15,6 +15,7 @@
  */
 package com.google.javascript.jscomp;
 
+import static com.google.common.base.Strings.isNullOrEmpty;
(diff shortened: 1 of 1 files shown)


lauraharker:
The compiler shouldn't be throwing an exception for this, yes.

Can you add some unit tests to https://github.com/google/closure-compiler/blob/b33fc4f/test/com/google/javascript/jscomp/SourceMapResolverTest.java? 

It looks like the compiler allows some invalid source map entries right now, but crashes on yours. Possibly we should still return a SourceFile instead of null from SourceMapResolver.getRelativePath for any paths passed in, or should more consistently handle invalid paths.

brad4d:
Created internal issue b/122075799

PR #2680, opened 2017-10-19
Update History API extern.

axls:
Added members missed according to [1]:

* length;
* go;
* back;
* forward;

[1] http://w3c.github.io/html/browsers.html#the-history-interface


axls:
Closing. I should have run tests localy before making PR. 
I didn't notice that the History API is extended in ie_dom.js.

PR #2694, opened 2017-10-30
Allow Module Load errors to be suppressed

ChadKillingsworth:
Recent changes promoted the ModuleLoader warnings for missing modules to errors. I need these to be suppressible now for certain cases. This reworks the commonJsModuleLoad diagnostic group to be just a general moduleLoad group so that they can be managed.

### DIFF
--- a/src/com/google/javascript/jscomp/DiagnosticGroups.java
+++ b/src/com/google/javascript/jscomp/DiagnosticGroups.java
@@ -20,6 +20,7 @@ import com.google.common.annotations.GwtIncompatible;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
+import com.google.javascript.jscomp.deps.ModuleLoader;
 import com.google.javascript.jscomp.lint.CheckArrayWithGoogObject;
 import com.google.javascript.jscomp.lint.CheckDuplicateCase;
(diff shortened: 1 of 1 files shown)


src/com/google/javascript/jscomp/DiagnosticGroups.java
```diff
@@ -149,8 +150,9 @@ public DiagnosticGroup forName(String name) {
           + "underscore, "
           + "visibility";
 
-  public static final DiagnosticGroup COMMON_JS_MODULE_LOAD =
-      DiagnosticGroups.registerGroup("commonJsModuleLoad",
```
MatrixFrog: If anyone already has `@suppress {commonJsModuleLoad}` in their code then they'll have to change it, right? We should mention that in the release notes when we bring this in.

ChadKillingsworth: I did consider that - just didn't like even more groups.

Alternatively I can add a new one and remove the commonJs specific one from the advertised list.

MatrixFrog: Sounds good, we can just note it in the release notes.

MatrixFrog:
As it happens, I recently had to add `options.setWarningLevel(DiagnosticGroup.forType(ModuleLoader.LOAD_WARNING), CheckLevel.OFF);` to fix an internal tool. Now we can change that to just reference this group :)

PR #2691, opened 2017-10-28
Refactor CommonJS rewriting to be a normal pass

ChadKillingsworth:
Reworks CommonJS module rewriting as well as imported script rewriting into normal passes. Moves them into the default pass configuration.

Rewriting JSON files will be moved in a separate PR.

PR #2695, opened 2017-10-30
If a commonjs module import cannot be located, provide a fallback

ChadKillingsworth:
If a `require(path)` call cannot be located by the module loader, provide a fallback module name so that the require call is still rewritten. Useful for per-file rewriting for analysis.


brad4d:
LGTM functionally, but please add a test case.

ChadKillingsworth:
@brad4d We'll have to wait for #2691 for a test. Module load problems are now errors and it's not easy to suppress in a test.

brad4d:
@ChadKillingsworth looks like #2691 is closed.
Is a test case possible now?

ChadKillingsworth:
@brad4d Test case added. Had to rebase to top-of-tree.

brad4d:
Looks good.
Imported and sent for internal testing and review.

brad4d:
Submitted internally.
Next push to github should close this PR.

PR #2699, opened 2017-10-31
Support "let" in persistent mode, and don't re-lookup the root ast

niloc132:
Without this, the following dependency is never resolved correctly (in the case where the `SourceAst` isn't a `JsAst`, or if `JsFileParser.isSupported()` is false):

   let someModule = goog.require('my.module');

### DIFF
--- a/src/com/google/javascript/jscomp/CompilerInput.java
+++ b/src/com/google/javascript/jscomp/CompilerInput.java
@@ -291,7 +291,7 @@ public class CompilerInput implements SourceAst, DependencyInfo {
         return SimpleDependencyInfo.EMPTY;
       }
 
-      finder.visitTree(getAstRoot(compiler));
+      finder.visitTree(root);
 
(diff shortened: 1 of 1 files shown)


brad4d:
@blickly appears to be most familiar with this code.
@niloc132 please add appropriate test cases to `CompilerInputTest.java`

blickly:
I thought that we required goog.modules with ES6 to use `const` instead of `let`, since it more closely matches the import semantics of ES6 modules.

gkdn:
@blickly Unless we consider this an error, shouldn't we still handle it correctly? I will update J2CL but seems like there are around 200+ let goog.require in google3.

blickly:
Hmm, I was under the (mistaken) impression that we already banned this, but the warning in question appears to only be linter-level:
https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/ClosureCheckModule.java#L86-L89

PR #2700, opened 2017-10-31
At the top of every bundle, print all defines, if any

niloc132:
When in `--compilationLevel=BUNDLE`, any `--define` flags are ignored, and the output may not be helpful to run in a browser. In particular, `goog.ENABLE_DEBUG_LOADER=false` seems necessary for BUNDLE to make sense, but other defines can be helpful as well.

### DIFF
--- a/src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
+++ b/src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
@@ -41,6 +41,7 @@ import com.google.javascript.jscomp.CompilerOptions.OutputJs;
 import com.google.javascript.jscomp.CompilerOptions.TweakProcessing;
 import com.google.javascript.jscomp.deps.ModuleLoader;
 import com.google.javascript.jscomp.deps.SourceCodeEscapers;
+import com.google.javascript.rhino.IR;
 import com.google.javascript.rhino.Node;
 import com.google.javascript.rhino.TokenStream;
 import com.google.protobuf.CodedOutputStream;
@@ -69,6 +70,7 @@ import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Set;
 import java.util.logging.Level;
 import java.util.zip.ZipEntry;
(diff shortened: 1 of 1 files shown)


MatrixFrog:
I'm not certain if we'll want this. There might be things depending on the fact that the bundle *doesn't* include the defines. But if we do then we'll probably want some tests.

src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
```diff
@@ -2009,6 +2011,18 @@ void printManifestTo(Iterable<CompilerInput> inputs, Appendable out)
   void printBundleTo(Iterable<CompilerInput> inputs, Appendable out)
       throws IOException {
 
+    // First, print all defines passed in
+    Set<Entry<String, Node>> defines = compiler.getOptions().getDefineReplacements().entrySet();
+    if (!defines.isEmpty()) {
+      Node[] defineNodes = defines.stream()
```
brad4d: We cannot yet use the `.stream()` method or arrow functions.
These are java 8 features and we currently have limitations that prevent us from using them in most of our code.

niloc132: Sorry about that, I'll back it down - the pom's seem to indicate java8, since otherwise my IDE should not have allowed me to do this...

niloc132:
I can understand the reticence to add something like this, but on the other hand, avoiding defines in a bundle is easy - don't add defines.

On the other hand, there is no warning when you add a define that it will be ignored, your output is just silently wrong. The "fix" for those who need these is to wait until output is generated (since base.js is always first if you use `goog.module`, etc), and then rewrite the file with this same JS at the top. Not terribly slow, but my demo app so far in BUNDLE is 5.5mb, while only 40kb in ADVANCED_OPTIMIZATIONS, so skipping that full copy is certainly nice to have.

src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
brad4d: Sorry the poms misled you.

We're already using the java 8 compiler, but we have google-specific limitations that prevent us using java 8 features in the code. I'm working on fixing that, but it will be awhile yet.

niloc132: Makes sense. I'll get all of these updated as soon as I am able, thanks for you and your team's quick replies.

brad4d: FYI, Java 8 can now be used for code in `javascript/jscomp`

brad4d:
I think it would be good for someone more familiar than I am with bundle output to decide whether this is a good change or not.

brad4d:
@jplaisted, It looks like you recently modified the `printBundleTo` method.
Do you have an opinion on this change?

jplaisted:
`--define` is more than just `goog.define`s. You can also have non-Closure `@define`s (like `COMPILED` in base.js). A bundle won't replace these and it doesn't make sense for them to go in `CLOSURE_[UNCOMPILED_]DEFINES`. I think it would be confusing if `--define` always set the Closure defines as a result.

There are work arounds, obviously. You can pass an input to the compiler that sets `CLOSURE_[UNCOMPILED_]DEFINES` and it will 1) output that to the bundle and 2) respect that in a compile. Then it is a lot clearer that those are Closure defines only.

PR #2701, opened 2017-10-31
General use of persistent input store

niloc132:
Allow this newly added class to be used from other packages, instead of requiring that all tool authors write code in `com.google...` packages. Additionally, rather than assume that only `--jszip` params which end with ".js.zip" might be persistent, allow any ".zip" file (as `--jszip` itself does, via `setMixedJsSources()`).


brad4d:
I've assigned this to @blickly since he reviewed all past changes related to `PersistentInputStore`.
@niloc132 could you please add appropriate test cases to `PersistentInputStoreTest`?

Thanks.

blickly:
This change looks good to me. Could you add a unit test of the ".js.zip" change to make sure that we don't break that feature in the future, and then assign to me?

brad4d:
Created internal issue b/122075134

PR #2706, opened 2017-11-06
Add the ability to output compiler error information in a JSON format

ChadKillingsworth:
In external command line tools, it's often preferred to receive the error information as raw data rather than a formatted string. This allows better usage of the native error reporting mechanisms in tools like Webpack.

### DIFF
--- a/src/com/google/javascript/jscomp/CommandLineRunner.java
+++ b/src/com/google/javascript/jscomp/CommandLineRunner.java
@@ -28,6 +28,7 @@ import com.google.common.collect.ImmutableMultimap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Multimap;
 import com.google.common.io.Files;
+import com.google.javascript.jscomp.AbstractCommandLineRunner.CommandLineConfig.ErrorFormatOption;
 import com.google.javascript.jscomp.CompilerOptions.IsolationMode;
 import com.google.javascript.jscomp.SourceMap.LocationMapping;
 import com.google.javascript.jscomp.deps.ModuleLoader;
@@ -790,6 +791,9 @@ public class CommandLineRunner extends
     )
     private String packageJsonEntryNames = null;
 
+    @Option(name = "--error_format", usage = "Specifies format for error messages.")
+    private ErrorFormatOption errorFormat = ErrorFormatOption.STANDARD;
+
     @Option(name = "--renaming",
         handler = BooleanOptionHandler.class,
         usage = "Disables variable renaming. Cannot be used with ADVANCED optimizations.")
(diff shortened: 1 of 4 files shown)


src/com/google/javascript/jscomp/PrintStreamJSONErrorManager.java
```diff
@@ -0,0 +1,158 @@
...
+      jsonWriter.beginObject();
+      jsonWriter.name("level").value("info");
+      jsonWriter.name("description").value(summaryBuilder.toString());
+      jsonWriter.endObject();
+
+      jsonWriter.endArray();
+      jsonWriter.flush();
+      jsonWriter.close();
+    } catch (UnsupportedEncodingException e) {
+
```
brad4d: I believe you should re-throw the caught exception wrapped in a `RuntimeException` so errors aren't just swallowed.

src/com/google/javascript/jscomp/PrintStreamJSONErrorManager.java
```diff
@@ -0,0 +1,158 @@
...
+    } catch (IOException e) {
+
+    }
+    stream.append(bufferedStream.toString());
+
+    // Restore the messages since some tests assert the values after generating the report.
+    messages.addAll(list);
+  }
+
+  @Override
```
brad4d: Please add comments to these empty methods something like:

```java
// This class overrides generateReport(), so nothing will call println().

**Comment by ChadKillingsworth** [comment]
Done. Merged and squashed to top of tree to handle conflicts.

File: src/com/google/javascript/jscomp/CommandLineRunner.java [hunk 3]
```diff
@@ -790,6 +791,9 @@
     )
     private String packageJsonEntryNames = null;
 
+    @Option(name = "--error_format", usage = "Specifies format for error messages.")
```
fredj: Minor but can the possible values be added in the usage?
Something like: `Specifies format for error messages. Options: STANDARD (default), JSON`

ChadKillingsworth: I believe args4j adds them automatically for us.

fredj: ok, I didn't know that. I've added this comment because other options in this file (compilation_level for example) list all the possible values.

ChadKillingsworth: It only works for ENUM types. Compilation Level has some very non-standard parsing logic around it which requires us to manually list the values.

brad4d:
imported for internal review and test
sent to @blickly for internal review

brad4d:
submitted internally

Issue #2690, opened 2017-10-27
Expose an option to turn off mangling


gaearon:
Hey! I chatted a bit with @MatrixFrog, and he asked me to file an issue.

We recently started using GCC for building [React](https://github.com/facebook/react) and are loving it! Unfortunately, we can only use GCC for the open source builds, and not the ones we use at Facebook.

For the Facebook builds, we could only use GCC if there was an option to turn off name mangling in simple mode. We have our own minifier, and we can’t check in minified code into the repository as it makes debugging traces very difficult. Even if we checked in sourcemaps, those wouldn’t work with our internal tooling.

Currently, we use Uglify for internal builds. It provides a `mangle: false` option that keeps dead code elimination and other optimizations, but doesn’t transform the names. I am asking to provide a similar option for GCC so that we can use excellent other GCC optimizations while keeping the production build source human-readable. Our ideal setup right now would be GCC in simple mode with turned off mangling + Prettier for formatting the output nicely.

I recognize options are a huge maintenance burden and totally understand if that’s not something you would consider. We would very much appreciate even an “unofficial” flag though.

Thanks.

ChadKillingsworth:
By name mangling, you mean any type of local variable renaming?

ChadKillingsworth:
And one more question, are you using any module rewriting? (ES6, CommonJS, etc)

gaearon:
Yes, I mean renaming of variables and functions in the simple mode.

>are you using any module rewriting? (ES6, CommonJS, etc)

We are compiling CommonJS (and soon ES6) with Rollup, and running GCC on the output.

ChadKillingsworth:
@gaearon Can you test a custom build? In theory we should just be able to set the renaming policies to "OFF".

If you build from this branch: https://github.com/ChadKillingsworth/closure-compiler/tree/renaming-flag you should be able to test out a `--renaming=false` flag.

If it works for you, I'll PR it to the main repository and probably even publicly advertise it.

gaearon:
I haven't built from source yet but can give it a try on Monday. We're currently using the JS version but as I understand this is the upstream so it needs to be added here first.

ChadKillingsworth:
I can make a build for you. Just email me and I'll send you a link - but it will be the Java version. Adding flags is a bit different for the JS version.

gaearon:
Emailed. If this works for us and you get it merged, what steps do we need to do to expose it on JS version? I'm happy to spend a few days myself if that's something I could contribute.

[...]

PR #2707, opened 2017-11-07
Allow to disable renaming for the JS compiler

gaearon:
Fixes https://github.com/google/closure-compiler/issues/2690.

I used the approach suggested by @ChadKillingsworth in https://github.com/google/closure-compiler/issues/2690#issuecomment-340186286, but apply it to `GwtRunner` (since I need this to be exposed in the JS version) as suggested by @anmonteiro in https://github.com/google/closure-compiler/issues/2690#issuecomment-340212450.

**This adds a new `renaming` option to the JS version.**

It defaults to `true`:

screen shot 2017-11-07 at 4 07 33 pm

When set to `false`, it produces a build that doesn't minify variable or property names:

screen shot 2017-11-07 at 4 06 42 pm

It can only be set to `false` for `SIMPLE` mode. Doing so in `ADVANCED` mode throws:

screen shot 2017-11-07 at 4 08 21 pm

I need this option to move React builds completely to Google Closure Compiler. I’m happy to also include changes to expose this option to Java consumers, if you’d like, but I figured it’s best to send a PR with the minimal scope.

I’m sending this PR on the individual behalf not on company time. The code belongs to me.

### DIFF
--- a/src/com/google/javascript/jscomp/gwt/client/GwtRunner.java
+++ b/src/com/google/javascript/jscomp/gwt/client/GwtRunner.java
@@ -37,9 +37,11 @@ import com.google.javascript.jscomp.DependencyOptions;
 import com.google.javascript.jscomp.DiagnosticType;
 import com.google.javascript.jscomp.JSError;
 import com.google.javascript.jscomp.ModuleIdentifier;
+import com.google.javascript.jscomp.PropertyRenamingPolicy;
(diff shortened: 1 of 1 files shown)


dimvar:
LGTM. I will bring this internally for testing.

gaearon:
Thank you!

PR #2712, opened 2017-11-09
Fix parameter type of ES6 Number.* functions

gberaudo:
The standard accepts any type, not just numbers and always returns a boolean.
I fixed the references to point to the standard instead of the more vague Mozilla page.

### DIFF
--- a/externs/es6.js
+++ b/externs/es6.js
@@ -1394,34 +1394,34 @@ Number.parseInt = function(string, radix) {};
 Number.parseFloat = function(string) {};
 
 /**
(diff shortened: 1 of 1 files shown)


dimvar:
We have tightened the types of these functions intentionally.

gberaudo:
An easy workaround is code like `Number.isFinite(myVar || 0)`.

@dimvar, for my knowledge, what is the rationale for this restriction?

gberaudo:
Thanks @dimvar for the explanation.

PR #2719, opened 2017-11-14
Recognize entry points that are specified as goog.module or provide

### LINKED ISSUE #2692: CommonJS rewriting is being skipped for explicit deps entry points
CommonJS module rewriting isn't happening in particular cases. I'm finding `require("...")` calls left in the output.

In this case I'm using the Compiler via the Java API in a continuous build+test process, in checks-only mode primarily to identify developer errors but also using the bundled output as part of a jasmine unit test. This process of ours detects developer edits and only builds those compilation units (and their dependencies) that were edited. In the example below, TestGoogModOnly.js was edited by a developer, and thus triggering compilation with TestGoogModOnly as the explicit entry point.

I've narrowed the test case down to:
* Dep management: prune, drop moochers
* Entry point in DependencyOptions: `es6modulesdemo.sub.TestGoogModOnly`
* Enable CJS modules rewriting

TestEs6Only.js:
``` javascript
export default "TestEs6Only";
```
TestGoogModOnly.js:
``` javascript
goog.module("es6modulesdemo.sub.TestGoogModOnly");
goog.module.declareLegacyNamespace();
let t = require("./TestEs6Only");
```

Output:
``` javascript
var es6modulesdemo = {};
es6modulesdemo.sub = {};
es6modulesdemo.sub.TestGoogModOnly = {};
var module$contents$es6modulesdemo$sub$TestGoogModOnly_t = require("./TestEs6Only");
```

I've narrowed down the problem to the logic in `Compiler.findDependenciesFromEntryPoints`. Here, TestGoogModOnly is present in `inputsByProvide` keyed by `es6modulesdemo.sub.TestGoogModOnly`. But in [Compiler:1957](https://github.com/google/closure-compiler/blob/a71036ea696ba5ec600f9dcb9c2c787084032986/src/com/google/javascript/jscomp/Compiler.java#L1957) it's being looked up by `module$...` internal path-based provide name. As such it never appears in entryPoints, and thus never in orderedInputs, and skipped from `ProcessCommonJSModules`. Effectively it's skipping any files from CJS rewriting if they can only be found from entry points given explicitly by dependency options.

I'm not sure if the fix is to lookup in inputsByProvide, or to change the end of the method to also iterate through skipped inputs, for CJS rewriting.

Perhaps @ChadKillingsworth will know about this?

ChadKillingsworth:
When finding dependencies, recognize entry points specified in the `goog:` style.

Fixes #2692

### DIFF
--- a/src/com/google/javascript/jscomp/Compiler.java
+++ b/src/com/google/javascript/jscomp/Compiler.java
@@ -1934,7 +1934,10 @@ public class Compiler extends AbstractCompiler implements ErrorHandler, SourceFi
       }
     }
     for (ModuleIdentifier moduleIdentifier : options.getDependencyOptions().getEntryPoints()) {
-      CompilerInput input = inputsByIdentifier.get(moduleIdentifier.toString());
+      CompilerInput input = inputsByProvide.get(moduleIdentifier.toString());
+      if (input == null) {
+        input = inputsByIdentifier.get(moduleIdentifier.toString());
+      }
       if (input != null) {
         entryPoints.add(input);
       }
--- a/test/com/google/javascript/jscomp/CompilerTest.java
+++ b/test/com/google/javascript/jscomp/CompilerTest.java
@@ -1199,6 +1199,37 @@ public final class CompilerTest extends TestCase {
     assertThat(result.errors).isEmpty();
   }
 
+  // https://github.com/google/closure-compiler/issues/2692
+  public void testGoogNamespaceEntryPoint() throws Exception {
+    List<SourceFile> inputs =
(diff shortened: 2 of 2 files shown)


brad4d:
A delightfully simple PR.
Could you add a test case, though?

ChadKillingsworth:
Added

brad4d:
imported for internal testing and review
got approval from @blickly 
will be submitted unless some test failure is discovered

supersteves:
Many thanks. I can now explore using CommonJS requires in our large mixed interop project.

PR #2721, opened 2017-11-15
Allow the renaming policies to be disabled by a flag

ChadKillingsworth:
Companion to #2707

Ability to disable renaming in SIMPLE mode.


src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
```diff
@@ -2681,6 +2681,13 @@ public CommandLineConfig setJsonStreamMode(JsonStreamMode mode) {
       return this;
     }
 
+    private boolean renaming = true;
```
MatrixFrog: I think we don't need this, since we set the values on the CompilerOptions directly in CommandLineRunner.

PR #2726, opened 2017-11-23
Walk dep tree visit exports fix for export {x, y as z} from 'module-identifier'

gregmagolan:
I needed the following change for a build to work that contained exports such as:

```
export {x, y as z} from 'some-module';
```

Files referenced on lines like this were not being included in the bundle. I took a look at the Es6ModuleRewrite code and used that as an example to update FindModuleDependencies.java so that the files referenced in export lines like the one above are visited and end up included in the bundle.

Originally PR'd against the walk-dep tree branch: https://github.com/ChadKillingsworth/closure-compiler/pull/2.

@ChadKillingsworth has taken a look at it

### DIFF
--- a/src/com/google/javascript/jscomp/FindModuleDependencies.java
+++ b/src/com/google/javascript/jscomp/FindModuleDependencies.java
@@ -21,6 +21,7 @@ import static com.google.common.base.Preconditions.checkArgument;
 import com.google.javascript.jscomp.CompilerInput.ModuleType;
 import com.google.javascript.jscomp.Es6RewriteModules.FindGoogProvideOrGoogModule;
 import com.google.javascript.jscomp.deps.ModuleLoader;
+import com.google.javascript.rhino.IR;
 import com.google.javascript.rhino.Node;
 import com.google.javascript.rhino.Token;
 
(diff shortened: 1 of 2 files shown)


src/com/google/javascript/jscomp/FindModuleDependencies.java
```diff
@@ -97,6 +98,17 @@ public void visit(NodeTraversal t, Node n, Node parent) {
 
     if (supportsEs6Modules && n.isExport()) {
       moduleType = ModuleType.ES6;
+      if (n.getBooleanProp(Node.EXPORT_DEFAULT)) {
+        // export default
+      } else if (n.getBooleanProp(Node.EXPORT_ALL_FROM)) {
+        // export * from 'moduleIdentifier';
+        // not yet supported by closure ES6 transpilation: 'Wildcard export' is not yet implemented.
+      } else if (n.hasTwoChildren()) {
```
ChadKillingsworth: This is the only case we care about since I believe the only way an export has more than one child is if it exports from another module. There's no need to special case `export *`. While transpilation may not handle this case yet, we still want to add it to the module graph.

Can you simplify this case to that single if test?

src/com/google/javascript/jscomp/FindModuleDependencies.java
```diff
@@ -97,6 +98,17 @@ public void visit(NodeTraversal t, Node n, Node parent) {
...
+        // export {x, y as z} from 'moduleIdentifier';
+        Node moduleIdentifier = n.getLastChild();
+        Node importNode = IR.importNode(IR.empty(), IR.empty(), moduleIdentifier.cloneNode());
+        visit(t, importNode, parent);
```
ChadKillingsworth: Rather than cloning a node and revisiting, can you refactor out the ES6 import case to a function `getModuleNameFromImportPath`? Then you can just call that with the 2nd child and add the result to the graph.

brad4d:
@ChadKillingsworth thanks for doing the review.
@gregmagolan I believe we're awaiting your responses to Chad's comments.
@blickly would probably be best to import this and do the internal review.

gregmagolan:
@ChadKillingsworth Thanks for the comments. I'll work on your recommended changes today.

gregmagolan:
@ChadKillingsworth Pushed the refactored changes. Ended up adding two functions to capture the common functionality for import and export statements: `addEs6ModuleImportToGraph` and `getEs6ModuleNameFromImportNode`. Not sure if the naming is ideal.

Tested it against https://github.com/gregmagolan/closure-export-example.

Also tested with an `export * from 'path'` statement to ensure the compile still failed with: `ERROR - ES6 transpilation of 'Wildcard export' is not yet implemented.`

ChadKillingsworth:
@gregmagolan can you squash your commits? Importing multiple commits is known to cause problems.

gregmagolan:
> @gregmagolan can you squash your commits? Importing multiple commits is known to cause problems.

Done

blickly:
Is it possible to add a test of this?
I think that the exiting CompilerTest.testProperEs6ModuleOrdering is the most similar existing test. See:
https://github.com/google/closure-compiler/blob/master/test/com/google/javascript/jscomp/CompilerTest.java#L1466-L1525

gregmagolan:
> Is it possible to add a test of this?

Added testProperEs6ModuleOrderingWithExport

PR #2729, opened 2017-11-28
Support proper interop between CommonJS and ES6 modules

### LINKED ISSUE #2634: Using code from CJS module in ES6 module will try to access non-existing default property
Found a problem with React-leaflet (which is written in ES6) which uses prop-types (CommonJS). The code converted from ES6 module, tries to access `default` property in the converted CommonJS module, which will fail, as CommonJS module doesn't export `default` property:

```
❯ cat default-problem.js  
import Foo from './foo';

Foo.oneOfType();

❯ cat foo.js              
module.exports = {
  oneOfType: function() {
    return true;
  }
};

❯ java -jar target/closure-compiler-1.0-SNAPSHOT.jar --js foo.js default-problem.js  --process_common_js_modules --module_resolution NODE

var module$foo={oneOfType:function(){return!0}};
var module$default_problem={};
module$foo.default.oneOfType();
//          ^
//          Problem, should bee just module$foo.oneOfType
```

ChadKillingsworth:
The interop for ES6 and CommonJS is defined here: https://github.com/nodejs/node-eps/blob/master/002-es-modules.md

- When a `require('foo')` statement is importing an ES6 module, the import is the module object.
- When a `import bar from 'foo'` statement is importing a CommonJS module, the module export is treated as the default property.

This PR reworks CommonJS rewriting such that CommonJS modules are written to a `default` property on an export. A map of module names to module types is created by the compiler so that rewriting can vary based off what type of module is being imported. If an imported module cannot be located, it is assumed to be of the same type as the host.

Fixes #2634

Note this PR contains quite a few other CommonJS related fixes. It had become too difficult to separate them into standalone PRs.

cc @jplaisted

### DIFF
--- a/src/com/google/javascript/jscomp/Compiler.java
+++ b/src/com/google/javascript/jscomp/Compiler.java
@@ -140,8 +140,8 @@ public class Compiler extends AbstractCompiler implements ErrorHandler, SourceFi
   // The JS source inputs
   private List<CompilerInput> inputs;
 
-  // The JS source inputs
-  private Map<String, CompilerInput.ModuleType> moduleTypesByName;
+  // Map of module names to module types - used for module rewriting
+  private final Map<String, CompilerInput.ModuleType> moduleTypesByName;
 
   // error manager to which error management is delegated
   private ErrorManager errorManager;
--- a/src/com/google/javascript/jscomp/Compiler.java
+++ b/src/com/google/javascript/jscomp/Compiler.java
@@ -140,6 +140,9 @@ public class Compiler extends AbstractCompiler implements ErrorHandler, SourceFi
   // The JS source inputs
   private List<CompilerInput> inputs;
 
+  // The JS source inputs
+  private Map<String, CompilerInput.ModuleType> moduleTypesByName;
+
   // error manager to which error management is delegated
   private ErrorManager errorManager;
 
@@ -290,6 +293,7 @@ public class Compiler extends AbstractCompiler implements ErrorHandler, SourceFi
   public Compiler(PrintStream outStream) {
     addChangeHandler(recentChange);
     this.outStream = outStream;
+    this.moduleTypesByName = new HashMap<>();
   }
 
   /**
@@ -1986,6 +1990,7 @@ public class Compiler extends AbstractCompiler implements ErrorHandler, SourceFi
     if (wasImportedByModule && input.getJsModuleType() == CompilerInput.ModuleType.NONE) {
       input.setJsModuleType(CompilerInput.ModuleType.IMPORTED_SCRIPT);
     }
+    this.moduleTypesByName.put(input.getPath().toModuleName(), input.getJsModuleType());
 
     for (String requiredNamespace : input.getRequires()) {
       CompilerInput requiredInput = null;
@@ -2141,6 +2146,7 @@ public class Compiler extends AbstractCompiler implements ErrorHandler, SourceFi
       input.setCompiler(this);
       // Call getRequires to force regex-based dependency parsing to happen.
       input.getRequires();
+      input.setJsModuleType(CompilerInput.ModuleType.ES6);
     }
     return filteredInputs;
   }
@@ -3696,5 +3702,12 @@ public class Compiler extends AbstractCompiler implements ErrorHandler, SourceFi
     for (CompilerInput input : this.externs) {
       input.reset();
     }
+  }
+
+  /** Returns the module type for the provided namespace. */
+  @Override
+  @Nullable
+  CompilerInput.ModuleType getModuleTypeByName(String moduleName) {
+    return moduleTypesByName.get(moduleName);
   }
 }
--- a/src/com/google/javascript/jscomp/ProcessCommonJSModules.java
+++ b/src/com/google/javascript/jscomp/ProcessCommonJSModules.java
@@ -44,7 +44,7 @@ public final class ProcessCommonJSModules extends NodeTraversal.AbstractPreOrder
   private static final String EXPORTS = "exports";
   private static final String MODULE = "module";
   private static final String REQUIRE = "require";
-  private static final String exportPropertyName = "default";
+  private static final String EXPORT_PROPERTY_NAME = "default";
 
   public static final DiagnosticType UNKNOWN_REQUIRE_ENSURE =
       DiagnosticType.warning(
@@ -155,7 +155,7 @@ public final class ProcessCommonJSModules extends NodeTraversal.AbstractPreOrder
       return moduleName;
     }
 
-    return moduleName + "." + exportPropertyName;
+    return moduleName + "." + EXPORT_PROPERTY_NAME;
   }
 
   /**
@@ -652,7 +652,7 @@ public final class ProcessCommonJSModules extends NodeTraversal.AbstractPreOrder
       builder.recordConstancy();
       initModule.setJSDocInfo(builder.build());
       if (directAssignments == 0) {
-        Node defaultProp = IR.stringKey(exportPropertyName);
+        Node defaultProp = IR.stringKey(EXPORT_PROPERTY_NAME);
         defaultProp.addChildToFront(IR.objectlit());
         initModule.getFirstFirstChild().addChildToFront(defaultProp);
         builder = new JSDocInfoBuilder(true);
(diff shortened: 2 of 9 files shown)


test/com/google/javascript/jscomp/CommonJSIntegrationTest.java
```diff
@@ -59,105 +59,34 @@ public void testCrossModuleTypeAnnotation() {
...
         new String[] {
-          LINE_JOINER.join("/** @constructor */ function Hello() {}", "module.exports = Hello;"),
-          LINE_JOINER.join("var Hello = require('./i0');", "/** @type {!Hello} */ var hello = 1;")
+          "/** @constructor */ function Hello() {} module.exports = Hello;",
+          "var Hello = require('./i0'); /** @type {!Hello} */ var hello = 1;"
         },
         TypeValidator.TYPE_MISMATCH_WARNING);
   }
 
-  public void testMultipleExportAssignments1() {
```
brad4d: Are we no longer supporting these use-cases?

ChadKillingsworth: Well I thought not, but now I'm 2nd guessing myself. Let me go check ...

brad4d:
Imported and sent to @blickly  for internal review.
Standard presubmit testing is in progress.

src/com/google/javascript/jscomp/Compiler.java
```diff
@@ -140,6 +140,9 @@
   // The JS source inputs
   private List<CompilerInput> inputs;
 
+  // The JS source inputs
+  private Map<String, CompilerInput.ModuleType> moduleTypesByName;
```
brad4d: 1. Could you make this `final`?
2. (From @blickly) I think it would be easier to follow if this map were indexed by CompilerInput rather than by String.  It's not as clear to me what the "canonical" name of a file should be, and it looks like the CompilerInput is always what is used to compute the names that are currently used.

src/com/google/javascript/jscomp/ProcessCommonJSModules.java
```diff
@@ -897,25 +1106,11 @@ public void visit(NodeTraversal t, Node n, Node parent) {
...
-
-      String moduleName;
-      if (modulePath == null) {
-        // The module loader will issue an error, but use a fallback
-        moduleName = ModuleIdentifier.forFile(requireName).getModuleName();
-      } else {
-        moduleName = modulePath.toModuleName();
-      }
-      Node moduleRef = IR.name(moduleName).srcref(require);
+      String requireName = getCommonJsImportPath(require);
```
brad4d: `requireName` is unused

src/com/google/javascript/jscomp/ProcessCommonJSModules.java
```diff
@@ -943,66 +1138,95 @@ private void visitExport(NodeTraversal t, Node export) {
...
       }
 
+      moduleName = moduleName + "." + EXPORT_PROPERTY_NAME;
+
       Node updatedExport =
-          NodeUtil.newName(compiler, moduleName, export, export.getQualifiedName());
+          NodeUtil.newQName(compiler, moduleName, export.node, export.node.getQualifiedName());
+      boolean exportIsConst =
+          defaultExportIsConst
+              && getBasePropertyImport(getModuleName(t.getInput()))
```
brad4d: Can be replaced with simpler, equivalent code.
*Caveats*: This version no longer throws NPE if the first object in comparison is null.

Suggested replacement:
```java
              && updatedExport.matchesQualifiedName(
                  getBasePropertyImport(getModuleName(t.getInput())))
```

src/com/google/javascript/jscomp/ProcessCommonJSModules.java
```diff
@@ -943,66 +1138,95 @@ private void visitExport(NodeTraversal t, Node export) {
...
+
+      Node changeScope = null;
 
       if (root.matchesQualifiedName("module.exports")
           && rValue != null
-          && t.getScope().getVar("module.exports") == null
+          && export.scope.getVar("module.exports") == null
           && root.getParent().isAssign()) {
-        if (root.getParent().getParent().isExprResult()) {
+        if (root.getParent().getParent().isExprResult() && moduleInitialization == null) {
```
brad4d: Can be replaced with simpler, equivalent code.
*Caveats*: This version no longer throws NPE if the immediate parent is null.

Suggested replacement:
```java
        if (root.getGrandparent().isExprResult() && moduleInitialization == null) {
```

test/com/google/javascript/jscomp/ProcessCommonJSModulesTest.java
```diff
@@ -933,12 +980,70 @@ public void testIssue2616() {
...
+            "})('Fingerprint2', this, function() {",
+            "  var Fingerprint2 = function() {",
+            "    if (!(this instanceof Fingerprint2)) { return new Fingerprint2(); }",
+            "  };",
+            "  return Fingerprint2$$module$test;",
+            "})"),
+        lines(
+            "/** @const */ var module$test = {};",
+            "var Fingerprint2$$module$test = function() {",
+            "  if (!(this instanceof Fingerprint2$$module$test)) {",
```
brad4d: From @blickly 
I think this should be Fingerprint2

src/com/google/javascript/jscomp/Compiler.java
```diff
@@ -140,6 +140,9 @@
   // The JS source inputs
   private List<CompilerInput> inputs;
 
+  // The JS source inputs
```
brad4d: this comment should be updated

src/com/google/javascript/jscomp/Compiler.java
ChadKillingsworth: For 2: The main point is that the module type needs looked up for import statements - and you can't get the compiler input for that. You get a module path (and can easily get module name). See https://github.com/google/closure-compiler/pull/2729/files#diff-d1b3eef5af5a700763a2fce367ab27ddR153

test/com/google/javascript/jscomp/ProcessCommonJSModulesTest.java
ChadKillingsworth: I'm not following ... The rewriting looks correct.

test/com/google/javascript/jscomp/CommonJSIntegrationTest.java
ChadKillingsworth: @brad4d Those tests give errors like:

`JSC_DUP_VAR_DECLARATION. variable module$i0.default redefined, original definition at i0.js:3 at i0.js line 1 : 29 `

And that warning looks pretty correct to me. I could of course expect the warning, but I'm not sure it's really testing anything meaningful.

test/com/google/javascript/jscomp/ProcessCommonJSModulesTest.java
ChadKillingsworth: Oh I think this was meant for line 1008

src/com/google/javascript/jscomp/Compiler.java
ChadKillingsworth: To better illustrate:

```
const foo = require('bar');
```

We need to know what type of module `'bar'` is. That's the primary purpose of this map.

ChadKillingsworth: FYI I'm open to other suggestions on where to put this and how to utilize this information. I'm not 100% happy with this, but I don't have any better suggestions either.

brad4d:
I've imported Chad's commits and sent them for another round of testing and internal review.

src/com/google/javascript/jscomp/Compiler.java
blickly: Ah, OK, thanks for the detailed explanation, Chad.  That makes sense. Somehow I had forgotten that the ModuleLoader is all path-based and doesn't have access to the module contents.

brad4d:
submitted internally

jplaisted:
Hey-o! I'm not sure what other work is going on here / if this is a work in progress but as it stands doesn't look like the interop actually works.

Take a look at testES6ImportOfCJS:
https://github.com/google/closure-compiler/blob/3f954c107c79bc4305ef522f522fd2edb9b20e35/test/com/google/javascript/jscomp/CommandLineRunnerTest.java#L1809-L1835

Feel free to paste the expected output in the console to show it doesn't work. `import * as m from 'cjs'; m.foo()` gets transpiled to `const m = module$cjs; m.foo();`. Problem is CJS modules now only have a `default` property, so it fails.

Looking at the proposed node implementation this seems to be the expected output. But we then maybe we should error or warn if you don't import a CJS module with default semantics (it seems like we'll throw a less-clearer error at type checking time...)? Any particular reason this was done over other approaches?

PR #2739, opened 2017-12-04
Replace CommonJS module.id references with the full module path

ChadKillingsworth:
In a CommonJS module, `module.id` is supposed to return a value that another module would be able to use to require it. In our case, simply replace references to module.id with the full module path.


blickly:
Importing this now

PR #2740, opened 2017-12-04
Collapse module export properties in SIMPLE mode when an output wrapper is used

### LINKED ISSUE #1954: Collapse Properties has massive impact on ES6 module code size
I have not tested this recently, but it is pretty bad, since this cannot (or can it?) be set via the command line. Without collapse properties, DCE basically doesn't work.

Could the generated code be changed to not rely on this?

ChadKillingsworth:
When a function wrapper is assumed, dead code elimination can be significantly improved by collapsing module export properties. This is particularly effective when utilizing ES modules with multiple exports as any unused exports can be dead-code eliminated.

Fixes #1954

cc @cramforce

### DIFF
--- a/src/com/google/javascript/jscomp/CommandLineRunner.java
+++ b/src/com/google/javascript/jscomp/CommandLineRunner.java
@@ -1687,7 +1687,7 @@ public class CommandLineRunner extends
       level.setTypeBasedOptimizationOptions(options);
     }
 
-    if (flags.assumeFunctionWrapper) {
+    if (flags.assumeFunctionWrapper || flags.isolationMode == IsolationMode.IIFE) {
       level.setWrappedOutputOptimizations(options);
     }
 
--- a/src/com/google/javascript/jscomp/CompilerOptions.java
+++ b/src/com/google/javascript/jscomp/CompilerOptions.java
@@ -75,6 +75,12 @@ public class CompilerOptions implements Serializable {
     NONE
   }
 
+  public enum PropertyCollapseLevel {
+    ALL,
+    NONE,
+    MODULE_EXPORT
+  }
+
   // TODO(nicksantos): All public properties of this class should be made
   // package-private, and have a public setter.
 
@@ -631,11 +637,15 @@ public class CompilerOptions implements Serializable {
     renamePrefixNamespaceAssumeCrossModuleNames = assume;
   }
 
-  private boolean collapseProperties;
+  private PropertyCollapseLevel collapsePropertiesLevel;
 
   /** Flattens multi-level property names (e.g. a$b = x) */
   public boolean shouldCollapseProperties() {
-    return collapseProperties;
+    return collapsePropertiesLevel != PropertyCollapseLevel.NONE;
+  }
+
+  public PropertyCollapseLevel getPropertyCollapseLevel() {
+    return collapsePropertiesLevel;
   }
 
   /** Split object literals into individual variables when possible. */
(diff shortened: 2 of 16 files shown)


blickly:
Very cool, FYI @jplaisted on upcoming changes to CollapseProperties

test/com/google/javascript/jscomp/CollapsePropertiesTest.java
```diff
@@ -35,26 +37,45 @@
       + "/** @constructor */ function String() {};\n"
       + "var arguments";
 
+  private boolean processCommonJsModules = false;
```
blickly: Looks like this is unused here.

src/com/google/javascript/jscomp/CommandLineRunner.java
```diff
@@ -1687,7 +1687,7 @@ protected CompilerOptions createOptions() {
       level.setTypeBasedOptimizationOptions(options);
     }
 
-    if (flags.assumeFunctionWrapper) {
+    if (flags.assumeFunctionWrapper || flags.isolationMode == IsolationMode.IIFE) {
```
blickly: Nice catch!

src/com/google/javascript/jscomp/CompilerOptions.java
```diff
@@ -634,6 +634,9 @@ void setRenamePrefixNamespaceAssumeCrossModuleNames(boolean assume) {
   /** Flattens multi-level property names (e.g. a$b = x) */
   public boolean collapseProperties;
 
+  /** Flattens property names created by module rewriting (e.g. a$b = x) */
+  public boolean collapseModuleExportProperties;
```
blickly: I'd prefer to make collapseProperties an enum over adding a new boolean for this.  I just sent out a refactoring to make the regular collapseProperties private to enable you to change it here.

test/com/google/javascript/jscomp/CollapsePropertiesTest.java
```diff
@@ -35,26 +37,45 @@
...
   }
 
   @Override
   protected void setUp() throws Exception {
     super.setUp();
     enableNormalize();
+    if (this.collapseModuleExportProperties) {
+      this.setAcceptedLanguage(LanguageMode.ECMASCRIPT_2015);
+      this.setLanguageOut(LanguageMode.ECMASCRIPT5);
+    }
```
blickly: I think this part is redundant with the settings in setupModuleExportsOnly below, isn't it?

src/com/google/javascript/jscomp/CompilerOptions.java
ChadKillingsworth: I actually waffled back and forth on this. It was just easier to do it this way. That refactor will help though.

blickly: Ok, pushed my change out in https://github.com/google/closure-compiler/commit/36544c6f941fdb860eadfeedc6a932dbc28b8ea2

ChadKillingsworth:
@blicky Addressed review issues and rebased to top of tree.

blickly:
I think it may be easier to just deprecated the CompilerOptions.setCollapsePropeties(boolean) rather than remove it in this PR, but I can make that while importing.   Sending this out for internal review now.

src/com/google/javascript/jscomp/debugger/common/CompilationParam.java
```diff
@@ -682,7 +683,7 @@ public boolean isApplied(CompilerOptions options) {
   COLLAPSE_PROPERTIES(ParamGroup.TYPE_CHECKING_OPTIMIZATION) {
     @Override
     public void apply(CompilerOptions options, boolean value) {
-      options.setCollapseProperties(value);
+      options.setCollapsePropertiesLevel(PropertyCollapseLevel.ALL);
```
MatrixFrog: This should be 

```
if (value) {
  options.setCollapsePropertiesLevel(PropertyCollapseLevel.ALL);
} else {
  options.setCollapsePropertiesLevel(PropertyCollapseLevel.NONE);
}
```

ChadKillingsworth: I definitely made a guess there ...

PR #2745, opened 2017-12-07
Add webkitRelativePath to the File interface

vobruba-martin:
This property exists in the latest Chrome, Firefox and Edge. It is needed for directory upload.

See https://wicg.github.io/entries-api/#dom-file-webkitrelativepath

### DIFF
--- a/externs/browser/fileapi.js
+++ b/externs/browser/fileapi.js
@@ -336,6 +336,12 @@ File.prototype.lastModifiedDate;
 File.prototype.lastModified;
 
 /**
+ * @see https://wicg.github.io/entries-api/#dom-file-webkitrelativepath
+ * @type {string}
+ */
+File.prototype.webkitRelativePath;
+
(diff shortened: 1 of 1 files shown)


brad4d:
@vobruba-martin thanks for this PR.

Imported and sent to @MatrixFrog for internal review.

MatrixFrog:
In general, the world should be moving away from webkit-prefixed properties. It would be nice to add a comment indicating which browsers need the prefix (if you know off the top of your head). If not, it's not a big deal.

vobruba-martin:
In this case all browsers are using webkit prefix and there is no version of this property without the prefix.  You can inspect `File.prototype` in Chrome, FF and Edge.

brad4d:
@vobruba-martin 
Thanks for the info.
I'm submitting this change internally now.

vobruba-martin:
Thanks!

Issue #2683, opened 2017-10-26
Contributing using Eclipse+Windows is hard


supersteves:
This is an umbrella issue for the hurdles developers who want to develop in the Closure Compiler codebase using Eclipse on Windows will encounter. It's just taken me, a moderately seasoned external contributor, about half a day to get developing properly from scratch.

First, ideally the "mvn eclipse:eclipse" command would create eclipse configs that just worked, but I don't know how to solve that. I've found that you also need to:

1. Edit/add to .classpath:
```
<classpathentry excluding="com/google/debugging/sourcemap/super|com/google/javascript/jscomp/debugger/gwt/DebuggerGwtMain.java|com/google/javascript/jscomp/gwt/|com/google/jscomp/resources/super-gwt" kind="src" path="src"/>
<classpathentry kind="src" path="target/generated-sources/annotations"/>
```
2. Add to .project:
```
  <filteredResources>
    <filter>
      <id>1509014600173</id>
      <name></name>
      <type>30</type>
      <matcher>
        <id>org.eclipse.ui.ide.multiFilter</id>
        <arguments>1.0-projectRelativePath-matches-false-false-src/com/google/debugging/sourcemap/super</arguments>
      </matcher>
    </filter>
    <filter>
      <id>1509014600181</id>
      <name></name>
      <type>30</type>
      <matcher>
        <id>org.eclipse.ui.ide.multiFilter</id>
        <arguments>1.0-projectRelativePath-matches-false-false-src/com/google/javascript/jscomp/resources/super-gwt</arguments>
      </matcher>
    </filter>
  </filteredResources>
```
3. Edit the Java Compiler settings in Eclipse, changing to Java 1.8
4. Run `mvn -DskipTests` to ensure AutoValues are generated. Refresh eclipse.
5. Import Google code style `eclipse-java-google-style.xml`

Then, if you want to run tests, either via `mvn` or individually from within Eclipse, you get a bunch of failures mostly relating to filesystem absolute paths, line separators, and one relating to case insensitivity:
```
Failed tests:
  DepsGeneratorTest.testWithDepsAndSources:165 expected:<.../my-project/deps1.js[
goog.addDependency('../prescanned1/file1.js', ['dep.string'], []);
goog.addDependency('../prescanned1/file2.js', [], []);

// Included from: /base/my-project/deps2.js]
goog.addDependency(...> but was:<.../my-project/deps1.js[
goog.addDependency('../prescanned1/file1.js', ['dep.string'], []);
[...]

PR #2753, opened 2017-12-11
Updated classpath instructions for Eclipse

supersteves:
Tiny PR to update the README to have the current steps I use to get Eclipse working with the tip of the compiler github repo.

I've updated just the "Using Eclipse" section so it now includes the missing steps mentioned in #2683. It would be better if the `mvn eclipse:eclipse` target could do some of these automatically, but I'm no maven expert.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,17 +53,21 @@ unit tests too).
 
 ### Using [Eclipse](https://www.eclipse.org/)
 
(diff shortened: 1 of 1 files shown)


brad4d:
@supersteves, thanks for this update.

Imported and sent to @MatrixFrog for internal review.

PR #2762, opened 2017-12-18
Adds support for webpack module resolution

ChadKillingsworth:
* Adds a webpack specific module resolution mode to lookup modules by their webpack id. Requires passing in webpack module ids with the associated source.
* Adds support for dynamically loaded modules by treating them as entry points in the module graph. Late loaded webpack modules are the only recognized form so far, but the dynamic `import()` statement can use the same mechanism when supported.
* ProcessCommonJSModules is updated to recognize and rewrite webpack specific require calls.

This is the last PR in a long chain to get full webpack support.

### DIFF
--- a/src/com/google/javascript/jscomp/ProcessCommonJSModules.java
+++ b/src/com/google/javascript/jscomp/ProcessCommonJSModules.java
@@ -15,6 +15,7 @@
  */
 package com.google.javascript.jscomp;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Preconditions.checkState;
 
@@ -44,6 +45,7 @@ public final class ProcessCommonJSModules extends NodeTraversal.AbstractPreOrder
   private static final String EXPORTS = "exports";
   private static final String MODULE = "module";
   private static final String REQUIRE = "require";
+  private static final String WEBPACK_REQUIRE = "__webpack_require__";
   private static final String EXPORT_PROPERTY_NAME = "default";
 
   public static final DiagnosticType UNKNOWN_REQUIRE_ENSURE =
@@ -78,6 +80,10 @@ public final class ProcessCommonJSModules extends NodeTraversal.AbstractPreOrder
     if (n.isRoot()) {
       return true;
     } else if (n.isScript()) {
+      if (compiler.getOptions().getModuleResolutionMode() == ModuleLoader.ResolutionMode.WEBPACK) {
+        removeWebpackModuleShim(n);
+      }
+
       FindImportsAndExports finder = new FindImportsAndExports();
       NodeTraversal.traverseEs6(compiler, n, finder);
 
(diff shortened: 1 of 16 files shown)


src/com/google/javascript/jscomp/deps/WebpackModuleResolver.java
```diff
@@ -0,0 +1,68 @@
...
+import javax.annotation.Nullable;
+
+/**
+ * Resolution algorithm for Webpack. Modules are located by a map of numeric ids to module paths.
+ *
+ * <p>As the compiler normally locates modules by path string, webpack numeric ids are converted to
+ * strings.
+ */
+public class WebpackModuleResolver extends NodeModuleResolver {
+  private final ImmutableMap<String, String> modulesById;
```
blickly: Shouldn't this map be keyed by Integer rather than String?

src/com/google/javascript/jscomp/deps/ModuleLoader.java
```diff
@@ -103,7 +105,21 @@ public ModuleLoader(
       case NODE:
         this.moduleResolver =
             new NodeModuleResolver(
-                this.modulePaths, this.moduleRootPaths, packageJsonMainEntries, this.errorHandler);
+                this.modulePaths, this.moduleRootPaths, lookupMap, this.errorHandler);
+        break;
+      case WEBPACK:
```
blickly: I'm starting to think these constructors are complicated enough that they deserve to move to factory methods instead, but I'm not sure if that change makes sense to include now.

There aren't any callers of these constructors outside of Closure Compiler, FYI.

src/com/google/javascript/jscomp/ProcessCommonJSModules.java
```diff
@@ -157,24 +163,56 @@ public String getBasePropertyImport(String moduleName) {
...
-  public static String getCommonJsImportPath(Node requireCall) {
+  public String getCommonJsImportPath(Node requireCall) {
+    return getCommonJsImportPath(requireCall, compiler.getOptions().getModuleResolutionMode());
+  }
+
+  public static String getCommonJsImportPath(
+      Node requireCall, ModuleLoader.ResolutionMode resolutionMode) {
+    if (resolutionMode == ModuleLoader.ResolutionMode.WEBPACK) {
+      if (requireCall.getSecondChild().isNumber()) {
+        return String.valueOf(Double.valueOf(requireCall.getSecondChild().getDouble()).intValue());
```
blickly: Here and below, we can simplify
   Double.valueOf(n.getDouble()).intValue()
->
  (int) n.getDouble()

src/com/google/javascript/jscomp/ProcessCommonJSModules.java
```diff
@@ -221,6 +260,62 @@ public static boolean isCommonJsExport(NodeTraversal t, Node export) {
...
+    if (fnc.getParent() == null) {
+      return false;
+    }
+
+    Node callParent = fnc.getParent();
+    if (!callParent.isCall()) {
+      return false;
+    }
+
+    if (callParent.getFirstChild() != null
```
blickly: This line can be:
    callParent.hasChildren()

test/com/google/javascript/jscomp/CommandLineRunnerTest.java
```diff
@@ -2152,6 +2152,45 @@ public void testAssumeFunctionWrapper() {
         "function a() {return 'hi'}alert(a);alert(a)");
   }
 
+  public void testWebpackModuleIds() {
+    String inputString =
+        LINE_JOINER.join(
+            "[",
+            "  {\"src\": \"__webpack_require__(2);\", \"path\":\"foo.js\", \"webpackId\": \"1\"},",
```
blickly: Is there a reason why the webpackId is represented as a string rather than a number?

ChadKillingsworth: In webpack itself, they are stored as strings. I'm asking "why" on their slack right now. Maybe we can change it.

src/com/google/javascript/jscomp/deps/ModuleLoader.java
ChadKillingsworth: > There aren't any callers of these constructors outside of Closure Compiler, FYI.

Well that gives me a little hope. I cringe every time I have to make a non-trivial change to the deps package.

But can we handle that as a separate PR?

ChadKillingsworth:
While testing I discovered a bug with destructuring imports (they weren't being renamed) which is fixed in this PR. I also disabled a goog.module check for webpack as webpack always includes a module loader.

test/com/google/javascript/jscomp/CommandLineRunnerTest.java
blickly: Awesome, nice to catch something in review!

src/com/google/javascript/jscomp/deps/ModuleLoader.java
blickly: SGTM

blickly:
Importing this PR now...

PR #2788, opened 2018-01-16
Add IRegExpResult

morille:
Interface that matches the return value of String.prototype.match and RegExp.prototype.exec
Currently we need to use a cast to access the "index" and "input" properties.

### DIFF
--- a/externs/es3.js
+++ b/externs/es3.js
@@ -142,6 +142,18 @@ function IArrayLike() {}
 IArrayLike.prototype.length;
 
 /**
+ * @record
+ * @extends {IArrayLike.<string>}
+ */
+function IRegExpResult() {};
+
+/** @type {number} */
+IRegExpResult.prototype.index;
+
+/** @type {string} */
+IRegExpResult.prototype.input;
+
+/**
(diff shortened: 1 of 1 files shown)


brad4d:
Restarted the Travis build, since I suspect it was a flaky timeout.

externs/es3.js
```diff
@@ -141,6 +141,18 @@ function IArrayLike() {}
 /** @type {number} */
 IArrayLike.prototype.length;
 
+/**
+ * @record
+ * @extends {IArrayLike.<string>}
```
brad4d: Using IArrayLike is actually a problem here, because the result is really an Array and client code should be able to call array methods on it.

I think the only way to do this will be:

```javascript
/**
 * @constructor
 * @extends {Array<string>}
 */
function IRegExpResult() {};
```

This makes it look like you can call `new IRegExpResult()` when you really can't, but I think that's OK in this case.

brad4d:
Actually this seems to have already been done.
https://github.com/google/closure-compiler/blob/master/externs/es3.js#L2098

Issue #2793, opened 2018-01-18
ownerDocument.createNodeIterator not defined on Document


syranide:
NPM version: 20180101.0.0

So apparently `createNodeIterator` is defined on Document, but not when accessing it via `ownerDocument`, despite still being of type Document.

```js
// OK
document.createNodeIterator(document.body, NodeFilter.SHOW_COMMENT);
```

```js
// WARNING - Property createNodeIterator never defined on document.body.ownerDocument of type Document
document.body.ownerDocument.createNodeIterator(document.body, NodeFilter.SHOW_COMMENT);
```

```
java -XX:+TieredCompilation -jar node_modules/google-closure-compiler/compiler.jar -O ADVANCED --js \"src/**.js\" --entry_point \"src/index\" --js_output_file \"output/bundle.js\" --create_source_map \"output/bundle.js.map\" --source_map_include_content --output_wrapper \"(function(){%output%}).call(this)\" --language_out ECMASCRIPT5_STRICT --new_type_inf --env BROWSER --module_resolution NODE --warning_level=VERBOSE --jscomp_warning=extraRequire
```

syranide:
Or is `createNodeIterator` specific to only `HTMLDocument` and I have to explicitly cast it? I don't understand that specific W3C spec well enough to know (it says it's part of `DocumentTraversal`).

MatrixFrog:
At least according to our externs, yes, it's only on `HTMLDocument` not on every `Document`: https://github.com/google/closure-compiler/blob/master/externs/browser/w3c_dom2.js#L249

You can definitely use a cast to get yourself unblocked, but maybe the type of the `ownerDocument` property should be `HTMLDocument` instead of `Document`.

syranide:
@MatrixFrog OK. However, I just checked. `Document.prototype.createNodeIterator` has the function in both Chrome and Edge... browsers not following specs?

MatrixFrog:
It is perhaps more likely that our externs are not following the specs :)

syranide:
Do you want a PR or prefer to do it yourselves?

PR #2794, opened 2018-01-18
Externs createNodeIterator and createTreeWalker exists on Document

syranide:
For #2793

`createNodeIterator` and `createTreeWalker` are defined on `Document` according to https://dom.spec.whatwg.org/#interface-document, not on `HTMLDocument` as it is currently in your externs file.

https://www.w3.org/TR/2000/REC-DOM-Level-2-Traversal-Range-20001113/traversal.html#Traversal-Document defines them to exists on `DocumentTraversal` which I'm not sure what that means in practice, but somewhere in the documentation it was mentioned that a `Document` can simply be cast to `DocumentTraversal` to access them. Anyway, this spec just confuses me, it doesn't seem to relate to JS specifically.

I've left both URLs in as comments to both reflect the DOM level 2 status and as a source for them existing on `Document`.

cc @MatrixFrog

### DIFF
--- a/externs/browser/w3c_dom2.js
+++ b/externs/browser/w3c_dom2.js
@@ -52,6 +52,32 @@ Document.prototype.createAttributeNS =
     function(namespaceURI, qualifiedName) {};
 
 /**
+ * @param {Node} root
+ * @param {number=} whatToShow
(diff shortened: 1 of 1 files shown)


syranide:
I signed it

PR #2823, opened 2018-02-18
Fix Array.prototype.includes

### LINKED ISSUE #2820: Array#includes polyfill doesn't conform to the spec
The current [impl](https://github.com/google/closure-compiler/blob/e6dfb88c41eec9d5172eb0294228da5266b84496/src/com/google/javascript/jscomp/js/es6/array/includes.js) doesn't conform to [the spec](https://www.ecma-international.org/ecma-262/8.0/#sec-array.prototype.includes)

- null and undefined (`[null].includes(undefined)`: expected: false, but actual: true)
- negative index (`[0,1,2].includes(0, -1)`: expected: false, actual: true)
- access to each element twice (in the spec, `Get(O, ! ToString(k))` is called only once)

teppeis:
Fix #2820

### DIFF
--- a/src/com/google/javascript/jscomp/js/es6/array/includes.js
+++ b/src/com/google/javascript/jscomp/js/es6/array/includes.js
@@ -38,8 +38,13 @@ $jscomp.polyfill('Array.prototype.includes', function(orig) {
       array = /** @type {!IArrayLike} */ (String(array));
     }
     var len = array.length;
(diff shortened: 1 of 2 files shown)


brad4d:
I've imported, reviewed, and submitted this change.
It should be in today's push to github.

teppeis:
@brad4d thanks!

PR #2829, opened 2018-02-25
Add SourceMapObjectParserJs.getLineCount implementation

arichiardi:
The implementation was missing but used in the non-JVM version when building
  source maps.  The proposed implementation just returns the value of the
  lineCount property and fixes the problem.


arichiardi:
I am aware of the weird branch name, I can close and reopen in case.

MatrixFrog:
I don't think the branch name matters.

arichiardi:
I hoped so :wink:

PR #2831, opened 2018-03-01
HTMLElement.style.setAttribute is old API and need to be removed

vegegoku:
fix for https://github.com/google/elemental2/issues/39

### DIFF
--- a/externs/browser/w3c_css.js
+++ b/externs/browser/w3c_css.js
@@ -520,15 +520,6 @@ CSSStyleDeclaration.prototype.removeExpression = function(name) {};
 
 /**
  * @param {string} name
- * @param {*} value
- * @param {number=} opt_flags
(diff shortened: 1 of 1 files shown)


brad4d:
Removing an interface like this risks breaking existing code.
I see from the issue google/elemental2#39 that at least in some environments trying to use this method is a problem, but presumably it still works in some places.
A quick search through internal Google source definitely finds some references to it.

If you'd like to change this PR to just mark the method as deprecated, that would be fine.
Otherwise, removing it doesn't really seem worth the risk of breaking code outside Google and the effort of breaking and then fixing code inside Google.

Sorry.

brad4d:
Please feel free to file a new PR that marks the method as deprecated with an explanation.

vegegoku:
I will file a new PR very soon.

Issue #2005, opened 2016-09-07
TypeScript keyword in ES6 export list causes a parse error


pygy:
~~`type` is a valid ES6 identifier yet the Closure compiler will refuse to compile files that try to use it as a variable.~~

~~It would be nice if the restriction was either lifted in ES6 input mode or at least documented. The docs [currently say](https://developers.google.com/closure/compiler/docs/limitations):~~

> - ~~**The Compiler only recognizes ECMAScript.**~~
>   
>   ~~ECMAScript 5 is the version of JavaScript supported almost everywhere. However the compiler also supports many of the features in ECMAScript 6. The compiler only supports official language features.~~

~~[Searching for "TypeScript"](https://developers.google.com/s/results/?q=typescript&p=%2Fclosure%2Fcompiler%2F) doesn't return any page that mentions the language.~~

Edit: This bug may have a narrower scope than I initially thought... The problematic construct is 

``` JS
export {type};
```

I'll check later today if there are other issues but right now I must leave.

pygy:
Using `type` in import statements is also problematic:

Using the `closure-compiler-js` (but AFAICT it is a core issue):

```JS
let compile = require("google-closure-compiler-js").compile;
console.log(compile({jsCode: [
  {
    path:"./foo.js",
    src:"const type = 5; export {type};"
  }, {
    src:"import {type} from './foo.js';"
  }
]});
```

gives 

```JS
{ compiledCode: '',
  errors:
   [ { file: './foo.js',
       description: 'Parse error. cannot use keyword \'type\' here.',
       type: 'JSC_PARSE_ERROR',
       lineNo: 1,
       charNo: 24 },
     { file: 'Input_1',
       description: 'Parse error. \'as\' expected',
       type: 'JSC_PARSE_ERROR',
       lineNo: 1,
       charNo: 12 } ],
  warnings: [] }
```

dimvar:
Repro in the debugger:
[...]

PR #2835, opened 2018-03-03
Allow typed ES6 keywords in import and export statements

jbouwman:
Default behavior forbids the use of TypeScript keywords in import and export statements when ES6 typed features are inactive: modify scanner to emit IdentifierTokens when parseTypeSyntax is false.

Export decl parser coerces identifiers in exports to keywords: check there also.

Addresses #2005

### DIFF
--- a/test/com/google/javascript/jscomp/parsing/ParserTest.java
+++ b/test/com/google/javascript/jscomp/parsing/ParserTest.java
@@ -3657,6 +3657,15 @@ public final class ParserTest extends BaseJSTypeTestCase {
     parseError("export * as s from './someModule';", "'from' expected");
   }
 
+  public void testImportExportTypescriptKeyword() {
+    mode = LanguageMode.TYPESCRIPT;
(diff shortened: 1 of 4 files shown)


test/com/google/javascript/jscomp/parsing/ParserTest.java
```diff
@@ -3657,6 +3657,14 @@ public void testExport() {
     parseError("export * as s from './someModule';", "'from' expected");
   }
 
+  public void testExportTypescriptKeyword() {
```
MatrixFrog: Can you add a test for: `"import { namespace } from './input0.js';"`? Thanks @lauraharker for catching that.

jbouwman: Ok, done. Lemme know if you need anything else or there are style/doc concerns.

MatrixFrog:
Sorry, I think our import script is getting confused since there are multiple commits. Can you squash this down to a single commit?

jbouwman:
Squashed!

PR #2856, opened 2018-03-20
Fix charset in inline data url sourcemaps

tadeegan:
I noticed that babel produces inline source maps of the form:

`//# sourceMappingURL=data:application/json;charset=utf-8;base64,eyJ2ZXJzaW9uIjozLCJzb3VyY2VzIjpbImZyb250ZW5kL2FwcC9yZWFjdC1jb21wb25lbnRzL2FwcC5qc3giXSwibmFtZXMiOltdLCJtYXBwaW5ncyI6IkFBQUE7Ozs7QUFJQSxTQUFTLE9BQVQsUUFBd0IsWUFBeEI7QUFDQSxPQUFPLEdBQVAsTUFBZ0IsVUFBaEI7QUFDQSxPQUFPLFlBQVAsTUFBeUIsbUJBQXpCO0FBQ0EsU0FBUyxTQUFULFFBQTBCLGFBQTFCO0FBQ0EsT0FBTyxNQUFQLE1BQW1CLFdBQW5COztBQUVBO0FBQ0EsTUFBTSxlQUFOLFNBQThCLE1BQU0sU0FBcEMsQ0FBOEM7QUFDMUM7QUFDQSxnQkFBWSxLQUFaLEVBQW1CLE9BQW5CLEVBQTRCO0FBQ3hCLGNBQU0sS0FBTixFQUFhLE9BQWI7QUFDQSxhQUFLLEtBQUwsR0FBYTtBQUNULDJCQUFlLGVBRE4sRUFDdUI7QUFDaEMsNkJBQWlCO0FBQ2IsMEJBQVU7QUFERyxhQUZSO0FBS1QsMkJBQWU7QUFDWCwwQkFBVTtBQURDLGFBTE47QUFRVCw0Q0FBaUMsSUFBSSxJQUFKLEVBQUQsQ0FBVyxPQUFYLEVBUnZCLENBUTZDO0FBUj......`

This is not recognized by the SourceMapResolver because of the 'charset=utf-8'

See acceptable data url format: https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs

### DIFF
--- a/src/com/google/javascript/jscomp/SourceMapResolver.java
+++ b/src/com/google/javascript/jscomp/SourceMapResolver.java
@@ -20,14 +20,21 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.io.BaseEncoding;
 import java.net.URI;
 import java.net.URISyntaxException;
-import java.nio.charset.StandardCharsets;
 import java.nio.file.FileSystems;
 import javax.annotation.Nullable;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
 
 /** Utility class for resolving source maps and files referenced in source maps. */
 @GwtIncompatible("Accesses the file system")
 public class SourceMapResolver {
-  private static final String BASE64_URL_PREFIX = "data:application/json;base64,";
+  private static final String BASE64_URL_PREFIX = "data:";
+  private static final String BASE64_START = "base64,";
+  // For now only accept utf-8. We could use the actual charset information in the future.
+  private static final String[] acceptedMetadata = {
+          "application/json;charset=utf-8;",
+          "application/json;" // default is utf-8 if unspecified.
+  };
 
   /**
    * For a given //# sourceMappingUrl, this locates the appropriate sourcemap on disk. This is use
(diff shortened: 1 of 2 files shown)


src/com/google/javascript/jscomp/SourceMapResolver.java
```diff
@@ -53,6 +60,28 @@ static SourceFile extractSourceMap(
...
+  private static String extractBase64String(String url) {
+    if (url.startsWith(BASE64_URL_PREFIX)) {
+      if (url.contains(BASE64_START)) {
+        int base64StartIndex = url.indexOf(BASE64_START);
+        String metadata = url.substring(BASE64_URL_PREFIX.length(), base64StartIndex);
+        for (String md : acceptedMetadata) {
+          if (md.equals(metadata)) {
+            byte[] data = BaseEncoding.base64().decode(
+                    url.substring(base64StartIndex + BASE64_START.length()));
+            return new String(data, StandardCharsets.UTF_8);
```
MatrixFrog: nit: Let's static-import `UTF_8` and remove `StandardCharsets.` here.

src/com/google/javascript/jscomp/SourceMapResolver.java
```diff
@@ -53,6 +60,28 @@ static SourceFile extractSourceMap(
     return getRelativePath(jsFile.getName(), sourceMapURL);
   }
 
+  /**
+   * Based on https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs
+   * @param url
+   * @return String or null.
+   */
```
MatrixFrog: add `@Nullable` annotation since we explicitly expect this to return null in some cases.

MatrixFrog:
otherwise lgtm

tadeegan:
done

MatrixFrog:
Can you squash this into a single commit as well? (We have a bug open to fix our script so it can accept multi-commit PRs but I don't expect we'll actually fix it anytime soon, if I'm being honest)

tadeegan:
done.

> but I don't expect we'll actually fix it anytime soon, if I'm being honest

classic

PR #2864, opened 2018-03-26
Add activeElement attribute to the document object

vegegoku:
fix for https://github.com/google/elemental2/issues/42


MatrixFrog:
We have this at https://github.com/google/closure-compiler/blob/master/externs/browser/ie_dom.js#L805 -- can you just add the proper `@type` there instead?

PR #2865, opened 2018-03-26
Missing Document.activeElement

vegegoku:
fix for https://github.com/google/elemental2/issues/40

### DIFF
--- a/externs/browser/ie_dom.js
+++ b/externs/browser/ie_dom.js
@@ -800,6 +800,7 @@ Document.prototype.loadXML;
 // http://msdn.microsoft.com/en-us/library/ms531073(VS.85).aspx
 
 /**
+ * @type {Element}
  * @see http://msdn.microsoft.com/en-us/library/ms533065(VS.85).aspx
  */
 Document.prototype.activeElement;


MatrixFrog:
Note that this may take a day or three to submit, since we will need to do some Google-wide testing to make sure this doesn't introduce new type errors somewhere.

vegegoku:
That is perfect, for the time being i have a workaround for this.
Thank you. :+1:

vegegoku:
@MatrixFrog looking more into this i found that the activeElement is not and ie only, and it should be supported by most of the browsers according to MDN

https://developer.mozilla.org/en-US/docs/Web/API/DocumentOrShadowRoot/activeElement
i have also tested this on both firefox 59.0.1 and chrome 65.0.3325.181 and worked on both.

so maybe this does not belong to ie only anymore.

so what do you suggest?

MatrixFrog:
It being in `ie_dom.js` doesn't mean it's IE only, it just means that most likely it was in IE _first_ so that's where it got added. We should perhaps move it to `w3c_dom1.js` or something, but since most projects just use all the default externs it doesn't actually matter too much which file it's in.

Also, I've been busy this week with various things, and was out for a day and a half, so sorry my estimate of "a day or three" was inaccurate, will be more like a week.

vegegoku:
First i would like to thank you for the efforts, we dont need to harry on this.

my need is to have `document.activeElement` available in [elemental2](https://github.com/google/elemental2) which depends on the closure externs, looking at elemental2 which does not use browser specific externs like in the dom module
https://github.com/google/elemental2/blob/676d1b937b068f078b9b6e57781fba26e3a12064/java/elemental2/dom/BUILD#L88

having the `document.activeElement` in the ie extern wont make part of the elemental2 api, same thing goes for other api element
like `document.exeCommand` which supposed to be supported by most of the modern browsers
https://caniuse.com/#search=execCommand but is still appears only in the gecko externs.

so the question is, is it safe to move such api to none browser specific externs?

MatrixFrog:
There are a few breakages that would be easily fixed if the type were HTMLElement rather than just Element. I don't suppose that's correct though, since the active element could be, say, an SVG element.

MatrixFrog:
> is it safe to move such api to none browser specific externs?

Most projects just load in all the default externs which includes ie_dom.js, w3c_dom.js, etc. So in practice it doesn't much matter which file something is defined in.

MatrixFrog:
This is landed internally, will be out on the next push to Github.

niloc132:
Following @jDramaix's guidance (from https://github.com/google/elemental2/issues/40), I'm looking to take the next step here and move some members like `activeElement` and `outerHTML` from the IE-specific file to a w3c-based on. How does one pick the correct one? MDN gives me these two references:

* `activeElement` https://w3c.github.io/webcomponents/spec/shadow/#extensions-to-the-documentorshadowroot-mixin (I'll note that this is present in html5.js, as part of `ShadowRoot`, though the linked spec seems to be clear that this is part of both Document and ShadowRoot)
* `outerHTML` https://w3c.github.io/DOM-Parsing/#x2-3-extensions-to-the-element-interface

Furthering my confusion, I'm fairly sure that w3c_dom1.js is not just things from the DOM level 1, such as NodeList.forEach (and so on for a handful of items in each file), so perhaps there is another strategy for organization at work, and the `@fileoverview` is somewhat misleading as to where contents should come from.

Apologies for misusing this closed pull request in this way, but I'm hoping to submit a complete patch, rather than use it as a mechanism for trial-and-error to understand how to help.

jDramaix:
I planned to spent some time (but not right now) to clean up all these extern files and merging all w3c_domX.js in one file. Try to pick up one that make sense but it doesn't really matter if the apis is added to the wrong w3c_dom.js file.

niloc132:
Thanks @jDramaix, do let us know if there is anything we can do to help out, work you can carve out for us, etc.

ChadKillingsworth:
The extern filenames are misleading - they are not intended to be "browser specific". Rather often times that indicates where documentation for the symbol might have first been found.

In the intervening years, things have been added to seemingly random places. The intent is that users will use all the externs in an environment. ("browser" in this case).

The compiler has the concept of additional environments with distinct extern sets (say like for workers), but nobody has done the work to define such externs and maintain them.

We also had the idea that externs could potentially be generated from Chrome IDL files so that they did not have to be manually managed. But again - nobody has done that work.

PR #2889, opened 2018-04-15
Remove the deprecated replaceWholeText api

vegegoku:
this PR Removes the Text.replaceWholeText which is according to MDN is obsolete
https://developer.mozilla.org/en-US/docs/Web/API/Text/replaceWholeText

### DIFF
--- a/externs/browser/w3c_dom3.js
+++ b/externs/browser/w3c_dom3.js
@@ -559,13 +559,6 @@ Element.prototype.setIdAttributeNS = function(namespaceURI, localName, isId) {};
 Text.prototype.wholeText;
 
 /**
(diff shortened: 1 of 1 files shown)


MatrixFrog:
Nice!

brad4d:
looks like there's no Google code that uses this method, so it should be totally safe to remove.

brad4d:
imported PR
I've just started testing to confirm it doesn't breaking any internal Google projects.

brad4d:
something went wonky with the test run, so I'm having to start it again.

brad4d:
submitted internally - should be in Monday's push to GitHub

PR #2900, opened 2018-04-19
Implement string literal trim() folding

puddly:
The ECMAScript spec requires implementations at least trim off Unicode 3.0 whitespace characters plus the BOM. I've included all of code points with the [`White_Space` property](https://www.unicode.org/Public/UNIDATA/PropList.txt) from Unicode 10.0.0.


brad4d:
@puddly 

I think this PR looks OK to accept, but we do need the CLA signature to do so.

puddly:
Alright, I've signed it.

brad4d:
imported and sent for internal testing and review

brad4d:
submitted internally.
should be in Monday's push to GitHub

PR #2901, opened 2018-04-22
Updated TranspileResultTest to have separate tests for case sensitive and case insensitive files systems.

chetmichals:
Updated TranspileResultTest to have separate tests for case sensitive and case insensitive files systems. Test would previously fail when run on Windows Operating Systems. Made it so the test is OS agnostic, but check for the behavior of the underlying file system.

The test was changed from a Junit3 Style test to a Junit4 Style test so that the Assume functionality would work correctly with the project's build is setup.

This test failure was one of the ones reported in this issue:
https://github.com/google/closure-compiler/issues/2683

### DIFF
--- a/test/com/google/javascript/jscomp/transpile/TranspileResultTest.java
+++ b/test/com/google/javascript/jscomp/transpile/TranspileResultTest.java
@@ -20,13 +20,27 @@ import static com.google.common.truth.Truth.assertThat;
 
 import com.google.common.testing.EqualsTester;
 import java.nio.file.Paths;
-import junit.framework.TestCase;
+import org.junit.Assume;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
 
 /** Tests for {@link TranspileResult}. */
+@RunWith(JUnit4.class)
+public final class TranspileResultTest {
 
-public final class TranspileResultTest extends TestCase {
+  /*
+   * Java currently lacks a built in method for checking if a the file system is
+   * case sensitive or not, so check if two different cased paths are equivalent
+   * or not, to determine which of the two testEquals tests to run.
+   */
 
-  public void testEquals() {
+  private static final boolean CASE_SENSITIVE_FILE_SYSTEM =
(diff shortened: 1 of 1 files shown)


brad4d:
@chetmichals 
Changing from JUnit3 to JUnit4 may make it difficult for us to accept this PR.
We'll have to make changes to our internal build setup to handle that, and we really have bigger fish to fry right now.

Would you consider dropping the JUnit 4 changes?

chetmichals:
I can drop the JUnit 4 changes if needed, I would just have to not use the Assume functionality and use an if statement instead (Since Assume works via throwing a specific exception which Junit 4+ handle to mean the test isn't applicable, which the JUnit 3 test runner just sees as a failure).

One thing to note, is that there is other tests that use the JUnit 4 runner currently, like [RecoverableJsAstTest](https://github.com/google/closure-compiler/tree/master/test/com/google/javascript/jscomp/RecoverableJsAstTest.java) which is in the 'parent' package, and most of the tests in the [com.google.javascript.refactoring package](https://github.com/google/closure-compiler/tree/master/test/com/google/javascript/refactoring)

But there may be reasons I am not aware of for why the test runner shouldn't be changed. Just let me know that you still want me to change it back to a JUnit 3 style test even if there are other JUnit 4 styles tests, and I can get the pull request updated.

brad4d:
I'm curious to see how easy/hard it will be to upgrade this single test to JUnit 4, so I'll take this on.

FYI, I'll be on travel April 30 - May 3, so I don't expect to have time to work on this until after that.

brad4d:
@chetmichals could you resolve conflicts so I can import this cleanly?
Thanks

chetmichals:
When looking over the merge conflict, it appears that the changes made to how the transpiler class works makes these changes to the tests unnecessary. The URL class is case sensitive inherently, which eliminates the platform specific behavior.

Specifically, the changes made in this commit:
https://github.com/google/closure-compiler/commit/18c9a695fcb2781ef8f25bd04dd27e9a383b1a4c

PR #2903, opened 2018-04-22
Fix filename of Math.expm1 polyfill

teppeis:
Currently the file name is `math/exp1m`.
`resources.json` may need to be updated.


brad4d:

PR #2905, opened 2018-04-24
Fix: Map/Set polyfill should store -0 in keys as +0

teppeis:
- https://www.ecma-international.org/ecma-262/8.0/#sec-map.prototype.set
- https://www.ecma-international.org/ecma-262/8.0/#sec-set.prototype.add
- https://kangax.github.io/compat-table/es6/#test-Map_-0_key_converts_to_+0
- https://kangax.github.io/compat-table/es6/#test-Set_-0_key_converts_to_+0

### DIFF
--- a/src/com/google/javascript/jscomp/js/es6/map.js
+++ b/src/com/google/javascript/jscomp/js/es6/map.js
@@ -142,6 +142,8 @@ $jscomp.polyfill('Map',
 
   /** @override */
   PolyfillMap.prototype.set = function(key, value) {
+    // normalize -0/+0 to +0
+    key = key === 0 ? 0 : key;
     var r = maybeGetEntry(this, key);
     if (!r.list) {
       r.list = (this.data_[r.id] = []);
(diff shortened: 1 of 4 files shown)


brad4d:
@teppeis are you sure this change is really needed?
I mean, did you actually see a case in running code where our polyfill for Map or Set treated +0 and -0 differently?

@shicks what do you think?

teppeis:
@brad4d 
In actual code, `-0` is rarely used.
However, I think that polyfill should conform to the specification as much as possible if it doesn't have performance or other issues obviously.

brad4d:
@teppeis what I'm saying is I suspect our code actually handles `-0` correctly already even though that isn't visible in the code.
Can you show otherwise?

teppeis:
@brad4d These tests are failed in current Closure Compiler polyfill.

```js
const map = new Map().set(-0, 'foo');
assertEquals(Infinity, 1 / map.keys().next().value);

const set = new Set([-0]);
assertEquals(Infinity, 1 / set.keys().next().value);
```

brad4d:

brad4d:
added a few more test cases
starting internal submission

teppeis:
@brad4d @lauraharker thanks!

PR #2911, opened 2018-04-30
Fix Array.from polyfill: 2nd arg of mapFn is missing

teppeis:
- https://www.ecma-international.org/ecma-262/8.0/#sec-array.from
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from


brad4d:
imported PR and running through internal testing and review

teppeis:
@brad4d thanks!

PR #2914, opened 2018-05-06
Prevent an internal compiler error when generating TypeScript code fo…

bowenni:
…r optional parameters


bowenni:
Closing this PR. I will send a change internally instead.

PR #2921, opened 2018-05-12
When composing source maps, utilize source content provided in an input map

ChadKillingsworth:
I discovered that with the JS version, source maps are not composed. This turned out to be from a bug where we only use the sourcesContent from an input source map to display errors and we only load the content from disk.

In addition, we only composed source maps if the content of the original file could be loaded. This wasn't needed.

Adds support to source maps where content provided in an input source map can be forwarded to the output source map. Also allows the GWT version to compose source maps when the input source map provides the original content in the `sourcesContent` field.

### DIFF
--- a/src/com/google/javascript/jscomp/Compiler.java
+++ b/src/com/google/javascript/jscomp/Compiler.java
@@ -69,6 +69,7 @@ import java.io.PrintStream;
 import java.io.Serializable;
 import java.util.AbstractSet;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


ChadKillingsworth:
Fixed. Let me know if you want me to squash the commits.

blickly:
LGTM, let's squash and I'll bring it in! Thanks Chad

ChadKillingsworth:
Hmm 200824732982ab6032e53de74eee59a49c775bc0 did some of the same work. I'll have to see what of my changes still needs applied.

ChadKillingsworth:
@blickly squashed. This is significantly simpler due to 2008247 handling many of the same concerns.

ChadKillingsworth:
@blickly I prefer to have the tests, but they were not actually failing (at least when I ran it). I'm not sure why. A lot of things changed with the other commit and I haven't had a chance to fully understand the implications.

PR #2922, opened 2018-05-13
Add more flags to the GWT version

ChadKillingsworth:
In order to move the GWT version of the compiler into the same distribution package as the java version, new flags needed added.

These changes should be still compatible with the https://github.com/google/closure-compiler-js repo as well as the https://github.com/google/closure-compiler-npm/tree/include-js-version.

There are no direct tests for GWT here, but the built jscomp.js file can be copied to each of the other repos and the tests there run.

Depends on #2921 to properly compose source maps.

cc @samthor

### DIFF
--- a/src/com/google/javascript/jscomp/gwt/client/GwtRunner.java
+++ b/src/com/google/javascript/jscomp/gwt/client/GwtRunner.java
@@ -106,8 +106,13 @@ public final class GwtRunner implements EntryPoint {
     boolean useTypesForOptimization;
     String tracerMode;
     String moduleResolutionMode;
+    String jsOutputFile;
+    String[] formatting;
+    boolean sourceMapIncludeContent;
+    boolean parseInlineSourceMaps;
 
     // These flags do not match the Java compiler JAR.
+    @Deprecated
     File[] jsCode;
     File[] externs;
     boolean createSourceMap;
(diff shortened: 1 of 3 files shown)


src/com/google/javascript/jscomp/gwt/client/GwtRunner.java
```diff
@@ -242,32 +254,57 @@ private static native JavaScriptObject createError(String file, String descripti
...
       }
-      marker = OUTPUT_MARKER;
+      postfix = flags.outputWrapper.substring(pos + marker.length());
     }
-
-    String prefix = outputWrapper.substring(0, pos);
-    SourceMap sourceMap = compiler.getSourceMap();
-    if (sourceMap != null) {
-      sourceMap.setWrapperPrefix(prefix);
+    if (flags.createSourceMap) {
```
samthor: double-conditional?

src/com/google/javascript/jscomp/gwt/client/GwtRunner.java
```diff
@@ -461,8 +520,7 @@ private static void disableUnsupportedOptions(CompilerOptions options) {
         if (path == null) {
           path = unknownPrefix + i;
         }
-        path += ".map";
-        SourceFile sf = SourceFile.fromCode(path, file.sourceMap);
+        SourceFile sf = SourceFile.fromCode(path + ".map", file.sourceMap);
         inputSourceMaps.put(path, new SourceMapInput(sf));
```
samthor: nice catch. I guess we were specifying that the source map was for the `file.js.map` file :|

src/com/google/javascript/jscomp/gwt/client/GwtRunner.java
```diff
@@ -242,32 +254,57 @@ private static native JavaScriptObject createError(String file, String descripti
...
+          // ignore
+        }
+        file.sourceMap = b.toString();
+      }
     }
-    return prefix + code + outputWrapper.substring(pos + marker.length());
+
+    file.src = prefix + code + postfix;
+    outputFiles.add(file);
+    output.compiledFiles = outputFiles.toArray(new File[outputFiles.size()]);
```
samthor: Will this only be of size one right now? Should we update the JS wrappers to recognize multiple file output rather than relying on compiledCode?

src/com/google/javascript/jscomp/gwt/client/GwtRunner.java
ChadKillingsworth: That was only one of the problems with source maps ...

src/com/google/javascript/jscomp/gwt/client/GwtRunner.java
samthor: I think the existing wrappers just assume a single file out and get the source via `compiledCode`. I wrote this originally to be somewhat filename agnostic.
Just keep an eye on it when you merge in the gulp/grunt/etc wrappers.

PR #2925, opened 2018-05-15
Rename code splitting flags from "module" to "chunk"

ChadKillingsworth:
To avoid confusion with proper JS modules, rename the code splitting flags to use the name "chunk". Adds aliases for the old flag names for backwards compatibility.


blickly:
Nice, I like it!

PR #2940, opened 2018-05-19
Add generics support to the reflectObject polyfill

ChadKillingsworth:
The polymer pass injects an object reflection polyfill. With NTI, the return type was coerced to match expectations. But with OTI, it causes warnings. By making the reflectObject polyfill use generics, the warning is removed.

The type of code this effects looks something like:

```js
class FooElement extends Polymer.Element {
  static get is() { return 'foo-element'; }
  /** @return {Object<string, (!Function|{type: !Function, value: *})>} */
  static get properties() {
    return $jscomp.reflectObject(FooElement, {
      myProp: {
        type: String,
        value: 'foo'
      }
    });
  }
}
```

### DIFF
--- a/src/com/google/javascript/jscomp/js/util/reflectobject.js
+++ b/src/com/google/javascript/jscomp/js/util/reflectobject.js
@@ -19,7 +19,7 @@
 /**
  * Definition for object reflection. See goog.reflect.object.
  *
- * @param {!Function} type Type to cast to.
+ * @param {?Object} type Type to cast to.
  * @param {T} object Object literal to cast.
  * @return {T} The object literal.
  * @template T
--- a/src/com/google/javascript/jscomp/js/util/reflectobject.js
+++ b/src/com/google/javascript/jscomp/js/util/reflectobject.js
@@ -20,8 +20,9 @@
  * Definition for object reflection. See goog.reflect.object.
  *
  * @param {!Function} type Type to cast to.
- * @param {Object} object Object literal to cast.
- * @return {Object} The object literal.
+ * @param {T} object Object literal to cast.
+ * @return {T} The object literal.
+ * @template T
  */
 $jscomp.reflectObject = function(type, object) {
   return object;
--- a/test/com/google/javascript/jscomp/PolymerPassTest.java
+++ b/test/com/google/javascript/jscomp/PolymerPassTest.java
@@ -109,8 +109,9 @@ public class PolymerPassTest extends CompilerTestCase {
           "/** @const */ $jscomp.scope = {};",
           "/**",
           " * @param {!Function} type",
-          " * @param {Object} object",
-          " * @return {Object}",
+          " * @param {T} object",
+          " * @return {T}",
+          " * @template T",
           " */",
           "$jscomp.reflectObject = function (type, object) { return object; };");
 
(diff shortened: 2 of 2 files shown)


brad4d:
imported and sent for internal testing

brad4d:
internal testing looks good
sent to @blickly for internal review

src/com/google/javascript/jscomp/js/util/reflectobject.js
```diff
@@ -20,8 +20,9 @@
  * Definition for object reflection. See goog.reflect.object.
  *
  * @param {!Function} type Type to cast to.
```
brad4d: @concavelenz has pointed out that the return type should be related to the `type` parameter not to the  `object` parameter.

Ideally I think we should have this, but I suspect it doesn't work.
```javascript
 * @param {function(new:T, ...var_args)} type
 * @param {?Object}
 * @param {?T}
```

test/com/google/javascript/jscomp/PolymerPassTest.java
```diff
@@ -109,8 +109,9 @@
           "/** @const */ $jscomp.scope = {};",
           "/**",
           " * @param {!Function} type",
-          " * @param {Object} object",
-          " * @return {Object}",
+          " * @param {T} object",
+          " * @return {T}",
+          " * @template T",
```
brad4d: This seems to just echo back the text of the polyfill.
Is it possible to add some kind of test to show that this actually fixes the problem you're trying to address?

ChadKillingsworth: Currently the only way I know how to test this is with a full integration test including property renaming. That would be very brittle.

Once the type checker understands ES6 code, the `disableTypeCheck()` calls can be removed in the PolymerPassTest.java file and it should automatically be tested.

concavelenz: Just have it return ? then.

ChadKillingsworth: I think we might lose substantial type checking if we use the unknown type. For instance, the original code would look like:

```js
/**
 * @typedef {{
 * type: !Function,
 * value: *,
 * readOnly: (boolean | undefined),
 * computed: (string | undefined),
 * reflectToAttribute: (boolean | undefined),
 * notify: (boolean | undefined),
 * observer: (string | function(*,*) | undefined)
 * }}
 */
let PolymerElementPropertiesMeta;

/**
 * @typedef {Object<string, !PolymerElementPropertiesMeta>}
 */
let PolymerElementProperties;

class FooElement extends Polymer.Element {
  static get is() { return 'foo-element'; }
  /** @return {!PolymerElementProperties} */
  static get properties() {
    return {
      myProp: {
        type: String,
        value: 'foo'
      }
    };
  }
}
```

The PolymerPass then wraps the anonymous object with the reflection call. I still want type checking on the anonymous object itself, if possible. Don't we keep that with generics?

brad4d: Sounds like I can consider this comment thread as closed with "works as intended" then.
@ChadKillingsworth do you see how to fix the issue I pointed out on `reflectobject.js` line 20?

src/com/google/javascript/jscomp/js/util/reflectobject.js
ChadKillingsworth: I missed this question. It needs to match [goog.reflect.object](https://google.github.io/closure-library/api/goog.reflect.html) which is set as `?Object`. I'll see if that works.

ChadKillingsworth: @brad4d This is now fixed

brad4d:
Imported latest changes.
Discussed with @concavelenz 
Added improvements to JSDoc comments to help reduce future confusion.
I'm submitting this internally now.

PR #2959, opened 2018-06-01
Fix IE Symbol errors when used with output wrapper

### LINKED ISSUE #2957: 'Symbol' undefined on IE 11 when using --output_wrapper
In https://github.com/google/shaka-player/issues/1455#issuecomment-393250035 we discovered that references to `Symbol` in ES6 polyfills don't work inside our wrapper.  This seems similar to #1227.

To fix this for our project, I was able to recompile the release of the compiler we use (v20180402) after going through `src/com/google/javascript/jscomp/js/es6/` and changing references to `Symbol` to `$jscomp.global.Symbol`.

The same problem seems to be present in the latest compiler sources on `master`, as well.  Would a PR for this be welcome?

joeyparrish:
When using an output wrapper, the polyfill for "Symbol" does not
necessarily appear in "window".  So all internal references to
"Symbol" need to go through "$jscomp.global".

Closes #2957


joeyparrish:
We will be fixing our output wrapper instead.  Thanks for the tips about $jscomp.global!

PR #2961, opened 2018-06-02
Let HTMLOptionsCollection extend HTMLCollection

TimvdLippe:
Per https://developer.mozilla.org/en-US/docs/Web/API/HTMLOptionsCollection the HTMLOptionsCollection is a subclass of HTMLCollection.

I discovered this issue when writing a `for-of`-loop which reported the following error:
```
src/patch-accessors.js:312: WARNING - actual parameter 1 of $jscomp.makeIterator does not match formal parameter
found   : HTMLOptionsCollection
required: (Arguments<?>|Iterable<?>|Iterator<?>|string)
          for (const option of select.options) {
                               ^^^^^^^^^^^^^^
```
By fixing this subclass definition I think this error should be resolved as well.

### DIFF
--- a/externs/browser/w3c_dom2.js
+++ b/externs/browser/w3c_dom2.js
@@ -128,25 +128,32 @@ HTMLCollection.prototype.namedItem = function(name) {};
 
 /**
  * @constructor
- * @implements {IObject<(string|number),HTMLOptionElement>}
- * @implements {IArrayLike<!HTMLOptionElement>}
- * @see http://www.w3.org/TR/DOM-Level-2-HTML/html.html#HTMLOptionsCollection
+ * @extends {HTMLCollection<HTMLOptionElement>}
+ * @see https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#htmloptionscollection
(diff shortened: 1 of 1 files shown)


TimvdLippe:
Hm, this change seems to have broken some tests. I will need to investigate later why they are broken.

blickly:
Would you mind squashing the commits into one to make it easier for us to import?

After that, Tiago, could you import this one?

TimvdLippe:
@blickly Done!

TimvdLippe:
Apparently this was merged into https://github.com/google/closure-compiler/commit/bb02a47b73e7fde19f44742801f53afd4693aebb but the PR was not closed. Was this an issue with MOE @tjgq ?

blickly:
Oops, looks like Tiago forgot the "Closes issues
https://github.com/google/closure-compiler/pull/2961" stanza that makes the
issue auto-close.

TimvdLippe:
Ah okay, not a detrimental mistake in MOE luckily :smile: Thanks for the confirmation!

PR #2967, opened 2018-06-06
Better handling of CommonJS exports in complex expressions

### LINKED ISSUE #2932: Using the result of an assignment to module.exports causes an error
Simple code:
```
const width = 800;
const vwidth = exports.vwidth = width;
```
compiled with `--process_common_js_modules` results in an error:
```
a.js:2: ERROR - variable width is undeclared
const vwidth = exports.vwidth = width;
                                ^^^^^
```
With `--jscomp_warning=undefinedVars` and without optimization we get the following output:
```
var module$experimental$users$cyriln$js$a = {"default":{}};
module$experimental$users$cyriln$js$a["default"].vwidth = 800;
var vwidth$$module$experimental$users$cyriln$js$a = module$experimental$users$cyriln$js$a["default"].vwidth = width;
```
Where did `width` go?

ChadKillingsworth:
Update handling of cases where a CommonJS export assignment is used in a larger expression. Replaces the export assignment with just a reference to the export itself.

Fixes #2932

### DIFF
--- a/src/com/google/javascript/jscomp/ProcessCommonJSModules.java
+++ b/src/com/google/javascript/jscomp/ProcessCommonJSModules.java
@@ -1374,6 +1374,7 @@ public final class ProcessCommonJSModules extends NodeTraversal.AbstractPreOrder
       }
 
       if (root.getParent().isAssign()
+          && root.getGrandparent().isExprResult()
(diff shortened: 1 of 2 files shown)


blickly:
I'll try to import this now

PR #2978, opened 2018-06-16
Recognize UMD patterns when define is referenced as a property of window

ChadKillingsworth:
The fingerprint2 library references `define` as `window.define` in the UMD pattern. Update the common js module processing to recognize this variant.


brad4d:
@blickly could you handle this PR?
Thanks

PR #2979, opened 2018-06-19
Add support for code splitting to the GWT version

ChadKillingsworth:
Adds full code splitting functionality to the GWT version.

Since the whole of AbstractCommandLineRunner is marked as GWTIncompatible, copy out common methods to a new CommandLineRunnerUtils class that can be shared with the GWT runner.

### DIFF
--- a/src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
+++ b/src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
@@ -111,7 +111,6 @@ import javax.annotation.Nullable;
  *
  * @author bolinfest@google.com (Michael Bolin)
  */
-@GwtIncompatible("Unnecessary")
 public abstract class AbstractCommandLineRunner<A extends Compiler,
     B extends CompilerOptions> {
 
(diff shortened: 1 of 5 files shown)


brad4d:
@blickly Could you take this PR?
Thanks

blickly:
It's a bit tricky to migrate all the references to the moved types
that were in AbstractCommandLineRunner, like FlagUsageException and
JsModuleSpec.

Would it be possible to move the GWT-incompatible bits out of
AbstractCommandLineRunner instead?

ChadKillingsworth:
@blickly That's a lot of annotations, but done.

Since we are going this route, I added the rest of the flags needed to this same PR. It reworks the flag logic in the GWT runner to match the order and logic in CommandLineRunner. The GWT version can be fully tested with https://github.com/google/closure-compiler-npm/tree/gwt-chunks - just copy the `jscomp.js` file over and run `yarn test` (you'll need to disable compiler downloading though).

This also now is fully supported by the webpack plugin: https://github.com/webpack-contrib/closure-webpack-plugin/tree/js-version

This PR will now officially let us deprecate the js standalone distribution.

PR #2987, opened 2018-06-22
Fixed SpeechRecognition to be an EventTarget

phistuck:


externs/browser/w3c_speech.js
```diff
@@ -72,6 +72,7 @@ HTMLInputElement.prototype.onwebkitspeechchange;
 // W3C Web Speech API implemented in Chrome M23
 /**
  * @constructor
+ * @extends {EventTarget}
```
brad4d: Since we cannot make `EventTarget` a constructor at this time, perhaps you mean `@implements`?

brad4d:
@phistuck after changing `@extends` to `@implements`, could you confirm that the maven tests pass?
Thanks.

phistuck:
They will probably fail because I will need to add the appropriate methods first... I might get to it later this week.

phistuck:
@brad4d - is this approved?

brad4d:
This is stalled, because I've been too busy with other stuff to work on it.

brad4d:
I've now imported that change and sent it for internal tests and review with @nreid260

brad4d:
Submitted internally. Should get pushed to GitHub tomorrow.

PR #2988, opened 2018-06-23
Made EventTarget a constructor

phistuck:
Firefox and Chrome supports this.

### DIFF
--- a/externs/browser/w3c_event.js
+++ b/externs/browser/w3c_event.js
@@ -25,7 +25,7 @@
 
 
 /**
- * @interface
+ * @constructor
  * @see https://dom.spec.whatwg.org/#interface-eventtarget
  */
 function EventTarget() {}


phistuck:
Perhaps I should add @override to any addEventListener/removeEventListener/dispatchEvent?

phistuck:
This needs a broader change (change @implements to @extend everywhere as well as probably add @override), I see.
Since it is a constructor now, maybe we can just remove all of those (newly turned) overrides?

brad4d:
According to MDN Safari still doesn't support `EventTarget` as a constructor, and it's unknown whether Edge does.
https://developer.mozilla.org/en-US/docs/Web/API/EventTarget

Landing this change will be extremely difficult, and the cost cannot be justified while there are still major browser vendors who don't support it.
Sorry.

freshp86:
> According to MDN Safari still doesn't support EventTarget as a constructor, and it's unknown whether Edge does.

This has traditionally not been a blocker in adding/updating the compiler's externs.

>  Landing this change will be extremely difficult, and the cost cannot be justified 

Can you elaborate a bit on why this is considered costly? Is there no backwards compatible update we can come up with that can eliminate that cost of changing @implements to @inherits? How about introducing a new class `EventTargetBase` or something to that effect?

brad4d:
There are approximately 300 occurrences of either `@implements {EventTarget}` or `@extends {EventTarget}` in internal Google code that would have to be updated.
Now that I look that's a lower number than I expected, but it's still a significant amount of effort
to get all of them changed and deal with whatever errors crop up as a result.

It isn't worth that effort for something that isn't even safe to use in at least one major browser.

freshp86:
For internal Google usage 300 occurrences means that there are quite a few re-implementations of native `EventTarget`. Providing an alternative, like proposed in previous comment, would be a stepping stone in slowly removing unnecessary implementations in favor of the default one, once browser support is there (or already for code that only deals with Chrome, like internal Chrome extensions for example).

Are you opposed to adding `EventTarget2` or `EventTargetBaseClass` or something to that effect, so that new code can use that already? Otherwise there is no path forward.

freshp86:
@brad4d: Can you comment on the previous suggestion about providing a path forward on Closure compiler users who want to use `new EventTarget()`? I am trying to determine what is the best patch forward for our case in Chromium, see https://bugs.chromium.org/p/chromium/issues/detail?id=854268.

brad4d:
Fixing this will be really tricky, because we'll have to fix existing code that defines interfaces that extend `EventTarget`. If it isn't an interface anymore, that becomes invalid.

If you really want to use it right now, I think your best bet would be to do something like this in your code.

```javascript
/**
 * @constructor
 * @implements {EventTarget}
 */
var NativeEventTarget = goog.global()['EventTarget'];
```

Then use / extend `NativeEventTarget`.

phistuck:
@brad4d - it either does not work, or I am holding it wrong -
https://closure-compiler.appspot.com/home#code%3D%252F%252F%2520%253D%253DClosureCompiler%253D%253D%250A%252F%252F%2520%2540compilation_level%2520ADVANCED_OPTIMIZATIONS%250A%252F%252F%2520%2540output_file_name%2520default.js%250A%252F%252F%2520%2540language_out%2520ES5%250A%252F%252F%2520%253D%253D%252FClosureCompiler%253D%253D%250A%250A%252F**%2520%2540constructor%250A%2520%2520%2520%2520%2540implements%2520%257BEventTarget%257D%2520*%252F%250Avar%2520NativeEventTarget%2520%253D%2520self%255B%2522EventTarget%2522%255D%253B%250A%250Aclass%2520Foo%2520extends%2520NativeEventTarget%250A%257B%250A%257D

PR #2992, opened 2018-06-28
Default for empty-message replacement.

salsolatragus:
Replacing a call node in an empty message leads to an NoSuchElementException. This fix adds a check and defaults to an empty string message in this case, since the same happens when replacing function nodes. Alternatively, an exception with a higher-level message should be thrown, to explain the problem.

This problem was identified by our automated detector [MUDetect](https://github.com/stg-tud/MUDetect/). We would very much appreciate your feedback on our patch. Thank you very much!


salsolatragus:
I signed it!

brad4d:
imported for internal testing and review with @lauraharker

PR #3000, opened 2018-07-02
Replace more references to "module" in CommonJS modules.

### LINKED ISSUE #2999: Dangling module reference after processing CommonJS modules
With https://github.com/google/closure-compiler/commit/3edafa4c1168d29894ad2ca6021eb72be6e73af5, revisions were made to rewriting of code containing `module` during CommonJS rewriting.

There appears to be a bit of code in Lodash (https://github.com/lodash/lodash/blob/ab960bdc964fa1b1c46630f6d6681e55f6af862c/isBuffer.js#L7) that can trip up this logic, leaving an undefined `module` reference after code is rewritten.

This might be a good repro (or at least sufficient to illustrate the issue):

```
$ java -jar target/closure-compiler-1.0-SNAPSHOT.jar --process_common_js_modules true
The compiler is waiting for input via stdin.
var x = module && !module.nodeType && module;
;; Return and Ctrl-D
var x=!module.nodeType&&{};
```

Prior to this change, there was evidently no attempt to rewrite the `module` references (presumably they were OK). For example:

```
$ java -jar closure-compiler-v20180610.jar --process_common_js_modules true
The compiler is waiting for input via stdin.
var x = module && !module.nodeType && module;
;; Return and Ctrl-D
var x=module&&!module.nodeType&&module;
```

I suspect that a bona fide repro would need to involve more files and context to truly illustrate the full problem. Please let me know if this is needed.

The context where this was discovered is in automated tests involving running the ClojureScript compiler's test suite using the latest Closure Compiler snapshot. Here is the associated ClojureScript ticket: https://dev.clojure.org/jira/browse/CLJS-2796 with this downstream regression being associated with the Closure Compiler commit https://github.com/google/closure-compiler/commit/3edafa4c1168d29894ad2ca6021eb72be6e73af5.

ChadKillingsworth:
Rewrites additional uses of "module" in CommonJS modules.

Fixes #2999

### DIFF
--- a/src/com/google/javascript/jscomp/ProcessCommonJSModules.java
+++ b/src/com/google/javascript/jscomp/ProcessCommonJSModules.java
@@ -1368,10 +1368,19 @@ public final class ProcessCommonJSModules extends NodeTraversal.AbstractPreOrder
           }
 
         case GETPROP:
-          if (n.matchesQualifiedName(MODULE + ".id")) {
+          if (n.matchesQualifiedName(MODULE + ".id")
+              || n.matchesQualifiedName(MODULE + ".filename")) {
             Var v = t.getScope().getVar(MODULE);
             if (v == null || v.isExtern()) {
               n.replaceWith(IR.string(t.getInput().getPath().toString()).useSourceInfoFrom(n));
+            }
+          } else if (allowFullRewrite
+              && n.getFirstChild().isName()
+              && n.getFirstChild().getString().equals(MODULE)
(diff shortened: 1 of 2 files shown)


mfikes:
Thank-you @ChadKillingsworth! I've confirmed that, with this change, the ClosureScript downstream tests pass.

lauraharker:
Imported the PR and sent to @jplaisted for review

PR #3001, opened 2018-07-02
If the compiler output is a JSON stream, don't restrict chunk naming

ChadKillingsworth:
With JSON streams, it's the responsibility of the caller to create the file structure.

Allows chunk names to contain path segments and other non-identifier characters.

### DIFF
--- a/src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
+++ b/src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
@@ -1162,8 +1162,11 @@ public abstract class AbstractCommandLineRunner<A extends Compiler,
       if (isInTestMode()) {
         modules = modulesSupplierForTesting.get();
       } else {
-        for (JsModuleSpec m : jsModuleSpecs) {
-          checkModuleName(m.getName());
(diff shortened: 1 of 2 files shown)


lauraharker:
Imported and sent for review

src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
```diff
@@ -1162,8 +1162,11 @@ protected int doRun() throws IOException {
       if (isInTestMode()) {
         modules = modulesSupplierForTesting.get();
       } else {
-        for (JsModuleSpec m : jsModuleSpecs) {
-          checkModuleName(m.getName());
+        if (config.jsonStreamMode == JsonStreamMode.IN
```
lauraharker: Per comment from internal reviewer, can this use ```equals()```? Google's internal Java practices guide suggests ```equals()```  for enums unless both sides are nullable

lauraharker:
Chad, one other internal review comment - do you know if there are any tests for this? If not, is it reasonable to add a unit test?

ChadKillingsworth:
Review issue addressed. Test added. Commits squashed.

PR #3007, opened 2018-07-05
Synchronize even more flags between the GwtRunner and CommandLineRunner

ChadKillingsworth:
Adds the ability to output a disclaimer for documentation when the GwtRunner doesn't support a flag.

Matches more flags in the GwtRunner to the CommandLineRunner version.

I verified that the Gwt version passes all the tests in the npm package with these changes.


lauraharker:
Imported and sent to @jplaisted for review.

Issue #3016, opened 2018-07-09
closure-compiler skips all checks if input and output are set to ECMASCRIPT2017


nicks:
Create the following `main.js`:

```
/** @return {string} */
var main = function() { return 2; };
```

And run the following commandline:

```
java -jar closure-compiler-v20180610.jar --js main.js --jscomp_error checkTypes --language_in ECMASCRIPT_2017 --language_out ECMASCRIPT_2017
```

Expectation:
closure-compiler should fail with a type error "inconsistent return type"

Actual:
closure-compiler prints `WARNING: Skipping pass checkTypes` (!!!) but otherwise passes successfully

Note that this works as expected if you pass `--language_out ECMASCRIPT5`

nicks:
It's also worth noting that this weird bug is the default behavior in the API because of the way the API defaults shake out

lauraharker:
We're still working on making the typechecker handle ES2015 syntax, so all the typechecking passes and some of the optimizations are disabled in ES2015+ output right now. 

The goal is to have full support for ES2015 output by the end of Q3.

The warning is coming from https://github.com/google/closure-compiler/blob/407cb25bcf5c4afd23737ebc1414c04f84a31829/src/com/google/javascript/jscomp/PhaseOptimizer.java#L284

nicks:
Could this be an error that makes the whole compiler job fail?

The problem is that we're using the API. The alternative for us us to reverse engineering the API, and properly report an error for all the invalid combinations.

Also notice that the default ConpilerOptions is es2017 -> es2017. It seems really bad for the default behavior to be broken.

lauraharker:
hmm @ChadKillingsworth what are your thoughts on making this an error? Given https://github.com/google/closure-compiler/pull/2998

ChadKillingsworth:
No it can't be an error. For instance, if you have `language_in=ECMASCRIPT_2017` and `language_out=ECMASCRIPT5`, you can get skipped pass warnings for the `RewriteGoogJsImport` pass as it doesn't yet support ECMASCRIPT_2017 (or newer). These errors are not correctable by a user either without losing language features.

What needs to happen is language out needs to be limited to the highest level supported. Since the highest language out level is ECMASCRIPT_2015, anything higher should trigger an error.

This particular issue is pretty much only seen by the API as the open source command line runner sets both the input and output language level defaults.

I'm really surprised @nicks uses the API though ...

nicks:
[...]

PR #3021, opened 2018-07-17
Add an IntegrationTest demonstrating issue #3016

nicks:


brad4d:
I appreciate the benefit of this PR for demonstrating #3016, but I don't see any benefit in actually importing it.
Thanks.

nicks:
:+1: thanks!

PR #3026, opened 2018-07-19
Support multiline ES6 import statements

### LINKED ISSUE #3025: Support multiline ES6 `export ... from` syntax
### Description of problem

When parsing ES6 code of the form `export ... from`, the compiler fails to generate the correct `goog.require` statement.

### Steps to reproduce

Given the following files:

```js
//greet.js
export default function greet(m) {
    document.write("\nHello, " + m);
};
```
```js
//hello.js
export {
    default as helloGreet
} from "./greet";
```

Execute:
```
java -jar $JAR -O BUNDLE --formatting PRETTY_PRINT --module_resolution NODE --js_output_file=/dev/stdout -W VERBOSE --debug --dependency_mode STRICT --entry_point hello.js greet.js hello.js | grep -A 11 -E 'greet|hello'
```

### Expected outcome

Closure compiler should detect `greet.js` as a dependency of `hello.js` and output them both under strict dependency mode.

```js
//greet.js
$jscomp.registerAndLoadModule(function($$require, $$exports, $$module) {
  "use strict";
  Object.defineProperties($$exports, {"default":{enumerable:true, get:function() {
    return greet;
  }}});
  function greet(m) {
    document.write("\nHello, " + m);
  }
}, "greet.js", []);

//hello.js
$jscomp.registerAndLoadModule(function($$require, $$exports, $$module) {
  "use strict";
  Object.defineProperties($$exports, {helloGreet:{enumerable:true, get:function() {
    return module$greet["default"];
  }}});
  var module$greet = $$require("greet.js");
}, "hello.js", ["greet.js"]);

```

### Actual outcome

[`JsFileParser`](/src/com/google/javascript/jscomp/deps/JsFileParser.java) fails compute the list of requires.

```js
//hello.js
$jscomp.registerAndLoadModule(function($$require, $$exports, $$module) {
  "use strict";
  Object.defineProperties($$exports, {helloGreet:{enumerable:true, get:function() {
    return module$greet["default"];
  }}});
  var module$greet = $$require("greet.js");
}, "hello.js", ["greet.js"]);

```

### Workaround/patches

Condensing the export to a single line works around the issue but may be infeasible for third party libs (e.g. from NPM).

Avoiding `JsFileParser` also fixes the problem:

```patch
diff --git a/src/com/google/javascript/jscomp/CompilerInput.java b/src/com/google/javascript/jscomp/CompilerInput.java
index 62609562b..4e4d3e3dc 100644
--- a/src/com/google/javascript/jscomp/CompilerInput.java
+++ b/src/com/google/javascript/jscomp/CompilerInput.java
@@ -285,7 +285,7 @@ public class CompilerInput extends DependencyInfo.Base implements SourceAst {
 
     // If the code is a JsAst, then it was originally JS code, and is compatible with the
     // regex-based parsing of JsFileParser.
-    if (ast instanceof JsAst && JsFileParser.isSupported()) {
+    if (false && ast instanceof JsAst && JsFileParser.isSupported()) {
       // Look at the source code.
       // Note: it's OK to use getName() instead of
       // getPathRelativeToClosureBase() here because we're not using
```

### Analysis

Due to the line parsing nature of `JsFileParser`, it fails to look ahead when faced with a line that starts with export (or import) but doesn't contain a semicolon. In fact, under shortcut mode, it appears to halt the parser.

au-phiware:
Prior to this change, due to the line parsing nature of
JsFileLineParser, the parser would fail to match import/export
statements that span multiple lines.

This change tracks prior lines when JsFileParser encounters an import or
export statement without a terminating semicolon, then parses the set of
lines when a semicolon is encountered.

resolves #3025

### DIFF
--- a/src/com/google/javascript/jscomp/deps/JsFileParser.java
+++ b/src/com/google/javascript/jscomp/deps/JsFileParser.java
@@ -21,6 +21,7 @@ import static com.google.common.base.Preconditions.checkState;
 import com.google.common.annotations.GwtIncompatible;
 import com.google.common.base.CharMatcher;
 import com.google.javascript.jscomp.CheckLevel;
+import com.google.javascript.jscomp.DiagnosticType;
 import com.google.javascript.jscomp.ErrorManager;
 import com.google.javascript.jscomp.JSError;
 import com.google.javascript.jscomp.deps.DependencyInfo.Require;
@@ -80,6 +81,31 @@ public final class JsFileParser extends JsFileLineParser {
           // Finally, this should be the entire statement, so ensure there's a semicolon.
           + "\\s*;");
 
+  /** Pattern for matching the beginning multi-line import statements. */
+  private static final Pattern ES6_MODULE_START_PATTERN =
+      Pattern.compile("^(im|ex)port\\b");
+
+  /** Pattern for matching a failed multi-line import statements.
+   *
+   * A multi-line import statement can fail because it is missing a semicolon at the end.
+   * Note that it might not be an import statement at all (it could be a valid export
+   * statement), so a failed match will not raise an error.
+   *
+   * We consider a multi-line failed when the parser encounters a line beginning with a commonly
+   * used keyword or the beginning of a property in a object literal (which is common in an export
+   * statement).
+   */
+  private static final Pattern ES6_MODULE_FAIL_PATTERN =
+      Pattern.compile(
+          // Ignore whitespace at the beginning of the line.
+          "^\\s*(?:"
+          // Then match one of these commonly used keyword.
+          + "(?:return|if|var|function|for|while|let|const)\\b"
+          // Alternatively match a common property name in an object literal.
+          + "|(?:[a-zA-Z_$][0-9a-zA-Z_$]*\\s*:)"
+          // Ignore the rest of the line.
+          + ")");
+
   /**
    * Pattern for 'export' keyword, e.g. "export default class ..." or "export {blah}".
    * The '\b' ensures we don't also match "exports = ...", which is not an ES6 module.
@@ -92,11 +118,18 @@ public final class JsFileParser extends JsFileLineParser {
   /** The start of a bundled goog.module, i.e. one that is wrapped in a goog.loadModule call */
   private static final String BUNDLED_GOOG_MODULE_START = "goog.loadModule(function(";
 
+  /** Multi-line import statement length limit */
+  private static final int ES6_MODULE_LIMIT = 1024 - 100;
+  private static final DiagnosticType PARSE_LONG_WARNING = DiagnosticType.warning(
+      "DEPS_PARSE_WARNING", "Statement length limit reached ({1} chars) after reading line {0}");
+
   /** Matchers used in the parsing. */
   private final Matcher googMatcher = GOOG_PROVIDE_REQUIRE_PATTERN.matcher("");
 
   /** Matchers used in the parsing. */
   private final Matcher es6Matcher = ES6_MODULE_PATTERN.matcher("");
+  private final Matcher es6StartMatcher = ES6_MODULE_START_PATTERN.matcher("");
+  private final Matcher es6FailMatcher = ES6_MODULE_FAIL_PATTERN.matcher("");
 
   /** The info for the file we are currently parsing. */
   private List<String> provides;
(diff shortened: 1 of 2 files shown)


ChadKillingsworth:
I think this will fail horribly if anyone follows the "no semicolons" philosophy.

au-phiware:
I agree; I can think of a number of options to deal with 'no semicolon'. But I note the semicolon is currently a requirement in the regex. So, support for the 'no semicolon approach' should be a separate feature request. This PR addresses a bug that I am facing now.

Having said that, I'm happy to put a cap on the buffer length or terminate when a valid `from` clause is found or something else. Please, let me know what you think the best approach is.

As an aside, during my debugging process, I tried removing the JsFileParser and just used the AST which works. Is JsFileParser present for performance reasons? In which case, should the 'no semicolon approach' be supported at all? Maybe a new compiler option could be provided for those that need it.

Again, I suggest that it is a separate issue and shouldn't be a reason to delay this PR/bugfix.

au-phiware:
The bug is #3025, sorry if it wasn't clear, I only mention it in the commit message.

au-phiware:
@ChadKillingsworth, are you suggesting that `JsFileParser` can be dropped?

ChadKillingsworth:
Ahh it's the bundle mode. I have not used that yet, but in general the regex parsing has been nothing but troublesome in the compiler.

I'll leave this to someone else to review, but I would highly recommend bounding the algorithm so that it can't consume an entire file.

au-phiware:
Thanks for your suggestion @ChadKillingsworth, I shall look into it.

Can you please suggest another person to review.

au-phiware:
I note that the public interface of `JsFileParser` is essentially consists of little more than `parseFile(String, String, String)`, which takes the file contents as a string. So the entire file already resides in memory!

brad4d:
I think @blickly is the right person to take this one.
He'll be back from vacation on Thursday, August 26.

blickly:
Would you mind squashing these into a single commit?  That just makes it a bit easier for us to merge

blickly:
It seems like some of the new assumptions clash with code that we have written at Google.  I'll try to boil down to some reasonable examples.

blickly:
Looks like the issues I'm seeing are related to using relative path imports, like e.g.

```javascript
import {nonDefaultExport} from './subdir/otherFile.js';
```

For these files, I'm hitting [UNKNOWN_PATH_IMPORT](https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/deps/DepsGenerator.java#L81) errors where we weren't before, for imports that look to be correct.  Can you think of why this might be the case?

If it helps, I think that the entry point looks to be [DepsGenerator.computeDependencyCalls](https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/deps/DepsGenerator.java#L132)

au-phiware:
Superficially, I don't see why relative path imports would cause a problem, indeed I added your example as a test case and it passes: [`JsFileParserTest.java`, line 203](https://github.com/google/closure-compiler/pull/3026/commits/0fa360bc7362558dd683f38a4127956d720b6ee6#diff-be89b1e9bec31f969a0fc18a383f658cR203).

Unfortunately I'm unable to reproduce the error that you've encountered and I'm not familiar with the code beyond the files that lead me to this point (`CompilerInput` and `JsFileParser`). So the following is highly speculative...

It's possible that this change produces a more complete list of `Require` objects for some files. (Note that this richer dependency information was not needed for successful compilation since the compiler previously obtained the files through some other mechanism.) The consequence of having this richer information is that it is made available for analysis at some earlier stage; a stage that is perhaps missing some critical information pertaining to the new information or possibly a stage that contains some unknown bugs of its own. This argument is severely lacking in evidence and I'm afraid I don't have the experience or knowledge with the code base to fill in the blanks.

Is it possible distil a minimal reproduction case from your codebase, @blickly?

Having already said I can't reproduce the error; I have encountered problems of mismatched symbols in the past. I.e. a file is emitted as providing a particular symbol, yet the code the requires that file is using a slightly different symbol to require it. The result is that the name/symbol is not found. I have no idea if that class of problem is related to this UNKNOWN_PATH_IMPORT error and I have assumed that there was something in the inputs or options to the compiler that was wrong (i.e. user error).

ChadKillingsworth:
It's interesting that those files don't even have multi-line import / export statements: https://github.com/webcomponents/shadycss/blob/master/src/apply-shim-utils.js

blickly:
Sorry for taking so long to get back to this, and thanks for trying all the avenues to reproduce the issue yourself.

I'm looking at minimizing the example now into something that can be reproduced in a unit test, and I'll report back here when I get something working.

blickly:
Closing since there's been no response for about a year.

PR #3033, opened 2018-07-21
Added support for new URLSearchParam({foo: 1})

phistuck:
This is implemented everywhere (modern).

### DIFF
--- a/externs/browser/url.js
+++ b/externs/browser/url.js
@@ -24,7 +24,7 @@
 
 /**
  * @constructor
- * @implements {Iterable<!Array<string>>}
+ * @implements {Iterable<!Array<string>>|Object<string, ?>}
  * @param {(string|!URLSearchParams)=} init
  */
 function URLSearchParams(init) {}
--- a/externs/browser/url.js
+++ b/externs/browser/url.js
@@ -23,7 +23,7 @@
  */
 
 /**
- * @typedef {Iterable<string>}
+ * @typedef {Array<string>}
  */
 var URLSearchParamsTupleType;
 
@@ -31,12 +31,14 @@ var URLSearchParamsTupleType;
  * @see https://url.spec.whatwg.org/#interface-urlsearchparams
  * @constructor
  * @implements {Iterable<!Array<string>>}
- * @param {(string|!Iterable<!URLSearchParamsTupleType>|!Object<string,string>)=} init
+ * @param {(string|!Array<!URLSearchParamsTupleType>|!Object<string,string>)=} init
  * when |init| is a string, it is basically parsed as a query string (a=b&b=c).
- * when |init| is an iterable of iterables of string ([["a", "b"], ["b", "c"]]),
+ * when |init| is an array of arrays of string ([["a", "b"], ["b", "c"]]),
  *  it must contain pairs of strings,
  *  where the first item in the pair will be interpreted as a key
  *  and the second as a value.
+ *  Note - the specification uses Iterable rather than Array,
+ *  but this is not supported in Edge 17 - 18.
  * when |init| is an object, keys and values will be interpreted as such
  *  ({a: "b", b: "c"}).
  */
--- a/externs/browser/url.js
+++ b/externs/browser/url.js
@@ -24,7 +24,7 @@
 
 /**
  * @constructor
- * @implements {Iterable<!Array<string>>|Object<string, ?>}
+ * @implements {Iterable<!Array<string>>|Object<*, *>}
  * @param {(string|!URLSearchParams)=} init
  */
 function URLSearchParams(init) {}
--- a/externs/browser/url.js
+++ b/externs/browser/url.js
@@ -23,19 +23,22 @@
  */
 
 /**
- * @typedef {IArrayLike<string>}
+ * @typedef {Iterable<string>}
  */
 var URLSearchParamsTupleType;
 
 /**
- * @typedef {IArrayLike<!URLSearchParamsTupleType>|Iterable<!URLSearchParamsTupleType>}
- */
-var URLSearchParamsTupleSequenceType;
-
-/**
+ * @see https://url.spec.whatwg.org/#interface-urlsearchparams
  * @constructor
  * @implements {Iterable<!Array<string>>}
- * @param {(string|!URLSearchParamsTupleSequenceType|!Object<string,string>)=} init
+ * @param {(string|!Iterable<!URLSearchParamsTupleType>|!Object<string,string>)=} init
+ * when |init| is a string, it is basically parsed as a query string (a=b&b=c).
+ * when |init| is an iterable of iterables of string ([["a", "b"], ["b", "c"]]),
+ *  it must contain pairs of strings,
+ *  where the first item in the pair will be interpreted as a key
+ *  and the second as a value.
+ * when |init| is an object, keys and values will be interpreted as such
+ *  ({a: "b", b: "c"}).
  */
 function URLSearchParams(init) {}
 
--- a/externs/browser/url.js
+++ b/externs/browser/url.js
@@ -24,8 +24,8 @@
 
 /**
  * @constructor
- * @implements {Iterable<!Array<string>>|Object<*, *>}
- * @param {(string|!URLSearchParams)=} init
+ * @implements {Iterable<!Array<string>>}
+ * @param {(string|!URLSearchParams|Object<*, *>)=} init
  */
 function URLSearchParams(init) {}
 
--- a/externs/browser/url.js
+++ b/externs/browser/url.js
@@ -23,9 +23,19 @@
  */
 
 /**
+ * @typedef {IArrayLike<string>}
+ */
+var URLSearchParamsTupleType;
+
+/**
+ * @typedef {IArrayLike<!URLSearchParamsTupleType>|Iterable<!URLSearchParamsTupleType>}
+ */
+var URLSearchParamsTupleSequenceType;
+
+/**
  * @constructor
  * @implements {Iterable<!Array<string>>}
- * @param {(string|!URLSearchParams|Object<*, *>)=} init
+ * @param {(string|!URLSearchParamsTupleSequenceType|!Object<string,string>)=} init
  */
 function URLSearchParams(init) {}
 


externs/browser/url.js
```diff
@@ -25,7 +25,7 @@
 /**
  * @constructor
  * @implements {Iterable<!Array<string>>}
- * @param {(string|!URLSearchParams)=} init
+ * @param {(string|!URLSearchParams|Object<*, *>)=} init
```
brad4d: Hmmm, I'm looking at the spec here.
https://url.spec.whatwg.org/#interface-urlsearchparams
And the MDN docs here
https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/URLSearchParams

The `init` argument is very flexible, type-wise, so I think something like this is needed.

```javascript
/**
 * @typedef {IArrayLike<string>}
 */
var URLSearchParamsTupleType;

/**
 * @typedef {IArrayLike<!UrlSearchParamsTupleType>|Iterable<!URLSearchParamsTupleType>}
 */
var URLSearchParamsTupleSequenceType;

/**
 * @constructor
 * @implements {Iterable<!Array<string>>}
 * @param {(string|!URLSearchParamsTupleSequenceType|!Object<string,string>)=} init
 */
function URLSearchParams(init) {}
```
Since `URLSearchParams` implements `Iterable<!Array<string>>` it is a subtype of the `URLSearchParamsTupleSequenceType` I defined above. We don't need to explicitly name it as an option.

BTW, I notice that MDN says `URLSearchParams` is "experimental" and the compatibility table at the bottom of the doc page says it definitely doesn't work in IE and has unknown status in Safari.

phistuck: The table seems to be a bit outdated/missing some information (it also uses the old compatibility data format, which probably means it was not touched in a while). It works in all of the latest versions of all of the modern browsers - I checked using BrowserStack.
(Note - Internet Explorer 11, released in 2013, is not modern ;))

brad4d: I'm willing to back off of the more elaborate code I suggested above.

What use-cases are you actually wanting to enable with this PR?
Could you provide sample code showing what they would look like.
Then we can work from there to get annotations that work.
I find the spec for this method and the MDN docs rather obtuse,
so I think concrete examples would be very helpful.

brad4d:
@phistuck are you still interested in getting this PR merged?
If so, could you please update your branch to our current head and respond to the question I posed in the review comments?

externs/browser/url.js
phistuck: The use case is very simple,
```
new URLSearchParams({parameter1: 1, parameter2: 2})
```


Some routing code I have -
```
 async goTo(path, queryParameters)
 {
  const canNavigate = await this._isNavigationAllowed();
  if (canNavigate === false)
  {
   return;
  }

  const url = new URL(path, this._locationProvider.href);
  if (queryParameters && queryParameters instanceof URLSearchParams)
  {
   url.search = queryParameters.toString();
  }
  else if (typeof queryParameters === kECMAScript.basicTypes.object)
  {
   url.search = new URLSearchParams(queryParameters).toString();
  }
  this._commitURL(null, null, url);
  this._handleLocationChange();
 }
```

And usage -
```
router.goTo(
 kApplication.routePaths.settings,
 {error: kApplication.errors.authorizationFailed});
```

phistuck:
Sorry for taking the time with this, I got too busy to even be doing my own things. :(

I do not mind the elaborate code (as long as it matches all of the needed features implemented in modern browsers and complies with the specification, of course). I will try to get into looking back into it soon.

externs/browser/url.js
phistuck: I realize that the use case does not justify `Object<*, *>`. I basically want numbers and others to be able to function as stringified keys here (in order to be able to also create a query string result of `1=8&2=9&true=k&toStringOfObject`).
Maybe this can be more explicit...
Object<(string|number|boolean), (string|number|boolean)>

On the other hand, if you choose Closure Compiler with advanced optimizations, you usually prefer to be type accurate... So maybe Object<string, string> is the right way to go.

phistuck:
Done according to your suggestion (fixed a casing issue, though). I tested my application and it fixed the issue. Thank you!

externs/browser/url.js
```diff
@@ -22,10 +22,20 @@
...
+/**
+ * @typedef {IArrayLike<!URLSearchParamsTupleType>|Iterable<!URLSearchParamsTupleType>}
+ */
+var URLSearchParamsTupleSequenceType;
+
 /**
  * @constructor
  * @implements {Iterable<!Array<string>>}
- * @param {(string|!URLSearchParams)=} init
+ * @param {(string|!URLSearchParamsTupleSequenceType|!Object<string,string>)=} init
```
brad4d: Please add:
`@see https://url.spec.whatwg.org/#interface-urlsearchparams`

externs/browser/url.js
```diff
@@ -22,10 +22,20 @@
  * @author rdcronin@google.com (Devlin Cronin)
  */
 
+/**
+ * @typedef {IArrayLike<string>}
+ */
```
brad4d: Please add this descriptive comment or a more correct version.

`* In practice this is an array with 2 strings. The first is a parameter name. The second is its value.`

externs/browser/url.js
brad4d: Could you add some comments explaining how `init` is used.
What does it mean for it to be a string?
What does it mean for it to be an `Object`?
What does it mean for it to be a sequence of tuples?

externs/browser/url.js
brad4d: obsoleted by my comment above.

externs/browser/url.js
```diff
@@ -22,10 +22,20 @@
...
+var URLSearchParamsTupleType;
+
+/**
+ * @typedef {IArrayLike<!URLSearchParamsTupleType>|Iterable<!URLSearchParamsTupleType>}
+ */
+var URLSearchParamsTupleSequenceType;
```
brad4d: Per my reasoning in the comment above, I think we should just drop this definition.

The `@param` below then becomes.
`{(string|!Iterable<!URLSearchParamsTupleType>|!Object<string,string>)=} init`

externs/browser/url.js
```diff
@@ -22,10 +22,20 @@
  * @author rdcronin@google.com (Devlin Cronin)
  */
 
+/**
+ * @typedef {IArrayLike<string>}
```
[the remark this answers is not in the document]
phistuck: @brad4d - is there a way to tell the compiler that this array/iterable must have a length of 2? Chrome throws an exception otherwise, `Sequence initializer must only contain pair elements`.
I think there will be other cases where the length matters. For example, I assume the upcoming Object.fromEntries would require it.

And yes, it must be an `Iterable`.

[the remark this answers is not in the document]
brad4d: Sadly there's no way to say "must be array/iterable of length 2".

Someday maybe we'll finally have time to implement tuple types. *(wistful sigh)*

FYI, I'm a bit concerned that the spec is defined purely in terms of `sequence`, when I couldn't find a hard definition of what that translates to in JS terms.

I'm inclined to go with this PR as it currently exists, but I do think we have this possible failure case.

1. code input to the compiler uses a non-Array Iterable object for the individual tuples or the list of tuples.
2. after compilation this code runs on an older browser that assumed it would be getting an Array, so it fails, because it doesn't know to use the `[Symbol.iterator]` property for iteration and just tries to use a counting loop with a `length` property.

This doesn't seem terribly likely, since I expect people will just always use Arrays.
That's what's most convenient, after all.

brad4d:
I've imported this and sent it for testing to see what breaks.

externs/browser/url.js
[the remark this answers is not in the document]
phistuck: Seems pretty defined to me -
https://heycam.github.io/webidl/#es-sequence

<strike>Also, I believe `URLSearchParams` was implemented by the time browsers already supported iterables (and even more so with this constructor overload that expects a sequence as a parameter), so it is not possible non-array iterables are not supported. Either sequences are not supported at all, or they are fully supported.</strike>
However... I just tested Edge 17 and 18 (going down soon, though. And 16 did not support `URLSearchParams` at all) - Edge does not support iterables, only arrays -
```js
new URLSearchParams((a=new Map(),a.set('a', '1'),a)).toString()
// "" in Edge
// "a=1" in Chrome 71, Firefox 64 and Safari 12
new URLSearchParams([(a=new Set(),a.add('a'),a.add('b'),a)]).toString()
// Invalid argument in Edge
// "a=b" in Chrome 71, Firefox 64 and Safari 12
```

Should I change it to array of arrays to cater for poor, poor Edge?

phistuck:
@brad4d - did you see my question?
https://github.com/google/closure-compiler/pull/3033#discussion_r251741611

externs/browser/url.js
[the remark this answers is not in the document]
brad4d: No, I don't think we should change it to an array of arrays.

As you said, Edge 16 doesn't support URLSearchParams at all.

We don't polyfill web-standard stuff, just core JS features, so it is expected that when you write code using web-standard stuff you still have to be concerned about what your customers' browsers actually support. Sadly we just can't do everything. :)

brad4d:
@phistuck sorry for the late reply. I think I answered your question in my head but failed to type it into GitHub.

Current status is that I have internal reviewer approval, but I have one failing test related to clutz that has to get sorted out before submission.

externs/browser/url.js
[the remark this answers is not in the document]
phistuck: @brad4d - I am not sure how the fact that Edge 16 does not support this at all affects the decision. Edge 17 and 18 (but not the future Chromium based Edge, obviously) do not support Iterable in this case at all which is exactly what you were worried about before in https://github.com/google/closure-compiler/pull/3033#discussion_r251611405. Edge 17 and 18 will be here for a while (with a low market share, but still).

(I have not tested old versions of other browsers, so old versions may have had Array and not Iterable support in this case, too, at some point)

I am not necessarily advocating Array and not Iterable here (I do have a slight preference for lessening the footguns by supporting only Array), I am just verifying that you reconsidered your previous worries in light of the new information about Edge turning out supporting exactly what you were worried about.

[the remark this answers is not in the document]
phistuck: @brad4d - also, this is not a polyfill. The externs do try to reflect the lowest common denominator in many cases in order to prevent footguns by naïve developers.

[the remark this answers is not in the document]
brad4d: @phistuck You've convinced me.

I think the small amount of pain felt by a few developers who really want to use Iterables here is more than balanced by the very large amount of potential pain caused by people using Iterables and having their code mysteriously fail on Edge.

Of course, their Array-using code will mysteriously fail on even earlier versions of Edge already, because those don't support `URLSearchParam` at all, but, why make things even more dangerous?

I will await your update for import.
Thanks.

brad4d:
imported updates
also tweaked the formatting of the JSDoc comment to make it look prettier

had to send to an additional internal reviewer, but I hope to submit this by Monday

phistuck:
@brad4d - is it fine to ignore the Travis CI failure?
```
Tests run: 7, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.041 sec <<< FAILURE! - in com.google.javascript.jscomp.SourceFileTest
testCachingZipFile(com.google.javascript.jscomp.SourceFileTest)  Time elapsed: 0.023 sec  <<< FAILURE!
com.google.common.truth.ComparisonFailureWithFacts: 
expected: // new content new content new content
but was : // content content content
	at com.google.javascript.jscomp.SourceFileTest.testCachingZipFile(SourceFileTest.java:126)
```

It does not seem related to my change, but I am just making sure you saw it.

phistuck:
@brad4d - or maybe just restart the job...

tjgq:
@phistuck We have been seeing this particular failure lately - I believe @shicks has been trying to track it down. Feel free to ignore it.

brad4d:
@phistuck the failure is definitely a known flake.
We have someone looking at fixing it, but it's totally unrelated to this PR.

I have a change ready to submit internally with all testing looking good.
I'm just waiting for one more reviewer approval that I need.

brad4d:
FYI, my change submitted internally.
I expect it to be pushed to GitHub tomorrow.

phistuck:
Yay, thank you, @brad4d.
By the way, is it worth sending a pull request for removing domainToASCII and domainToUnicode from url.js?
```
/**
 * @param {string} domain
 * @return {string}
 */
URL.domainToASCII = function(domain) {};

/**
 * @param {string} domain
 * @return {string}
 */
URL.domainToUnicode = function(domain) {};
```

Those are not in the specification and apparently only Firefox and Node.js supports them. Maybe worth moving to the Node.js externs?

brad4d:
I don't think that PR would be hard to land, because it looks like no Google internal code is using those methods.

Still, unless these are getting in your way, it's probably not worth your time to do the PR.

PR #3036, opened 2018-07-22
Add support for the --js and --externs flags to the GWT version

ChadKillingsworth:
When the gwt version is running in a NodeJS environment, support loading source files from disk via the `--js` and `--externs` flags.

CC @samthor

### DIFF
--- a/src/com/google/javascript/jscomp/gwt/client/GwtRunner.java
+++ b/src/com/google/javascript/jscomp/gwt/client/GwtRunner.java
@@ -104,6 +104,7 @@ public final class GwtRunner {
     String[] entryPoint;
     String env;
     boolean exportLocalPropertyDefinitions;
+    Object[] externs;
     String[] extraAnnotationName;
     String[] forceInjectLibraries;
     String[] formatting;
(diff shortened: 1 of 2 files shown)


ChadKillingsworth:
I tested this both in a browser and in node.

brad4d:
imported and sent for internal test and review

src/com/google/javascript/jscomp/gwt/client/GwtRunner.java
```diff
@@ -104,13 +104,15 @@
     String[] entryPoint;
     String env;
     boolean exportLocalPropertyDefinitions;
+    Object[] externs;
```
brad4d: Why is `externs` typed as `Object[]` when `js` is `String[]`?

ChadKillingsworth: Unfortunately, past versions of the GWT compiler handled externs differently - users were expected to pass an array of objects:

```json
[{
  path: '/path/to/extern.js',
  src: 'function foo() {}'
}]
```

In order for this not to be a breaking change, the parsing handles either an array of strings or an array of objects.

See https://github.com/google/closure-compiler/pull/3036/files#diff-0420e060726b261dbc58a5ca6e98613aR704

brad4d: OK, passed that info on to internal reviewer.
I think this will land tomorrow.

PR #3039, opened 2018-07-31
update node definitions for os.tmpdir

MidnightWonderer:
`os.tmpDir` has been deprecated and `os.tmpdir` is available for at least way back to node 0.10.  
the definition managed to survive the update in the Closure compiler definitions until these day.


blickly:
Seems like there's no way this change could break anyone.  @lauraharker, could you try to import this when you get a chance?

Issue #3046, opened 2018-08-07
Function Inlining loses captures of "super"


ChadKillingsworth:
Rewriting Async methods produces code such as:

```js
class Foo extends Bar {
  baz() {
    const jscomp$get$super$baz = () => super.baz;
    const jscomp$this = this;
    return function() {
      jscomp$get$super$baz().call(jscomp$this);
    }
  }
}
```

Currently, function inling will inline the `super.baz` return value from the arrow function to produce an invalid reference to `super`. This only seems to show up with ES2015 out as a language.

MatthewMerrill:
Submitted fix internally

PR #3047, opened 2018-08-07
Prevent inlining for "super" captures in RewriteAsyncFunctions

### LINKED ISSUE #3015: Super inside an async method does not produce valid ES2015 output
Using a language out of `ECMASCRIPT_2015`, the following code produces invalid output.

*Original:*
```js
class Foo {
  async bar() {
    console.log('bar');
  }
}

class Baz extends Foo {
  async bar() {
    await Promise.resolve();
    super.bar();
  }
}
```

Compiled with:
```
java -jar compiler.jar -O=SIMPLE --language_out=ECMASCRIPT_2015 \
    --renaming=false --formatting=PRETTY_PRINT --js test.js
```

*Produces:*
```js
// Generator polyfills ...
class Foo {
  bar() {
    return $jscomp.asyncExecutePromiseGeneratorFunction(function*() {
      console.log("bar");
    });
  }
}
class Baz extends Foo {
  bar() {
    const $jscomp$async$this = this;
    return $jscomp.asyncExecutePromiseGeneratorFunction(function*() {
      yield Promise.resolve();
      super.bar.call($jscomp$async$this); // super is not a valid keyword here
    });
  }
}
```

ChadKillingsworth:
When rewriting async functions, an arrow function is used to capture a reference to `super` methods. However function inlining is currently overwriting that change

Closes #3015

Alternatively, fixing #3046 would render this PR unnecessary.


MatthewMerrill:
Working on a fix for all `super`s internally

PR #3066, opened 2018-09-04
Check iterator immediately before access.

salsolatragus:
It is not immediately obvious that the Iterator queried in `constructAddOrStringNode()` always has a next element to fetch, since the `hasNext()` check occurs outside the querying method. The refactoring pulls this check closer to the query, in order to clarify that the access is always safe.

This problem was identified by our automated detector [MUDetect](https://github.com/stg-tud/MUDetect/). We would very much appreciate your feedback on our patch. Thank you very much!

### DIFF
--- a/src/com/google/javascript/jscomp/ReplaceMessages.java
+++ b/src/com/google/javascript/jscomp/ReplaceMessages.java
@@ -180,9 +180,7 @@ final class ReplaceMessages extends JsMessageVisitor {
     checkNode(oldBlockNode, Token.BLOCK);
 
     Iterator<CharSequence> iterator = message.parts().iterator();
-    Node valueNode = iterator.hasNext()
-        ? constructAddOrStringNode(iterator, argListNode)
-        : IR.string("");
+    Node valueNode = constructAddOrStringNode(iterator, argListNode);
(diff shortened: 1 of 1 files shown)


src/com/google/javascript/jscomp/ReplaceMessages.java
```diff
@@ -211,40 +209,44 @@ private void updateFunctionNode(JsMessage message, Node functionNode)
   private static Node constructAddOrStringNode(Iterator<CharSequence> partsIterator,
                                                Node argListNode)
       throws MalformedException {
-    CharSequence part = partsIterator.next();
-    Node partNode = null;
-    if (part instanceof JsMessage.PlaceholderReference) {
-      JsMessage.PlaceholderReference phRef =
-          (JsMessage.PlaceholderReference) part;
+    if (partsIterator.hasNext()) {
```
blickly: Seems like we could avoid extra nesting (and make the diff smaller) by making the bailout condition return early.

tjgq:
Can you squash your pull request into a single commit to make it easier for our import tools to merge? Thanks.

salsolatragus:
> Can you squash your pull request into a single commit to make it easier for our import tools to merge? Thanks.

Sure! Done.

PR #3070, opened 2018-09-10
[FIX] CSS property user-select is not fully supported

### LINKED ISSUE #3069: CSS property user-select is not fully supported
The CSS property `user-select` is currently not fully supported:
Working:
- WebKitUserSelect
- MozUserSelect

Not working:
- UserSelect
- MsUserSelect

I know, that this property is still a working draft but it is already supported by most browsers (with vendor prefix)
See https://developer.mozilla.org/en-US/docs/Web/CSS/user-select

marci4:
Fixes #3069


lauraharker:
Imported the PR for internal testing

PR #3073, opened 2018-09-17
Set parameters in Node methods to required

TimvdLippe:
While working on a polyfill which patches these methods, I discovered that the default externs did not specify the arguments are required. However, per https://dom.spec.whatwg.org/#ref-for-dom-node-insertbefore and the other 3 methods, they are required. Therefore, update the parameter signature to reflect that update.

### DIFF
--- a/externs/browser/w3c_dom1.js
+++ b/externs/browser/w3c_dom1.js
@@ -195,7 +195,7 @@ Node.prototype.parentNode;
 Node.prototype.previousSibling;
 
 /**
- * @param {Node} newChild
+ * @param {!Node} newChild
  * @return {!Node}
  * @see http://www.w3.org/TR/1998/REC-DOM-Level-1-19981001/level-one-core.html#method-appendChild
  */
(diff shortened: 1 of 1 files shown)


tjgq:
Can you squash your pull request into a single commit to make it easier for our import tools to merge? Thanks.

TimvdLippe:
@tjgq Done :smile:

shicks:
This would be nice, but it's currently not feasible.  There's way too much code depending on these externs, and turning off type checking widely enough to get this through is a worse trade-off than simply having inaccurate externs.

shicks:
A better approach would be to move the code in question to TypeScript and take advantage of the more accurate lib.dom.d.ts, which does specify these correctly.  There's better targetted-refactoring tools for TS source code, and it's trivial to add `!` where necessary, while `/** @type {!Node} */ (...)` is quite a bit more difficult to add correctly, and `/** @suppress {checkTypes} */` is outright dangerous.

brad4d:
Sadly completing this PR just isn't feasible. There's too much code that would have to be cleaned up to land it.
Sorry

PR #3074, opened 2018-09-17
Fix types of Document.prototype.importNode

TimvdLippe:
Per https://dom.spec.whatwg.org/#ref-for-dom-document-importnode the first argument must always be a Node and the return type is always a Node. Morever, `deep` is optional.

Found together with #3073 while working on some polyfills.


brad4d:
Created internal issue b/122074150 to track this

PR #3094, opened 2018-09-30
jQuery.prototype.attr

tdelmas:
According to the documentation ( https://api.jquery.com/attr/#attr2 )
Since at least jQuery 1.6:
If arg1 is a string, arg2 is "String or Number or Null"
If arg1 is an object, arg2 is "Function( Integer index, String attr ) => String or Number" and "this is the current element"

In this PR:
- Improved first parameter object definition
- Second parameter can't be a boolean
- Improved second parameters function prototype


lauraharker:
Imported and sent for internal review.

PR #3100, opened 2018-10-11
 Keep original identifier when its missing in input source maps

WearyMonkey:
The TypeScript compiler does not rename any symbols, so does not include any `names` in its output sourcemaps.

The closure compiler when reading mappings from an input sourcemap, applies name from the input sourcemap mapping without checking it is defined.

This results in the closure compiler sourcemap output missing many mapping names.

This fix checks if the input mapping name is defined before applying it.

### DIFF
--- a/src/com/google/javascript/jscomp/SourceMap.java
+++ b/src/com/google/javascript/jscomp/SourceMap.java
@@ -180,7 +180,10 @@ public final class SourceMap {
         sourceFile = sourceMapping.getOriginalFile();
         lineNo = sourceMapping.getLineNumber();
         charNo = sourceMapping.getColumnPosition();
-        originalName = sourceMapping.getIdentifier();
+        String identifier = sourceMapping.getIdentifier();
+        if (identifier != null && !identifier.isEmpty()) {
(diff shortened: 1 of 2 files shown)


ChadKillingsworth:
Looks like a test needs updated: https://github.com/google/closure-compiler/blob/826b0fcdfa4b9f18a448279af8a2f6654cbfa36d/test/com/google/javascript/jscomp/CommandLineRunnerTest.java#L2288-L2295

WearyMonkey:
@ChadKillingsworth tests updated

WearyMonkey:
@ChadKillingsworth squashed 👍

ChadKillingsworth:
Yes - but it may take a few days. A Google Engineer will import this into their main repo, run tests, then push out the changes.

blickly:
Thanks for the contribution. @jplaisted, can you merge this one?

WearyMonkey:
@ChadKillingsworth @blickly @jplaisted any update on this?

jplaisted:
Thanks for the ping, this did fall off my radar. I'll take a look tomorrow.

PR #3103, opened 2018-10-11
Update async await transpilation to avoid captures of `super`

### LINKED ISSUE #3101: ES6 Out: Captures of Super Fail in MS Edge - breaks async/await transpilation
The following code will infinitely loop and alert `"Child foo"` in MS Edge. All other browsers seem to handle this correctly.

```js
class Base {
  foo() {
    alert('base foo');
  }
}

class Child extends Base {
  foo() {
    const parent = () => super.foo;
    alert('Child foo');
    parent().call(this);
  }
}

(new Child()).foo();
```

This breaks async / await transpilation with ES2015 out.

ChadKillingsworth:
Closes #3101 

MS Edge 17 does not properly honor captures of `super` in arrow functions. Avoid this pattern when transpiling async/await.


brad4d:
I don't think we can make this change.
I believe we initially tried just saving the value of `super.foo`, but ran into runtime failures because the value of `super.foo` changed between the point where we'd saved it and the point where we needed to evaluate it.

Is there some workaround for the MS edge bug we could generate here?
What does it do when you reference super() inside an arrow function?
Does an exception get thrown or is the value "wrong" in some way?

ChadKillingsworth:
@brad4d I think I may have found a workable solution: Use `Object.getPrototypeOf`.

I only switch to this technique if classes are not being transpiled. This avoids a dependence on `Object.getPrototypeOf` for IE < 10.

ChadKillingsworth:
I tested this on both Chrome and MS Edge in a project where I'm trying to land ES6 out.

test/com/google/javascript/jscomp/RewriteAsyncFunctionsTest.java
```diff
@@ -142,6 +142,74 @@ public void testInnerSuperReference() {
...
+            "class A {",
+            "  m() {",
+            "    return this;",
+            "  }",
+            "}",
+            "class X extends A {",
+            "  m() {",
+            "    const $jscomp$async$this = this;",
+            "    const $jscomp$async$super$get$m =",
+            "        () => Object.getPrototypeOf(this.constructor).prototype.m;",
```
brad4d: Couldn't we just get the prototype directly from `this`?
```javascript
() => Object.getPrototypeOf(this).m;
```

brad4d:
@ChadKillingsworth 
I like your solution. Very clean.

Is there some reason why we can't just do `Object.getPrototypeOf(this)` for the non-static case?
See also comment on code.

test/com/google/javascript/jscomp/RewriteAsyncFunctionsTest.java
ChadKillingsworth: Unfortunately not. See the comments in the following snippet:

```js
class Base {
  static foo() { console.log('base foo'); }
  bar() { console.log('base bar'); }
}

class Child extends Base {
  static foo() {
    console.log(Object.getPrototypeOf(this), Object.getPrototypeOf(this).foo);
    Object.getPrototypeOf(this).foo.call(this);
  }

  bar() {
    console.log(Object.getPrototypeOf(this), Object.getPrototypeOf(this).bar);
    Object.getPrototypeOf(this).bar(); // logs out "base bar" appropriately
    Object.getPrototypeOf(this).bar.call(this); // infinitely loops and locks up the Chrome tab!
  }
}
```

ChadKillingsworth:
Classes and super references are tricky to transpile. I was able to simplify static references to `super` though.

test/com/google/javascript/jscomp/RewriteAsyncFunctionsTest.java
brad4d: Ah, I see my mistake.
I meant to say `Object.getPrototypeOf(Object.getPrototypeOf(this))`

I wonder which is the safer choice in the face of code that mucks around with prototypes?
Is it more likely that `this.constructor` isn't pointing at the right thing or that an additional object has been inserted into the prototype chain?

brad4d: Discussed this with @concavelenz 
We agree that it seems more correct and slightly safer to call `Object.getPrototypeOf()` twice to get up to the superclass prototype rather than relying on `this.constructor`.

Would you be willing to make that change?

src/com/google/javascript/jscomp/RewriteAsyncFunctions.java
```diff
@@ -243,10 +243,35 @@ private void convertAsyncFunction(NodeTraversal t, LexicalContext functionContex
       newBody.addChildToBack(IR.constNode(IR.name(ASYNC_ARGUMENTS), IR.name("arguments")));
       NodeUtil.addFeatureToScript(t.getCurrentFile(), Feature.CONST_DECLARATIONS);
     }
+    boolean needsSuperTranspilation = compiler.getOptions().needsTranspilationFrom(FeatureSet.ES6);
     for (String replacedMethodName : functionContext.replacedSuperProperties) {
+      // MS Edge 17 cannot properly capture references to "super" in an arrow function.
```
brad4d: Do you have any link we could insert here pointing to the MS Edge bug?
We'd really like to be able to track when the MS Edge bug gets fixed, or if it's already fixed and we're just stuck because of installed browsers not getting the fix.

ChadKillingsworth: I created one and added the link

test/com/google/javascript/jscomp/RewriteAsyncFunctionsTest.java
ChadKillingsworth: Of course - done.

brad4d:
Imported and sent for internal review and testing.

PR #3109, opened 2018-10-14
Add a no-arg constructor for the JsonFileSpec class to support Graal

ChadKillingsworth:
For reflection support, Graal requires classes to which GSON de-serializes have a no-args constructor. Add a private no-args constructor to the `JsonFileSpec` class.

Fixes https://github.com/oracle/graal/issues/680


blickly:
Thanks! @EatingW, can you merge this one?

Issue #3146, opened 2018-11-17
Make RescopeGlobalSymbols usable in a node environment


thheller:
I'm using the `RescopeGlobalSymbols` compiler pass via the `options.setRenamePrefixNamespace` compiler option in `shadow-cljs`. This works great for Browser-targetted builds but `node.js` targetted builds break since the compiler is rewriting all of the node "specials".

```
module.exports -> window.module.exports
require("foo") ->  (0,window.require)("foo")
global -> window.global
```

This could be fixed by either making `SPECIAL_EXTERNS` configurable or just including the node-specific globals by default.

https://github.com/google/closure-compiler/blob/4c97e69df965da4179b639ea3ee2020c98cc3621/src/com/google/javascript/jscomp/RescopeGlobalSymbols.java#L60-L82

Happy to open a PR for this just wanted to check in first about the preferred way.

The added externs would be `global, process, require, module, exports, __filename, __dirname`.

brad4d:
Created Google internal issue b/119770296

brad4d:
@thheller I think your suggestion is a reasonable one, but I'm not personally familiar with this pass, so I could be wrong.

We're unlikely to work directly on this ourselves anytime soon, but if you're willing to create a PR we can review it and merge it if it seems safe to do so.

zavr-1:
`Buffer` is also needed as an extern for the node environment.

thheller:
Indeed. Here is a more complete List: https://nodejs.org/api/globals.html

thheller:
In https://github.com/google/closure-compiler/pull/3154 I opted to make this configurable instead adding another hardcoded fixed set of names.

ChadKillingsworth:
The root issue here is that `RescopeGlobalSymbols` isn't designed for a NodeJS environment. For it to work correctly, the global object needs to be `global` instead of `window`.

That's still not going to fix `require` however, because it's not really global. It's provided as an argument to an IIFE wrapper that node adds around all CommonJS module sources. We don't have a good way to model that in the compiler right now.

thheller:
> We don't have a good way to model that in the compiler right now.

I don't understand that comment. The way it is currently modeled in the compiler is via the `SPECIAL_EXTERNS` constant I initially linked. The PR I created just makes that configurable. As you mentioned `require` is not a global and the `SPECIAL_EXTERNS` just exempts anything in that Set from getting rescoped at all, which is exactly what we want for `require`.

[...]

PR #3154, opened 2018-11-24
Add configuration options for RescopeGlobalSymbols

thheller:
Making it usable in a node.js environment that does not use "window". The original behavior is preserved and can now be overridden selectively.

I struggled coming up with reasonable names for the new configuration options. It might also be useful to adjust some other option names.

### DIFF
--- a/src/com/google/javascript/jscomp/CompilerOptions.java
+++ b/src/com/google/javascript/jscomp/CompilerOptions.java
@@ -47,6 +47,7 @@ import java.io.Serializable;
 import java.nio.charset.Charset;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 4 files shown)


thheller:
I would much prefer to have this completely open from the Java API directly, I may need to add custom globals due to the way I'm packaging the code in shadow-cljs.

I have never used the CLI version but using `--env` as a preset with the "default" globals could be done on top of the changes I already added. Happy to add it to `CommandLineRunner/createOptions`. Any thoughts on the names?

thheller:
Oops, clicked the wrong button.

ChadKillingsworth:
You should be able to add any custom global by just defining an extern and passing it to the compiler. You are currently trying to solve 2 problems:

 1. NodeJS specific externs
 2. RescopeGlobalSymbols using `window` as the global object.

If RescopeGlobalSymbols is switched to use `global` rather than `window` when `--env=NODE` is set, then any global extern will also be defined as a property: `global.Buffer = Buffer;`.

The `--env` flag is also settable from the Java API.

thheller:
I'm specifically trying to prevent rescoping to a global in my case, so externs would not work. I'm also using this in a `webpack` environment that will eventually run in the browser but requires support for `module.exports` and `exports` which I'm currently hacking by regexing the code after compilation and replacing the rescope.

https://github.com/thheller/shadow-cljs/blob/1c9cc1500afd12132e11d8bb2f66336d0552d78c/src/main/shadow/build/closure.clj#L863-L880

thheller:
I was not concerned about import/require, the point is that those don't get touched at all.

I'm specifically trying to produce code that is consumable by other JS tools after it has been optimized by Closure without actually modifying those JS tools (or node). I'm going for Compatibility here, not the best possible result. This works fine already. I just don't like regexing the code after compilation to undo something `RescopeGlobalSymbols` did it shouldn't have.

EatingW:
Internal Google issue b/120032161 created.

thheller:
Solved via https://github.com/google/closure-compiler/issues/3146

PR #3158, opened 2018-12-01
Update CanvasImageSource typedef

dsanders11:
Pulling this back closer to the [current version of the spec](https://html.spec.whatwg.org/multipage/canvas.html#canvasimagesource).


dsanders11:
Build failure is due to `OffscreenCanvas` not being defined, which I've created #3160 for. There's also an error that `SVGImageElement` is not defined, but it does exist in `contrib/externs/svg.js`. Should this extern not reference it? I can drop the reference to `SVGImageElement`, but it is part of the spec.

ChadKillingsworth:
Re-triggering Travis

PR #3165, opened 2018-12-06
Add support for webpack4

ChadKillingsworth:
Recognize calls to __webpack_require__.t as namespace imports.
Recognize the Promise.all calls for the webpack4 require.ensure loading

### DIFF
--- a/src/com/google/javascript/jscomp/ProcessCommonJSModules.java
+++ b/src/com/google/javascript/jscomp/ProcessCommonJSModules.java
@@ -46,6 +46,9 @@ public final class ProcessCommonJSModules extends NodeTraversal.AbstractPreOrder
   private static final String MODULE = "module";
   private static final String REQUIRE = "require";
   private static final String WEBPACK_REQUIRE = "__webpack_require__";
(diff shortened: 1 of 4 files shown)


EatingW:
Internal Google issue b/120606843 created.

src/com/google/javascript/jscomp/ProcessCommonJSModules.java
```diff
@@ -287,7 +304,11 @@ public static boolean isCommonJsDynamicImportCallback(
    * Recognizes __webpack_require__ calls that are the .then callback of a __webpack_require__.e
    * call. Example:
    *
-   * <p>__webpack_require__.e(0).then(function() { return __webpack_require__(4); })
+   * <ul>
+   *   <li>__webpack_require__.e(0).then(function() { return __webpack_require__(4); })
+   *   <li>Promise.all([__webpack_require__.e(0)]).then(function() { return
```
blickly: Is this new format just the new way that webpack4 translates
```
import("module").then(callback);
```
or are we supporting a new source pattern here?

More generally, do you have an idea of what our policy should be on doing more dynamic things like saving the promise to a variable or using import outside the global scope?

PR #3166, opened 2018-12-06
Fix inclusion of input source maps sourcesContent in JSModule compilations

WearyMonkey:
It looks like `sourceFileContentMap` was incorrectly being cleared. `sourceFileContentMap` is populated during the compile initialisation, not when the sourceMaps are generated. So clearing it was preventing `sourcesContent` from being included in the source map.

Our pattern for jsmodule compilations is:

```
compiler.compileModules(EXTERNS, modules, options); // sourceFileContentMap populated here
for (JSModule module : modules) {
   compiler.getSourceMap().reset(); // sourceFileContentMap dropped here
   String source = compiler.toSource(module);
   SourceMap = compiler.getSourceMap();
   writeOutput(source, sourceMap);
}
```

### DIFF
--- a/src/com/google/debugging/sourcemap/SourceMapGeneratorV3.java
+++ b/src/com/google/debugging/sourcemap/SourceMapGeneratorV3.java
@@ -137,7 +137,6 @@ public final class SourceMapGeneratorV3 implements SourceMapGenerator {
     mappings.clear();
     lastMapping = null;
     sourceFileMap.clear();
-    sourceFileContentMap.clear();
     originalNameMap.clear();
     lastSourceFile = null;
     lastSourceFileIndex = -1;
--- a/test/com/google/debugging/sourcemap/SourceMapGeneratorV3Test.java
+++ b/test/com/google/debugging/sourcemap/SourceMapGeneratorV3Test.java
@@ -17,14 +17,17 @@
 package com.google.debugging.sourcemap;
 
 import static com.google.common.truth.Truth.assertThat;
+import static com.google.common.truth.Truth.assertWithMessage;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import com.google.debugging.sourcemap.SourceMapGeneratorV3.ExtensionMergeAction;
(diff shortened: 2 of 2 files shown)


EatingW:
Internal Google issue b/120608357 created.

src/com/google/debugging/sourcemap/SourceMapGeneratorV3.java
```diff
@@ -137,7 +137,6 @@ public void reset() {
     mappings.clear();
     lastMapping = null;
     sourceFileMap.clear();
-    sourceFileContentMap.clear();
```
shicks: This doesn't seem right.  It's very surprising that it would reset everything *except* this map.

I don't quite understand your assertion about being reset and added at the wrong time.  Once the source map is reset, all the mappings need to be re-added.  Would it be possible to simply ensure that, along with adding mappings, we also add the sources contents?

ChadKillingsworth:
Re-triggering Travis

src/com/google/debugging/sourcemap/SourceMapGeneratorV3.java
shicks: Can the life cycle be changed to be more consistent?  Can we initialize `sourceFileContentMap` during source map generation?  Or pass the content map in as a separate argument later?  I'm just not comfortable having a `reset` method that doesn't actually reset everything.

ChadKillingsworth: I just ran into this as well and need it fixed.

@WearyMonkey If you aren't comfortable making this change, I have an idea of how to approach it. It'll need handled in the GWT runner too.

ChadKillingsworth:
Not hearing anything on this PR, I created #3166 as an alternative to address @shicks concerns.

WearyMonkey:
Thanks @ChadKillingsworth , I'm travelling at the moment so can't follow up on this.

shicks:
Let's go with #3172.  Thanks for bringing this issue to our attention!

PR #3172, opened 2018-12-13
When resetting source maps, re-add input source maps.

ChadKillingsworth:
When code splitting, the source map is reset between generation of every output file. However, input source maps were not re-applied. So the original source content from an input source map was not copied to the output source map.

Adds a helper method to the compiler to simply reapply those input source maps anytime the source map needs reset.

Alternative to #3166

### DIFF
--- a/test/com/google/javascript/jscomp/CommandLineRunnerTest.java
+++ b/test/com/google/javascript/jscomp/CommandLineRunnerTest.java
@@ -2460,6 +2460,33 @@ public final class CommandLineRunnerTest {
     testSame("alert('hello world')");
   }
 
+  @Test
+  public void testOutputSourceMapContainsSourcesContent() {
+    String inputString = "[{\"src\": \"\", \"path\": \"base.js\"},"
+        + "{\"src\": \"alert('foo');\", \"path\":\"foo.js\"}]";
+    args.add("--json_streams=BOTH");
+    args.add("--chunk=m1:1");
+    args.add("--chunk=m2:1:m1");
+    args.add("--source_map_include_content");
+
+    CommandLineRunner runner =
+        new CommandLineRunner(
+            args.toArray(new String[] {}),
+            new ByteArrayInputStream(inputString.getBytes(UTF_8)),
+            new PrintStream(outReader),
+            new PrintStream(errReader));
+
+    lastCompiler = runner.getCompiler();
+    try {
+      runner.doRun();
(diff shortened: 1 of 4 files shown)


blickly:
@shicks, could you review this one?

brad4d:
Internal Google issue b/121273647 created and assigned to @shicks

brad4d:
FYI @shicks is OOO until Dec. 27

brad4d:
I've imported this and sent it for internal review.
However, I believe we should have some tests before we actually submit it.

@ChadKillingsworth could you add a test or two to confirm that this is doing what you intended it to do?
Thanks.

ChadKillingsworth:
Adding a test for this has turned out to be a bit tricky and I found another bug. Whomever did the original work for including source map content didn't take into account code splitting.

ChadKillingsworth:
Alright - failing test added. For code splitting, all of the compilation sources have to be added to the source map as well.

test/com/google/javascript/jscomp/CommandLineRunnerTest.java
```diff
@@ -2460,6 +2460,33 @@ public void testWarningGuardQuotedValue() {
...
+
+    lastCompiler = runner.getCompiler();
+    try {
+      runner.doRun();
+    } catch (IOException e) {
+      e.printStackTrace();
+      assertWithMessage("Unexpected exception " + e).fail();
+    }
+    String output = new String(outReader.toByteArray(), UTF_8);
+    assertThat(output).contains("\\\"sourcesContent\\\":");
```
brad4d: Don't we care what value is associated with "sourcesContent"?
Would it be reasonable to assert on the exact content of the output?
As a reader not intimately familiar with what the output *should* look like,
it would help a lot for the test to tell me what it should look like.

I realize it would be more prone to "spurious" breakages, but those would be pretty easy to fix when they happen.

ChadKillingsworth: I was kind of assuming that source maps in general were tested elsewhere. The issue was that with code splitting, only the first emitted chunk had any sources content. I just really wanted to ensure it was present.

That being said though, I can add the exact snippet there.

ChadKillingsworth: Added.

brad4d:
imported and sent for internal testing and review

PR #3183, opened 2018-12-31
Disable collapse properties for module exports in simple mode

### LINKED ISSUE #3182: Collapsing Module Exports in SIMPLE mode Causes Undefined References
A year ago, I submitted #2740 and thought it to be safe.

However, I now have found a code pattern which seems to violate an assumption and breaks compilations in SIMPLE mode:

```js
Promise.all([import('/path1'), import('path2')])
  .then(([module1, module2]) => {
    console.log(module1.default); // Property collapsing breaks this reference
    console.log(module2.default); // Property collapsing breaks this reference
  });
```

With webpack, the code gets transpiled and rewritten to be:

```js
Promise
  .all([
    __webpack_require__.e(outputChunk).then(() => module$path1),
    __webpack_require__.e(outputChunk).then(() => module$path2)
  ])
  .then(([module1, module2]) => {
    console.log(module1.default); // Property collapsing breaks this reference
    console.log(module2.default); // Property collapsing breaks this reference
  });
```

Somehow this pattern prevents the CollapseProperties pass from recognizing the alias and backing off.

ChadKillingsworth:
With dynamic import statements, this optimization is not safe. Property collapsing will leave the following with an undefined reference.

```js
import('./path/to/module.js').then(moduleNamespace => console.log(moduleNamespace.default));
```

Fixes #3182.

### DIFF
--- a/src/com/google/javascript/jscomp/CompilationLevel.java
+++ b/src/com/google/javascript/jscomp/CompilationLevel.java
@@ -241,7 +241,7 @@ public enum CompilationLevel {
       case SIMPLE_OPTIMIZATIONS:
         // Enable global variable optimizations (but not property optimizations)
         options.setVariableRenaming(VariableRenamingPolicy.ALL);
-        options.setCollapsePropertiesLevel(PropertyCollapseLevel.MODULE_EXPORT);
+        options.setCollapsePropertiesLevel(PropertyCollapseLevel.NONE);
         options.setCollapseAnonymousFunctions(true);
         options.setInlineConstantVars(true);
         options.setInlineFunctions(Reach.ALL);


brad4d:
Internal issue http://b/122264392 created.

ChadKillingsworth:
FYI I would love to have any alternative fix for this. It is a pretty significant loss to dead code elimination in simple mode.

concavelenz:
Can we simply not collapse dynamically imported modules?  Is this through Webpack?

brad4d:
Since the change is really simple, I've created an internal version and am running it through testing in case we decide to do it.

brad4d:
FWIW the internal testing didn't find any failures caused by this PR.

jplaisted:
We don't support `import()`, right? Even in the parser, I get a syntax error...

brad4d:
@ChadKillingsworth 
The open questions on this PR as I understand them are:

1. Could we simply not collapse the dynamically imported modules?
2. Is this only an issue for Webpack?
3. Wait, do we actually support dynamic imports yet? Is this intended to remove a blocker to supporting them?

PR #3184, opened 2019-01-04
Add property reflection for Polymer computed properties and observers

ChadKillingsworth:
Polymer observers and computed properties use string reflection. When property renaming is enabled, add property reflection calls to these strings such that they are renamed consistently.

Example:

```js
class FooElement extends Polymer.Element {
  static get properties() {
    return {
      foo: {
        type: String,
        compute: '_computeFoo(bar)'
      },
      bar: {
        type: String,
        value: 'bar'
      }
    };
  }
  _computeFoo(bar) {
    return 'foo' + bar;
   }
}
```

The `'_computeFoo(bar)'` string will now be parsed and reflection calls added like:

```js
JSCompiler_renameProperty('_computeFoo', /** @type {!FooElement} */ ({})) + '(' +
    JSCompiler_renameProperty('bar', /** @type {!FooElement} */ ({})) + ')'
```

### DIFF
--- a/src/com/google/javascript/jscomp/PolymerClassRewriter.java
+++ b/src/com/google/javascript/jscomp/PolymerClassRewriter.java
@@ -288,7 +288,8 @@ final class PolymerClassRewriter {
         sinkFunction.putBooleanProp(Node.FREE_CALL, true);
 
         Node stmt = NodeUtil.getEnclosingStatement(clazz);
-        stmt.getParent().addChildToBack(IR.exprResult(sinkFunction).useSourceInfoFromForTree(stmt));
+        stmt.getParent().addChildAfter(
+            IR.exprResult(sinkFunction).useSourceInfoFromForTree(stmt), stmt);
         compiler.reportChangeToChangeScope(sinkFunction.getFirstChild());
         compiler.reportChangeToEnclosingScope(stmt);
       }
--- a/src/com/google/javascript/jscomp/PolymerClassRewriter.java
+++ b/src/com/google/javascript/jscomp/PolymerClassRewriter.java
@@ -233,11 +233,14 @@ final class PolymerClassRewriter {
       jsDocInfoNode.setJSDocInfo(classInfo.build());
     }
 
+    Node insertAfterReference = NodeUtil.getEnclosingStatement(clazz);
     if (block.hasChildren()) {
       removePropertyDocs(cls.descriptor, cls.defType);
-      Node stmt = NodeUtil.getEnclosingStatement(clazz);
-      stmt.getParent().addChildrenAfter(block.removeChildren(), stmt);
-      compiler.reportChangeToEnclosingScope(stmt);
+      Node newInsertAfterReference = block.getLastChild();
+      insertAfterReference.getParent()
+          .addChildrenAfter(block.removeChildren(), insertAfterReference);
+      compiler.reportChangeToEnclosingScope(insertAfterReference);
+      insertAfterReference = newInsertAfterReference;
     }
 
     addReturnTypeIfMissing(cls, "is", new JSTypeExpression(IR.string("string"), VIRTUAL_FILE));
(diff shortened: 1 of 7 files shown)


ChadKillingsworth:
cc @aomarks

lauraharker:
Internal issue http://b/122359057 created

test/com/google/javascript/jscomp/PolymerPassTest.java
```diff
@@ -3681,6 +3689,498 @@ public void testExportsUniqueMethodsFromLegacyElementAndBehaviors() {
...
+            "    return $jscomp.reflectObject(XElement, {",
+            "      user_: Object,",
+            "      pets: Array,",
+            "      name: {",
+            "        type: String,",
+            "        computed: $jscomp.reflectProperty(",
+            "            '_computeName', /** @type {!XElement} */ ({})) + '(' +",
+            "            '\"user\"' + ',' + '12.0' + ',' + ",
+            "            $jscomp.reflectProperty('user_', /** @type {!XElement} */ ({})) + '.' + ",
+            "            $jscomp.reflectProperty('id', /** @type {!XElement} */ ({}).user_) + ',' + ",
```
aomarks: Just confirming that the compiler is able to understand the getprops in second parameter here? (i.e. that the result should correspond to `XElement.user_.id`?)

ChadKillingsworth:
I'm on vacation this week. I'll address the feedback next week.

test/com/google/javascript/jscomp/PolymerPassTest.java
ChadKillingsworth: Yes this works. It's not safe in general to do this - but since the reference is only used for type information and will not exist in runtime code, it will be safe here.

ChadKillingsworth:
@aomarks I've made the changes you requested

lauraharker:
Importing for internal review

ChadKillingsworth:
@lauraharker @aomarks I have been unable to recreate the remaining issue. I've re-arranged the code a bit on a hunch and also updated the source references so that the warning should point to what exactly is triggering the warning.

If you can try this again with the latest changes, that will at least help me narrow down what's causing the type warning. It's a bit odd because I'm not changing any type signatures with this code.

lauraharker:
@ChadKillingsworth here's the new warning I get:

```
polymer/v2/polymer/lib/elements/array-selector.js:51: ERROR - variable ArraySelectorMixin.prototype.items redefined with type (Array|null), original definition at polymer/v2/polymer/lib/elements/array-selector.js:59 with type Array
  class ArraySelectorMixin extends elementBase {
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Here's `array-selector.js` immediately after the PolymerPass with --pretty_print and --preserve_type_annotations:
[array-selector-after-polymer.txt](https://github.com/google/closure-compiler/files/2810432/array-selector-after-polymer.txt)

ChadKillingsworth:
Thanks - that helps.

Here's my best guess: the compiler is treating the `Element.prototype.propName;` statements being used for a property sink as a nullability redefinition. That's surprising to me, but at least plausible.

I've refactored the code to add the synthetic sink statements directly rather than relying on the CheckSideEffects pass to add them. Hopefully this will solve the issue.

lauraharker:
That fixed it!

I'm running another set of internal tests now.

lauraharker:
Or no, that was a false positive.

I now get a different error:
```
{SyntheticVarsDeclar}: ERROR - Variable JSCOMPILER_PRESERVE declared more than once. First occurrence: {SyntheticVarsDeclar}
```

src/com/google/javascript/jscomp/PolymerClassRewriter.java
```diff
@@ -254,8 +258,36 @@ void rewritePolymerClassDeclaration(
...
-      addObjectReflectionCall(cls, cls.descriptor);
+      convertSimpleObserverStringsToReferences(cls);
+      List<Node> propertySinks = new ArrayList<>();
+      if (polymerExportPolicy != PolymerExportPolicy.EXPORT_ALL) {
+        propertySinks.addAll(addComputedPropertiesReflectionCalls(cls));
+        propertySinks.addAll(addComplexObserverReflectionCalls(cls));
+      }
+
+      if (propertySinks.size() > 0) {
+        if (traversal.getScope().getVar(CheckSideEffects.PROTECTOR_FN) == null) {
```
lauraharker: This check might be what's causing the crash. NodeTraversal doesn't automatically regenerate parent scopes when they change, so if this is called twice while in the same traversal, it will add the extern both times.

ChadKillingsworth: Oh interesting - I had no idea.

I pushed a property check up to keep track and hopefully ensure it is only injected once.

lauraharker:
The latest commit fixes the error I was seeing. :+1: 
I'm doing another round of global testing now.

Issue #3198, opened 2019-01-16
Incorrect Destructuring Optimisation When Object is Cloned With { ...object } Down The Line


zavr-1:
```js
const Detect = () => {
  const detected = detect()
  // THIS BELOW causes destructuring to be .map((a)=>({from:a}=a))
  const f = detected.map((ff) => {
    const newF = { ...ff }
    return newF
  })
  return f 
  // if ABOVE commented out, it will work.
  // return detected 
}

const detect = () => {
  const matches = getMatches() // supposed to be array of strings
  const [m] = matches
  if (m !== 'number') throw new Error(`m is ${JSON.stringify(m)}`)
  return matches
}

const getMatches = () => {
  const r = [
    { test: Infinity, from: 'number' },
    { test: 'string', from: 'string' },
    { test: false, from: 'boolean' },
  ]
  const res = r.map(({ from }) => from)
  return res
}

console.log(Detect())
```
With the cloning commented out, it is possible to obtain the result:
```sh
[ 'number', 'string', 'boolean' ]
```

```sh
'use strict';const d=()=>{const a=c(),[b]=a;if("number"!==b)throw Error(`m is ${JSON.stringify(b)}`);return a},c=()=>[{test:Infinity,from:"number"},{test:"string",from:"string"},{test:!1,from:"boolean"}].map((a)=>({from:a}=a));console.log((()=>{const a=d();a.map((a)=>Object.assign({},a));return a})());
                                                            ^

Error: m is {"test":null,"from":"number"}
```
```
-jar /Users/zavr/node_modules/google-closure-compiler-java/compiler.jar --compilation_level ADVANCED--language_in ECMASCRIPT_2018 --language_out ECMASCRIPT_2017
  --js t/mismatch.js
```

zavr-1:
This is also true for `.filter` and probably other functional programming calls on arrays:
```js
const items = [
  { internal: true },
  { entry: 'hello' },
]

const fi = items.filter(({ entry }) => entry)
// fi.map((a) => {
//   return { ...a, mapped: 1 }
// })
console.log(fi)
```

```sh
-jar /Users/zavr/node_modules/google-closure-compiler-java/compiler.jar --language_in ECMASCRIPT_2018 --language_out ECMASCRIPT_2017
```
** CORRECT **
```js
'use strict';const items=[{internal:!0},{entry:"hello"}],fi=items.filter(({entry:a})=>a);console.log(fi);
```
```
[ { entry: 'hello' } ]
```
[...]

PR #3200, opened 2019-01-18
Prevent assign collapse with destructuring

zavr-1:
Addresses #3198


lauraharker:
Created internal issue http://b/123083204

Thanks for looking into this! Let us know when you've signed the CLA.

Issue #2770, opened 2017-12-22
[FEATURE] Support dynamic import


dtruffaut:
Support dynamic import :

https://developers.google.com/web/updates/2017/11/dynamic-import

module.js :
```
export const foo = _ => {
  console.log('foo');
}
```

main.js :
```
const main = async _ => {

  const m = await import('module.js');
  m.foo();

  (await import('module.js')).foo();

}

main();
```

Should display :
```
foo
foo
```

The thing is about supporting 'export' and 'import' keywords.

blickly:
Thanks for filing the tracking bug.

Just want to confirm that the feature we're tracking here should be the [currently stage 3 dynamic import syntax](https://github.com/tc39/proposal-dynamic-import), not [top-level await syntax](https://gist.github.com/Rich-Harris/0b6f317657f5167663b493c722647221).

dtruffaut:
Yes !

ChadKillingsworth:
@blickly There is already both Chrome and Safari support for this feature. The hard part is figuring out exactly what the compiler should do here. But parsing would be a good first step...

dtruffaut:
Any news on this ?
At least, just allow the parsing ?

MaxMilton:
Would be amazing to see parser support for `import()`.

Spent the morning trying to add Closure Compiler as the minifier in <https://github.com/MaxMilton/sapper-template-rollup/tree/feat/use-closure-compiler> but in the end it turns out the import() can't be parsed and the compiler bails completely. In this project the import function is polyfilled so we don't need the compiler to do anything special. This is now a blocker for adding Closure Compiler support :sweat:

lauraharker:
Created Google internal issue http://b/122961838 for this request.

Tonsil:
Has there been any movement on this recently? Asking because we're getting into a project for which it would be extremely useful. I figured I'd ask before we decide to fork and fix it ourselves.

dtruffaut:
Most of JS developers now use Rollup instead of Closure Compiler.

Rollup does a great job in handling await import instructions and moving all your imports at the same level on the disk. 

As a result, code splitting leads to faster load times.

File size matters, but code splitting really is on another level. The benefits are impressive.

In addition, Rollup understands ESNext and respects property names, which leads to fewer bugs.

As of July 2019, I would recommend Rollup over Closure Compiler.

MaxMilton:
[...]

PR #3204, opened 2019-01-21
Add initial support for dynamic module import

ChadKillingsworth:
Adds basic parsing support and code generation for the dynamic import call.  See https://tc39.github.io/proposal-dynamic-import/

The specification severely limits usage of `import(specifier)` type calls. Aliasing, `.call`, `.apply`, `.bind` and many other JS function semantics are not supported.

At the moment, the Refastor tests are failing, but I assume that is because certain passes are being skipped now. I'll need some help in figuring out how to work around that.

First step for full support of #2770


brad4d:
@ChadKillingsworth 
I've been looking through the PR and thinking about what we need here.

I believe the code you're proposing is modeling `import(module)` as a CALL node with an IMPORT node as the first child (callee) and the single module name argument as its second child.

Although it looks like a CALL, it is different in many significant ways that I think will be likely to trip up existing compiler code.

Could you change this implementation to instead define a new `TokenType` for `Node`, possibly `DYNAMIC_IMPORT` or `IMPORT_CALL`? This will ensure that existing code doesn't trip on it and will avoid some of the special casing you had to add in your current version to deal with an IMPORT that has no children.

Thanks.

ChadKillingsworth:
@brad4d Switched to using a new node type `DYNAMIC_IMPORT` with a single argument.

Also verified in Chrome that only a single argument is expected (not an argument list).
`import('foo', 'bar')` doesn't parse.

brad4d:
found the source of at least some of the test failures.
We need to put the new feature in `ES_UNSUPPORTED` (which didn't exist when this PR was created) instead of `ES_NEXT`.

I've done that in my internal version.
I don't plan to push my changes back to this PR's branch unless I feel I need more input from Chad, though.

brad4d:
had another small internal breakage, which I've fixed today
sent to an additional internal reviewer to get approval for that fix

brad4d:
I'm submitting this change internally now.
If nothing fails, then it will appear in GitHub tomorrow.

Issue #3206, opened 2019-01-23
SourceFileTest.testSourceFileResolvesZipEntries failes for Windows


lgemeinhardt:
SourceFileTest.testSourceFileResolvesZipEntries failes for Windows

java.nio.file.NoSuchFileException: C:\Users\LGEMEI~1\AppData\Local\Temp\test4171789207027663828.js.zip!\foo.js
	at sun.nio.fs.WindowsException.translateToIOException(WindowsException.java:79)
	at sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:97)
	at sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:102)
	at sun.nio.fs.WindowsFileSystemProvider.newByteChannel(WindowsFileSystemProvider.java:230)
	at java.nio.file.Files.newByteChannel(Files.java:361)
	at java.nio.file.Files.newByteChannel(Files.java:407)
	at java.nio.file.spi.FileSystemProvider.newInputStream(FileSystemProvider.java:384)
	at java.nio.file.Files.newInputStream(Files.java:152)
	at java.nio.file.Files.newBufferedReader(Files.java:2784)
	at com.google.javascript.jscomp.SourceFile$OnDisk.getCodeReader(SourceFile.java:624)
	at com.google.javascript.jscomp.SourceFile$OnDisk.getCode(SourceFile.java:602)
	at com.google.javascript.jscomp.SourceFileTest.testSourceFileResolvesZipEntries(SourceFileTest.java:148)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
[...]

PR #3207, opened 2019-01-23
SourceFileTest.testSourceFileResolvesZipEntries failes for Windows #3206

lgemeinhardt:
And fixes parts of "Support the workflow under Windows. #9" (https://github.com/google/j2cl/issues/9) from j2cl

### DIFF
--- a/src/com/google/javascript/jscomp/SourceFile.java
+++ b/src/com/google/javascript/jscomp/SourceFile.java
@@ -46,6 +46,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Enumeration;
 import java.util.List;
+import java.util.regex.Pattern;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipFile;
 
@@ -358,13 +359,13 @@ public class SourceFile implements StaticSourceFile, Serializable {
   private static final String JAR_URL_PREFIX = "jar:file:";
 
   private static boolean isZipEntry(String path) {
-    return path.contains(".zip!/") && (path.endsWith(".js") || path.endsWith(".js.map"));
+    return path.contains(".zip!" + File.separator) && (path.endsWith(".js") || path.endsWith(".js.map"));
   }
 
   @GwtIncompatible("java.io.File")
   private static SourceFile fromZipEntry(String zipURL, Charset inputCharset) {
     checkArgument(isZipEntry(zipURL));
-    String[] components = zipURL.split(BANG_SLASH);
+    String[] components = zipURL.split(Pattern.quote(BANG_SLASH.replace("/",  File.separator)));
     try {
       String zipPath = components[0];
       String relativePath = components[1];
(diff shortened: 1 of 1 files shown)


src/com/google/javascript/jscomp/SourceFile.java
```diff
@@ -359,13 +359,13 @@ public String toString() {
...
   private static boolean isZipEntry(String path) {
-    return path.contains(".zip!" + File.separatorChar) && (path.endsWith(".js") || path.endsWith(".js.map"));
+    return path.contains(".zip!" + File.separator) && (path.endsWith(".js") || path.endsWith(".js.map"));
   }
 
   @GwtIncompatible("java.io.File")
   private static SourceFile fromZipEntry(String zipURL, Charset inputCharset) {
     checkArgument(isZipEntry(zipURL));
-    String[] components = zipURL.split(Pattern.quote(BANG_SLASH.replace('/',  File.separatorChar)));
+    String[] components = zipURL.split(Pattern.quote(BANG_SLASH.replace("/",  File.separator)));
```
tjgq: Instead of doing a replace, can you fix the `BANG_SLASH` definition to

`private static final String BANG_SLASH = "!" + File.separatorChar;`?

src/com/google/javascript/jscomp/SourceFile.java
```diff
@@ -378,7 +379,7 @@ private static SourceFile fromZipEntry(String zipURL, Charset inputCharset) {
   public static SourceFile fromZipEntry(
       String originalZipPath, String absoluteZipPath, String entryPath, Charset inputCharset)
       throws MalformedURLException {
-    String zipEntryPath = JAR_URL_PREFIX + absoluteZipPath + BANG_SLASH + entryPath;
+    String zipEntryPath = JAR_URL_PREFIX + absoluteZipPath + BANG_SLASH + entryPath.replace(File.pathSeparatorChar,  '/');
```
tjgq: Shouldn't this be `File.separatorChar` instead of `File.pathSeparatorChar`?

My understanding is that `separatorChar` is the path component separator (`/` on Unix, `\` on Windows), while `pathSeparatorChar` is the `PATH` environment variable separator (`:` on Unix, `;` on Windows).

tjgq:
Created Google internal issue b/123291586

src/com/google/javascript/jscomp/SourceFile.java
lgemeinhardt: This was also my first idea, but BANG_SLASH is used 3 times, but only one is windows (path) dependend. Only using another constant (like BANG_SLASH_FS_SPECIAL) would be a way to go, but I think this more confusing.

src/com/google/javascript/jscomp/SourceFile.java
```diff
@@ -359,13 +359,13 @@ public String toString() {
   private static final String JAR_URL_PREFIX = "jar:file:";
 
   private static boolean isZipEntry(String path) {
-    return path.contains(".zip!" + File.separatorChar) && (path.endsWith(".js") || path.endsWith(".js.map"));
```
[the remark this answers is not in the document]
lgemeinhardt: It simple does not exists in the emulation / super source.

src/com/google/javascript/jscomp/SourceFile.java
lgemeinhardt: This was fixed with the second commit.

lgemeinhardt:
> Once you address the review comments, can you please squash the two commits into a single one?

Squash is done!

PR #3208, opened 2019-01-23
Fix the return type for message handler to correctly return void

realityforge:


realityforge:
This change was ultimately motivated by generating correct [Elemental2](https://github.com/google/elemental2) bindings for the property. It was originally submitted as https://github.com/google/elemental2/pull/77

externs/browser/html5.js
```diff
@@ -2981,7 +2981,7 @@ WebSocket.prototype.onopen;
 /**
  * An event handler called on message event.
  * TODO(blickly): Change this to MessageEvent<*> and add casts as needed
- * @type {?function(!MessageEvent<?>)}
+ * @type {?(function(!MessageEvent<?>): void)}
```
blickly: I think that this extra set of parenthesis is unnecessary.  Otherwise LGTM

tjgq:
Created Google internal issue b/123290434

externs/browser/html5.js
realityforge: It is possible that it is unnecessary for closure compiler .... however when it was passed through https://github.com/google/jsinterop-generator the `?` would bind to `function(!MessageEvent<?>)`and would produce incorrect output. 🤷‍♂️ I am not confident enough to track down where the issue lies but this seems to work

blickly: Ah, OK. Got it.

In that case, can you file an issue there and put a comment here to mention that, like:
```javascript
/**
  * Once https://github.com/google/jsinterop-generator/issues/XYZ is fixed, we can remove the 
  * extra parenthesis here.
  */
```

Mostly I want to avoid the situation where someone inadvertently remove the parens in a "refactoring" and break something unintentionally.

realityforge: So it turns out removing the brackets works fine so I have re-pushed the change without them.

FWIW: It turns out that the problem is somehow related to incremental Bazel builds in https://github.com/google/elemental2 that is the issue. Next rabbit hole to wander down ;)

realityforge:
@tjgq friendly ping ;)

tjgq:
Sorry, this fell through the cracks. Importing for internal review.

realityforge:
No problem - thanks!

realityforge:
Thanks!

PR #3209, opened 2019-01-23
Correct the return type for the AbortController.abort() method

realityforge:


realityforge:
This change was ultimately motivated by generating correct elemental2 bindings for the method. It was originally submitted as google/elemental2#76

tjgq:
Created Google internal issue b/123291771

PR #3210, opened 2019-01-24
Add missing extern, chrome.bluetoothPrivate.connect

dtsengchromium:
The chrome.bluetoothPrivate api has diverged from this extern definition (committed in 2014).

This proposes its removal for the following reasons:
1. the extern is wrong and incomplete
2. chrome.bluetoothPrivate is indeed private
3. extensions wishing to use it should be part of chrome's main repopsitory and utilize the externs generated there which are kept up-to-date

### DIFF
--- a/contrib/externs/chrome_extensions.js
+++ b/contrib/externs/chrome_extensions.js
@@ -10518,6 +10518,16 @@ chrome.bluetoothPrivate.TransportType = {
 
 
 /**
+ * Connects to the given device. This will only throw an error if the device
+ * address is invalid or the device is already connected. Otherwise this will
+ * succeed and invoke |callback| with ConnectResultType.
+ * @param {string} deviceAddress
+ * @param {function(!chrome.bluetoothPrivate.ConnectResultType):void=} callback
+ */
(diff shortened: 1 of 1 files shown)


tjgq:
Internal Google issue b/123406744 created.

tjgq:
For reference, the corresponding Chromium externs are at https://cs.chromium.org/chromium/src/third_party/closure_compiler/externs/bluetooth_private.js.

tjgq:
The PR was merged, but had to be rolled back because some internal projects depend on those externs.

dtsengchromium:
Apologies for the breakage. Updated my commit to update one missing extern, |chrome.bluetoothPrivate.connect|. PTAL.

dtsengchromium:
Friendly ping on this. To be clear, I now only add the missing extern instead of trying to remove chrome.bluetoothPrivate* entirely for projects that already depend on it. Since this is only additive, it shouldn't break anyone, but please let me know and I'll resolve those breaks if any.

tjgq:
It has been submitted internally. It should be pushed out today.

PR #3214, opened 2019-01-25
Prevent module collapsing of dynamically imported modules in SIMPLE mode

### LINKED ISSUE #3182: Collapsing Module Exports in SIMPLE mode Causes Undefined References

ChadKillingsworth:
With dynamic module imports, module export properties should not be collapsed in SIMPLE mode as all references to the property are not directly trackable and other properties are not collapsed. Adds a map of modules which were dynamically imported and prevents collapsing any property on the module namespace.

Closes #3182

### DIFF
--- a/src/com/google/javascript/jscomp/CollapseProperties.java
+++ b/src/com/google/javascript/jscomp/CollapseProperties.java
@@ -32,6 +32,7 @@ import com.google.javascript.rhino.Node;
 import com.google.javascript.rhino.Token;
 import com.google.javascript.rhino.TokenStream;
 import com.google.javascript.rhino.jstype.JSType;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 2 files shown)


tjgq:
Created Google internal issue b/123409370

PR #2902, opened 2018-04-22
Fix Array.prototype.copyWithin polyfill

teppeis:
Improvement:
- should handle negative params
- should coerce params to integer
- fix bug in deleting props of ArrayLike
- should throw if `this` is null or undefined
- should throw if impossible to delete props

Also this fixes `.name` and `.length` with `defineProperty`.
But if the policy of Closure Compiler for them is "won't fix", I will revert them.

I don't know how to run polyfill tests (https://github.com/google/closure-compiler/issues/1504), but I confirmed this passed [all Test262 tests for copyWithin](https://github.com/tc39/test262/tree/c55d2ab7c344e35a3ceb93cf1d4d30019584db82/test/built-ins/Array/prototype/copyWithin). (https://gist.github.com/teppeis/c8a88f653cdd70ed3e3d6597ec1ee30c)

Ref:
- https://www.ecma-international.org/ecma-262/8.0/#sec-array.prototype.copywithin
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/copyWithin

PR #3216, opened 2019-01-26
Fix ES6 polyfills to use ToInteger and ToLength

teppeis:
from https://github.com/google/closure-compiler/pull/2902#discussion_r

This change improves compatibility with ECMAScript spec.

Use [`ToInteger`](https://www.ecma-international.org/ecma-262/9.0/#sec-tointeger) to support coercing, null, undefined, float, NaN and larger-than-Int32.
Also use [`ToLength`](https://www.ecma-international.org/ecma-262/9.0/#sec-tolength) to handle negative values and larger-than-2^53.

### DIFF
--- a/src/com/google/javascript/jscomp/js/es6/array/fill.js
+++ b/src/com/google/javascript/jscomp/js/es6/array/fill.js
@@ -15,6 +15,8 @@
  */
 
 'require util/polyfill';
+'require util/tointeger';
+'require util/tolength';
 
 $jscomp.polyfill('Array.prototype.fill', function(orig) {
   if (orig) return orig;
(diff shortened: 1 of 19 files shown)


test/com/google/javascript/jscomp/runtime_tests/polyfill_tests/string_fromcodepoint_test.js
```diff
@@ -58,6 +58,7 @@ testSuite({
     assertFails(RangeError, () => String.fromCodePoint(noCheck({})));
     assertFails(RangeError, () => String.fromCodePoint(NaN));
     assertFails(RangeError, () => String.fromCodePoint(noCheck(/./)));
+    assertFails(RangeError, () => String.fromCodePoint(noCheck(undefined)));
```
teppeis: Although it is already working, I added a test just in case.

src/com/google/javascript/jscomp/js/es6/array/fill.js
```diff
@@ -31,14 +33,19 @@ $jscomp.polyfill('Array.prototype.fill', function(orig) {
...
+    var length = $jscomp.toLength(this.length);
+    var start = $jscomp.toInteger(opt_start);
+    if (start < 0) {
+      start = Math.max(0, length + start);
     }
-    if (opt_end == null || opt_end > length) opt_end = length;
-    opt_end = Number(opt_end);
-    if (opt_end < 0) opt_end = Math.max(0, length + opt_end);
-    for (var i = Number(opt_start || 0); i < opt_end; i++) {
+    var end = opt_end === undefined ? length : $jscomp.toInteger(opt_end);
```
shicks: There's a lot of repetitive code here and in other files.  What do you think about any of the following:

* $jscomp.toInteger takes a second optional argument for what to return on undefined (may not be quite as smooth as just checking isNaN tho since we still want NaN to map to zero even if undefined maps to length - so not sure if it ends up being worth it).
* add a $jscomp.toPosition(index, length) that handles negative indices and clamping all in one place (but would be nice to handle undefined here as well, and sometimes that's 0 and sometimes it's length - but at least we could just call $jscomp.toPosition(opt_start !== undefined ? opt_start : length, length))

lauraharker:
Closing this since there have been no comments for over a year.

PR #3226, opened 2019-02-07
Fix the return type for onmessage handler

realityforge:


tjgq:
Imported for review and testing. Created internal issue b/124051620.

realityforge:
Thanks again!

PR #3227, opened 2019-02-07
Make sure that callback returns void to match WebIDL

realityforge:


tjgq:
Imported for internal review and testing. Created internal issue b/124052479.

realityforge:
Thanks!

PR #3228, opened 2019-02-11
Fix contentEditable definition

RReverser:
It has always been a string and not a boolean, and is being extended with new values in a W3C proposal, which now fail the check.

isContentEditable, on the other hand, is a boolean but was defined without a type.

### DIFF
--- a/externs/browser/ie_dom.js
+++ b/externs/browser/ie_dom.js
@@ -1025,7 +1025,7 @@ Element.prototype.componentFromPoint = function(iCoordX, iCoordY) {};
 
 
 /**
- * @type {string}
+ * @type {string|boolean}
  * @see http://msdn.microsoft.com/en-us/library/ms533690(VS.85).aspx
  */
 Element.prototype.contentEditable;
--- a/externs/browser/ie_dom.js
+++ b/externs/browser/ie_dom.js
@@ -1025,7 +1025,7 @@ Element.prototype.componentFromPoint = function(iCoordX, iCoordY) {};
 
 
 /**
- * @type {boolean}
+ * @type {string}
  * @see http://msdn.microsoft.com/en-us/library/ms533690(VS.85).aspx
  */
 Element.prototype.contentEditable;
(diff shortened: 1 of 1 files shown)


tjgq:
Imported for internal testing. Created internal issue b/124236940.

externs/browser/ie_dom.js
```diff
@@ -1025,7 +1025,7 @@ Element.prototype.componentFromPoint = function(iCoordX, iCoordY) {};
 
 
 /**
- * @type {boolean}
+ * @type {string}
```
tjgq: This is one of those unfortunate cases where we can't fix the type easily because too much Google-internal code has come to depend on the current definition.

The best that can be done at the moment is to make this a `string|boolean`. Then, in a future where all internal code has been fixed to use strings, we can tighten the definition.

RReverser: *sigh* I was worried that would be the case, but hoped it would be treated as a bugfix and so the code would be fixed instead of definitions.

I can change it to the broader type.

RReverser: Done.

> Then, in a future where all internal code has been fixed to use strings, we can tighten the definition.

Is this tracked by the internal bug you mentioned? As in, do you think there's a chance of this being removed someday?

RReverser:
@tjgq Is this looking good now? I'm not sure what's going on on the CI...

tjgq:
I don't think the Travis failure is related to this PR. Let's ignore it for now.

This version is looking a lot better in internal testing. There are still a couple of internal projects that need fixing, but I expect to be able to submit it before the end of the week.

RReverser:
Sounds good, thanks. Out of curiosity - what are the cases still failing looking like?

For the initial version I expected just `smth.contentEditable = true` and similar to be popular enough to be a problem, but for the new one it seems trickier to trigger...

tjgq:
The problem is with `isContentEditable`, not `contentEditable`. It has to do with the fact that `?` (the unknown type) is contagious, and can be legally used in place of another type.

So if you have a return statement like `return str && element.isContentEditable`, and the function where it appears is `@return {boolean}`, the code type-checks before your changes (the type of the entire expression is `?`, since `isContentEditable` was `?` by omission), but not after (the type of the expression becomes `string|boolean`).

Fortunately, it's pretty easy to fix by rewriting the code as `return !!str && element.isContentEditable`.

RReverser:
Ah ok, that makes sense, thanks for explanation.

PR #3241, opened 2019-02-28
Use the PropertyCollapseLevel to guard for adding passes

ChadKillingsworth:
Prevents a module collapse level from adding passes which are only safe when all properties can be potentially collapsed.

Mark the `shouldCollapse` helper method as deprecated to prevent further use.


ChadKillingsworth:
Retriggering travis

brad4d:
Created internal issue http://b/126929590

PR #3245, opened 2019-03-06
Correct the return type of the WebSocket opopen and onclose event handers

realityforge:


blickly:
Created internal issue: http://b/127705524

realityforge:
Thanks!

PR #3248, opened 2019-03-08
Fix the constructor so that it can accept an array of strings

realityforge:
This is needed to align with the spec/WebIDL.

See

https://www.w3.org/TR/websockets/#handler-websocket-onmessage

### DIFF
--- a/externs/browser/html5.js
+++ b/externs/browser/html5.js
@@ -2919,7 +2919,7 @@ TimeRanges.prototype.end = function(index) { return 0; };
  * @see http://dev.w3.org/html5/websockets/
  * @constructor
  * @param {string} url
- * @param {string=} opt_protocol
+ * @param {(string|!Array<string>)=} opt_protocol
  * @implements {EventTarget}
  */
 function WebSocket(url, opt_protocol) {}


blickly:
Created internal issue http://b/128033107

externs/browser/html5.js
```diff
@@ -2919,7 +2919,7 @@ TimeRanges.prototype.end = function(index) { return 0; };
  * @see http://dev.w3.org/html5/websockets/
  * @constructor
  * @param {string} url
- * @param {string=} opt_protocol
+ * @param {(string|Array<string>)=} opt_protocol
```
lauraharker: Please add a `!` (indicating non-nullability):
``` js
 * @param {(string|!Array<string>)=} opt_protocol
```

https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler#nullable-type

Otherwise LGTM.

realityforge: Done and re-pushed as a single commit

lauraharker:
Running internal tests now.

realityforge:
Thanks

PR #3249, opened 2019-03-08
Mark some constant fields with @const

realityforge:


blickly:
Created internal issue: http://b/128033140

realityforge:
Thanks

PR #3250, opened 2019-03-08
Make it possible to pass a Blob to webSocket.send()

realityforge:
This is required to support scenario where websocket.binaryType = 'blob'


blickly:
Created internal issue: http://b/128030553

realityforge:
Tah

PR #3251, opened 2019-03-08
Explicitly specify the possible types for websocket message payloads

realityforge:


blickly:
Created internal issue: http://b/128033880

brad4d:
Added an update to the WebSocket spec, imported, and sent for internal review and testing.

There is some risk that some internal Google tests will break, because they rely on the unknown (`?`) type currently used in the definition of `onmessage`.

realityforge:
Thanks for putting in the time with this!

brad4d:
This change causes an error for closure-library:

```
javascript/closure/net/websocket.js:327: ERROR - assignment to property onmessage of WebSocket
found   : function((MessageEvent<string>|null)): undefined
required: (function(MessageEvent<(ArrayBuffer|Blob|null|string)>): undefined|null)
  this.webSocket_.onmessage = goog.bind(this.onMessage_, this);
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

realityforge:
Eyeballing it ... it seems that the fix may be updating the closure-library onMessage_ to something like

```javascript
**
 * Called when a new message arrives from the server.
 *
 * @param {MessageEvent<string>} event The web socket message event.
 * @return {void}
 * @private
 */
goog.net.WebSocket.prototype.onMessage_ = function(event) {
  var message = event.data;
  this.dispatchEvent(new goog.net.WebSocket.MessageEvent(message));
};
```
Does that seem correct? Would you like me to submit a PR to that library to do so?

externs/browser/html5.js
```diff
@@ -3040,8 +3040,7 @@ WebSocket.prototype.onopen;
 
 /**
  * An event handler called on message event.
- * TODO(blickly): Change this to MessageEvent<*> and add casts as needed
- * @type {?function(!MessageEvent<?>): void}
+ * @type {?function(!MessageEvent<string|ArrayBuffer|Blob>): void}
```
brad4d: I don't think the spec allows `null`, so the `MessageEvent` type should be `!MessageEvent<string|!ArrayBuffer|!Blob>`.
Is that right?

brad4d:
I would appreciate it if you would do a PR for closure-library.

I think the correction would look something like this, though.

```javascript
**
 * Called when a new message arrives from the server.
 *
 * @param {MessageEvent<string|!ArrayBuffer|!Blob>} event The web socket message event.
 * @return {void}
 * @private
 */
goog.net.WebSocket.prototype.onMessage_ = function(event) {
  // This code only uses string messages
  var message = assertString(event.data);
  this.dispatchEvent(new goog.net.WebSocket.MessageEvent(message));
};
```

externs/browser/html5.js
realityforge: That does seem right. I will push the change. Would you prefer if I rebased and squashed the commit?

realityforge:
PR submitted to closure-library. Assuming it passes review I assume you have some way to coordinate the changes at your end. 

Thanks!

externs/browser/html5.js
brad4d: Thanks for asking, but I don't need the rebase/squash.
My personal workflow for importing PRs will actually work better if you don't do that.

realityforge:
@brad4d The change has landed in closure-library, kick me if there is anything else you need me to do.

brad4d:
@realityforge Thanks for the closure-library fix.
I've pulled your latest change and sent the result for testing again.

realityforge:
Thanks!

brad4d:
PR submitted internally.
It should be in tomorrow's push to GitHub.

realityforge:
Great - thanks

PR #3252, opened 2019-03-08
Align the API with the standardized WebIDL

realityforge:
Historically some browsers (i.e. Gecko) returned a boolean from the
send method to indicate whether the message was successfully
buffered and/or sent. The modern specification specifies a void
return type [1] which is how it is implemented in modern browsers.

[1] https://html.spec.whatwg.org/multipage/web-sockets.html#the-websocket-interface


blickly:
Yiting, can you merge this one?

blickly:
Created internal issue: http://b/128034259

PR #3254, opened 2019-03-11
Add WebSocket.prototype.onerror extern

realityforge:
This is defined as part of the current WebIDL

https://html.spec.whatwg.org/multipage/web-sockets.html#the-websocket-interface


blickly:
Internal issue created http://b/128352435

PR #3255, opened 2019-03-11
Explicitly type the CloseEvent passed to WebSocket.onclose event handler

realityforge:
This aligns the externs with the current WebIDL

https://html.spec.whatwg.org/multipage/web-sockets.html#the-closeevent-interface


blickly:
Created internal issue http://b/128352373

realityforge:
Thanks!

PR #3265, opened 2019-03-14
Add Buffer.alloc to nodejs externs (fixes #3263)

### LINKED ISSUE #3263: contrib/nodejs/buffer.js is outdated
This 

https://github.com/google/closure-compiler/blob/master/contrib/nodejs/buffer.js 

is outdated wrt to 

https://npmdoc.github.io/node-npmdoc-buffer/build/apidoc.html#apidoc.element.buffer.Buffer.alloc

resulting in `JavaScript exception: TypeError: Buffer.nb is not a function` error (`JavaScript exception: TypeError: Buffer.$alloc$ is not a function` when compiled with `--debug`) in my code. The following addition (a partial hack) fixed it for me:

```
/**
 * @param {number=} size
 * @return {Buffer}
 */
buffer.Buffer.alloc;
```

Note that I get the same error when using https://github.com/dcodeIO/node.js-closure-compiler-externs/blob/master/buffer.js, which is more complete, but even older and so similarly outdated, so instead I report here.

Mikolaj:
See #3263.

### DIFF
--- a/contrib/nodejs/buffer.js
+++ b/contrib/nodejs/buffer.js
@@ -55,3 +55,11 @@ buffer.SlowBuffer.prototype.slice;
  * @return {string}
  */
 buffer.SlowBuffer.prototype.toString;
+
+/**
+ * @param {number} size
+ * @param {(string|Buffer|number)=} fill
+ * @param {number=} encoding
+ * @return {Buffer}
+ */
+buffer.Buffer.alloc;
--- a/contrib/nodejs/buffer.js
+++ b/contrib/nodejs/buffer.js
@@ -58,8 +58,8 @@ buffer.SlowBuffer.prototype.toString;
 
 /**
  * @param {number} size
- * @param {(string|Buffer|number)=} fill
- * @param {number=} encoding
- * @return {Buffer}
+ * @param {(string|!Buffer|number)=} fill
+ * @param {string=} encoding
+ * @return {!Buffer}
  */
 buffer.Buffer.alloc;


Mikolaj:
I signed it.

contrib/nodejs/buffer.js
```diff
@@ -55,3 +55,11 @@ buffer.SlowBuffer.prototype.slice;
  * @return {string}
  */
 buffer.SlowBuffer.prototype.toString;
+
+/**
+ * @param {number} size
+ * @param {(string|Buffer|number)=} fill
```
tjgq: Write `!Buffer` to make it non-nullable.

contrib/nodejs/buffer.js
```diff
@@ -55,3 +55,11 @@ buffer.SlowBuffer.prototype.slice;
...
+ * @param {number=} encoding
```
tjgq: According to https://nodejs.org/api/buffer.html#buffer_class_method_buffer_alloc_size_fill_encoding, I think this should be `string=`?

contrib/nodejs/buffer.js
```diff
@@ -55,3 +55,11 @@ buffer.SlowBuffer.prototype.slice;
...
+ * @return {Buffer}
```
tjgq: Write `!Buffer` to make it non-nullable.

contrib/nodejs/buffer.js
Mikolaj: Oh, yes, good catch.

contrib/nodejs/buffer.js
Mikolaj: Will do.

contrib/nodejs/buffer.js
Mikolaj: Done. Let's see if this fixes travis.

tjgq:
Created Google internal issue b/128616787. Imported for internal review and testing.

Let's ignore the Travis failure for now - I believe it's related to https://github.com/google/closure-compiler-npm/issues/130 which @ChadKillingsworth is trying to fix.

PR #3268, opened 2019-03-18
Migrate the standardized Console to a separate top-level extern

realityforge:
The purpose is to allow projects that only want to support the standardized part of the API to pull in that single extern. Currently google/elemental2 has copy pasted the relevant section into the downstream project. Non-standardized aspects will continue to be present in `webkit_dom.js`. This patch also renames some parameter names to align with the names in the spec and orders the declarations as they appear in the spec.

### DIFF
--- a/externs/browser/webkit_dom.js
+++ b/externs/browser/webkit_dom.js
@@ -94,86 +94,6 @@ ScriptProfile.prototype.uid;
 ScriptProfile.prototype.head;
 
 /**
- * @constructor
- * @see https://console.spec.whatwg.org/
- */
-function Console() {};
-
-/**
- * @param {*} condition
- * @param {...*} var_args
- * @return {undefined}
- */
-Console.prototype.assert = function(condition, var_args) {};
-
-/**
- * @param {...*} var_args
- * @return {undefined}
- */
-Console.prototype.error = function(var_args) {};
-
-/**
- * @param {...*} var_args
- * @return {undefined}
- */
-Console.prototype.info = function(var_args) {};
-
-/**
- * @param {...*} var_args
- * @return {undefined}
- */
-Console.prototype.log = function(var_args) {};
-
-/**
- * @param {...*} var_args
- * @return {undefined}
- */
-Console.prototype.warn = function(var_args) {};
-
-/**
- * @param {...*} var_args
- * @return {undefined}
- */
-Console.prototype.debug = function(var_args) {};
-
-/**
- * @param {*} value
- * @return {undefined}
- */
-Console.prototype.dir = function(value) {};
-
-/**
- * @param {...*} var_args
- * @return {undefined}
- */
-Console.prototype.dirxml = function(var_args) {};
-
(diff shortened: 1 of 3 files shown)


realityforge:
It is unclear to me why this build is failing ... I am assuming it is because `resources.json` is not rebuilt which seems to be something built inside google. If that is not the case then feel free to kick it back to me and I will try and figure out what the root cause is.

tjgq:
This PR is doing too much at once. Can you please split it into:

* One PR that moves the console externs into a separate file, without changing any of the externs definitions or adding new ones;

* One PR with the rest of the changes (this one might still be too big to land at once, but we'll see how that goes).

This makes it a lot easier on us, as changes to the default externs tend to be disruptive and require fixes to internal Google projects.

realityforge:
No problem. I will break the commit into a bunch of small ones based on risk of breaking stuff and then one last one to split into a separate file.

Thanks

realityforge:
@tjgq I have updated this PR to just include the movement of externs to separate file and rename of function parameters to match the spec.

One thing that should be considered going forward is that `whatwg_console.js` could be used in a node environment and thus the duplicate extern declarations in `contrib/nodejs/globals.js` could be removed if some how this could be enabled. This is beyond my current understanding to do ... but maybe some one else understands enough and cares enough to do so ;)

EatingW:
Created Google internal issue http://b/128987595

tjgq:
How important is it to you that the standard externs be separate from the non-standard ones?

In general, if we're going to split the existing externs at all (which is a good idea!), I think we'd prefer to separate them by top-level API, rather than by version of the spec.

Guarding against uses of non-standard APIs would be better achieved by [conformance rules](https://github.com/google/closure-compiler/wiki/JS-Conformance-Framework) than by restricting the set of externs visible to the compiler.

realityforge:
It would be really, really, really nice to have them in separate files. The reason is that these externs will drive the generation of java classes and I would much prefer that non-standard operations/properties are just omitted. If they are not omitted, then someone will use them. Using Elemental2+J2CL+closure compiler would allow us to pick up the problems using the JS Conformance Framework but using Elemental2+GWT2.x would be unable to pick up the problem. 
Also it seems unkind to the consumers to provide an API and then emit an error if it is used. It . would be much nicer to just omit the API IMHO.

As to how to group the externs - I had considered moving the webkit specific externs to `externs/browser/whatwg_console_webkit.js` (My preference as it groups files about the same API) or `externs/browser/webkit_console.js` but left it as is as I was not sure this would be desired by the project but I am happy to do so if you think it would be better.

tjgq:
In the end, we decided to move the non-standard console extensions into a separate `nonstandard_console.js` file, and keep only the standard ones in `whatwg_console.js`.

You don't need to amend this pull request; I've made the changes internally to avoid additional back-and-forth. The updated version will be in our next push to GitHub.

realityforge:
Great - thanks

PR #3270, opened 2019-03-18
Add externs for countReset and timeLog

realityforge:


realityforge:
Note this has been extracted from #3268 to simplify merging

externs/browser/webkit_dom.js
```diff
@@ -194,6 +200,12 @@ Console.prototype.profileEnd = function(opt_title) {};
  */
 Console.prototype.time = function(name) {};
 
+/**
+ * @param {string} label
+ * @return {undefined}
+ */
+Console.prototype.timeLog = function(label) {};
```
tjgq: Per the spec, this needs an additional `@param {*...} data`.

https://console.spec.whatwg.org/#timing

realityforge: Interesting - I had not seen that before. It is not documented at https://developer.mozilla.org/en-US/docs/Web/API/Console/timeLog but I will try it out in the browser and if it works add it in.

realityforge: New commit pushed with change

tjgq:
Created Google internal issue b/128865976. Imported for internal review and testing.

PR #3273, opened 2019-03-19
Move extern Node.prototype.parentElement to w3c externs

realityforge:
It is defined as part of DOM Level 4 so should be in that extern.
See https://developer.mozilla.org/en-US/docs/Web/API/Node/parentElement

This also makes it available to downstream consumers such as elemental2


tjgq:
Created Google internal issue b/128936874. Imported for internal review and testing.

realityforge:
Thanks

PR #3274, opened 2019-03-19
Remove the @readonly annotation as it is not used

realityforge:


tjgq:
@jDramaix has already fixed this internally. It should land with the next push to GitHub.

realityforge:
Thanks

PR #3275, opened 2019-03-19
Correct the type of the children and link to the current spec

realityforge:


tjgq:
Created Google internal issue b/128937896. Imported for internal review and testing.

tjgq:
Unfortunately, this PR breaks a large number of internal projects (more than it is reasonable to manually fix).

The issue is that every reference to `Element#children` currently stored into a variable of type `NodeList`, or passed into a function that expects a `NodeList` argument, will now cause a type error.

One possible workaround is to make `Element#children` a union type `(NodeList|HTMLCollection)`. Given the quirks in how the compiler handles union types, that might be enough to fix most cases, or at least bring them down to a tractable number.

Another idea is to declare `HTMLCollection` as a subclass of `NodeList` (per the spec it isn't, but I believe the differences to be fairly immaterial). However, that would leave us in a bind if the definition of `NodeList` ever evolved into something that is no longer a subset of `HTMLCollection`.

I'm also open to other suggestions.

I intend to try these two approaches and report back, but given my limited time it's likely to take a few days.

realityforge:
One of the drives of this is to make the downstream artifacts generated from externs (i.e. the java classes generated via `google/elemental2`) as correct and useful as possible or failing that less likely to encourage incorrect behaviour. Basically so that if you use the generated code it should not produce errors.

Currently the value of children exposes some methods that are not present on the real children value (i.e. `NodeList.prototype.forEach`). Thus I think that making `HTMLCollection` extend `NodeList` is the wrong solution as it exposes all the same issues. A union type, while not ideal seems the least bad. Although if that proves problematic I can update `elemental2` to patch the externs prior to use, although that seems like a less ideal situation.

tjgq:
Apologies for the long turnaround on this.

I investigated both the union type and the subclass approach, and unfortunately they both require a nontrivial amount of work to fix internal code relying on the old type. The payoff doesn't seem to be large enough to justify that work, so we won't be able to accept this pull request, sorry :(

Since elemental2 currently needs to patch the externs and they're also a Google project, one other avenue is to inquire whether they'd be willing to do the internal cleanup themselves. I don't want to speak for elemental2, however, so you'd have to follow up with them separately.

realityforge:
No problem - thanks for looking into it.

PR #3276, opened 2019-03-19
Explicitly mark const values

realityforge:
If a value is defined as "const" in the WebIDL or the spec it
should be marked as such. The advantage is that this flows into
downstream consumers such as google/elemental2 allowing those
projects to define values as constants.


tjgq:
Created Google internal issue b/128939064. Imported for internal review and testing.

realityforge:
Fantastic - thanks!

PR #3278, opened 2019-03-20
Cleanup EventSource externs

realityforge:
* Mark withCredentials as @const to align with the spec [1]
* Updated spec url
* Corrected the return type of onopen, onmessage and onerror to be void
* Collapsed '@type {X} @const' declarations to '@const {X}'

[1] https://html.spec.whatwg.org/multipage/server-sent-events.html#the-eventsource-interface

### DIFF
--- a/externs/browser/w3c_eventsource.js
+++ b/externs/browser/w3c_eventsource.js
@@ -16,7 +16,7 @@
 
 /**
  * @fileoverview Definitions for W3C's EventSource API.
- * @see http://www.w3.org/TR/eventsource/
+ * @see https://html.spec.whatwg.org/multipage/server-sent-events.html#server-sent-events
  *
  * @externs
  */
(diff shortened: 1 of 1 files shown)


EatingW:
Created internal issue http://b/128987592

EatingW:
I'm not very familiar with EventHandlers, but it seems to me from https://html.spec.whatwg.org/multipage/webappapis.html#the-event-handler-processing-algorithm that there might be return values? Is there documentation elsewhere that says the return type of onopen, onmessage, and onerror is void?

realityforge:
I am away from my development machine atm so I can not confirm this but I am under the impression that it would effectively be a no-op to return a value as the event is not cancelable and there is no subsequent event handler and thus `returnValue` is also ignorable. It is mostly equivalent to invoking `event.preventDefault()` FWIU which is likewise a noop (again because there is no "default" action).

Note a similar change has been applied to `WebSocket` at https://github.com/google/closure-compiler/blob/master/externs/browser/html5.js#L3029-L3052

Cleaning this up has the advantage that downstream consumers of the externs such as google/elemental2 are forced to generate types with a return type if there is an any return type (which increases verbosity and code size) even though there is no real reason to allow this.

EatingW:
Ok, sounds good. Thanks for explaining. I've imported this for internal testing and review. It should be fine as long as no one else was depending on the previous behavior.

realityforge:
Thanks

Issue #3283, opened 2019-03-21
Incorrect peephole optimisation (all targets!)


zavr-1:
The compiler does not work correctly with `export { default as Module, OtherModule } from './mod'` which have side-effects (and capital letters are the key for some reason, because it's fine with lowercase). The `peepholeOptimizations` pass removes code that it should not remove. It affects all targets, including 2015. For example, let's have a library module:

```js
// sync.js
export default (callback) => {
  callback()
}
```

And further, a module that exports the initial function using `export { as }`:

```js
// s.js
export { default as Update } from './sync'
```

Then our source code would import the module in the following way, first correctly from the original initial function:

```js
// source.js
import Update from './sync'

let stack = 1
Update(() => {
  stack = 0
})

if (stack) throw new Error('The stack is not updated')
console.log('everything is fine')
```

When compiled in advanced mode, all is good:

```sh
java -jar /Users/zavr/node_modules/google-closure-compiler-java/compiler.jar
--compilation_level ADVANCED 
--language_out ECMASCRIPT_2015 
--module_resolution NODE 
--js source.js sync.js
```

```js
'use strict';
console.log("everything is fine");
```

However, when we change the source to import the method from the proxying library, everything becomes broken:

```js
import { Update } from './s'

let stack = 1
Update(() => {
  stack = 0
})

if (stack) throw new Error('The stack is not updated')
console.log('everything is fine')
```

```sh
java -jar /Users/zavr/node_modules/google-closure-compiler-java/compiler.jar 
--compilation_level ADVANCED 
--language_out ECMASCRIPT_2015 
--module_resolution NODE 
--js source.js s.js sync.js
```

```js
'use strict';
throw Error("The stack is not updated");
```

## peepholeOptimizations yields:

The problem was uncovered with `--print_source_after_each_pass`. At some point we have:

```js
// ************************************

// earlyInlineVariables yields:
// ************************************
'use strict';
var $jscompDefaultExport$$module$sync = callback$jscomp$48 => {
  callback$jscomp$48();
};
let stack$$module$source = 1;
({get Update() {
  return $jscompDefaultExport$$module$sync;
}}).Update(() => {
  stack$$module$source = 0;
});
if (stack$$module$source) {
  throw new Error("The stack is not updated");
}
console.log("everything is fine");

// ************************************

// peepholeOptimizations yields:
// ************************************
'use strict';
var $jscompDefaultExport$$module$sync = callback$jscomp$48 => {
  callback$jscomp$48();
[...]

PR #3286, opened 2019-03-22
Handle get side-effects in peephole optimisation.

zavr-1:
This is to address incorrect optimisations for `export { default as X } from Y` in https://github.com/google/closure-compiler/issues/3283. In sum, the peephole optimisation didn't care for `get` which could have side-effects and folded it like a normal argument, breaking code like 

```js
({ get b() { console.log(10) } }).b
```

 It's still a mystery to me why 
```js
({get Update() {
  return $jscompDefaultExport$$module$t$modules$sync;
}}).Update(function() {
  stack$$module$t$modules$index = 0;
});
```
got folded incorrectly, and the following didn't...
```js
({get update() {
  return $jscompDefaultExport$$module$t$modules$sync;
}}).update(function() {
  stack$$module$t$modules$index = 0;
});
``` 

--- 
In any way, the side effects of getters are handled better now. The diff might be a lot of red/green, but I just moved some code logic into separate functions. Cheers.


src/com/google/javascript/jscomp/NodeUtil.java
```diff
@@ -1461,89 +1461,95 @@ static boolean functionCallHasSideEffects(
...
+              return !param.getNext().isString();
+            } else if (STRING_REGEXP_METHODS.contains(method)) {
+              return false;
             }
           }
         }
       }
     }
-
     return true;
```
lauraharker: I suspect this `return true;`, assuming most property accesses have side effects, will block us from accepting this PR.

Right now @concavelenz is working on something with https://github.com/google/closure-compiler/blob/ab857cffd33d7114c58f47655d6f23795a27a610/src/com/google/javascript/jscomp/GatherGettersAndSetterProperties.java. This pass tracks which properties are potentially getters and setters.

I agree that the peephole optimization behavior is bad, but long-term we want to be more choosy about which property accesses we treat as potentially side-effectful. Does that sound reasonable?

zavr-1: Yeah I agree this is why (just want to be more specific than _suspect_ for myself)

If you open https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/NodeUtil.java#L1463 and follow down then you can see a many `return false`, `return false` and it's never returning true. If the function didn't return in that `if` block, then it falls through to return true. Therefore I just moved the code to a separate function, and no change to the `functionCallHasSideEffects` at all because it's pure refactoring. 

Then, the real functional change is the addition of `case GETPROP` to the `nodeTypeMayHaveSideEffects` because it is the one that is called by the peephole optimiser. When it started to return true, I saw the failure in [the test](https://github.com/google/closure-compiler/blob/master/test/com/google/javascript/jscomp/PeepholeRemoveDeadCodeTest.java#L1289) that did `await some.thing()` [coming from](https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/PureFunctionIdentifier.java#L606) the `PureFunctionIdentifier` class, saying "unhandled side effect". Therefore, any function would now assume a side-effect exactly as you say. Despite the fact that there is an attempt to handle it by `setSideEffectsForGetProp`, it does not seem to be efficient: even the simple call `some.thing()` will now mark the function as having global side effects:

Screenshot 2019-03-26 at 03 20 39

I don't know why the function `setSideEffectsForGetProp` is not that efficient, and if it could be better, but it does not seem reasonable to mark a function tainted just because it accesses a property in an external object (although there's a chance as with getters). I do agree that side-effects in getters is a horrible thing to do, but there're some styles of programming that do exactly that, I think I saw that in some assertion libraries. Perhaps a note in the README or WIKI could be made about that? Plus as in with the issue, it was generated internally by the compiler -- is there any other places it does that. Finally, the fact that the green diff for `setSideEffectsForGetProp` finishes on line 666 makes me reluctant to continue (on top a number of failing tests) 😆 maybe when that change you pointed out earlier is implemented, it would be possible to improve the `setSideEffectsForGetProp` function. 

I also commented on the issue regarding if the latest changes help. ty

zavr-1: would it still help to refactor the code into separate functions though?

lauraharker: We have documented the getters with side effects issue here: https://github.com/google/closure-compiler/wiki/Compiler-Assumptions. I added another link in the readme under "Reporting a bug".

It's kind of sad that the compiler is the one creating getters with side-effects, yes :)

PR #3287, opened 2019-03-22
Duplicate EventSource constants onto symbol

realityforge:
This aligns with the spec and ensures that Elemental2 generates these as static final fields rather than instance fields

### DIFF
--- a/externs/browser/w3c_eventsource.js
+++ b/externs/browser/w3c_eventsource.js
@@ -57,17 +57,32 @@ EventSource.prototype.withCredentials;
 /**
  * @const {number}
  */
-EventSource.prototype.CONNECTING = 0;
+EventSource.prototype.CONNECTING;
 
 /**
  * @const {number}
  */
-EventSource.prototype.OPEN = 1;
+EventSource.CONNECTING;
 
 /**
  * @const {number}
  */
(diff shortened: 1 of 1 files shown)


EatingW:
Created Google internal issue http://b/129273027.

lauraharker:

externs/browser/w3c_eventsource.js
```diff
@@ -57,17 +57,17 @@ EventSource.prototype.withCredentials;
 /**
  * @const {number}
  */
-EventSource.prototype.CONNECTING = 0;
+EventSource.CONNECTING = 0;
```
lauraharker: Looks like https://github.com/google/closure-compiler/pull/3293 already discusses values in externs, so I'll wait for that discussion to be resolved.

realityforge: Ok @lauraharker - I have removed values from this extern to follow discussion in #3293 and changed the PR so that rather than moving constants to symbol that they are instead on both as both are required

realityforge: It should be ready to re-review again.

lauraharker: Sounds good, thanks!

realityforge:
Thanks!

PR #3288, opened 2019-03-22
Fix the type of data payload for EventSource.onmessage

realityforge:


EatingW:
Created Google internal issue http://b/129274308.

externs/browser/w3c_eventsource.js
```diff
@@ -81,7 +81,7 @@ EventSource.prototype.readyState;
 EventSource.prototype.onopen = function(e) {};
 
 /**
- * @type {?function(!MessageEvent<*>): void}
+ * @type {?function(!MessageEvent<string>): void}
```
lauraharker: Could you point me to where the spec defines this as a string?

I can only find it defined as a `MessageEvent` [here](https://html.spec.whatwg.org/multipage/server-sent-events.html#the-eventsource-interface).

realityforge: The whatwg kind of describes it in the section [Interpreting an event stream](https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation) where it describes how to initialize the buffer as an empty string, parse lines as string and concat "data" lines using a joining LF character. It then describes how the user agent must interpret it

> 2. If the data buffer is an empty string, set the data buffer and the event type buffer to the empty string and return.
> 3. If the data buffer's last character is a U+000A LINE FEED (LF) character, then remove the last character from the data buffer.
> 4. Let event be the result of creating an event using MessageEvent, in the relevant Realm of the EventSource object.
> 5. Initialize event's type attribute to message, its data attribute to data, its origin attribute to the serialization of the origin of the event stream's final URL (i.e., the URL after redirects), and its lastEventId attribute to the last event ID string of the event source.

It may not explicitly indicated that it MUST be a string but data is essentially the result of a bunch of operations on UTF-8 string which I presume the browser never transforms to another type (I certainly have never seen it). The underlying protocol is UTF-8 text based protocol the easiest explanation is on the MDN site at [Event stream format](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#Event_stream_format).

lauraharker: Thanks, that sounds reasonable to me.

realityforge:
Tah!

PR #3289, opened 2019-03-23
Correctly mark parameter type as non-nullable

realityforge:


EatingW:
Created Google internal issue http://b/129274127.

EatingW:
Imported for internal testing and review.

PR #3290, opened 2019-03-23
Ensure that the error constants in MediaError are correctly annotated

realityforge:
Constant values extracted from https://developer.mozilla.org/en-US/docs/Web/API/MediaError

This fixes google/elemental2#71

### DIFF
--- a/externs/browser/html5.js
+++ b/externs/browser/html5.js
@@ -2389,27 +2389,27 @@ MediaError.prototype.message;
 /**
  * The fetching process for the media resource was aborted by the user agent at
  * the user's request.
- * @type {number}
+ * @const {number}
  */
 MediaError.MEDIA_ERR_ABORTED;
 
 /**
  * A network error of some description caused the user agent to stop fetching
  * the media resource, after the resource was established to be usable.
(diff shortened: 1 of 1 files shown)


EatingW:
Created Google internal issue http://b/129275520.

externs/browser/html5.js
```diff
@@ -2386,29 +2386,29 @@ MediaError.prototype.message;
 /**
  * The fetching process for the media resource was aborted by the user agent at
  * the user's request.
- * @type {number}
+ * @const {number}
  */
-MediaError.MEDIA_ERR_ABORTED;
+MediaError.MEDIA_ERR_ABORTED = 1;
```
EatingW: As Ben mentioned in #3293, we generally avoid putting values in the externs. Otherwise looks good.

realityforge: Ok - wasn't able to convince anyone that the code size optimization was worth including the values for constants so I have stripped them out and re-pushed the changeset.

EatingW:
Thanks! I'm importing this for internal testing and review.

PR #3291, opened 2019-03-23
Remove Node.XPATH_NAMESPACE_NODE as no such constant.

realityforge:
The correct location of this constant is XPathNamespace.XPATH_NAMESPACE_NODE
and it is correctly specified on this type in w3c_xml.js. The Node constants
are defined at https://www.w3.org/TR/1998/REC-DOM-Level-1-19981001/level-one-core.html#ID-1950641247


EatingW:
Created Google internal issue http://b/129273343 for tracking.

realityforge:
Thanks

PR #3292, opened 2019-03-23
Remove MediaStreamTrack.getSources as no modern browser supports it

realityforge:
See support at https://developer.mozilla.org/en-US/docs/Web/API/MediaStreamTrack

It was removed from Chrome a long time ago (in web terms). See https://www.chromestatus.com/features/4765305641369600


EatingW:
Created Google internal issue http://b/129274728.

realityforge:
Thanks

PR #3293, opened 2019-03-23
Annotate and specify values for constants on Node

realityforge:


EatingW:
Created Google internal issue http://b/129272816 for tracking.

externs/browser/w3c_dom1.js
```diff
@@ -312,10 +312,10 @@ Node.TEXT_NODE;
 Node.XPATH_NAMESPACE_NODE;
 
 /**
- * @type {number}
+ * @const {number}
  * @see http://www.w3.org/TR/1998/REC-DOM-Level-1-19981001/level-one-core.html#ID-1950641247
  */
-Node.NOTATION_NODE;
+Node.NOTATION_NODE = 12;
```
blickly: We generally avoid putting values in the externs, as that keeps them as more clearly a declaration of the "interface" of the external code.

The @const changes look good, though.

realityforge: I hope I can change your mind about that ;)There is a few reasons for supplying explicit values.

Firstly when the `google/elemental2` project reads these externs and sees both a `@const` and a value it is able to explicitly model these values as constant final values in java code. This is much more comfortable experience for elemental2 users to see that a property is constant rather representing it as mutable property in java that generates an error at runtime (GWT2.x) or compile time (j2cl+closure) when you assign a value to it.

Secondly (at least in GWT2.x) it leads to more optimised code. Essentially the compiler no longer needs to use the long symbol (i.e. `Node.NOTATION_NODE`) versus inlining the constant value. I would presume that this is true in closure pipeline as well although I have not verified it.

Thirdly, a lot of sample code or code in the wild actually inlines the constants rather than using constant symbol. (i.e. `WebSocket.readyState` is almost always checked against integers (i.e. `1`) rather than the symbols (i.e. `WebSocket.OPEN`). I am unclear on why this is. Probably people performing a manual optimization ... or copying code that does the same.

The only downsides to including this that I am aware of is the extra effort to add the constants and the possibility that you get something wrong.

Thoughts?

blickly:
In closure compiler, externs represent the types for something that is external to the compilation, and cannot, e.g. be inlined into the code.  As such, there should be no requirement that externs include a value in order for them to be considered final constant values; they should be considered final constants with just a @const annotation.

Any compiler that is consuming externs files with values, and changing behavior with respect to whether the externs has values or not is in my opinion showcasing a bug of it's own.

Look at for example at all the @enums inside:
https://github.com/google/closure-compiler/blob/master/contrib/externs/chrome_extensions.js#L2311
Since externs are intended to include only types, the values used there are not meaningful and should definitely not be depended on in any downstream tooling.

realityforge:
Interesting.

Although that is rather unfortunate as it did lead to a nice code size reduction. FWIW the `google/elemental2` project only relies on `externs/*`, `contrib/externs/svg.js` and `contrib/externs/webassembly.js` and none of the constant values in them seem wrong (although none of these use `@enum` that I am aware of. 

Would you consider accepting the PR with the values present or do they need to be stripped out? FWIW the current closure-compiler branch I am on has 147 other constant values present in the `externs/browser` directory alone. If it is considered "wrong" enough that this PR won't be accepted without the values stripped should I submit a PR stripping all the other values?

gkdn:
I agree with @blickly  on this. Externs shouldn't replicate the values from spec which my lead to potentially if used by any tooling including Elemental. Code size part is only relevant to old GWT compiler, and should be fixed on that side as we discussed earlier.

realityforge:
I have pushed a changeset that does not supply values.

@gkdn While inlining these values into the java code would work around the unnecessary code-size increases that are present in GWTs version of Elemental2, inlining them would also reduce the size of the code in J2CL/closure world as well.

I appreciate that it does increase the work load to specify the values, and risk occurs if the value is wrong or if the inlined value was ever changed (which seems unlikely for spec driven externs). However if the project does not want to do it on principal because you would then thats fine too.

PR #3297, opened 2019-03-26
Cleanup Geolocation API to align with spec

realityforge:
1. Rename the following symbols to align with the spec
   - GeolocationPositionError to PositionError
   - GeolocationPositionOptions to PositionOptions
   - GeolocationCoordinates to Coordinates
   - GeolocationPosition to Position
1. Move the UNKNOWN_ERROR, PERMISSION_DENIED and TIMEOUT properties from the PositionError prototype onto the PositionError type
1. Annotate the UNKNOWN_ERROR, PERMISSION_DENIED and TIMEOUT properties as @const
1. Extract typedefs for the success and error callbacks passed to `Geolocation.getCurrentPosition` and `Gelocation.watchPosition`. The typedefs are named using the names in the spec. This improves the types generated in downstream consumers ala Elemental2.
1. Correct the return values of callbacks to return void to align with spec.

### DIFF
--- a/externs/browser/w3c_geolocation.js
+++ b/externs/browser/w3c_geolocation.js
@@ -27,9 +27,19 @@
 function Geolocation() {}
 
 /**
- * @param {function(!GeolocationPosition)} successCallback
- * @param {(function(!GeolocationPositionError)|null)=} opt_errorCallback
- * @param {GeolocationPositionOptions=} opt_options
+ * @typedef {function(!Position): void}
+ */
+var PositionCallback;
+
+/**
+ * @typedef {function(!PositionError): void}
+ */
+var PositionErrorCallback;
+
+/**
+ * @param {PositionCallback} successCallback
+ * @param {PositionErrorCallback=} opt_errorCallback
+ * @param {PositionOptions=} opt_options
  * @return {undefined}
  */
 Geolocation.prototype.getCurrentPosition = function(successCallback,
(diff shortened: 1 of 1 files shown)


realityforge:
My guess is that this commit is too large to push through the google internal review process. But assuming that the changes seem fine in principle I can break up the commits in whatever way works for you.

blickly:
[Converted your description to a numbered list to help with the discussion]

My gut reaction here is that points 3-5 all sound convincing to me.

2 sounds off to me.  Don't they need to be on the prototype to be part of the @record?

1 sounds reasonable enough if we were starting from scratch, but I'm not convinced that the churn of trying to update everyone with references to the old names are worth it.

realityforge:
The WebIDL for `TIMEOUT` and friends declares them as constants on an interface. The WebIDL spec definition for [constants](https://heycam.github.io/webidl/#idl-constants) declares:

> Constants are not associated with particular instances of the interface on which they appear. It is language binding specific whether constants are exposed on instances.

In javascript that means that they are available on instances ... (yay prototype inheritance). 

So looking at the [WebIDL for PositionError](https://www.w3.org/TR/geolocation-API/#position_error_interface) ... err wait. The interface is marked with the `[NoInterfaceObject]` extended attribute so you are probably correct. I will experiment with this over the weekend.

Hmmm this does make the generated code in elemental2 less useful :(

Anyhoo - I will split out the other changes and send a PR.

blickly:
OK, sounds good.  Closing this one in favor of the smaller more manageable split-off PRs.

Thanks for all the contributions!

PR #3300, opened 2019-03-28
Cleanup the ApplicationCache typing

realityforge:
1. Update the callback methods to return void
2. Expose the constants on the type as well as the instance

### DIFF
--- a/externs/browser/html5.js
+++ b/externs/browser/html5.js
@@ -1176,10 +1176,24 @@ ApplicationCache.prototype.dispatchEvent = function(evt) {};
 ApplicationCache.prototype.UNCACHED = 0;
 
 /**
+ * The object isn't associated with an application cache. This can occur if the
+ * update process fails and there is no previous cache to revert to, or if there
+ * is no manifest file.
+ * @const {number}
+ */
+ApplicationCache.UNCACHED = 0;
+
+/**
  * The cache is idle.
  * @const {number}
  */
 ApplicationCache.prototype.IDLE = 1;
+
+/**
+ * The cache is idle.
+ * @const {number}
+ */
+ApplicationCache.IDLE = 1;
 
 /**
  * The update has started but the resources are not downloaded yet - for
@@ -1189,10 +1203,23 @@ ApplicationCache.prototype.IDLE = 1;
 ApplicationCache.prototype.CHECKING = 2;
 
 /**
+ * The update has started but the resources are not downloaded yet - for
+ * example, this can happen when the manifest file is fetched.
+ * @const {number}
+ */
+ApplicationCache.CHECKING = 2;
+
+/**
  * The resources are being downloaded into the cache.
  * @const {number}
  */
 ApplicationCache.prototype.DOWNLOADING = 3;
+
+/**
+ * The resources are being downloaded into the cache.
+ * @const {number}
+ */
+ApplicationCache.DOWNLOADING = 3;
 
 /**
  * Resources have finished downloading and the new cache is ready to be used.
@@ -1201,10 +1228,22 @@ ApplicationCache.prototype.DOWNLOADING = 3;
 ApplicationCache.prototype.UPDATEREADY = 4;
 
 /**
+ * Resources have finished downloading and the new cache is ready to be used.
+ * @const {number}
+ */
+ApplicationCache.UPDATEREADY = 4;
+
+/**
  * The cache is obsolete.
  * @const {number}
  */
 ApplicationCache.prototype.OBSOLETE = 5;
+
+/**
+ * The cache is obsolete.
+ * @const {number}
+ */
+ApplicationCache.OBSOLETE = 5;
 
 /**
  * The current status of the application cache.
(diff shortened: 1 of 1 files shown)


externs/browser/html5.js
```diff
@@ -1173,40 +1173,40 @@ DOMApplicationCache.prototype.dispatchEvent = function(evt) {};
...
  */
-DOMApplicationCache.prototype.UNCACHED = 0;
+DOMApplicationCache.UNCACHED = 0;
 
 /**
  * The cache is idle.
- * @type {number}
+ * @const {number}
  */
-DOMApplicationCache.prototype.IDLE = 1;
```
blickly: I'm still not convinced that removing these declarations from the prototype is correct.  Or did you intend to remove that part of the change from this PR?

blickly:
Created internal issue http://b/129496599

externs/browser/html5.js
realityforge: I believe that in this context it is correct. 

The WebIDL for application cache is at [Application cache API](https://html.spec.whatwg.org/multipage/offline.html#applicationcache). The constants are on an interface.

The WebIDL spec definition for [constants](https://heycam.github.io/webidl/#idl-constants) declares:

> Constants are not associated with particular instances of the interface on which they appear. It is language binding specific whether constants are exposed on instances.

In javascript that means that they are specified on the type but are available via instances ... (yay prototype inheritance). Unlike the `PositionError` from geolocation API, the interface is not marked with the `[NoInterfaceObject]` extended attribute which basically means there is a type backing this object.

If you open up a chrome console you should be able to do  run the following which should all be true.

```javascript
ApplicationCache. CHECKING == 2
window.applicationCache.__proto__ == ApplicationCache
```

You should also be able to run the following (indicating constants is available via the instance).

```javacript
window.applicationCache.CHECKING == 2
```

Hmmm ... but actually you are correct that this code will not work inside pure closure environment as the base name should match the spec. One moment and I will update the type from `DOMApplicationCache` to `ApplicationCache`

realityforge: Force pushed the change to address the above mentioned issue.

Sorry about that - I tend to test using downstream artifacts

blickly: Things available on the instance should be on the prototype, right?

I'd also prefer to keep the renaming changes in a separate PR, just since those tend to require cleanups of the existing usages that can sometimes be hard to land.

realityforge: I am not super familiar with the closure type system but in regular javascript this is perfectly acceptable;

```javascript
var X = {};
X.MYCONST = 1;
const instance = Object.create(X);
// All these resolve correctly
console.log(instance.MYCONST)
console.log(instance.__proto__.MYCONST)
console.log(X.MYCONST)
```

The fact that `MYCONST` is available from the instance is just because it is in parent prototype. I just assumed that this would type check correctly in closure compiler. It certainly generates correct code in the downstream libraries (i.e. Elemental2/GWT2.X).

If you don't think that this is the case I can look deeper. However if this is not the case then there is a LOT of externs in this project that are incorrectly defined :/

The renaming will need to happen first ... so I will raise a different PR for that first.

realityforge:
Ok - Once #3305 has been merged I will update this PR to remove naming aspects

externs/browser/html5.js
realityforge: Interestingly the Typescript WebIDL to .d.ts generated generates the properties on both the interface and well as the top level var. See https://github.com/Microsoft/TSJS-lib-generator/blob/master/baselines/dom.generated.d.ts#L1851-L1872

I will poke around and try and figure out exactly what is required for closure over the weekend...

realityforge: Ok - after doing some tests it seems that externs for both are required ;(

So what this does mean that you would need both such as

```javascript
/** @const {number} */
ApplicationCache.UPDATEREADY = 4;
/** @const {number} */
ApplicationCache.prototype.UPDATEREADY = 4;
```

Which is somewhat annoying. It seems that a lot of the `@const` vars fall into this category. Pretty much every property that is defined as const in WebIDL except those on  interfaces with the `[NoInterfaceObject]` extended attribute.

blickly: Yeah, properties available both on an instance and on the constructor add quite a bit of boilerplate to model correctly. Thanks for digging in and investigating.

blickly:
It looks like the renaming parts are still present here.  Can you reassign to me once they are extracted out?

realityforge:
Hi @blickly - the name change is a precursor to this PR as `DOMApplicationCache.UPDATEREADY = 4;` is not valid. So the rename has been extracted into #3305 - once that PR has merged I can rebase this changeset on the repository so that the diff contains the minimal set of changes.

However I am unable to reassign issues at all - I believe that only people with write access to the repository can make assignments.

So once #3305 has been addressed I will ping you again ;)

blickly:
Ah, OK. Makes sense. Thanks!

realityforge:
Hi @blickly - this should be ready to go now

blickly:
Sorry for the delay.
Importing now...

realityforge:
Thanks!

PR #3301, opened 2019-03-28
Define typedefs for callbacks, naming the typedefs as per the spec

realityforge:
The callbacks have also been fixed to return void as declared in the spec.

This PR has been extracted from #3297

### DIFF
--- a/externs/browser/w3c_geolocation.js
+++ b/externs/browser/w3c_geolocation.js
@@ -27,8 +27,18 @@
 function Geolocation() {}
 
 /**
- * @param {function(!GeolocationPosition)} successCallback
- * @param {(function(!GeolocationPositionError)|null)=} opt_errorCallback
+ * @typedef {function(!GeolocationPosition): void}
+ */
+var PositionCallback;
+
+/**
+ * @typedef {function(!GeolocationPositionError): void}
+ */
+var PositionErrorCallback;
+
+/**
+ * @param {PositionCallback} successCallback
+ * @param {PositionErrorCallback=} opt_errorCallback
  * @param {GeolocationPositionOptions=} opt_options
  * @return {undefined}
  */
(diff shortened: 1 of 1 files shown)


blickly:
Created internal issue http://b/129495921

externs/browser/w3c_geolocation.js
```diff
@@ -27,8 +27,18 @@
 function Geolocation() {}
 
 /**
- * @param {function(!GeolocationPosition)} successCallback
- * @param {(function(!GeolocationPositionError)|null)=} opt_errorCallback
+ * @typedef {function(!Position): void}
```
EatingW: I see in you original PR #3297 that you renamed GeolocationPosition to Position. Do you still have a PR for the renaming, since this new name won't be recognized.

realityforge: Not at this stage, this was just stupidity on my part. I hadn't noticed the TravisCI build failure before I generated the PR. Anyhoo I will force push the fix

externs/browser/w3c_geolocation.js
```diff
@@ -27,8 +27,18 @@
...
+var PositionCallback;
+
+/**
+ * @typedef {function(!GeolocationPositionError): void}
+ */
+var PositionErrorCallback;
+
+/**
+ * @param {PositionCallback} successCallback
+ * @param {PositionErrorCallback} opt_errorCallback
```
EatingW: {PositionErrorCallback=}? It should be optional (in the spec, and indicated by the opt_ name)
Same for the one below in function watchPosition.

realityforge: You are completely right. It seems it got lost as I was copying this code back from our integration branch to this PR.

So sorry for wasting your time.

Force pushed a hopefully correct patch

EatingW:
No problem at all! Sorry for not noticing that in the first review. 

Looks good now! I'll import it for internal testing and review.

realityforge:
Woohoo!

PR #3303, opened 2019-03-28
add ECMASCRIPT6, ECMASCRIPT_2016, ECMASCRIPT_2017 to --language_out options

hanguokai:
add new --language_out options values for wiki Flags and Options, because these options values are fully supported since v20190301.

### DIFF
--- a/src/com/google/javascript/jscomp/CommandLineRunner.java
+++ b/src/com/google/javascript/jscomp/CommandLineRunner.java
@@ -685,8 +685,8 @@ public class CommandLineRunner extends
         name = "--language_out",
         usage =
             "Sets the language spec to which output should conform. "
-                + "Options: ECMASCRIPT3, ECMASCRIPT5, ECMASCRIPT5_STRICT, "
-                + "ECMASCRIPT_2015, STABLE")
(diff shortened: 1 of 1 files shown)


src/com/google/javascript/jscomp/CommandLineRunner.java
```diff
@@ -685,8 +685,8 @@ private static DependencyMode toDependencyMode(DependencyModeFlag flag) {
         name = "--language_out",
         usage =
             "Sets the language spec to which output should conform. "
-                + "Options: ECMASCRIPT3, ECMASCRIPT5, ECMASCRIPT5_STRICT, "
-                + "ECMASCRIPT_2015, STABLE")
+                + "Options: ECMASCRIPT3, ECMASCRIPT5, ECMASCRIPT5_STRICT, ECMASCRIPT6, "
```
blickly: Since ES6 is just an old synonym for ES2015, should we omit that one here?

blickly:
Created internal issue http://b/129495613

src/com/google/javascript/jscomp/CommandLineRunner.java
tjgq: Since the requested changes are trivial, I'll import the PR and do the fix internally.

PR #3305, opened 2019-03-28
Rename DOMApplicationCache to ApplicationCache to match spec

realityforge:
Renaming to the name that matches the spec will allow the placement of static values on the type


realityforge:
This is an extraction of the renaming component of #3300

blickly:
Internal issue created http://b/129795862

realityforge:
🎉

PR #3309, opened 2019-04-01
Expose DiagnosticGroups.forName(name) as a static method

realityforge:


blickly:
Internal issue created http://b/129796346

brad4d:
@realityforge I don't have a problem merging this PR, but I presume that this is pre-work for solving some problem. Could you add some explanation here, hopefully with links, to explain why you want this change?

realityforge:
I am away from my development computer atm so this may lack in specifics but essentially [google/jsinterop-generator](https://github.com/google/jsinterop-generator) uses closure compiler programatically to read externs. If closure compiler detects an error it should fail but does not atm (See google/jsinterop-generator#28) and this resulted in some errors in downstream projects (i.e.  [google/elemental2](https://github.com/google/jsinterop-elemental2) that went undetected (Mostly as elemental2 patches and selectively imports externs).

To fix this I need to enable new warnings to pick up problems (in this specific case it is both an unknown type error as well as an `@override` being present without method on parent type and probably more when I get into it)

Currently I have to do something like

```java
    CompilerOptions options = new CompilerOptions();
     options.setWarningLevel(new DiagnosticGroups().forName( "checkTypes" ), CheckLevel.ERROR);
```

which is just cumbersome. Exposing the method as static seems reasonable as it is a static concept and the https://github.com/google/closure-compiler/commit/51fd0675a180c1b40f983c600b18131f7030e610#diff-657416cab37a79f6ee2356a5bb860d96L105 change does something similar to the getRegisteredGroups() 

HTH

brad4d:
@realityforge Thanks for the explanation.

PR imported and sent for internal testing and review.

brad4d:
Submitted internally.
Should be in tomorrow's push to GitHub.

PR #3310, opened 2019-04-01
Explicitly type the contents of the arrays to match the implements declarations

realityforge:


realityforge:
I am trying to figure out why the Array extern is not typed as in this PR. With such a fundamental type I presume there must be a reason but I can't seem to figure it out.

blickly:
I'm guessing that it breaks folks using the Array constructor in type-unsafe ways in untyped code.  That said, it would be great to land if it were possible to do so.

blickly:
Internal issue created http://b/129795774

lauraharker:
Like Ben said, this will probably break untyped code, e.g. the following case would warn:
``` javascript
const someArray = new Array(1);
someArray.push('a string');
```

I'm testing this internally now.

lauraharker:
Update:
I ran into trouble testing this because the compiler also defines `Array` (and some other symbols), and changing the externs causes a type error because the externs type is different from the compiler.

I'm trying again today to 
1. update the compiler
2. then see how many breakages that causes inside Google, including the compiler fix.

This will be harder to land than I thought, though, even if we don't have to fix a lot of code. Inside Google, some projects use a versioned compiler, but all projects use the head externs. We'd have to find a way to not break those projects until the next compiler release.

lauraharker:
Another problem I just ran into is the one-arg case `new Array(lengthOfArray)`. 

We don't want to infer the return type as `Array<number>`; that would break code assigning the array to an `Array<SomeOtherType>`.

lauraharker:
@realityforge out of a random sample of 1000 projects in Google, this change broke 580 with 45 unique errors messages. Most of these were related to people initializing their array with a default size.

It looks like typing `new Array(3)` is a blocker for landing this change. Any suggestions? Otherwise, I'll close this PR.

realityforge:
I am away from my development machine atm but is it valid to type the constructor parameters using a union ala `@param {number|...T} var_args`?

lauraharker:
Afraid not, you could write
``` js
@param {number|T=} possibleLength
@param {...T} var_args
```

but T is still inferred to be `number` in `new Array(0)`.

realityforge:
I had a poke around and could not figure out any solution .. so I will close it. If anything jumps out in the future I will try again ;)

realityforge:
Thanks for taking the time to look into it.

PR #3312, opened 2019-04-02
Remove values from externs

realityforge:
See #3293 for the reasoning but the short version is that closure
compiler does not use the values and reads the values at runtime.
If any tool were to start using it and the values were incorrect,
then incorrect behaviour will occur.

FWIW, while stripping the values I found that there is at least
one value (XPathException.INVALID_EXPRESSION_ERR) that was incorrectly
defined.


blickly:
Created internal issue http://b/129795868

blickly:
@jDramaix, it looks like importing this causes a conflict with some externs patching that elemental2 is doing.  Do you have any idea how to avoid that?

realityforge:
FWIW, my medium term goal is to try and minimize he amount of patching done by elemental2 by slowly evolving closures externs so it is not needed. It seems like it is unlikely to eliminate the need for patching but hopefully the incidence of conflicts will be reduced.

jDramaix:
@blickly I commented in your internal cl

PR #3314, opened 2019-04-04
Update jquery-3.3.js jQuery.noConflict() type

ComputerJy:
Having return type Object for the noConflict function causes Closure compiler to warn about object expressions not being callable.


ComputerJy:
I signed it

brad4d:
Internal issue http://b/129955426 created.

brad4d:
Imported and sent for internal review and approval.

brad4d:
@ComputerJy I've pushed back to your branch a simplified form of your change.
Could you confirm that it works for your use case before I submit it on my side?

Thanks.

ComputerJy:
> @ComputerJy I've pushed back to your branch a simplified form of your change.
> Could you confirm that it works for your use case before I submit it on my side?
> 
> Thanks.

This works just fine. Thanks

brad4d:
I'm submitting this change internally. It should be included in tomorrow's push to GitHub.

PR #3316, opened 2019-04-08
Improve w3c_batterystatus.js

realityforge:
With this PR I split the commits into small targeted commits. I am hoping this makes it easier to consume. If not I can split the PR or squash the commits.

### DIFF
--- a/externs/browser/gecko_dom.js
+++ b/externs/browser/gecko_dom.js
@@ -92,12 +92,6 @@ Window.prototype.frames;
 Window.prototype.fullScreen;
 
 /**
- * @return {!Promise<!BatteryManager>}
- * @see http://www.w3.org/TR/battery-status/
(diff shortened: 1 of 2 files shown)


realityforge:
It should be noted that I assume that when this lands in closure, then google/elemental2#108 will need to be simultaneously applied.

brad4d:
Created internal issue http://b/130163668

brad4d:
@jDramaix it looks like we would need to coordinate this PR and google/elemental2#108 to avoid breaking things.
Are you interested in doing that now?

jDramaix:
yep let coordinate. I've commented in the internal buganizer ticket.

brad4d:
With help from @jDramaix I've imported a combination of the two PRs and sent the resulting change for testing.

PR #3330, opened 2019-04-11
Migrate standardized History API to html5.js from ie_dom.js

realityforge:
IE10 and IE11 also support passing a string parameter as a parameter
to History.prototype.go function and passing an optional delta parameter
to History.prototype.back both of which are easily replaced with standards
equivalents (i.e. using "window.location.href = ..." or "history.go(delta)").

The externs for these non-standard parameters has been removed as part
of this PR as it seems very unlikely that there is limited if any use
of these parameters any more.

### DIFF
--- a/externs/browser/html5.js
+++ b/externs/browser/html5.js
@@ -3100,8 +3100,42 @@ WebSocket.prototype.binaryType;
 // HTML5 History
 /**
  * @constructor
+ * @see http://w3c.github.io/html/browsers.html#the-history-interface
  */
 function History() {}
+
+/**
+ * Goes back one step in the joint session history.
+ * If there is no previous page, does nothing.
+ *
+ * @return {undefined}
+ */
+History.prototype.back = function(opt_distance) {};
+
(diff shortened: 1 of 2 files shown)


lauraharker:
Created internal issue http://b/130353696

externs/browser/html5.js
```diff
@@ -3100,9 +3100,43 @@ WebSocket.prototype.binaryType;
...
+/**
+ * Goes forward one step in the joint session history.
+ * If there is no next page, does nothing.
+ *
+ * @return {undefined}
+ */
+History.prototype.forward = function() {};
+
+/**
+ * Returns the number of entries in the joint session history.
```
lauraharker: Suggested rephrasing: `The number of entries in the joint session history.`, as this is not a function.

realityforge: Done and force pushed the change. 

FWIW the text was an exact copy of the comments from the spec but it does make sense to use this terminology.

PR #3331, opened 2019-04-12
Collapse @type annotation into @const annotation where present

realityforge:
This change just simplifies the type annotations and should have no behavioural impact


lauraharker:
Created internal issue http://b/130418786

PR #3332, opened 2019-04-13
Only mark a module as CommonJS when an export statement is encountered

### LINKED ISSUE #3324: Error: Modules now detected as both CommonJS and ES6
As of the latest release, my projects are getting multiple errors such as:

`A file cannot be both a CommonJS module and an ES6 module.`

git bisect points to 0393c80ca01b6b861376dad7f91043a38bb37dc0

ChadKillingsworth:
The CommonJS "require" call can be used as a module loader from a script tag. The presence of a "require" call does not indicate that the file is a CommonJS module.

Fixes #3324


lauraharker:
Created internal issue http://b/130555143

PR #3333, opened 2019-04-13
Fail gracefully when a source map path contains invalid characters

ChadKillingsworth:
Prevents the compiler from crashing when a source map path contains invalid characters for the current os filesystem.


lauraharker:
Created internal issue https://b/130553999

PR #3334, opened 2019-04-15
Cleanup nullability annotations

realityforge:
* Remove ! if type is nonnull by default

### DIFF
--- a/externs/browser/window.js
+++ b/externs/browser/window.js
@@ -134,20 +134,20 @@ function alert(message) {}
 function clearImmediate(immediateID) {}
 
 /**
- * @param {number|undefined?} intervalID
+ * @param {number=} opt_intervalID
  * @see https://developer.mozilla.org/en/DOM/window.clearInterval
  * @suppress {duplicate}
  * @return {undefined}
  */
-function clearInterval(intervalID) {}
+function clearInterval(opt_intervalID) {}
 
 /**
- * @param {number|undefined?} timeoutID
+ * @param {number=} opt_timeoutID
  * @see https://developer.mozilla.org/en/DOM/window.clearTimeout
  * @suppress {duplicate}
  * @return {undefined}
  */
-function clearTimeout(timeoutID) {}
+function clearTimeout(opt_timeoutID) {}
 
 /**
  * @param {*} message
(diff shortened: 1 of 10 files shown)


lauraharker:
Created internal issue http://b/130555289

lauraharker:
Google's style guide mandates that all non-primitive or literal type references must have an explicit `!` or `?`.
https://google.github.io/styleguide/jsguide.html#jsdoc-nullability

I'd be happy to see a few of the changes in this PR land, like `!number` to `number` or `!function(...` to `function(...`, but we can't accept this as is.

@realityforge could you revert the changes to non-primitive or literal types?

realityforge:
@lauraharker I believe it is good to go. Holler if there is something more that needs to be done.

externs/browser/window.js
```diff
@@ -134,15 +134,15 @@ function alert(message) {}
 function clearImmediate(immediateID) {}
 
 /**
- * @param {number|undefined?} intervalID
+ * @param {number|?undefined} intervalID
```
lauraharker: `?undefined` should just be `undefined`.

externs/browser/window.js
```diff
@@ -134,15 +134,15 @@ function alert(message) {}
...
  * @see https://developer.mozilla.org/en/DOM/window.clearInterval
  * @suppress {duplicate}
  * @return {undefined}
  */
 function clearInterval(intervalID) {}
 
 /**
- * @param {number|undefined?} timeoutID
+ * @param {number|?undefined} timeoutID
```
lauraharker: Same here

externs/browser/window.js
realityforge: I was under the impression that this would change the semantics. i.e. `?undefined` allowed null while `undefined` just allowed the omission of parameter. However looking at the spec for [WindowOrWorkerGlobalScope](https://html.spec.whatwg.org/multipage/webappapis.html#windoworworkerglobalscope) it seems that it is probably better represented by `@param {number=} opt_intervalID` ... one moment

realityforge:
@lauraharker Alright ... next attempt ;)

externs/browser/window.js
lauraharker: Oops, you're right about `?undefined`. Thanks for noticing it's optional.

lauraharker:
Looks good to me. Importing for internal review.

realityforge:
Thanks!

PR #3335, opened 2019-04-15
Avoid accessing static method via instance

realityforge:


lauraharker:
Created internal issue http://b/130555041

Issue #3212, opened 2019-01-25
Optimization of Array.push()


dan1wang:
After I reported Issue #3211, I discovered Closure Compiler (ADVANCED_OPTIMIZATIONS) also doesn't optimize push().

I had
```
      obj.arrayProp.push( offsetSegment(mask, offsetX, 0) );
      obj.arrayProp.push( offsetSegment(mirroredMask, offsetX, 0) );
```
and I get
```
a.f.push(offsetSegment(c,d,0)),
a.f.push(offsetSegment(h,d,0)),
```
instead of 
`a.f.push(offsetSegment(c,d,0),offsetSegment(h,d,0)),`

tjgq:
Created Google internal issue b/123406223

ChadKillingsworth:
This would be a pretty good first contribution. It can be done with a peephole optimization.

cshung:
I wanted to attempt this one.

brad4d:
Once I looked more closely at the PR proposed to fix this issue, I realized it wasn't really an optimization we should add.

From the discussion on that PR:

> Overall this optimization doesn't strike me as a very useful one.
> It's not likely that a human author would put a lot of array.push(), right next to each other,
> so I'd think this would only be useful if other optimizations ended up putting them together, which
> then allowed this optimization to do something useful. That also seems likely to be a rare thing to me.
> 
> Execution time is precious for large projects.

Issue #3211, opened 2019-01-25
Optimization of Array.concat()


dan1wang:
It seems Closure Compiler (ADVANCED_OPTIMIZATIONS) is not very good at Array concatenate.

If I have the following (formatted for readability):
```
const template4 = [].concat(
        offsetSegment(array1, r1x, r1y),
        offsetSegment(array2, r4x, r4y),
        offsetSegment(array3, 0, obj.a),
        array4);
```

I would get
`  f=[].concat(offsetSegment(f,g,d),offsetSegment(m,g,n),offsetSegment(h,0,-a.b),k);`
instead of 
`  f=offsetSegment(f,g,d).concat(,offsetSegment(m,g,n),offsetSegment(h,0,-a.b),k);`

Similarly, if I have 
```
const template3 =
       array1
       .concat(offsetSegment(array2, r2x, r2y))
       .concat(offsetSegment(array2, r4x, r4y))
       .concat(offsetSegment(array3, 0, obj.a));
```

I would get something like
`  b=t.concat(offsetSegment(p,e,f)).concat(offsetSegment(m,n,o)).concat(offsetSegment(r,s,a.a,))`

In my externs file, I declared the various array functions with the following
```
/**
 * @nosideeffects
 * @param {Array} seg
 * @return {Array.<Array.<number>>}
 */
```

tjgq:
Optimizing `[].concat(f(), x)` into `f().concat(x)` is only safe if `f` is guaranteed to return a fresh array every time it's called. `@nosideffects` is not enough to assert that: a side effect-free function may still, for example, return a reference to a non-local variable.

dan1wang:
shouldn't x1 and x2 be the same regardless of whether f1() returns a fresh array?
```
x1 =[].concat(f1(), f2(), f3()...)
x2= f1().concat(f2(), f3(), ...)
```
If the code has a bug, wouldn't optimization just preserve the bug and not introduce a new one?

brad4d:
@dan1wang you're right.
When I chatted with @tjgq about this I forgot that `concat` always creates a new array.

So, I suppose it should be safe to save 3 bytes by dropping the initial `[].` in such cases.
I feel certain that I'm forgetting some other concern that leads to us keeping it, though.

Feel free to take a look at the code if you're interested in tweaking this.
I don't think the closure-compiler team is likely to make it a priority, though.

tjgq:
[...]

PR #3338, opened 2019-04-18
Introducing PeepholeCombineArrayPush

cshung:
Fixes issue #3212

### DIFF
--- a/src/com/google/javascript/jscomp/DefaultPassConfig.java
+++ b/src/com/google/javascript/jscomp/DefaultPassConfig.java
@@ -1726,6 +1726,7 @@ public final class DefaultPassConfig extends PassConfig {
     }
     optimizations.add(new PeepholeFoldConstants(late, useTypesForOptimization));
     optimizations.add(new PeepholeCollectPropertyAssignments());
+    optimizations.add(new PeepholeCombineArrayPush());
     return new PeepholeOptimizationsPass(compiler, passName, optimizations);
   }
 
--- a/src/com/google/javascript/jscomp/DefaultPassConfig.java
+++ b/src/com/google/javascript/jscomp/DefaultPassConfig.java
@@ -1726,7 +1726,9 @@ public final class DefaultPassConfig extends PassConfig {
     }
     optimizations.add(new PeepholeFoldConstants(late, useTypesForOptimization));
     optimizations.add(new PeepholeCollectPropertyAssignments());
-    optimizations.add(new PeepholeCombineArrayPush());
+    if (useTypesForOptimization) {
+      optimizations.add(new PeepholeCombineArrayPush());
+    }
     return new PeepholeOptimizationsPass(compiler, passName, optimizations);
   }
 
--- a/src/com/google/javascript/jscomp/PeepholeCombineArrayPush.java
+++ b/src/com/google/javascript/jscomp/PeepholeCombineArrayPush.java
@@ -40,48 +40,57 @@ final class PeepholeCombineArrayPush extends AbstractPeepholeOptimization {
     }
 
     boolean codeChanged = false;
-
-    for (Node child = subtree.getFirstChild(); child != null; child = child.getNext()) {
-      if (isArrayDotPush(child)) {
-        Node next = child.getNext();
-        if (isArrayDotPush(next)) {
-          Node firstCall = child.getFirstChild();
-          Node secondCall = next.getFirstChild();
-          String firstArrayName = firstCall.getFirstChild().getFirstChild().getString();
-          String secondArrayName = secondCall.getFirstChild().getFirstChild().getString();
-          if (firstArrayName.equals(secondArrayName)) {
-            Node secondArgument = secondCall.getFirstChild().getNext();
-            while (secondArgument != null) {
-              Node temp = secondArgument.getNext();
-              firstCall.addChildToBack(secondArgument.detach());
-              secondArgument = temp;
-            }
-            subtree.removeChild(next);
-            codeChanged = true;
-          }
-        }
+    for (Node statement = subtree.getFirstChild(); statement != null; statement = statement.getNext()) {
+      ArrayDotPushStatementSequence arrayDotPushStatementSequence = asArrayDotPushStatementSequence(statement);
+      if (arrayDotPushStatementSequence != null) {
+        arrayDotPushStatementSequence.combine(subtree);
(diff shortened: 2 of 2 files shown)


tjgq:
Created Google internal issue b/130801558.

Assigning to @brad4d for review.

src/com/google/javascript/jscomp/PeepholeCombineArrayPush.java
```diff
@@ -0,0 +1,93 @@
...
+
+  @Override
+  Node optimizeSubtree(Node subtree) {
+    if (!subtree.isScript() && !subtree.isBlock()) {
+      return subtree;
+    }
+
+    boolean codeChanged = false;
+
+    for (Node child = subtree.getFirstChild(); child != null; child = child.getNext()) {
```
brad4d: Rather than `child`, let's call the loop variable `statement` to make it a bit clearer what it represents.

src/com/google/javascript/jscomp/PeepholeCombineArrayPush.java
```diff
@@ -0,0 +1,93 @@
...
+      if (isArrayDotPush(child)) {
```
brad4d: Since `isArrayDotPush()` just returns a boolean, we end up having to re-fetch nodes that we already found in that method. If you instead defined a method that returned an object containing exactly the information you need you could simplify this logic.

```java
for (Node statement = subtree.getFirstChild(); statement != null; statement = statement.getNext()) {
  ArrayDotPushStatementSequence arrayDotPushStatement = asArrayDotPushStatementSequence(statement);
  if (arrayDotPushStatementSequence != null) {
    // This always combines following statements into the current one,
    // so statement.getNext() will still move to the new next statement when this is done.
    arrayDotPushStatementSequence.combine();
    codeChanged = true;
  }
}
```

`asArrayDotPushStatementSequence()` would know how to pull out all the nodes needed for the current statement and any following ones that can be combined and store them in the object it returns.

src/com/google/javascript/jscomp/PeepholeCombineArrayPush.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    if (codeChanged) {
+      reportChangeToEnclosingScope(subtree);
+    }
+    return subtree;
+  }
+
+  private boolean isArrayDotPush(Node child) {    
+    if (child != null && child.getToken() == Token.EXPR_RESULT) {
```
brad4d: child.isExprResult();

src/com/google/javascript/jscomp/PeepholeCombineArrayPush.java
```diff
@@ -0,0 +1,93 @@
...
+      Node callNode = child.getFirstChild();
+      if (callNode != null && callNode.getToken() == Token.CALL) {
```
brad4d: callNode.isCall()

src/com/google/javascript/jscomp/PeepholeCombineArrayPush.java
```diff
@@ -0,0 +1,93 @@
...
+
+    if (codeChanged) {
+      reportChangeToEnclosingScope(subtree);
+    }
+    return subtree;
+  }
+
+  private boolean isArrayDotPush(Node child) {    
+    if (child != null && child.getToken() == Token.EXPR_RESULT) {
+      Node callNode = child.getFirstChild();
```
brad4d: statement.getOnlyChild()
EXPR_RESULT only ever has one child, so it's nice to assert this

src/com/google/javascript/jscomp/PeepholeCombineArrayPush.java
```diff
@@ -0,0 +1,93 @@
...
+      if (callNode != null && callNode.getToken() == Token.CALL) {
+        Node arrayDotPushNode = callNode.getFirstChild();
+        if (arrayDotPushNode != null && arrayDotPushNode.getToken() == Token.GETPROP) {
```
brad4d: isGetProp()

src/com/google/javascript/jscomp/DefaultPassConfig.java
```diff
@@ -1726,6 +1726,7 @@ private static CompilerPass createPeepholeOptimizationsPass(
     }
     optimizations.add(new PeepholeFoldConstants(late, useTypesForOptimization));
     optimizations.add(new PeepholeCollectPropertyAssignments());
+    optimizations.add(new PeepholeCombineArrayPush());
```
brad4d: Since this optimization requires type information to work, don't add it to the list unless `useTypesForOptimization` is `true`.

brad4d:
Overall this optimization doesn't strike me as a very useful one.
It's not likely that a human author would put a lot of `array.push()`, right next to each other,
so I'd think this would only be useful if other optimizations ended up putting them together, which
then allowed this optimization to do something useful. That also seems likely to be a rare thing to me.

Execution time is precious for large projects. Do you have a use-case in which the code size savings from this optimization will justify the time spent in looking for places where it can be done?

cshung:
> Overall this optimization doesn't strike me as a very useful one.

I agree. I was hunting for an issue to contribute to a compiler related project, and this one seems like a challenge that I would like to get started on, so I did it. It was marked help wanted and was commented as a good first issue, so I took it on, without a second thought. In retrospect, I probably should have considered the cost/benefit for the optimization.

> Do you have a use-case in which the code size savings from this optimization will justify the time spent in looking for places where it can be done?

@dan1wang, can you help us to understand why do we want this optimization?

dan1wang:
I think the optimization is useful only if the original array is pretty large. I don't know of a use-case where that's the case (I did find this ugly code, though: https://github.com/nemac/seldon/blob/master/areas.js).

I think it would be more worthwhile to address issue #3211 (`string.concat` & `array.concat`) because it's more likely one of the string/array is fairly large.

brad4d:
Sounds like we've agreed to close this PR.
@cshung thanks for submitting it anyway.
Hopefully you still found the experience a valuable one and can apply it to creating another?
:)

cshung:
> Hopefully you still found the experience a valuable one and can apply it to create another?

@brad4d Can you please recommend an issue that is good?

brad4d:
@cshung As it happens I've just been looking at #3343 which would probably be a good first PR.

PR #3341, opened 2019-04-19
Remove @const annotation from method

realityforge:
AFAICT this does not really make any sense ... although I could be missing something


tjgq:
Created Google internal issue b/130887109

PR #3342, opened 2019-04-19
Prefer optional parameter syntax rather than allowing undefined param

realityforge:


tjgq:
Created Google internal issue b/130887990

PR #3344, opened 2019-04-22
Move the Location externs from ie_dom.js to html5.js

realityforge:
The Location interface has been standardized and can thus be moved to
the standardized externs. This means that google/elemental2 will
no longer need to to have local patche to include these externs.

The @see references have been updated to point at the new spec and the
documentation has been copied from the Elemental2 externs patch which was
ultimately sources from the spec.

The one type change is that `function Location()` was previously
annotated with @constructor but is now annotated with @interface to both
match the spec and generate better code in google/elemental2. Given that
this is not an object that a end-user can instantiate it is not expected
to have any impact.


tjgq:
Created Google internal issue b/131089637, importing for review and testing.

externs/browser/html5.js
```diff
@@ -3180,6 +3180,121 @@ History.prototype.scrollRestoration;
  */
 Window.prototype.history;
 
+/**
+ * @interface
```
tjgq: FYI, this needs to be a constructor because it's an actual property of `window` and we need to be able to write `x instanceof Location`.

I will make the change internally to avoid the additional roundtrip.

realityforge:
Thanks

PR #3347, opened 2019-04-24
Move nonstandard property to a separate extern

realityforge:
This makes it possible to import the extern and only receive the standardized externs which stops downstream projects such as google/elemental2 from patching the extern.

This needs to be simultaneously applied with google/elemental2#113

### DIFF
--- a/externs/browser/nonstandard_navigation_timing.js
+++ b/externs/browser/nonstandard_navigation_timing.js
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2011 The Closure Compiler Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


blickly:
Internal issue created http://b/131363422

tjgq:
My understanding of https://www.w3.org/TR/navigation-timing is that the only standardized bits are `performance.timing` and `performance.navigation`. Should we also move the other nonstandard `Performance.prototype` properties defined in `w3c_navigation_timing.js`?

realityforge:
Hmm .. poking about the different standards that extend `Performance` I think that all of the methods besides those prefixed with `webkit` are part of a standard
* https://w3c.github.io/navigation-timing/#extensions-to-the-performance-interface
* https://w3c.github.io/performance-timeline/#extensions-to-the-performance-interface
* https://w3c.github.io/resource-timing/#sec-extensions-performance-interface
* https://w3c.github.io/user-timing/#extensions-performance-interface

One good way to figure these out is to look at the idl from the browsers. i.e. https://github.com/mozilla/gecko-dev/blob/master/dom/webidl/Performance.webidl 

I moved the `webkit` prefixed methods to non standard externs and fixed up a few other differences from the standard.

tjgq:
Thanks for checking. Importing for review and testing.

PR #3349, opened 2019-04-24
Improve the typing of the callback parameter for the timer APIs

realityforge:
This needs to be applied at the same time as google/elemental2#114

### DIFF
--- a/externs/browser/window.js
+++ b/externs/browser/window.js
@@ -144,22 +144,38 @@ function prompt(message, opt_value) {}
 function setImmediate(callback, callbackParams) {}
 
 /**
- * @param {Function|!TrustedScript|string} callback
+ * A function to invoke when a timer expires.
+ *
+ * NOTE: The typing is explicitly loose in order to allow a function with any signature (params and return type) to be
+ * used. Having an explicit typedef for use in the declarations of the timer methods below instead of just using
+ * Function there helps Elemental2 to do a better job synthesizing Java code for this interface.
+ *
+ * @typedef {Function}
+ */
+var TimerCallbackFunction;
+
+/**
+ * @typedef {!TimerCallbackFunction|!TrustedScript|string}
+ */
+var TimerCallback;
+
+/**
+ * @param {?TimerCallback} callback
  * @param {number=} opt_delay
- * @param {...?} callbackParams
+ * @param {...*} callbackParams
  * @return {number}
  * @see https://developer.mozilla.org/en/DOM/window.setInterval
- * @see https://html.spec.whatwg.org/multipage/webappapis.html#timers
+ * @see https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#timers
  */
 function setInterval(callback, opt_delay, callbackParams) {}
 
 /**
- * @param {Function|!TrustedScript|string} callback
+ * @param {?TimerCallback} callback
  * @param {number=} opt_delay
  * @param {...*} callbackParams
  * @return {number}
  * @see https://developer.mozilla.org/en/DOM/window.setTimeout
- * @see https://html.spec.whatwg.org/multipage/webappapis.html#timers
+ * @see https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#timers
  */
 function setTimeout(callback, opt_delay, callbackParams) {}
 


externs/browser/window.js
```diff
@@ -191,9 +191,9 @@ function setImmediate(callback, callbackParams) {}
 function setInterval(callback, opt_delay, callbackParams) {}
 
 /**
- * @param {Function|!TrustedScript|string} callback
+ * @param {function(...?):undefined|!TrustedScript|string} callback
  * @param {number=} opt_delay
- * @param {...*} callbackParams
```
blickly: Why this loosening?

blickly:
Created internal issue http://b/131364248

externs/browser/window.js
realityforge: The short answer is that, this is how it appeared in Elemental2 patch. The longer answer is that I was trying to make it consistent with the typing of the parameters passed to the callback function as well as consistent with the other timing related functions (i.e. `setInterval` and `setImmediate`) which are both typed as `{...?}`.

In the context of a parameter accepted by a function defined in an extern I did not think it would have any impact on the code that type checks. Although I have been wrong in the past ;-)

realityforge: Also I believe it has an impact on the code generated in elemental2 ... and moving it to `{...?}` means we could eliminate the patch altogether.

brad4d: Using `*` vs. `?` makes no difference in type checking of calls made to this method.
It will make a difference if you ever try to define a method that is supposed to implement this API.
If you use `*`, then that implementation won't be allowed to access any properties on the `callbackParams` items,
but it will if you use `?`.

`*` is the more correct choice here, since `setTimeout` shouldn't do anything with the `callbackParams` other than pass them to the `callback` function. Picking the wrong one may not break anything now, but could bite someone later on.

Regardng the patch, can't you still eliminate the patch by changing the elemental2 side to expect/define `*` instead of `?`?

externs/browser/window.js
```diff
@@ -191,9 +191,9 @@ function setImmediate(callback, callbackParams) {}
 function setInterval(callback, opt_delay, callbackParams) {}
 
 /**
- * @param {Function|!TrustedScript|string} callback
+ * @param {function(...?):undefined|!TrustedScript|string} callback
```
blickly: Investigated a bit more, and apparently there are places that pass in a nullable type here, which this patch forbids.

Example (inside base.js!):
javascript/closure/base.js:2927: ERROR - actual parameter 1 of global this.setTimeout does not match formal parameter
found   : (function(): undefined|null)
required: (TrustedScript|function(...?): undefined|string)
        goog.global.setTimeout(cb, 0);
                               ^^

externs/browser/window.js
realityforge: I moved some code around within the extern (essentially introducing appropriate typedefs) and I believe `*` now works and I also applied this typing to `setInterval`. 

Hopefully that addresses your concerns.

I must admit I never even considered downstream code implementing the code defined by externs ;)

externs/browser/window.js
realityforge: Hmm ... I had not considered that `Function` was nullable. Certainly the browser allows passing null. I will make it so by extracting another typedef and prefixing it with `?`. Hopefully that addresses this.

externs/browser/window.js
```diff
@@ -181,22 +181,32 @@ function prompt(message, opt_value) {}
 function setImmediate(callback, callbackParams) {}
 
 /**
- * @param {Function|!TrustedScript|string} callback
+ * @typedef {function(...*):undefined}
+ */
+var TimerCallbackFunction;
```
blickly: This is just a style nit, but I don't find this particular typedef worth it if it's just used in the one place below.

realityforge: This is true but introducing this typedef generates much better code in Elemental2. With a typedef, Elemental2 will generate a java interface named `elemental2.dom.TimerCallbackFunction`. Without the typedef it generates multiple synthetic types based on the usage location. In this case it would generate `elemental2.dom.DomGlobal.SetIntervalCallbackFn` even when passing to `setTimeout` function. And if we reordered the definitions in the extern it would generate different code (i.e. `elemental2.dom.DomGlobal.SetTimeoutCallbackFn`). Having a stable name is a much better developer experience IMHO.

However if this needs to be change to get the PR accepted, I can do it.

externs/browser/window.js
```diff
@@ -181,22 +181,32 @@ function prompt(message, opt_value) {}
 function setImmediate(callback, callbackParams) {}
 
 /**
- * @param {Function|!TrustedScript|string} callback
+ * @typedef {function(...*):undefined}
```
blickly: An even more basic issue is that tightening up callback type is catching lots of new errors, which aren't immediately clear if they are valid or not.  For example, in closure-library, we have errors like:

closure/goog/module/modulemanager.js:646: ERROR - inconsistent return type
found   : Error
required: undefined
          () => new Error('Module already loaded: ' + id), 0);
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

closure/goog/module/modulemanager.js:820: ERROR - actual parameter 1 of Window.setTimeout does not match formal parameter
found   : function(*): undefined
required: (TrustedScript|function(...*): undefined|null|string)
      window.setTimeout(goog.bind(callbackWrapper.execute, callbackWrapper), 0);
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

I somewhat suspect that the new type is to restrictive.  Can you investigate?

realityforge: I will investigate over the weekend

realityforge: The first issue highlighted seems to be a bug in `modulemanager.js`. Looking at the [goog.module.ModuleManager.prototype.processModulesForLoad_](https://github.com/google/closure-library/blob/0f3a8a51bb4897086642bf9934270cd3d6a23ff2/closure/goog/module/modulemanager.js#L638-L647) I believe that they are trying to throw an error on the next task. Similar code exists at [goog.async.throwException](https://github.com/google/closure-library/blob/2159b6717c6b6869a780eafc6ae0fb0737599f40/closure/goog/async/nexttick.js#L43-L46). In this scenario I believe the type checker has identified a bug as the code as it stands is effectively a noop.

Looking into the second scenario now ...

realityforge: The second scenario is because I did not understand the typing system correctly it seems. I had assumed that `function(...*):undefined` would accept a function typed with `function():undefined` but that seems not to be the case. I ended up having to add both `function(...*):undefined` and `function():undefined` into type union. (Although it is possible that I am doing something else wrong here...).

Anyhoo the test code I experiemented with was

**foo.extern.js:**
```javascript
/**
 * @fileoverview BLAH
 * @externs
 */

/**
 * @typedef {function(...*):undefined}
 */
var MyTimerCallbackFunction;

/**
 * @typedef {function():undefined}
 */
var MySimpleCallbackFunction;

/**
 * @typedef {!MyTimerCallbackFunction|!TrustedScript|string|!MySimpleCallbackFunction}
 */
var MyTimerCallback;

/**
 * @param {?MyTimerCallback} callback
 * @param {number=} opt_delay
 * @param {...*} callbackParams
 * @return {number}
 */
function mySetTimeout(callback, opt_delay, callbackParams) {}

```
**foo.js:**
```javascript
/** @type {function(): undefined} */
const myCallback = () => console.log('hi');

/** @type {function(?): undefined} */
const myCallback2 = (count) => console.log(count);

const foo = {};

const myOtherCallback = myCallback2.bind(foo, 3)

mySetTimeout(myCallback, 0);
mySetTimeout(myCallback2, 0, 33);
mySetTimeout(myOtherCallback, 0);
```

If you remove `MySimpleCallbackFunction` from the extern then `mySetTimeout(myCallback2, 0, 33)` will not type check with.

```
foo.js:12: WARNING - actual parameter 1 of mySetTimeout does not match formal parameter
found   : function(?): undefined
required: (TrustedScript|function(...*): undefined|null|string)
mySetTimeout(myCallback2, 0, 33);
```

It is not entirely clear why it behaves this way ...

jDramaix:
@blickly What is the status of this PR?

externs/browser/window.js
blickly: I'm assuming that the requirement has to do with contravariance and not wanting to allow a function that can only accept one argument to substitute for one that can handle arbitrarily many arguments.  (See http://closuretools.blogspot.com/2012/06/subtyping-functions-without-poking-your.html for some good background)

So I think that adding MySimpleCallbackFunction only hides the issue for the 1-argument case, but not for all the other cases (1-argument, 2-argument, etc).

So I don't see a feasible annotation to put here that is better than !Function, unfortunately.

realityforge: Right. I will rollback to `!Function` and leave this as something that is patched downstream in Elemental2.

Thanks for looking into it.

realityforge:
@blickly Ready for re-review.

realityforge:
@blickly What is the status of this issue?

externs/browser/window.js
brad4d: Could you add a comment to help the reader understand the point of this declaration?
I think it would look like this.

```javascript
/**
 * A function to invoke when a timer expires.
 *
 * NOTE: The typing is explicitly loose in order to allow a function with any signature (params and return type) to be
 * used. Having an explicit typedef for use in the declarations of the timer methods below instead of just using
 * Function there helps Elemental2 to do a better job synthesizing Java code for this interface.
 */
var TimerCallbackFunction;
```

brad4d:
@blickly is OOO for several weeks, so I'll take it on.

I've asked for a comment change.

brad4d:
I've imported this change as-is for the moment so I can run it through internal tests to see if it breaks anything.

realityforge:
@brad4d Thanks for picking this up. Commented change applied and change re-pushed. Thanks!

brad4d:
Got lots of breakages in internal testing.
I think they indicate that there is Java code depending on the existence of the class `elemental2.dom.DomGlobal.SetIntervalCallbackFn` which ends up getting eliminated by this change.

It doesn't make sense for closure-compiler team to drive fixing something like that.
I've added a comment to this effect on the related elemental2 PR https://github.com/google/elemental2/pull/114

If the elemental2 team decides to accept that PR, then they'll be able to make the corresponding closure-compiler change as part of their internal-to-google change, and that makes more sense than us trying to do it, since the purpose of this change is to benefit them and the breakages are all in things related to them.

I'm closing this PR, because there's not really any benefit in keeping it open.
The work it represents can be considered part of https://github.com/google/elemental2/pull/114

realityforge:
Thanks anyway

PR #3350, opened 2019-04-25
Remove live connect externs and navigator global

realityforge:
This needs to be applied at the same time as google/elemental#115

### DIFF
--- a/externs/browser/window.js
+++ b/externs/browser/window.js
@@ -75,44 +75,6 @@ var isSecureContext;
  */
 var visualViewport;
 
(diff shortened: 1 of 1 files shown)


blickly:
Created internal issue http://b/131364800

lauraharker:
There are a few internal projects referencing `java` (although none of the other deleted externs are used). I'll see if we can remove those.

realityforge:
Thanks for looking into it

lauraharker:
I've sent this for internal review.

realityforge:
Thanks!

Issue #3346, opened 2019-04-23
Symbol.toPrimitive is not reserved


blooddy:
input:
```javascript
class A {
  [ Symbol.toPrimitive ]() {
    return 5;
  }
}
console.log( 5 + new A() );
```
output:
```javascript
class a {
  [Symbol.a]() {
    return 5;
  }
}
console.log(5 + new a);
```

tjgq:
Created Google internal issue b/131172228.

This should be fixed by adding an externs declaration for `Symbol.toPrimitive` in https://github.com/google/closure-compiler/blob/master/externs/es3.js#L70. Would you like to send us a PR?

qqingvarqq:
I found that this issue more general
closure-compiler polyfill "Symbol" and part of well-known symbols are inside es3 definition file -> https://github.com/google/closure-compiler/blob/master/externs/es3.js#L64
Including:
``` Javascript
Symbol.iterator;
Symbol.toStringTag;
Symbol.unscopables;  
/* Missing Symbol.toPrimitive */
```
and es6 definition file -> https://github.com/google/closure-compiler/blob/master/externs/es6.js#L1771
Including only
``` Javascript
Symbol.asyncIterator;
```
Other well-known Symbols  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol#Well-known_symbols are not defined in the es6 definition file, which causes errors when we use ADVANCED_OPTIMIZATION and transpile to  ECMASCRIPT6;
Input:
``` Javascript
console.log(Symbol.hasInstance);
console.log(Symbol.isConcatSpreadable);
console.log(Symbol.species);

/* closure-compiler warn as, that these Property species never defined on Symbol, but not truncate those properties */
console.log(Symbol.match);
console.log(Symbol.matchAll);
console.log(Symbol.replace);
console.log(Symbol.search);
console.log(Symbol.split);
```
Output:
``` Javascript
'use strict';
console.log(Symbol.a);
console.log(Symbol.b);
console.log(Symbol.c);
console.log(Symbol.match);
console.log(Symbol.matchAll);
console.log(Symbol.replace);
console.log(Symbol.search);
console.log(Symbol.split);
```
[...]

PR #3352, opened 2019-04-26
Update definition for well-known symbols

qqingvarqq:
This PR define well-known Symbols, which is not defined previously, and fix this issue: #3346
Also, I added the @see annotation to previously defined symbols

### DIFF
--- a/externs/es3.js
+++ b/externs/es3.js
@@ -63,13 +63,82 @@ Symbol.keyFor = function(sym) {};
 
 // Well known symbols
 
-/** @const {symbol} */
+/**
+ * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/asyncIterator
+ * @const {symbol}
+ */
+Symbol.asyncIterator;
+
+/**
+ * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/hasInstance
+ * @const {symbol}
+ */
+Symbol.hasInstance;
+
+/**
+ * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/isConcatSpreadable
+ * @const {symbol}
+ */
+Symbol.isConcatSpreadable;
+
+/** 
+ * @const {symbol}
+ * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/iterator
+ */
 Symbol.iterator;
 
-/** @const {symbol} */
+/**
+ * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/match
+ * @const {symbol}
+ */
+Symbol.match;
+
+/**
+ * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/matchAll
+ * @const {symbol}
+ */
+Symbol.matchAll;
+
+/**
+ * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/replace
+ * @const {symbol}
+ */
+Symbol.replace;
+
+/**
+ * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/search
+ * @const {symbol}
+ */
+Symbol.search;
+
+/**
+ * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/species
+ * @const {symbol}
+ */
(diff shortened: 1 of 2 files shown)


externs/es3.js
```diff
@@ -63,13 +63,28 @@ Symbol.keyFor = function(sym) {};
...
+ * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/toStringTag
+ */
 Symbol.toStringTag;
 
-/** @const {symbol} */
+/** 
+ * @const {symbol}
+ * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/toPrimitive
+ */
+Symbol.toPrimitive;
```
blickly: Any reason why this one is being added here rather than to es6.js?

(I'm actually not totally clear why `Symbol.toStringTag` and `Symbol.unscopables` were defined here either, to be honest).

blickly:
Internal issue created http://b/131438461

externs/es3.js
qqingvarqq: To be honest, I put `Symbol.toPrimitive` to es3 file because @tjgq say it's the right place for that Symbol https://github.com/google/closure-compiler/issues/3346#issuecomment-485939158,
And when I asked about other well-known symbols you agreed that es6 definition file it's the right place for other not defined symbols `Symbol.hasInstance, Symbol.isConcatSpreadable ...`
***
I found it out, why those Symbols `Symbol.iterator; Symbol.toStringTag; Symbol.unscopables;` previously are was put in the es3 definition file.
Explanation: in this commit message:
"Move symbol and iterator, etc to es3.js so that Array can be iterable. While these definitions are not part of EcmaScript 3 this is a better solution than forking the externs." commitId: 42037f6292cf151a4fa2886d2b4fbd0fb791b351
***
So, I assume all well-known symbols should go to es3 definition file, instead es6 or you know a better solution?
if you agree with it, I have a question.
This means that `Symbol.asyncIterator` should go to es3 definition file too (because he defined in es6 previously https://github.com/google/closure-compiler/blob/12f092b42fec07b9c1ff115503e00d1eea4a1ac0/externs/es6.js#L1771 
Should I change `@see` link from
https://tc39.github.io/proposal-async-iteration/
 to
 https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/asyncIterator  for this symbol?

blickly: OK, thanks for digging up the discussion.

It seems like even though it's possible to define some of the well-known symbols in es6.js, there are some that must be in es3.js (like Symbol.iterator) and it's easier to find them if we keep them all together.

I don't have a strong opinion one way or the other.  I also agree with you that:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/asyncIterator
is a better reference link for Symbol.asyncIterator.

Please use your best judgement and then ping this thread.

qqingvarqq: Good, I think the best solution is:
1. Put definition for well-known Symbols in the es3 file;
_I agree that is easier to find/move/refactor those definitions if we keep them together._
2. Add a comment in the es6 file, that those definitions (`Symbol, IIterableResult, Iterable, IteratorIterable, Iterator, IteratorIterable`) are placed in the es3 file.
_Because it's very confusing not to find these definitions in the es6 file, and if next ES standards define new properties or well-known symbols it's not clear where to define them._

If you agree I will make appropriate changes to this PR

blickly: Ok, that sounds like a reasonable end state to me.  Thanks!

qqingvarqq: updated
by the way @blickly  if you have some issues which require beginner experience to this project, I would be working on them

brad4d: FYI, I've just created a "good first issue" label for closure-compiler issues, and added a couple of issues to it.

https://github.com/google/closure-compiler/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22

qqingvarqq: Any status update on this PR?

externs/es3.js
```diff
@@ -63,13 +63,82 @@ Symbol.keyFor = function(sym) {};
...
+ * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/species
+ * @const {symbol}
+ */
+Symbol.species;
+
+/**
+ * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/split
+ * @const {symbol}
+ */
+Symbol.split;
```
blickly: For some reason, it looks like this Symbol.split definition creates some optimization backoff in closure library.  I've alerted them to the issue, but I didn't want it to block the other fixes, so I'm going to land all the other definitions first.

PR #3353, opened 2019-04-29
Mark some constants on Node as @const

realityforge:


blickly:
Internal issue created http://b/131631376

PR #3354, opened 2019-04-29
Mark some FileReader constants with @const

realityforge:


blickly:
Internal issue created: http://b/131631762

PR #3355, opened 2019-04-29
Specify void return type for onload and onload callbacks

realityforge:


blickly:
Internal issue created http://b/131631962

PR #3357, opened 2019-04-30
root in IntersectionObserverInit could be nulled

mk0x9:
Per https://w3c.github.io/IntersectionObserver/v2/#intersection-observer-init root could be nulled in case if pointing to the implicit root element.

### DIFF
--- a/externs/browser/intersection_observer.js
+++ b/externs/browser/intersection_observer.js
@@ -107,7 +107,7 @@ var IntersectionObserverCallback;
  *   threshold: (!Array<number>|number|undefined),
  *   delay: (number|undefined),
  *   trackVisibility: (boolean|undefined),
- *   root: (!Element|undefined),
+ *   root: (?Element|undefined),
  *   rootMargin: (string|undefined)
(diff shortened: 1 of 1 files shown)


externs/browser/intersection_observer.js
```diff
@@ -127,7 +127,7 @@ function IntersectionObserver(handler, opt_options) {};
  * The root Element to use for intersection, or null if the observer uses the
  * implicit root.
  * @see https://w3c.github.io/IntersectionObserver/#dom-intersectionobserver-root
- * @const {?Element}
+ * @const {?Element|null}
```
blickly: ?Element means that it includes null, so ?Element|null is redundant.

blickly:
Created internal issue http://b/131717411

externs/browser/intersection_observer.js
mk0x9: Thanks for noting! Reverted it back.

blickly:
Great, looks good! Importing...

PR #3367, opened 2019-05-09
Add a new pass to convert output chunks to ES Modules

ChadKillingsworth:
Here's some early work to convert output chunks to ESModules and use `import` and `export` statements to reference symbols across chunks. It's definitely not ready for main use yet, but I wanted some early feedback on the strategy.

My main concern here is mutability. How do we guard against:

```js
import {a} from './m0.js';
a = 4; // This didn't actually change A in m0, right?
```

### DIFF
--- a/src/com/google/javascript/jscomp/CompilerOptions.java
+++ b/src/com/google/javascript/jscomp/CompilerOptions.java
@@ -3104,9 +3104,14 @@ public class CompilerOptions implements Serializable {
     }
   }
 
+  /**
+   * Indicate whether output js files should be scripts where cross chunk references use
+   * global namespaces or ES modules where cross chunk references utilize import and export
+   * statements.
+   */
   public enum ChunkOutputType {
     GLOBAL_NAMESPACE,
-    ES_MODULES;
+    ES_MODULES
   }
 
   /**
--- a/src/com/google/javascript/jscomp/CompilerOptions.java
+++ b/src/com/google/javascript/jscomp/CompilerOptions.java
@@ -1277,6 +1277,8 @@ public class CompilerOptions implements Serializable {
     this.printConfig = printConfig;
   }
 
+  ChunkOutputType chunkOutputType;
+
   /**
    * Initializes compiler options. All options are disabled by default.
    *
@@ -1412,6 +1414,7 @@ public class CompilerOptions implements Serializable {
     colorizeErrorOutput = false;
     errorFormat = ErrorFormat.FULL;
     externExports = false;
+    chunkOutputType = ChunkOutputType.GLOBAL_NAMESPACE;
 
     // Debugging
     aliasHandler = NULL_ALIAS_TRANSFORMATION_HANDLER;
@@ -3099,6 +3102,11 @@ public class CompilerOptions implements Serializable {
           return null;
       }
     }
+  }
+
+  public enum ChunkOutputType {
+    GLOBAL_NAMESPACE,
+    ES_MODULES;
   }
 
   /**
--- a/src/com/google/javascript/jscomp/ConvertChunksToESModules.java
+++ b/src/com/google/javascript/jscomp/ConvertChunksToESModules.java
@@ -20,11 +20,11 @@ import com.google.javascript.jscomp.parsing.parser.FeatureSet;
 import com.google.javascript.rhino.IR;
 import com.google.javascript.rhino.Node;
 import com.google.javascript.rhino.Token;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
@@ -51,6 +51,10 @@ final class ConvertChunksToESModules implements CompilerPass {
   static final DiagnosticType ASSIGNMENT_TO_IMPORT = DiagnosticType.error(
       "JSC_IMPORT_ASSIGN",
       "Imported symbol \"{0}\" in chunk \"{1}\" cannot be assigned");
+
+  static final DiagnosticType UNABLE_TO_COMPUTE_RELATIVE_PATH = DiagnosticType.error(
+      "JSC_UNABLE_TO_COMPUTE_RELATIVE_PATH",
+      "Unable to compute relative import path from \"{0}\" to \"{1}\"");
 
   /**
    * Constructor for the ConvertChunksToESModules compiler pass.
@@ -140,12 +144,19 @@ final class ConvertChunksToESModules implements CompilerPass {
           importSpecs.addChildToBack(importSpec);
         }
         Node importStatement = new Node(Token.IMPORT);
-        String importPath =
-            this.resolveSibling(
-                jsModuleImports.getKey().getName() + ".js",
-                jsModuleImportsFromChunk.getKey().getName() + ".js");
-        if (!importPath.startsWith(".") && !importPath.startsWith("/")) {
-          importPath = "./" + importPath;
+        String importPath = jsModuleImportsFromChunk.getKey().getName();
+        try {
+          importPath =
+              this.relativePath(
+                  jsModuleImports.getKey().getName(),
+                  jsModuleImportsFromChunk.getKey().getName());
+        } catch (IllegalArgumentException e) {
+          compiler.report(
+              JSError.make(
+                  astRoot,
+                  UNABLE_TO_COMPUTE_RELATIVE_PATH,
+                  jsModuleImports.getKey().getName(),
+                  jsModuleImportsFromChunk.getKey().getName()));
         }
         importStatement.addChildToFront(IR.string(importPath));
         importStatement.addChildToFront(importSpecs);
@@ -215,25 +226,29 @@ final class ConvertChunksToESModules implements CompilerPass {
     }
   }
 
-  private String resolveSibling(String fromPath, String toPath) {
-    // output chunks are likely to be in the same folder. Specialize for that case.
-    List<String> fromPathParts = new ArrayList<>(Arrays.asList(fromPath.split("/")));
-    List<String> toPathParts = new ArrayList<>(Arrays.asList(toPath.split("/")));
-    if (fromPathParts.size() == toPathParts.size() &&
-        !(fromPathParts.get(0).equals(".") || fromPathParts.get(0).equals("..")) &&
-        !(toPathParts.get(0).equals(".") || toPathParts.get(0).equals(".."))) {
-      boolean equalFolderPaths = true;
-      for (int i = 0; i < fromPathParts.size() - 1; i++) {
-        if (!fromPathParts.get(i).equals(toPathParts.get(i))) {
-          equalFolderPaths = false;
-          break;
-        }
-      }
-      if (equalFolderPaths) {
-        return "./" + toPathParts.get(toPathParts.size() - 1);
-      }
+  /**
+   * Calculate the relative path between two URI paths. To remain compliant with
+   * ES Module loading restrictions, paths must always begin with a "./", "../" or "/"
+   * or they are otherwise treated as a bare module specifier.
+   *
+   * TODO(ChadKillingsworth): This method likely has use cases beyond this class
+   * and should be moved.
+   */
+  private String relativePath(String fromUriPath, String toUriPath) {
+    Path fromPath = Paths.get(fromUriPath);
+    Path toPath = Paths.get(toUriPath);
+    Path fromFolder = fromPath.getParent();
+
+    // if the from URIs are simple names without paths, they are in the same folder
+    // example: m0.js
+    if (fromFolder == null) {
+      return "./" + toUriPath;
     }
 
-    return Compiler.resolveSibling(fromPath, toPath);
+    String calculatedPath = fromFolder.relativize(toPath).toString();
+    if (calculatedPath.startsWith(".") || calculatedPath.startsWith("/")) {
+      return calculatedPath;
+    }
+    return "./" + calculatedPath;
   }
 }
--- a/src/com/google/javascript/jscomp/ConvertChunksToESModules.java
+++ b/src/com/google/javascript/jscomp/ConvertChunksToESModules.java
@@ -15,6 +15,7 @@
  */
 package com.google.javascript.jscomp;
 
+import static com.google.common.base.Preconditions.checkState;
 import com.google.javascript.jscomp.NodeTraversal.AbstractPostOrderCallback;
 import com.google.javascript.jscomp.parsing.parser.FeatureSet;
 import com.google.javascript.rhino.IR;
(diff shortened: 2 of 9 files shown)


EatingW:
Created internal issue http://b/132364485

src/com/google/javascript/jscomp/ConvertChunksToESModules.java
```diff
@@ -0,0 +1,192 @@
...
+ * Finds all references to global symbols in a different output chunk and add ES Module imports and
+ * exports for them.
+ *
+ * <p>Given the name of the global object is NS
+ *
+ * <pre> var a = 1; function b() { return a }</pre>
+ *
+ * becomes
+ *
+ * <pre> var a = 1; var b = function b() { return a }; export {a};</pre>
```
jplaisted: Fly on the wall comment: why did `b` switch to a `var` and why isn't it exported?

ChadKillingsworth: I'll address - it's a copy / paste error from the RescopeGlobalSymbols pass. It wasn't exported because it was not referenced in a different chunk.

src/com/google/javascript/jscomp/ConvertChunksToESModules.java
```diff
@@ -0,0 +1,192 @@
...
+      }
+      for (Node importStatement : importStatements) {
+        moduleBody.addChildToFront(importStatement);
+      }
+      compiler.reportChangeToChangeScope(astRoot);
+    }
+  }
+
+  /** Find names in a module that are defined in a different module. */
+  private class FindCrossChunkReferences extends AbstractPostOrderCallback {
```
brad4d: We'd never encourage people to write code this way, but perhaps the simplest thing we could do to break our big output script into ES-module-style chunks would be to do everything as side effects rather than trying to track all the variables shared across chunks.

If we required `--isolation_mode=IIFE` and `--rename_prefix_namespace=_` or similar, then we could just not bother generating any exports and use the JSModuleGraph to generate the appropriate imports for each chunk. It's not *quite* that simple, but it might still be better than generating potentially enormous import statements and export statements.

 WDYT?

ChadKillingsworth: If you chain RescopeGlobalSymbols and this pass, that's almost what you get.

However that's definitely not my goal. I'm trying to get proper ES modules as output and the single namespace export that you pass around using imports / exports violates the spirit of that. I want to run this on some real world projects, but I believe the potentially enormous import/export statements are going to be the minority of projects. I don't believe that will be a problem for the majority of the open source world.

I would also expect a pretty strong negative developer reaction to the namespace idea. I'm wanting to fully embrace ES module output here.

brad4d: OK, I can see that.

So, do you have a real-world project handy you can try it on?

brad4d:
Maybe you don't have to worry about mutability too much?

I just tried this little test example

```html
<!DOCTYPE html>
<meta charset="utf-8">
<head>
<title>Page Title</title>
</head>
<body>
  <script type="module">
    import {a, getA} from '/modules/lib1.js';
    console.log(`a = ${a}, getA() => ${getA()}`);
    a = 15;
    console.log(`a = ${a}, getA() => ${getA()}`);
  </script>
</body>
```
```javascript
// /modules/lib1.js
export let a = 1;
export function getA() {
  return a;
}
```

Chrome reported an error for the `a = 15` assignment.

src/com/google/javascript/jscomp/ConvertChunksToESModules.java
ChadKillingsworth: Sure do!

concavelenz:
Chad is it worth keeping this PR open?

ChadKillingsworth:
I finally was able to finish this and run real-world tests on a sizable project. The results were quite surprising. Across the board, this had almost no impact on the total size of the project. Some files did see significant changes in size, but overall that was not the case. A few files actually had a dramatic reduction in size - my guess is that the variable references vs the properties allowed more optimizations. I was prepared for this to increase the size so neutral was definitely a win.

Full results can be seen here: https://drive.google.com/file/d/16CfcUOrXdEZtezOPaA0tp30IgrPkLBHA/view?usp=sharing

The only optimization pass I have seen thus far that needed updated was ExtractPrototypeMemberDeclaration. The global variable reference that was shared threw exceptions as the imported symbol cannot be modified. I added an option for this pass to use a per-chunk variable rather than solely a global one.

brad4d:
Thanks for posting those results, @ChadKillingsworth 

Have you you done any testing to confirm that the output code behaves correctly?

ChadKillingsworth:
I haven't run the full end-to-end test suite, but I have run and manually tested the application locally. Everything worked as expected so far.

I added a check with an error for the case when an imported symbol is the LHS of an assign. My application had 4 errors for that - all in tests and were easy fixes.

ChadKillingsworth:
@brad4d this should be ready for review now

src/com/google/javascript/jscomp/ConvertChunksToESModules.java
```diff
@@ -0,0 +1,297 @@
...
+      if (chunk.getInputs().size() == 0) {
+        continue;
+      }
+
+      CompilerInput firstInput = null;
+      for (CompilerInput input : chunk.getInputs()) {
+        Node astRoot = input.getAstRoot(compiler);
+        FeatureSet scriptFeatures = NodeUtil.getFeatureSetOfScript(astRoot);
+        checkState(!scriptFeatures.contains(FeatureSet.ES6_MODULES));
+        if (firstInput == null) {
```
brad4d, line 112: Could you add comments near here to show what you intend the AST structure to look like when this is done?

Maybe:

```java
// Original AST:
// [ROOT] top JS sources root
//     [SCRIPT] // 1st chunk's 1st input
//         statements from 1st chunk's 1st script
//     [SCRIPT] // 1st chunk's 2nd input
//         statements from 1st chunk's 2nd script
//     ...
//     [SCRIPT] // 1st chunk's nth input
//     [SCRIPT] // 2nd chunk's 1st input
//     ...
//
// After this method
// [ROOT] top JS sources root
//     [SCRIPT] // 1st chunk's 1st input script
//         [MODULE_BODY] // new module body to contain the 1st chunk
//             // statements from all scripts in the first chunk
//             // All script nodes from the other inputs of the chunk are now empty
//     [SCRIPT] // 2nd chunk's 1st input script
//         [MODULE_BODY] // new module body to contain the 2nd chunk
//             // statements from all scripts in the first chunk
//             // All script nodes from the other inputs of the chunk are now empty
```

src/com/google/javascript/jscomp/ConvertChunksToESModules.java
```diff
@@ -0,0 +1,297 @@
...
+        exportSpec.putIntProp(Node.IS_SHORTHAND_PROPERTY, 1);
+        exportSpecs.addChildToBack(exportSpec);
+      }
+      Node export = IR.export(exportSpecs).useSourceInfoFromForTree(moduleBody);
+      moduleBody.addChildToBack(export);
+      compiler.reportChangeToEnclosingScope(moduleBody);
+    }
+  }
+
+  private static String GetChunkName(JSModule chunk) {
```
brad4d, line 159: How about `getChunkFileName()`?

src/com/google/javascript/jscomp/ConvertChunksToESModules.java
```diff
@@ -0,0 +1,297 @@
...
+          compiler.reportChangeToChangeScope(astRoot);
+        }
+      }
+    }
+  }
+
+  /**
+   * Add export statements to chunks
+   */
+  private void AddExportStatements() {
```
brad4d, line 140: Could you add a code comment to show what you expect the code generated by this to look like?

e.g.

```java
// Append one big export statement to the end of the whole chunk.
// ```javascript
// export { name1, name2, name3, ...., nameN };
// ```
```

src/com/google/javascript/jscomp/ConvertChunksToESModules.java
```diff
@@ -0,0 +1,297 @@
...
+
+      // For each distinct chunk where a referenced symbol is defined, create an import statement
+      // referencing the names.
+      for (Map.Entry<JSModule, Set<String>> importsByChunk : chunkImportsEntry.getValue().entrySet()) {
+        Node importSpecs = new Node(Token.IMPORT_SPECS);
+        for (String name : importsByChunk.getValue()) {
+          Node importSpec = new Node(Token.IMPORT_SPEC);
+          importSpec.addChildToFront(IR.name(name));
+          importSpec.addChildToFront(IR.name(name));
+          importSpec.putIntProp(Node.IS_SHORTHAND_PROPERTY, 1);
```
brad4d, line 183: use `importSpec.setShorthandProperty(true)` instead of `putIntProp(1)`
Also other places `IS_SHORTHAND_PROPERTY` is used.

src/com/google/javascript/jscomp/ConvertChunksToESModules.java
```diff
@@ -0,0 +1,297 @@
...
+    }
+  }
+
+  /** Find names in a module that are defined in a different module. */
+  private class FindCrossChunkReferences extends AbstractPostOrderCallback {
+    @Override
+    public void visit(NodeTraversal t, Node n, Node parent) {
+      if (n.isName()) {
+        String name = n.getString();
+        if ("".equals(name)) {
```
brad4d, line 221:
```java
// Function expressions without names have a NAME node containing an empty string.
```

src/com/google/javascript/jscomp/ConvertChunksToESModules.java
```diff
@@ -0,0 +1,297 @@
...
+          return;
+        }
+        Scope s = t.getScope();
+        Var v = s.getVar(name);
+        if (v == null || !v.isGlobal()) {
+          return;
```
brad4d, line 227:
```java
// No Var will be found for variables declared in extern files.
```

src/com/google/javascript/jscomp/ConvertChunksToESModules.java
```diff
@@ -0,0 +1,297 @@
...
+        }
+        CompilerInput input = v.getInput();
+        if (input == null) {
+          return;
```
brad4d, line 231: I don't think this should happen.

```java
CompilerInput input = checkNotNull(v.getInput();
```

src/com/google/javascript/jscomp/ConvertChunksToESModules.java
```diff
@@ -0,0 +1,297 @@
...
+        }
+        // Compare the chunk where the variable is declared to the current
+        // chunk. If they are different, the variable is used across modules.
+        JSModule definingChunk = input.getModule();
+        JSModule referencingChunk = t.getModule();
+        if (definingChunk != referencingChunk) {
+          if (NodeUtil.isLhsOfAssign(n)) {
```
brad4d, line 238: This will miss some cases. I believe you want `NodeUtil.isLValue(n)`.

src/com/google/javascript/jscomp/CompilerOptions.java
```diff
@@ -3101,6 +3104,11 @@ public static InstrumentOption fromString(String value) {
     }
   }
 
+  public enum ChunkOutputType {
```
brad4d, line 3107: Could you add JavaDoc to explain what this enum is for and what its values mean?

brad4d (changes requested):

src/com/google/javascript/jscomp/ConvertChunksToESModules.java
```diff
@@ -0,0 +1,297 @@
+/*
+ * Copyright 2011 The Closure Compiler Authors.
```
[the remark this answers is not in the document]
ChadKillingsworth: fixed

src/com/google/javascript/jscomp/ConvertChunksToESModules.java
```diff
@@ -0,0 +1,297 @@
...
+    // example: export {};
+    for (JSModule chunk : compiler.getModuleGraph().getAllModules()) {
+      if (!crossChunkExports.containsKey(chunk) &&
+          !crossChunkImports.containsKey(chunk) &&
+          chunk.getInputs().size() > 0) {
+        crossChunkExports.put(chunk, new HashSet<>());
+      }
+    }
+
+    ConvertChunkSourcesToModules();
```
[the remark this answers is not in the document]
ChadKillingsworth: fixed - sorry I work on so many different projects with very different style guides.

test/com/google/javascript/jscomp/ConvertChunksToESModulesTest.java
```diff
@@ -0,0 +1,254 @@
+/*
+ * Copyright 2011 The Closure Compiler Authors.
```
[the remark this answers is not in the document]
ChadKillingsworth: fixed

src/com/google/javascript/jscomp/ConvertChunksToESModules.java
```diff
@@ -0,0 +1,297 @@
...
+   */
+  private void ConvertChunkSourcesToModules() {
+    for (JSModule chunk : compiler.getModuleGraph().getAllModules()) {
+      if (chunk.getInputs().size() == 0) {
+        continue;
+      }
+
+      CompilerInput firstInput = null;
+      for (CompilerInput input : chunk.getInputs()) {
+        Node astRoot = input.getAstRoot(compiler);
```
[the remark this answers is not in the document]
ChadKillingsworth: done

src/com/google/javascript/jscomp/ConvertChunksToESModules.java
ChadKillingsworth: done

src/com/google/javascript/jscomp/ConvertChunksToESModules.java
ChadKillingsworth: Much better - fixed

src/com/google/javascript/jscomp/ConvertChunksToESModules.java
ChadKillingsworth: done

src/com/google/javascript/jscomp/ConvertChunksToESModules.java
ChadKillingsworth: Didn't know about that one - much nicer

src/com/google/javascript/jscomp/ConvertChunksToESModules.java
ChadKillingsworth: added

src/com/google/javascript/jscomp/ConvertChunksToESModules.java
ChadKillingsworth: I removed it - tests all still pass and no errors on my real internal project so I think you are correct.

src/com/google/javascript/jscomp/ConvertChunksToESModules.java
ChadKillingsworth: Fixed

src/com/google/javascript/jscomp/CompilerOptions.java
ChadKillingsworth: Added

src/com/google/javascript/jscomp/ConvertChunksToESModules.java
ChadKillingsworth: Added. I actually wasn't sure what made this check trigger so that helps.

ChadKillingsworth:
All comments addressed. I ran this on my internal project again just to make sure things were still working.

As a reminder, #3724 must be merged as well or this pass doesn't work well. I get errors like:

```
node_modules/google-closure-library/closure/goog/uri/uri.js:1165:0: ERROR - [JSC_IMPORT_ASSIGN] Imported symbol "h" in chunk "js/bannoweb-login.js" cannot be assigned
  1165| goog.Uri.QueryData.prototype.getCount = function() {
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  1166|   this.ensureKeyMapInitialized_();
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  1167|   return this.count_;
        ^^^^^^^^^^^^^^^^^^^^^
  1168| };
```

brad4d:
imported and sent for internal review and testing

brad4d:
> imported and sent for internal review and testing

Sorry, that was intended for #3724

brad4d:
imported, edited as needed to match internal requirements, sent to @blickly for internal review

brad4d:
I'm trying to submit this now after making quite a lot of changes.
Here's a couple of particularly interesting ones for educational purposes.

1. Replaced all uses of `HashSet` and `HashMap` with `LinkedHashSet` and `LinkedHashMap` - This is necessary to ensure that changes to the ordering of the Java hash implementation won't cause the compiler's output to change in subtle ways.
1. Had to add super source (j2cl fakes) for a couple of `Path` methods that weren't previously by our code.
1. Use `computeIfAbsent()` to add to a possibly-not-there-yet item in a map.
1. Prefer `.isEmpty()` over `.size == 0`

brad4d:
This was merged, but somehow didn't get marked as such.
Maybe that's because I had to make such significant changes to land it?

PR #3368, opened 2019-05-09
Add objectFit and objectPosition to CSSProperties externs

mk0x9:
[caniuse.com](https://caniuse.com/#search=object%20fit) reports that those properties are supported in 90% of browsers unprefixed.

### DIFF
--- a/externs/browser/w3c_css.js
+++ b/externs/browser/w3c_css.js
@@ -2053,6 +2053,20 @@ CSSProperties.prototype.userSelect;
 
 // http://www.w3.org/TR/cssom-view/#extensions-to-the-window-interface
 
+// CSS 3 Images
+
+/**
+ * @type {string}
+ * @see https://www.w3.org/TR/css3-images/#the-object-fit
+ */
+CSSProperties.prototype.objectFit;
+
+/**
+ * @type {string}
+ * @see https://www.w3.org/TR/css3-images/#object-position
+ */
+CSSProperties.prototype.objectPosition;
+
 /**
  * @param {string} media_query_list
  * @return {!MediaQueryList}


externs/browser/w3c_css.js
```diff
@@ -2053,6 +2053,20 @@ CSSProperties.prototype.userSelect;
 
 // http://www.w3.org/TR/cssom-view/#extensions-to-the-window-interface
 
+// CSS 3 Images
+
+/**
+ * @type {string}
+ * @see https://drafts.csswg.org/css-images-3/#the-object-fit
```
EatingW: I think you want to use https://www.w3.org/TR/css3-images/#object-fit?
The drafts.csswg.org seems suspicious. 
I'm not very familiar with Candidate Recommendation vs Editor's Draft. Do you have more context on that?

externs/browser/w3c_css.js
```diff
@@ -2053,6 +2053,20 @@ CSSProperties.prototype.userSelect;
...
+ */
+CSSProperties.prototype.objectFit;
+
+/**
+ * @type {string}
+ * @see https://drafts.csswg.org/css-images-3/#the-object-position
```
EatingW: https://www.w3.org/TR/css3-images/#object-position?
See above.

EatingW:
Created internal issue http://b/132360319

externs/browser/w3c_css.js
mk0x9: Yes indeed!

I'm afraid there is no context, I have blindly followed the link from the MDN page.

externs/browser/w3c_css.js
mk0x9: Fixed

EatingW:
Imported for internal testing and review.

PR #3369, opened 2019-05-10
Remove values from externs

realityforge:
See #3293 for the reasoning but the short version is that closure
compiler does not use the values and reads the values at runtime.
If any tool were to start using it and the values were incorrect,
then incorrect behaviour will occur.


realityforge:
This strips a few more values that were missed in #3312 as overlapped with other PRs that were in flight

EatingW:
Created internal issue http://b/132448102

EatingW:
Imported for internal review and testing.

realityforge:
Thanks!

PR #3370, opened 2019-05-10
Extract the externs for the Selection API into separate files

realityforge:
The standardized Selection API has been extracted into the w3c_selection.js
file while the non standardized elements have been moved to
nonstandard_selection.js

Also added some TODO notes where the typing of the standardized API
does not align with the standard.


realityforge:
It is likely that you will need to merge google/elemental2#117 at the same time as this PR is merged

EatingW:
Could you separate this pull request into smaller pull requests? In particular, right now its hard to see what is just being moved, versus being modified and moved. 

Could you separate this into at least 3 pull requests (feel free to break down further):
- All the changes above "The following changes may impact type checking". 
- All the changes below "The following changes may impact type checking", since these might require more internal testing to see if there are current code that depends on the previous behavior
- Extracting definitions into w3c_selection.js

realityforge:
Will do. As they will overlap and conflict I will try to get the merged one at a time. Thanks

realityforge:
Ok the two PRs that were extracted from this one (namely #3389 and #3374) have been merged and this PR has been rebased on master.

It should be ready for re-review.

tjgq:
Created Google internal issue b/132796901.

EatingW:
Effort to tighten the nullability of types were made in #3389 but unfortunately I wasn't able to get the internal version to land due to cascading places that need fixing (for example, other externs that use these). Leaving some TODOs might be a good idea though.

realityforge:
Done. I also changed the type of `Element.prototype.onselectionchange` from `?function(Event): void` to `?function(!Event): void`. I don't believe this should be an issue as the symbol was only added in #3374

tjgq:
Thanks, I will import it for review now.

PR #3373, opened 2019-05-11
Update Window.prototype.scrollTo extern to accept dictionary options

### LINKED ISSUE #3371: Update window.scroll signature to accept an options object
There is an alternative way of using `window.scroll` (and `window.scrollTo`) that allows to pass an object instead of the x and y coordinates.

See the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Window/scroll) for reference.

A change in the externs is needed around these lines:

https://github.com/google/closure-compiler/blob/964833f1af60a733cc6821f6c0737ac18fba9339/externs/browser/w3c_css.js#L2099-L2113

I'd be happy to help with a PR, if given some assistance in writing the externs correctly.

mk0x9:
Also updates:
- Window.prototype.scroll
- Window.prototype.scrollBy

Looks like closure-compiler doesn't support function overloading so I set second param optional.

Closes #3371

### DIFF
--- a/externs/browser/w3c_css.js
+++ b/externs/browser/w3c_css.js
@@ -2097,24 +2097,34 @@ Window.prototype.scrollY;
 Window.prototype.pageYOffset;
 
 /**
- * @param {number} x
- * @param {number} y
+ * @typedef {{
+ *   left: (number|undefined),
+ *   top: (number|undefined),
+ *   behavior: (string|undefined)
+ * }}
+ * @see https://www.w3.org/TR/cssom-view/#dictdef-scrolltooptions
+ */
+var ScrollToOptions;
+
+/**
+ * @param {number|!ScrollToOptions} x
+ * @param {number=} y
  * @see http://www.w3.org/TR/cssom-view/#dom-window-scroll
  * @return {undefined}
  */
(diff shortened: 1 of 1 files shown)


externs/browser/w3c_css.js
```diff
@@ -2097,24 +2097,34 @@ Window.prototype.scrollY;
...
+ *   left: (number|undefined),
+ *   top: (number|undefined),
+ *   behavior: (string|undefined)
+ * }}
+ * @see https://www.w3.org/TR/cssom-view/#dictdef-scrolltooptions
+ */
+var ScrollToOptions;
+
+/**
+ * @param {number|!ScrollToOptions} x
```
EatingW: Instead of x, I'll rename this to scrollToOptionsOrX, just to make it clear what it is. 

Similar to https://github.com/google/closure-compiler/blob/2375110b30ccb83b0a550c8c3d46c526e3896d62/externs/browser/w3c_css.js#L2322

No need to change it on your side, I'll just change it in the internal version.

EatingW:
Created internal issue http://b/132635210. Imported for internal review and testing.

mk0x9:
Cool, thanks!

PR #3374, opened 2019-05-11
Update documentation for Selection API and minor forward compatible API fixes

realityforge:
* Updated @see links to refer to the current spec and add missing documentation
  where missing for all standardized elements.
* Updated parameter names to match those in the spec.
* Explicitly document nullable types with a `?` rather than relying on implicit nullability.
  This modified the typing for the following symbols:
  - The return value of `Window.prototype.getSelection`
  - `Selection.prototype.anchorNode`
  - `Selection.prototype.focusNode`
  - The `node` parameter in `Selection.prototype.collapse`
* Made parameters optional to match the spec
  - The `offset` parameter of `Selection.prototype.extend`
  - The `offset` parameter of `Selection.prototype.collapse`
  - The `allowPartialContainment` parameter of `Selection.prototype.containsNode`
* Added missing extern for `Selection.prototype.setPosition`
* Added missing extern for `Element.prototype.onselectionchange`

### DIFF
--- a/externs/browser/gecko_dom.js
+++ b/externs/browser/gecko_dom.js
@@ -285,9 +285,9 @@ Window.prototype.forward = function() {};
 Window.prototype.getAttention = function() {};
 
 /**
- * @return {Selection}
- * @see https://developer.mozilla.org/en/DOM/window.getSelection
+ * @return {?Selection}
  * @nosideeffects
+ * @see https://w3c.github.io/selection-api/#dom-window-getselection
  */
 Window.prototype.getSelection = function() {};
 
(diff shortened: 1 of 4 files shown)


realityforge:
This is the first PR that extracts functionality from #3370 to make it easier to merge. It is expected that this branch should cause no type checking changes

realityforge:
This has been submitted as one commit but I am happy to break it up into N bytes size commits to improve traceability if you prefer.  (i.e. one for documentation updates, one for each new externs, one for `?` addition one for marking parameters as optional)

EatingW:
I think this is fine, since it's unlikely to change type checking behavior. Created internal issue http://b/132633849, imported for internal testing and review.

PR #3383, opened 2019-05-14
Move Node.prototype.insertAdjacentElement from ie_dom.js to html5.js

### LINKED ISSUE #2715: Element externs missing insertAdjacentElement/Text
insert-adjacent is supported by all evergreen browsers, and should be included in the browser externs (needed for a j2cl project)

https://caniuse.com/#feat=insert-adjacent

If project owners are okay with inclusion, I can submit a pull request (not sure which dom file to best put it in).

It would technically be nice to have outerHTML as well, but I'll ask for one thing at a time ^_^

realityforge:
The @see annotation was updated to refer current spec and the parameter
names have been updated to align with the names in the spec.

Fixes #2715

### DIFF
--- a/externs/browser/html5.js
+++ b/externs/browser/html5.js
@@ -55,6 +55,17 @@ Node.prototype.contains = function(n) {};
 Node.prototype.isConnected;
 
 /**
+ * Inserts the given HTML Element into the node at the location.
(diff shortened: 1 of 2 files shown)


lauraharker:
Created internal issue http://b/132798763

lauraharker:
Importing for internal testing

PR #3385, opened 2019-05-15
Move atob and btoa to html5.js

realityforge:
The methods have been standardized and can be moved out of the
vendor specific extern into the standardized extern. This makes
these methods available in google/elemental2

See https://html.spec.whatwg.org/multipage/webappapis.html#atob


lauraharker:
Created internal issue http://b/132797199

lauraharker:
Imported for internal testing

realityforge:
Thanks!

PR #3386, opened 2019-05-15
Array concat optimizations

qqingvarqq:
### This PR introduces 2 optimizations for an array concatenation described in this issue #3211
---
##### 1. Replacing empty array literal from the front of concatenation by the first argument of concat function call.
```javascript
[].concat(arr1,arr2,arr3) -> arr1.concat(arr2,arr3);
[].concat(funWithArrayReturnType(),funWithArrayReturnType()) -> funWithArrayReturnType().concat(funWithArrayReturnType())
```
This optimization can be performed only when we know that the first argument of concatenation is always an array.
for example here we can't perform this optimization:
```javascript
/**
 * @param {!Array<?>|string}
*/
function fn(param){
     return [].concat(param);
}
```
Overall this optimization ***saves 3 or 2 bytes***. And make sense to be added.

##### 2. Add folding of chained concat functions.
```javascript
arr0.concat(arr1).concat(arr2).concat(arr3) -> arr0.concat(arr1,arr2,arr3);
arr.concat(pureFnCall()).concat(1) -> arr.concat(pureFnCall(),1);
```
This one is very useful because it can potentially ***save 8\*chainCount bytes***
and this code pattern is very common, for example, code from gdrive minified source
```javascript
a.R.Lka.concat(a.R.NQ).concat(a.R.nla).concat(a.R.rla)
```
So this optimization can be performed everywhere, except when concat arguments can produce side effects.
Here we can't fold concatenation
```javascript
var a = [];
a.concat(1).concat(a.push(1)); // [1,1]
/*can't fold to */a.concat(1,a.push(1)); // [1,1,1]
```

### DIFF
--- a/src/com/google/javascript/jscomp/PeepholeReplaceKnownMethods.java
+++ b/src/com/google/javascript/jscomp/PeepholeReplaceKnownMethods.java
@@ -968,6 +968,9 @@ class PeepholeReplaceKnownMethods extends AbstractPeepholeOptimization {
     if (isSafeToRemoveArrayLiteralFromFrontOfConcat(n)) {
       return removeArrayLiteralFromFrontOfConcat(n);
     }
+    if (isSafeToFoldConcatChaining(n)) {
+      return foldConcatChaining(n);
+    }
     return n;
   }
 
@@ -986,7 +989,26 @@ class PeepholeReplaceKnownMethods extends AbstractPeepholeOptimization {
     reportChangeToEnclosingScope(n);
     return n;
   }
+  
+  /**
+   * perform folding of chained concat functions, so
+   * arr.concat(a).concat(b) will be fold into arr.concat(a,b);
+   */
+  private Node foldConcatChaining(Node n) {
+    checkArgument(isSafeToFoldConcatChaining(n), n);
 
+    Node previousConcatCall = n.getFirstFirstChild();
+    Node arg = n.getSecondChild();
+    while (arg != null) {
+      Node curentArg = arg;
+      arg = arg.getNext();
+      previousConcatCall.addChildToBack(curentArg.detach());
+    }
+    n.replaceWith(previousConcatCall.detach());
+    reportChangeToEnclosingScope(previousConcatCall);
+    return previousConcatCall;
+  }
+  
   /**
    * Check if we have this code pattern [].concat(exactlyArrayArgument,...*) so
    * removeArrayLiteralFromFrontOfConcat can be performed
@@ -1006,11 +1028,41 @@ class PeepholeReplaceKnownMethods extends AbstractPeepholeOptimization {
   }
 
   /**
+   * Check if we have this code pattern
+   * array.concat(...*).concat(sideEffectFreeArguments)
+   * so foldConcatChaining can be performed
+   */
+  private boolean isSafeToFoldConcatChaining(Node n) {
+    checkArgument(n.isCall(), n);
+    
+    if (!isConcatFunctionCall(n)) {
+      return false;
+    }
+    Node previousFunctionCall = n.getFirstFirstChild();
+    if (!previousFunctionCall.isCall() || !isConcatFunctionCall(previousFunctionCall)) {
+      return false;
+    }
+    // make sure that arguments in second concat function call can't change the array
+    // so we can fold chained concat functions
+    // to clarify, consider this code
+    // here we can't fold concatenation
+    // var a = [];
+    // a.concat(1).concat(a.push(1)); -> [1,1]
+    // a.concat(1,a.push(1)); -> [1,1,1]
+    for (Node arg = n.getSecondChild(); arg != null; arg = arg.getNext()) {
+      if (mayHaveSideEffects(arg)) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  /**
    * Check if this call node is Array.prototype.concat
    */
   private boolean isConcatFunctionCall(Node n) {
     checkArgument(n.isCall(), n);
-    
+
     Node callTarget = checkNotNull(n.getFirstChild());
     if (!callTarget.isGetProp()) {
       return false;
--- a/src/com/google/javascript/jscomp/PeepholeReplaceKnownMethods.java
+++ b/src/com/google/javascript/jscomp/PeepholeReplaceKnownMethods.java
@@ -24,6 +24,7 @@ import static com.google.common.base.Strings.isNullOrEmpty;
 import com.google.common.collect.ImmutableList;
 import com.google.javascript.rhino.IR;
 import com.google.javascript.rhino.Node;
+import com.google.javascript.rhino.jstype.JSType;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
@@ -59,16 +60,17 @@ class PeepholeReplaceKnownMethods extends AbstractPeepholeOptimization {
   private Node tryFoldKnownMethods(Node subtree) {
     // For now we only support string methods .join(),
     // .indexOf(), .substring() and .substr()
+    // array method concat()
     // and numeric methods parseInt() and parseFloat().
 
+    checkArgument(subtree.isCall(), subtree);
     subtree = tryFoldArrayJoin(subtree);
-
+    // tryFoldArrayJoin may return a string literal instead of a CALL node
     if (subtree.isCall()) {
-      Node callTarget = subtree.getFirstChild();
-      if (callTarget == null) {
-        return subtree;
-      }
-
+      subtree = tryToFoldArrayConcat(subtree);
+      checkState(subtree.isCall(), subtree);
+      Node callTarget = checkNotNull(subtree.getFirstChild());
+      
       if (NodeUtil.isGet(callTarget)) {
         if (isASTNormalized() && callTarget.getFirstChild().isQualifiedName()) {
           switch (callTarget.getFirstChild().getQualifiedName()) {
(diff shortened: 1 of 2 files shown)


lauraharker:
Created internal issue http://b/132798114

brad4d:
Could you move your implementation to `PeepholeReplaceKnownMethods.java`?
It would be a better fit there.
Thanks.

qqingvarqq:
Sure

src/com/google/javascript/jscomp/PeepholeReplaceKnownMethods.java
```diff
@@ -59,9 +60,11 @@ Node optimizeSubtree(Node subtree) {
   private Node tryFoldKnownMethods(Node subtree) {
     // For now we only support string methods .join(),
     // .indexOf(), .substring() and .substr()
+    // array method concat()
     // and numeric methods parseInt() and parseFloat().
 
     subtree = tryFoldArrayJoin(subtree);
+    subtree = tryToFoldArrayConcat(subtree);
```
qqingvarqq: Not sure if it the right place to put this function call,
because below here lots of nested if statement, and not clear if we should perform only one optimization during `tryFoldKnownMethods` subtree call or several optimizations or probably `tryFoldKnownMethods` may be refactored.

qqingvarqq:
@brad4d updated,
I find out one edge case which I added to the `PeepholeReplaceKnownMethodsTest.java` 
line 652. Also, I made `isConcatFunctionCall` more strict.
This edge case occurs when we add more 'concat' definitions to the extern file.
Example:
```javascript
function Foo(){};
Foo.prototype.concat = function(arg){
 console.log(arg);
 return this;
}
const obj = new Foo();
obj.concat(1).concat(2); //can't perform concat folding optimization
```

qqingvarqq:
@brad4d This PR still on review? or should I make some other changes?

src/com/google/javascript/jscomp/PeepholeReplaceKnownMethods.java
```diff
@@ -956,4 +958,135 @@ private Node tryFoldStringSplit(Node n, Node stringNode, Node arg1) {
...
     return arrayOfStrings;
   }
+
+  private Node tryToFoldArrayConcat(Node n) {
+    checkArgument(n.isCall(), n);
+
+    if (!isASTNormalized() || !useTypes) {
+      return n;
+    }
+    if (isSafeToRemoveArrayLiteralFromFrontOfConcat(n)) {
```
brad4d: It would be good inline the logic of `isConcatCall()` here and use it to pull apart `n` into the bits you'll need and early exit if it turns out `n` isn't a concat call.

Then the utility methods that do the work would take the pieces, which you could check the validity of with not-very-bright `checkArgument()` calls etc.
This reduces redundant logic and improves readability.

```java
Node calleeNode = checkNotNull(n.getFirstChild());
if (!calleeNode.isGetProp()) {
  return n; // not an `array.concat()` call
}
Node callTargetNode = calleeNode.getFirstChild();
String methodName = calleeNode.getLastChild().getString();

if (!"concat".equals(methodName) {
  return n; // not an `array.concat()` call
}
// etc.
```

brad4d:
@qqingvarqq 
Sorry, just been really busy with other stuff.

Thanks for your improvements.
Requested a bit more code reorganization.

test/com/google/javascript/jscomp/PeepholeReplaceKnownMethodsTest.java
```diff
@@ -608,6 +611,41 @@ public void testReplaceWithCharAt() {
     foldSameStringTyped("a.substr(0, 1)");
     foldSameStringTyped("''.substring(i, i + 1)");
   }
+  
+  @Test
+  public void testFoldArrayConcat() {
+    enableTypeCheck();
+
+    fold("[].concat([1,2,3],1)", "[1,2,3].concat(1)");
```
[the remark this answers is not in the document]
brad4d: Thanks!

src/com/google/javascript/jscomp/PeepholeReplaceKnownMethods.java
qqingvarqq: @brad4d  I get your point. Removed redundant logic, moved `isConcatCall` on level up, so if `n` is not concat we exit early.
But I'm not sure if we need to expose the logic of  `isConcatCall`,
because this method used in two places.
 1. To check if it actually concat function call. https://github.com/qqingvarqq/closure-compiler/blob/c9716b3185b90cb7599900907ce5d8752d503953/src/com/google/javascript/jscomp/PeepholeReplaceKnownMethods.java#L965
 2. To check if the previous function call in the chainig is concat function call (mean `.concat().concat()`.https://github.com/qqingvarqq/closure-compiler/blob/c9716b3185b90cb7599900907ce5d8752d503953/src/com/google/javascript/jscomp/PeepholeReplaceKnownMethods.java#L1036

So if I expose the logic it brings code duplication, which is not good.
But if you are sure we need it I can expose this logic.

brad4d: We can avoid both code duplication and redundant execution of code if we replace `isConcatFunctionCall` with a method that returns an object of type `ConcatFunctionCall` or `null` if the node it is passed isn't actually a call to `.concat()`.

```java
abstract private static class ConcatFunctionCall {
  private final Node callNode;
  private final Node calleeNode;
  private final Node firstArgumentNode;
  ConcatFunctionCall(Node callNode, Node calleeNode, Node firstArgumentNode) {
    this.callNode = checkNotNull(callNode);
    this.calleeNode = checkNotNull(calleeNode);
    this.firstArgumentNode = checkNotNull(firstArgumentNode);
  }
}

/**
 * If the argument node is a call to `Array.prototype.concat`, then return a `ConcatFunctionCall` object for it,
 * otherwise return `null`.
 */
@Nullable
private ConcatFunctionCall createConcatFunctionCallForNode(Node n) {
  // logic here to look into n and either return new ConcatFunctionCall or null
}
```

qqingvarqq: Thanks for the detailed explanation.
Updated
Hope now it can be sent on internal review 🤞
By the way, should I squash commits?

src/com/google/javascript/jscomp/PeepholeReplaceKnownMethods.java
```diff
@@ -956,4 +959,168 @@ private Node tryFoldStringSplit(Node n, Node stringNode, Node arg1) {
...
+    @Nullable
+    private final Node firstArgumentNode;
+
+    ConcatFunctionCall(Node callNode, Node calleeNode, Node firstArgumentNode) {
+      this.callNode = checkNotNull(callNode);
+      this.calleeNode = checkNotNull(calleeNode);
+      this.firstArgumentNode = firstArgumentNode;
+    }
+
+    public Node getCallNode() {
```
brad4d: Nit:

Since all the fields are private and final, there is no need to have accessor functions.
Leave them if you like, but I wouldn't bother with them.

src/com/google/javascript/jscomp/PeepholeReplaceKnownMethods.java
```diff
@@ -956,4 +959,168 @@ private Node tryFoldStringSplit(Node n, Node stringNode, Node arg1) {
...
+    concatCall.replaceWith(previousConcatCall.detach());
+    reportChangeToEnclosingScope(previousConcatCall);
+    return previousConcatCall;
+  }
+
+  /**
+   * Check if we have this code pattern `[].concat(exactlyArrayArgument,...*)` so
+   * `removeArrayLiteralFromFrontOfConcat` can be performed.
+   */
+  private boolean isSafeToRemoveArrayLiteralFromFrontOfConcat(
```
brad4d: I would really prefer to have this logic in the same place as the logic doing the removal.

Maybe have a `tryToRemoveArrayLiteralFromFrontOfConcat(concatFunctionCall)` method?
It would return either the same concatFunctionCall it was given, or a new one for the newly constructed call.

Just thought of something though.
Do you intend to change `[].concat(myArray)` into just `myArray`?
I don't think I see logic to handle that case or a test case for it, but I don't think I see logic checking to make sure you don't end up getting `myArray.concat()` either.

brad4d:
Added some more review comments.
Not ready to import it yet, I'm afraid.
No need to squash commits, though.

src/com/google/javascript/jscomp/PeepholeReplaceKnownMethods.java
qqingvarqq: Done
No, this optimization not intended to change `[].concat(myArray)` into just` myArray`.
There is no need for special checks for this case. 
Because if `myarray` is array node or function which return array and only array, its possible to do this `[].concat(myArray)`->`myArray.concat()` those js code are equivalent (and that's all that we can optimize). 
(concat only takes one optional parameter https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat)
And 2 test cases exist for this purpose
https://github.com/qqingvarqq/closure-compiler/blob/c9520c0851b4f80b957250fa0743077403cddfc1/test/com/google/javascript/jscomp/PeepholeReplaceKnownMethodsTest.java#L625
https://github.com/qqingvarqq/closure-compiler/blob/c9520c0851b4f80b957250fa0743077403cddfc1/test/com/google/javascript/jscomp/PeepholeReplaceKnownMethodsTest.java#L627

src/com/google/javascript/jscomp/PeepholeReplaceKnownMethods.java
```diff
@@ -956,4 +959,148 @@ private Node tryFoldStringSplit(Node n, Node stringNode, Node arg1) {
...
+    if (!isASTNormalized() || !useTypes) {
+      return n;
+    }
+    ConcatFunctionCall concatFunctionCall = createConcatFunctionCallForNode(n);
+    if (concatFunctionCall == null) {
+      return n;
+    }
+    concatFunctionCall = tryToremoveArrayLiteralFromFrontOfConcat(concatFunctionCall);
+    checkNotNull(concatFunctionCall);
+    if (isSafeToFoldConcatChaining(concatFunctionCall)) {
```
brad4d: The combined check-then-do-it logic looks great for `tryToRemoveArrayLiteralFromFrontOfConcat()`

Could you do the same here?

src/com/google/javascript/jscomp/PeepholeReplaceKnownMethods.java
```diff
@@ -956,4 +959,148 @@ private Node tryFoldStringSplit(Node n, Node stringNode, Node arg1) {
...
+    }
+    return n;
+  }
+
+  /**
+   * Check if we have this code pattern `[].concat(exactlyArrayArgument,...*)` and if yes replace
+   * empty array literal from the front of concatenation by the first argument of concat function
+   * call `[].concat(arr,1)` -> `arr.concat(1)`.
+   */
+  private ConcatFunctionCall tryToremoveArrayLiteralFromFrontOfConcat(
```
brad4d: s/Toremove/ToRemove/

brad4d:
I think once these last comments are addressed this will be ready for import and testing.

I'm about to go on a fairly long vacation.
I'm going to see if I can get another team member to take on this PR in the meantime.

qqingvarqq:
updated, and this PR ready for re-review

EatingW:
Since Bradford is on vacation I'll take over from here. Imported for internal testing and review.

PR #3387, opened 2019-05-16
Specify return types for form data methods

realityforge:


lauraharker:
Created internal issue http://b/132890934

realityforge:
Thanks!

PR #3388, opened 2019-05-16
Add WebGL 2.0 Compute externs

hujiajie:
See https://www.khronos.org/registry/webgl/specs/latest/2.0-compute/

### DIFF
--- a/externs/browser/webgl2_compute.js
+++ b/externs/browser/webgl2_compute.js
@@ -1,0 +1,1608 @@
+/*
+ * Copyright 2019 The Closure Compiler Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * @fileoverview Definitions for WebGL functions as described at
+ * http://www.khronos.org/registry/webgl/specs/latest/
+ *
+ * This file is current up to the WebGL 2.0 Compute spec.
+ *
+ * This relies on webgl2.js and html5.js being included for WebGL2, Canvas and
+ * Typed Array support.
+ *
+ * @externs
+ */
+
+
+/**
+ * @constructor
+ * @extends {WebGL2RenderingContext}
+ */
+function WebGL2ComputeRenderingContext() {}
+
+
+/** @const {number} */
+WebGL2ComputeRenderingContext.COMPUTE_SHADER;
+
+/** @const {number} */
+WebGL2ComputeRenderingContext.MAX_COMPUTE_UNIFORM_BLOCKS;
+
+/** @const {number} */
+WebGL2ComputeRenderingContext.MAX_COMPUTE_TEXTURE_IMAGE_UNITS;
+
+/** @const {number} */
+WebGL2ComputeRenderingContext.MAX_COMPUTE_IMAGE_UNIFORMS;
+
+/** @const {number} */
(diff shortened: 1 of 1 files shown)


lauraharker:
Created internal issue http://b/132890731

externs/browser/webgl2_compute.js
```diff
@@ -0,0 +1,1608 @@
...
+ */
+WebGL2ComputeRenderingContext.prototype.getProgramInterfaceParameter = function(
+    program, programInterface, pname) {};
+
+/**
+ * @param {!WebGLProgram} program
+ * @param {number} programInterface
+ * @param {number} index
+ * @param {!Array<number>} props
+ * @return {?Array<*>}
```
lauraharker: The return type doesn't seem to match the spec.

`any getProgramResource(WebGLProgram program, GLenum programInterface, GLuint index, GLenum props);`

externs/browser/webgl2_compute.js
```diff
@@ -0,0 +1,1608 @@
...
+ * @return {undefined}
+ */
+WebGL2ComputeRenderingContext.prototype.programUniform4f = function(
+    program, location, v0, v1, v2, v3) {};
+
+/**
+ * @param {?WebGLProgram} program
+ * @param {?WebGLUniformLocation} location
+ * @param {number} count
+ * @param {!Int32Array|!Array<number>|!Array<boolean>} data
```
lauraharker: Do you know where in the spec `Array<boolean>` is defined to be a subtype of `Int32List`? Asking since `Int32List` was defined as `typedef ([AllowShared] Int32Array or sequence<GLint>) Int32List;` in https://www.khronos.org/registry/webgl/specs/latest/1.0/. Is boolean generally permissible to pass as a `GLint`?

I'm not objecting to this line, since it agrees with the existing webgl externs after https://github.com/google/closure-compiler/commit/030192492b5774fb9d87050a68a049853e115800, but I'd like a more concrete reason if you have one.

externs/browser/webgl2_compute.js
```diff
@@ -0,0 +1,1608 @@
...
+ * This relies on webgl2.js and html5.js being included for WebGL2, Canvas and
+ * Typed Array support.
+ *
+ * @externs
+ */
+
+
+/**
+ * @constructor
+ * @extends {WebGL2RenderingContext}
```
lauraharker: Could you also link to https://www.khronos.org/registry/webgl/specs/latest/2.0-compute/#webgl2-compute-context here?

externs/browser/webgl2_compute.js
hujiajie: The right signature should be `sequence? getProgramResource(WebGLProgram program, GLenum programInterface, GLuint index, sequence props)` as described in https://www.khronos.org/registry/webgl/specs/latest/2.0-compute/#program-interface. We will fix the typo in the spec later. However, I'm not quite sure if `{?Array<*>}` is valid here.

externs/browser/webgl2_compute.js
hujiajie: Done. But isn't it duplicated with the `@fileoverview` part? I didn't see a similar line in `webgl2.js`.

externs/browser/webgl2_compute.js
hujiajie: I'm also not sure about this. Just following existing externs with no more concrete reason.

externs/browser/webgl2_compute.js
lauraharker: For me, it's helpful to have a link pointing directly to the `WebGL2ComputeRenderingContext` section, as well as to the spec as a whole. I acknowledge that's a matter of preference, though.

externs/browser/webgl2_compute.js
lauraharker: It looks correct to me given https://www.w3.org/TR/WebIDL-1/#es-sequence, although I think `?Array<?>` would be equivalent in practice. Unless there's something I'm missing.

lauraharker:

lauraharker:
This was fixed in https://github.com/google/closure-compiler/commit/7ca7bc1f1249fe2711ee1310805f5cfe943346e5, but I misconfigured the commit message scrubbing and so it didn't mark this issue as closed. Sorry for the delay.

PR #3389, opened 2019-05-16
Selection api nonnull updates

realityforge:
Specify values as nonnull in Selection API to align with the spec
    
This includes:
  - the return type of `Selection.prototype.getRangeAt`
  - the `range` parameter of `Selection.prototype.addRange`
  - the `range` parameter of `Selection.prototype.removeRange`
  - the `node` parameter of `Selection.prototype.extend`
  - the `node` parameter of `Selection.prototype.containsNode`
  - the `node` parameter of `Selection.prototype.selectAllChildren`
  - the `anchorNode` and `focusNode` parameters of `Selection.prototype.setBaseAndExtent`
  - the `event` parameter of the `Element.prototype.onselectstart` callback
    
The parameters in Selection.prototype.setBaseAndExtent have also be
updated to align with the names used in the spec.

The return type of `Element.prototype.onselectstart` was also marked as void

### DIFF
--- a/externs/browser/gecko_dom.js
+++ b/externs/browser/gecko_dom.js
@@ -506,7 +506,7 @@ Selection.prototype.isCollapsed;
 Selection.prototype.rangeCount;
 
 /**
- * @param {Range} range
+ * @param {!Range} range
  * @return {undefined}
  * @see https://w3c.github.io/selection-api/#dom-selection-addrange
  */
@@ -514,7 +514,7 @@ Selection.prototype.addRange = function(range) {};
 
 /**
  * @param {number} index
- * @return {Range}
+ * @return {!Range}
  * @nosideeffects
  * @see https://w3c.github.io/selection-api/#dom-selection-getrangeat
  */
@@ -549,7 +549,7 @@ Selection.prototype.collapseToEnd = function() {};
 Selection.prototype.collapseToStart = function() {};
 
 /**
- * @param {Node} node
+ * @param {!Node} node
  * @param {boolean=} allowPartialContainment
  * @return {boolean}
  * @nosideeffects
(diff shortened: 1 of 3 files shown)


realityforge:
This is the second PR extracted from google/closure-compiler#3370

lauraharker:
Created internal issue http://b/132988232

EatingW:
Imported for internal review and testing. It is very likely that some internal code might be depending on this behavior, or just have {Node} everywhere instead of {!Node} such that fixing one might cause a large cascade of fixes. I'll run some overnight tests to assess the scale of the problem and report back.

EatingW:
Update on the testing results I have seen so far:
It does seem that a lot of cascading type annotations need to be fixed in order to land this. The main problem is that some of these get consumed by other externs, so these externs need their types fixed as well, which might cause other problems down the road :/ I have not yet assessed the scale of things, so I can't be certain which might be able to get fixed. So far, issues I have seen are with these types:
- the `range` parameter of `Selection.prototype.addRange`
- the `node` parameter of `Selection.prototype.extend`
- the `node` parameter of `Selection.prototype.selectAllChildren`
- the `anchorNode` and `focusNode` parameters of `Selection.prototype.setBaseAndExtent`
- the `event` parameter of the `Element.prototype.onselectstart` callback

The rest I might be able to land, but this is really not much left. Would you like me to land these first?
  - the return type of `Selection.prototype.getRangeAt`
  - the `range` parameter of `Selection.prototype.removeRange`
  - the `node` parameter of `Selection.prototype.containsNode`

realityforge:
That would be great ... but whatever is easiest for you works for me too. Thanks

EatingW:
The smaller change that I mentioned in my previous comment landed internally, so you should be able to see that in the next github push. 

As for the rest, I experimented with changing them, and fixing some of the externs that are impacted by the return value change, but the list of things that needed fixing seems to go on and on... I'm afraid I'll have to give up on fixing those. Would that be ok with you? Sorry about not being able to do much here!

realityforge:
@EatingW Thanks for the effort you put in. If need be I can patch the externs in `google/elemental2` so that the java code is generated correctly.

Thanks again!

EatingW:
Sounds good, thanks for your understanding!

PR #3391, opened 2019-05-21
Remove extern for ViewCSS.prototype.getMatchedCSSRules

realityforge:
The api was deprecated in 2014, disabled by default since M63 and removed
Apr 2018.

See https://bugs.chromium.org/p/chromium/issues/detail?id=437569


lauraharker:
Created internal issue http://b/133254418

realityforge:
Thanks

PR #3392, opened 2019-05-21
Type ViewCSS as an interface to align with the spec

realityforge:
The incorrect typing results in the need to perform an unchecked cast
in downstream libraries. See google/elemental2#61 for details.

See https://www.w3.org/TR/DOM-Level-2-Style/css.html#CSS-ViewCSS


lauraharker:
Created internal issue http://b/133255448

lauraharker:
Importing for internal review

PR #3393, opened 2019-05-21
Move the standardized focus management APIs to html5.js

realityforge:
The @see annotations have also been updated to refer to
the whatwg specs. The parameter to Element.prototype.focus
was also changed to align with the spec.

This will fix google/elemental2#40


lauraharker:
Created internal issue http://b/133254399

realityforge:
Thanks

PR #3394, opened 2019-05-21
Move Element.prototype.id to w3c_dom1.js

realityforge:
Historically this id was defined on HTMLElement and SVGElement but
this is no longer the case - See [1]. As a result of moving id onto
Element it is now possible to remove the id property from these subtypes.

Fixes google/elemental2#38

[1] https://dom.spec.whatwg.org/index.html#interface-element


lauraharker:
Created internal issue http://b/133253965

realityforge:
Thanks!

PR #3395, opened 2019-05-21
Move the remaining window extension methods from cssom-view to w3c_css.js

realityforge:
These methods have been standardized and are no longer ie specific. Moving
them to w3c_css.js makes them available in google/elemental2

The @see jsdoc tags where updated with references to the spec and the
parameter names where updated to match those in the spec.


lauraharker:
Created internal issue http://b/133255329

PR #3396, opened 2019-05-21
Typedef the callback passed to requestAnimationFrame

realityforge:
Typedef the callback passed to requestAnimationFrame with a name that aligns with the name in the spec

Using a typedef ensures that the name generated by google/elemental2 is stable and meaningful.
This change also changes the return type to void for the callbacks in the non-standard
RequestAnimationFrame variants.

### DIFF
--- a/externs/browser/w3c_anim_timing.js
+++ b/externs/browser/w3c_anim_timing.js
@@ -23,7 +23,13 @@
  */
 
 /**
- * @param {function(number): undefined} callback
+ * @typedef {function(number): undefined}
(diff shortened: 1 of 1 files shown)


lauraharker:
Created internal issue http://b/133255230

realityforge:
I have updated this PR to use the typedef for all the different callbacks which _does_ change the type checking. This will also need to be simultaneously applied with google/elemental2#121

realityforge:
Thanks!

PR #3397, opened 2019-05-21
Move Element.prototype.outerHTML w3c externs as it is now standardized

realityforge:
See https://w3c.github.io/DOM-Parsing/#extensions-to-the-element-interface


lauraharker:
Created internal issue http://b/133253683

realityforge:
Thanks!

PR #3400, opened 2019-05-22
Fix return type of extern `Symbol.for()`

jey:
`Symbol.for()` returns `symbol` according to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/for

### DIFF
--- a/externs/es3.js
+++ b/externs/es3.js
@@ -47,7 +47,7 @@ Symbol.prototype.description;
 
 /**
  * @param {string} sym
- * @return {symbol|undefined}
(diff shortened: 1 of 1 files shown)


tjgq:
Thanks for the pull request! We need you to sign a CLA (see instructions in the comment above) before we can approve it. Can you sign it and comment here once it's done?

jey:
I signed it!

tjgq:
Created Google internal issue b/249885953.

PR #3402, opened 2019-05-24
Stop issuing global this warning for function with this type annotation

### LINKED ISSUE #3343: dangerous use of the global this object when function's type was annotated
I'm trying to make the function to have `this` by defining it in its type.

```js
/**
 * @type {function(this:{hello:string})}
 */
function test() {
  console.log(this.hello)
}

test.call({ hello: 'world' })
```
However, I get an error. Is this meant to happen despite me giving the hint as to what `this` is going to be ? 
```
t/this.js:5: WARNING - dangerous use of the global this object
  console.log(this.hello)
              ^^^^

```

 Basically my function must adhere to an interface but can have a custom `this`. When i define it like 
```js
/**
 * @this {{hello:string}}
 * @type {_functionInterface}
 */
function test() {
  console.log(this.hello)
}
```
I get another error because the annotations are incompatible (can't use 2 together). Just have to suppress the `globalThis` for now.

cshung:
Fixes #3343 

@brad4d

### DIFF
--- a/src/com/google/javascript/jscomp/CheckGlobalThis.java
+++ b/src/com/google/javascript/jscomp/CheckGlobalThis.java
@@ -19,6 +19,7 @@ package com.google.javascript.jscomp;
 import com.google.javascript.jscomp.NodeTraversal.Callback;
 import com.google.javascript.rhino.JSDocInfo;
 import com.google.javascript.rhino.Node;
+import com.google.javascript.rhino.JSTypeExpression;
 import com.google.javascript.rhino.Token;
 import com.google.javascript.rhino.jstype.FunctionType;
 import com.google.javascript.rhino.jstype.JSTypeNative;
@@ -101,11 +102,16 @@ final class CheckGlobalThis implements Callback {
         return false;
       }
 
-      if (n.getJSType() instanceof FunctionType) {
-        FunctionType functionType = (FunctionType)n.getJSType();
-        if (functionType.getTypeOfThis() != compiler.getTypeRegistry().getNativeType(JSTypeNative.UNKNOWN_TYPE))
-        {
-          return false;
+      if (jsDoc != null) {
+        JSTypeExpression functionType = jsDoc.getType();
+        if (functionType != null) {
+          Node functionNode = functionType.getRoot();
+          if (functionNode != null && functionNode.isFunction()) {
+            Node thisNode = functionNode.getFirstChild();
+            if (thisNode != null && thisNode.isThis()) {
+              return false;
+            }
+          }
         }
       }
 
--- a/src/com/google/javascript/jscomp/CheckGlobalThis.java
+++ b/src/com/google/javascript/jscomp/CheckGlobalThis.java
@@ -20,6 +20,8 @@ import com.google.javascript.jscomp.NodeTraversal.Callback;
 import com.google.javascript.rhino.JSDocInfo;
 import com.google.javascript.rhino.Node;
 import com.google.javascript.rhino.Token;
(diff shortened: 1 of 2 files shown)


EatingW:
Created internal issue http://b/133850941

src/com/google/javascript/jscomp/CheckGlobalThis.java
```diff
@@ -99,6 +101,14 @@ public boolean shouldTraverse(NodeTraversal t, Node n, Node parent) {
         return false;
       }
 
+      if (n.getJSType() instanceof FunctionType) {
+        FunctionType functionType = (FunctionType)n.getJSType();
+        if (functionType.getTypeOfThis() != compiler.getTypeRegistry().getNativeType(JSTypeNative.UNKNOWN_TYPE))
```
brad4d: I'm not sure this does what we want.

Could you add test cases to CheckGlobalThisTest to be sure?
Thanks.

cshung: Thanks for the feedback. Here is what I have done to test:

Content of pass.js
```js
/**
 * @type {function(this:{hello:string})}
 */
function test() {
  console.log(this.hello)
}

test.call({ hello: 'world' })
```

Content of fail.js
```js
/**
 * @type {function()}
 */
function test() {
  console.log(this.hello)
}

test.call({ hello: 'world' })
```

The goal of the issue, as I understand, is to get `pass.js` to pass and `fail.js` to fail. Here I am able to achieve the goal:

I ran:
`java -jar target/closure-compiler-1.0-SNAPSHOT.jar --compilation_level ADVANCED pass.js`

And the result is:
```js
console.log({a:"world"}.a);
```

Now I ran:
`java -jar target/closure-compiler-1.0-SNAPSHOT.jar --compilation_level ADVANCED fail.js`

And now the result is:
```
fail.js:5: WARNING - [JSC_USED_GLOBAL_THIS] dangerous use of the global this object
  console.log(this.hello)
              ^^^^

0 error(s), 1 warning(s), 83.3% typed
console.log({a:"world"}.a);
```

I was wondering about your comments in the issue around the `To fix this bug for the case where type checking is disabled` clause, so I attempted to disable type checking.

To my surprise, specifying `--use_type_for_optimization=false` doesn't change my experiment result, the pass.js still pass and fail.js still fail.

Without more information available, I just opened the pull request, thinking it is sufficient. I attempted to add this unit test and now it fails.

```Java
  public void testPass()
  {
    testSame(lines(
      "/**",
      " * @type {function(this:{hello:string})}",
      " */",
      "function test() {",
      "  console.log(this.hello)",
      "}"
    ));
  }
``` 

The test case failed because the `n.getJSType()` is null and so the logic to skip raising the warning is not triggered. This is probably the case you were talking about that I was unable to trigger.

Moving forward, it would be interesting to see how I can leverage the `jsdoc` available in line 95 to perform essentially the same check.

brad4d: Could you add your test cases to this file?

https://github.com/google/closure-compiler/blob/master/test/com/google/javascript/jscomp/CheckGlobalThisTest.java

Thanks.

brad4d:
FYI, I'm going on a week's vacation starting tomorrow, so I'm unlikely to respond to further updates until June 12.

cshung:
Hi @brad4d. it has been a week since 12th June, any updates?

brad4d:
imported and sent for internal testing

brad4d:
Submitted internally, so it should be in tomorrow's push to GitHub.
Sorry about the delay.

PR #3404, opened 2019-05-24
Update externs for CustomElementRegistry.

bicknellr:
I think I'll need to update test/com/google/javascript/jscomp/IntegrationTest.java too, but I don't have this building locally so I'll be debugging against Travis. :)


externs/browser/html5.js
```diff
@@ -5134,28 +5134,28 @@ Plugin.prototype.length;
...
+ * @param {!function(new:HTMLElement)} klass
  * @param {{extends: string}=} options
  * @return {undefined}
  */
 CustomElementRegistry.prototype.define = function (tagName, klass, options) {};
 
 /**
  * @param {string} tagName
- * @return {?function(new:HTMLElement)}
+ * @return {!function(new:HTMLElement)|undefined}
```
tjgq: Same as above.

externs/browser/html5.js
```diff
@@ -5134,28 +5134,28 @@ Plugin.prototype.length;
...
- * @see https://html.spec.whatwg.org/multipage/scripting.html#custom-elements
+ * @see https://html.spec.whatwg.org/multipage/custom-elements.html#customelementregistry
  * @constructor
  */
 function CustomElementRegistry() {}
 
 /**
  * @param {string} tagName
- * @param {function(new:HTMLElement)} klass
+ * @param {!function(new:HTMLElement)} klass
```
tjgq: Remove the `!` since `function` is already non-nullable. See https://google.github.io/styleguide/jsguide.html#jsdoc-nullability.

bicknellr: Updated.

externs/browser/html5.js
bicknellr: Also updated.

bicknellr:
There are references to CustomElementRegistry here, using the older types:
https://github.com/google/closure-compiler/blob/317bf98f6566ae9b796e1467ffaea898c5b0e9b7/test/com/google/javascript/jscomp/IntegrationTest.java#L828-L863
But neglecting to update them didn't fail the build on the last commit - should I do something there?

bicknellr:
Also https://github.com/google/closure-compiler/blob/317bf98f6566ae9b796e1467ffaea898c5b0e9b7/src/com/google/javascript/jscomp/resources.json

bicknellr:
I went ahead and updated IntegrationTest.java because it looks like that's meant to pretend to load a subset of the externs in html5.js?

tjgq:
Yes, it's a good idea to also update the copy of the externs in `IntegrationTest.java`. But `resources.json` is automatically generated, so it doesn't have to.

Created Google internal issue b/133518269. Imported for review and testing.

PR #3405, opened 2019-05-27
Move nonstandard indexdb symbols to separate extern

realityforge:
The intent is to only expose standard APIs to downstream consumers of google/elemental2


tjgq:
Created Google internal issue b/133796241.

PR #3406, opened 2019-05-27
Split nonstandard webkit symbols into separate extern

realityforge:
This results in only standard APIs being exposed via google/elemental2


tjgq:
Created Google internal issue b/133795465.

lauraharker:

lauraharker:
Update: I've been taking longer to import this because some projects in google depend directly on `w3c_rtc.js`, instead of just on the standard set of externs, and I have to update those projects individually. Probably this will be in next week.

realityforge:
Thanks for your effort to get this landed!

PR #3407, opened 2019-05-27
Move the nonstandard, browser specific Audio APIs to a separate extern

realityforge:


realityforge:
It should be noted that google/elemental2#123 needs to be applied at the same time that this is applied

tjgq:
Created Google internal issue b/133795291.

PR #3412, opened 2019-06-03
Move some standardized HTMLInputElement properties to html5.js

realityforge:
The change involves
* Moving `selectionStart`, `selectionEnd` and `setSelectionRange` symbols to `html5.js`
* Adding an additional optional parameter to `setSelectionRange` method defined by the spec and implemented by both chrome and gecko
* Adding the missing `HTMLInputElement.prototype.selectionDirection` symbol

The spec WebIDL is at https://html.spec.whatwg.org/#the-input-element while the gecko WebIDL is at https://github.com/mozilla/gecko-dev/blob/master/dom/webidl/HTMLInputElement.webidl#L121-L126

It should be noted that the existing typings used non-nullable properties for `selectionStart` and `selectionEnd` despite the spec indicating they are nullable. Chrome does not conform to the spec and always returns values for these properties (namely `0`) rather than `null`. It was decided to leave the typings as is to match browser implementations rather than aspirational typings defined by the specs.

### DIFF
--- a/externs/browser/gecko_dom.js
+++ b/externs/browser/gecko_dom.js
@@ -616,21 +616,6 @@ Element.prototype.click = function() {};
 Element.prototype.focus = function(focusOption) {};
 
 /** @type {number} */
-HTMLInputElement.prototype.selectionStart;
-
-/** @type {number} */
-HTMLInputElement.prototype.selectionEnd;
-
-/**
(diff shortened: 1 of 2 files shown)


EatingW:
Created internal issue http://b/134434056.

realityforge:
Thanks!

PR #3415, opened 2019-06-06
Enable support for polyfilling incomplete Promise implementations.

pope:
There are some TV browser devices that have partial support for Promises. Trying to use the Promise should let us know for sure if it's a reliable implementation or not.

Hiding this under a flag since most users won't need to worry about this.

### DIFF
--- a/src/com/google/javascript/jscomp/js/es6/promise/promise.js
+++ b/src/com/google/javascript/jscomp/js/es6/promise/promise.js
@@ -49,6 +49,7 @@ $jscomp.polyfill('Promise',
       return NativePromise;
     }
     try {
+      /** @suppress {checkTypes} */
       new NativePromise(function(resolve) {
         resolve();
       });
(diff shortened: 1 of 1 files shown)


pope:
Sorry for all of the small fixes here and there. But the outstanding build problems seems to be happening on a clean branch, not just my changes.

blickly:
Overall, the use case sounds quite similar to that for using
`--define=$jscomp.FORCE_POLYFILL_PROMISE=true` to me, so I think it's worth at least laying out the value that this approach has over that (and thus why you think it's worth adding the extra configuration for it).

blickly:
Created internal issue http://b/134706672

pope:
`--define=$jscomp.FORCE_POLYFILL_PROMISE=true` will force all cases to use the Polyfill. This new new flag will still rely on the native Promise except that it will run a little test code to make sure the Promise implementation is not incomplete. That said, I didn't want to introduce this check for all users of this Polyfill - since it really applies to the most dark corners of Browser support.

As a concrete example, there are some webOS TV devices were Promise is part of the global namespace, but it is dysfunctional. Doing this check should be good enough to expose the flaw and allow us to correct with a Polyfill.

Issue #3410, opened 2019-05-31
Support TypeScript bounded generic types notation


ikabirov:
We use TypeScript as type checker for Closure Compiler project. Typescript allows declare bounded generics in JSDoc. But it is not compatible with Closure Compiler ("WARNING - Bad type annotation.")

Is it possible to just basic template type? E.g. parse
```
/**
 * @param {T} obj
 * @template {{id: string}} T
 */
function hello(obj) {
	console.log(obj.id)
}
```
as
```
/**
 * @param {T} obj
 * @template T
 */
function hello(obj) {
	console.log(obj.id)
}
```

EatingW:
Created http://b/134435444 as a feature request.

PR #3416, opened 2019-06-06
Support TypeScript bounded generic types notation

ikabirov:
fix for https://github.com/google/closure-compiler/issues/3410

### DIFF
--- a/src/com/google/javascript/jscomp/parsing/JsDocInfoParser.java
+++ b/src/com/google/javascript/jscomp/parsing/JsDocInfoParser.java
@@ -974,6 +974,16 @@ public final class JsDocInfoParser {
             }
           }
 
+          String leftCurly = "{";
+          String rightCurly = "}";
+          int leftCurlyIndex = templateNames.indexOf(leftCurly);
+          int rightCurlyIndex = templateNames.indexOf(rightCurly);
+          if (leftCurlyIndex >= 0 && rightCurlyIndex > leftCurlyIndex)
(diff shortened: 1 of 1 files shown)


blickly:
Created internal issue http://b/134787937

blickly:
I think that the idea of creating a new type syntax that is ignored when type-checking is somewhat dangerous, in that folks using type-checking might get the impression that it's checking something that it isn't. Secondarily, changes like these can make it more difficult to land the full feature (real bounded generics) if people who were using the syntax have created code patterns that don't actually typecheck.

@shicks, do you have any other opinion on adding support for the syntax of TS-style bounded generics (See https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html#template) in advance of actually supporting bounded generics in the type system?

brad4d:
Assigned to @shicks to get an answer to the question asked by @blickly above

shicks:
I agree that we don't want to support the syntax until the semantics are also supported, for the same reason we don't want folks using type annotations that aren't actually checked - once they become checked, they're probably going to be wrong.  This is the reason Closure Templates had to invent a whole new syntax for their types when the time came, since too many folks were writing unchecked comments in the "first choice" format.

blickly:
OK, I think we're in agreement that we shouldn't add syntax only for this without type-system support.  @ikabirov, thanks for starting the discussion and sorry we didn't decide to merge the syntax-only support.

PR #3419, opened 2019-06-12
Remove the non-standard parameter opt_distance from History.prototype.back

### LINKED ISSUE #3418: opt_distance parameter in function definition for History.prototype.back
In pull request https://github.com/google/closure-compiler/pull/3330, the `opt_distance` parameter was included in the function definition for `History.prototype.back`

https://github.com/google/closure-compiler/blob/v20190528/externs/browser/html5.js#L3154

    /**
     * Goes back one step in the joint session history.
     * If there is no previous page, does nothing.
     *
     * @return {undefined}
     */
    History.prototype.back = function(opt_distance) {};

This causes the following warning when calling `History.prototype.back` with zero arguments

    WARNING - [JSC_WRONG_ARGUMENT_COUNT]
    Function History.prototype.back: called with 0 argument(s).
    Function requires at least 1 argument(s) and no more than 1 argument(s).

realityforge:
This was meant to be removed in #3330 as it is non-standard, not supported
in most browsers and easily replaceable with `history.go(-distance)`. Instead
#3330 incorrectly removed the type annotation but left the parameter in the
symbol definition, making this parameter required.

Fixes #3418

### DIFF
--- a/externs/browser/html5.js
+++ b/externs/browser/html5.js
@@ -3150,7 +3150,7 @@ function History() {}
  *
  * @return {undefined}
  */
-History.prototype.back = function(opt_distance) {};
+History.prototype.back = function() {};
 
 /**
  * Goes forward one step in the joint session history.


brad4d:
Internal issue created http://b/135116687

lauraharker:
Imported for internal review.

lauraharker:
Ok, it looks like there's one place in Google that would be broken by this. (It is explicitly targeting IE).

I can ask if that piece of code is still being used, but we may have to revert this to an optional parameter again.

lauraharker:
I've approved a change internally to add back the type annotation to make this optional, as the param being required was causing trouble for someone. So you might see that change in the next few days. I'm still in favor of landing this PR if the one internal blocker can be resolved.

realityforge:
Great - thanks

PR #3421, opened 2019-06-13
Cannot check the isFunctionType on null

arv:


brad4d:
Created internal issue http://b/135263512

PR #3423, opened 2019-06-22
Attempt to fix the travis builds

ChadKillingsworth:
Travis [switched the default linux distribution](https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment) and builds started failing. The error message wasn't helpful, but appeared to be the Node version being used.

Updating to Node 10 solves the problem.


lauraharker:
Created internal issue http://b/135950878 and sending for internal review.

Issue #3027, opened 2018-07-19
Stack overflow when using recursive type in combination with Array.prototype.map


cmacordex:
I get a stack overflow error when using a recursive type with Array.prototype.map.

If I replace the .map call with a loop that builds up the args in the `makeJsFunctionCall` function below, then the error goes away.

Tested locally and with appspot.

	// ==ClosureCompiler==
	// @compilation_level ADVANCED_OPTIMIZATIONS
	// @output_file_name default.js
	// ==/ClosureCompiler==

	/**
	 * @typedef {(string|number|boolean|null)}
	**/
	var JsonPrimitiveType;

	/**
	 * @typedef {Object<string,JsonType>!}
	**/
	var JsonObjectType;

	/**
	 * @typedef {Array<JsonType>!}
	**/
	var JsonArrayType;

	/**
	 * @typedef {(JsonPrimitiveType|JsonObjectType|JsonArrayType)}
	**/
	var JsonType;

	/**
	 * @param {string} func
	 * @param {Array<JsonType>!} args
	 * @return {string}
	**/
	function makeJsFunctionCall (func, args) {
		var inner = args.map(function (x) { return JSON.stringify(x); });
		return func + "(" + inner.join(", ") + ")";
	}

It produces this error:

23: java.lang.StackOverflowError
	at com.google.javascript.rhino.jstype.UnknownType.isSubtype(UnknownType.java:121)
	at com.google.javascript.rhino.jstype.UnknownType.isSubtype(UnknownType.java:115)
	at com.google.javascript.rhino.jstype.JSType.isSubtypeOf(JSType.java:1748)
	at com.google.javascript.jscomp.TypeInference.maybeResolveTemplatedType(TypeInference.java:1483)
	at com.google.javascript.jscomp.TypeInference.maybeResolveTemplatedType(TypeInference.java:1432)
	at com.google.javascript.jscomp.TypeInference.maybeResolveTemplatedType(TypeInference.java:1492)
	... many lines removed ...

brad4d:
@shicks could you take a look at this issue?

PR #3431, opened 2019-07-14
Allow recursive union type definitions

visj:
This PR fixes a few different cases of infinite recursion errors for templatized and union type definitions. 

Fixes https://github.com/google/closure-compiler/issues/3027, added test case for this as well. Also fixes existing issue b/112964849 (internal issue? not sure).

The `Set<JSType> seenTypes` already existed in `ModificationVisitor`, so I added functionality in other classes to conform to this previous convention. However - as this is something that affects many parts of TemplatizedTypes and UnionTypes - there might be a better way to abstract this than how I implemented it. I'll leave that for discussion.

Regarding travis CI build: I merged master branch into my own this morning and it seems like the closure-compiler-version-SNAPSHOT.jar no longer builds? Not sure if this is an issue upstream or just my fork, would appreciate advice on this.

### DIFF
--- a/src/com/google/javascript/jscomp/AbstractCompiler.java
+++ b/src/com/google/javascript/jscomp/AbstractCompiler.java
@@ -649,7 +649,7 @@ public abstract class AbstractCompiler implements SourceExcerptProvider, Compile
    * currently being compiled.
    */
   public AstAnalyzer getAstAnalyzer() {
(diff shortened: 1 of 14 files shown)


lauraharker:
Created internal issue http://b/137759265

concavelenz:
We appreciate your interest but we don't see this change as being the correct solution to this problem, if you would like to continue to persue this, please open an issue to discuss what might be done.

PR #3434, opened 2019-07-17
Remove Node.prototype.setUserData and supporting UserDataHandler

realityforge:
This has not been supported in browsers for a very long time. This change removes the functionality.

See https://developer.mozilla.org/en-US/docs/Web/API/Node/setUserData


lauraharker:
Created internal issue http://b/137762920

tjgq:
Importing for review and testing

realityforge:
Thanks

PR #3435, opened 2019-07-17
Remove RangeException as it is no longer supported in modern browsers

realityforge:


lauraharker:
Created internal issue http://b/137763037

PR #3440, opened 2019-07-24
Remove historic DOMConfiguration as no longer present in modern browsers

realityforge:


rishipal:
Created internal issue http://b/138396305

PR #3442, opened 2019-07-28
Remove historic Feature support.

realityforge:
This no longer seems to be support in any modern browser


rishipal:
Created internal issue http://b/138590048

PR #3446, opened 2019-07-30
[FIX] Element.hasPointerCapture is not supported

### LINKED ISSUE #3445: Element.hasPointerCapture is not supported
The function `Element.hasPointerCapture()` is currently not supported.

Relevant links:
https://www.w3.org/TR/pointerevents/#dom-element-haspointercapture
https://developer.mozilla.org/en-US/docs/Web/API/Element/hasPointerCapture

marci4:
Fixes #3445


rishipal:
Internal Google issue created http://b/138677191

PR #3463, opened 2019-09-06
Fix nullability of jQuery.makeArray extern

tdelmas:
Docs: https://api.jquery.com/jQuery.makeArray/
Code: https://code.jquery.com/jquery-3.3.0.js
```
	// results is for internal usage only
	makeArray: function( arr, results ) {
		var ret = results || [];

		if ( arr != null ) {
			if ( isArrayLike( Object( arr ) ) ) {
				jQuery.merge( ret,
					typeof arr === "string" ?
					[ arr ] : arr
				);
			} else {
				push.call( ret, arr );
			}
		}

		return ret;
	},
```

### DIFF
--- a/contrib/externs/jquery-3.3.js
+++ b/contrib/externs/jquery-3.3.js
@@ -1320,7 +1320,7 @@ jQuery.prototype.length;
 
 /**
  * @param {*} obj
- * @return {Array<*>}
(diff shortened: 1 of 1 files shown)


tjgq:
Created Google internal issue b/140769876. Importing for internal review and testing.

PR #3465, opened 2019-09-10
Remove some historic functions that not present in modern browsers

realityforge:
I can break this commit up if you prefer but given how old these properties are and how unlikely they are to be used, it seems like less work to perform change in a single commit

### DIFF
--- a/externs/browser/gecko_dom.js
+++ b/externs/browser/gecko_dom.js
@@ -281,6 +281,8 @@ Document.prototype.tooltipNode;
  */
 Document.prototype.vlinkColor;
 
+/** @type {number} */ Document.prototype.width;
+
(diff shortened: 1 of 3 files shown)


brad4d:
Well my first attempt at testing failed due to a test infrastructure thing, probably totally unrelated to this change. Trying again...

brad4d:
The good news is that only 2 properties had to be added back.

See the commit I've pushed back to your branch.

Assuming you're cool with that, I'll submit the internal change.

realityforge:
Looks good to me

brad4d:
Submitted internally.
Will be in tomorrow's push to GitHub.

realityforge:
Thanks!

PR #3467, opened 2019-09-10
Changing HashMap to LinkedHashMap for deterministic iterations

adinstesters:
`com.google.javascript.jscomp.TypeInferenceTest#testNew4` can fail due to a different iteration order of a HashMap. The test parses some JS function with a specified order for the parameters, but the parameters are eventually inserted into a `HashMap` in `FunctionTypeBuilder`. When bounding the parameters, the `HashMap` in question is iterated over, but this iteration order is not guaranteed. However, the logic for building the templates implicitly assumes the order to be insertion order, as demonstrated by the test.  The proposed fix is to change the `HashMap` in question into a `LinkedHashMap` where order is guaranteed to be based on order of insertion.

### DIFF
--- a/src/com/google/javascript/jscomp/FunctionTypeBuilder.java
+++ b/src/com/google/javascript/jscomp/FunctionTypeBuilder.java
@@ -49,6 +49,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


adinstesters:
@googlebot I signed it!

brad4d:
Created internal issue http://b/140866070

PR #3472, opened 2019-09-24
Allow Number.is[Nan|Finite|Integer|SafeInteger] to take * input

joelhockey:
Spec gives the strong impression that any type could be
expected as input for these functions
https://www.ecma-international.org/ecma-262/6.0/#sec-number.isinteger

And practically, it is useful to allow code such as:

/** @type {number|string} */
let foo;

if (Number.isInteger(foo)) {
...
}

Currently, this code must be written as:

if (typeof foo == 'number' && Number.isInteger(foo)) {
...
}

See https://chromium-review.googlesource.com/c/apps/libapps/+/1816332

### DIFF
--- a/externs/es6.js
+++ b/externs/es6.js
@@ -1524,7 +1524,7 @@ Number.parseInt = function(string, radix) {};
 Number.parseFloat = function(string) {};
 
 /**
- * @param {number} value
+ * @param {*} value
  * @return {boolean}
  * @nosideeffects
  * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN
@@ -1532,7 +1532,7 @@ Number.parseFloat = function(string) {};
 Number.isNaN = function(value) {};
 
 /**
- * @param {number} value
+ * @param {*} value
  * @return {boolean}
  * @nosideeffects
  * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isFinite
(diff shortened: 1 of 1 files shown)


blickly:
I think that the general philosophy for most JS optional type systems is err on the side of having the typing make the most sense for developers, rather than allowing everything that the spec allows.

For comparison, it looks like TypeScript uses a similar typing as us, and they provide some of their rationale on their issue tracker: https://github.com/microsoft/TypeScript/issues/4002

blickly:
Created internal issue http://b/141640588

vapier:
i grok the "lets keep it narrow when it makes sense position" and don't disagree with that position.  i think this is a case where it doesn't make sense: the point of the `Number.isInteger` API is specifically to validate the input is an integer.  it isn't a matter of the type being implicitly coerced from e.g. string to number before being checked for its integerness.

blickly:
OK, in that case, it's probably a bug in TypeScript's typings as well, right?

cschuller:
@joelhockey Is there a pull-request for TypeScript open/closed?

shicks:
https://github.com/microsoft/TypeScript/pull/24436

Unfortunately the conversation has been locked for the last year.

shicks:
It looks like that PR was actually merged, but then reverted shortly thereafter in https://github.com/microsoft/TypeScript/commit/2f73986b44a4ec10c7ebe7188e23863c3879d54e

blickly:
I'm closing for now, but please feel free to reopen when the corresponding TypeScript PR sticks

PR #3482, opened 2019-10-14
explicit nullable types in externs

ikabirov:
We use custom closure compiler build with non-nullable types by default. So we can't use current externs with implicit nullable types


lauraharker:
@ikabirov could you fix the conflict in html5.js? Thanks!

Overall this looks good to me

lauraharker:
Importing this for internal review now

brad4d:
I've imported this PR and resolved its conflicts in the imported version.

brad4d:
I'm afraid I have to reject this PR.

Although it may seem like it should be a no-op change, in fact it causes breakages in a lot of projects.
Also, since this is a broad change, touching lots of totally unrelated declarations, it causes a very wide range of types of breakage.

I don't think the benefit we would get from landing this as it is would justify the large amount of time I believe would be spent going through Google's code base and rooting out all of the individual cases and fixing them.

I believe we could accept a series of smaller, much more targeted PRs, where each one changes only a handful of closely-related declarations.

If there are particular declarations that you'd especially like to see fixed, I'd suggest you try that.
Sorry.

PR #3485, opened 2019-10-19
Prevent renaming of the FS module method

ChadKillingsworth:
The j2cl build does not recognize the readFileSync method utilized to read files when running in a Node environment. Use quoted syntax to prevent renaming of the method.


lauraharker:
Created internal Google issue http://b/143616189.

nreid260:
Is there a test I can used to verify the fix?

PR #3512, opened 2019-11-27
Check for JsonNull and null in getJavaStringArray()

kwivix:
If gson encouters a `null` value it returns a `JsonNull` object instead of `null`.


blickly:
Internal Google issue http://b/145525915 created

PR #3531, opened 2020-01-03
Add extern for `Blob.arrayBuffer()`.

inker610566:
The API is documented here:
http://www.w3.org/TR/FileAPI/#arraybuffer-method-algo


inker610566:
We add annotation to our own extern files. No longer need this CL.

ShikChen:
Please keep this open. We would like to have it in upstream Closure Compiler since it's a standard API.

PR #3564, opened 2020-02-28
Move webassembly.js from "contrib" to "browser"

### LINKED ISSUE #3552: WebAssembly externs likely incorrectly located in "contrib"
As noted in #2533 by @RReverser:

> An up-to-date Closure Compiler is complaining about variable WebAssembly being undeclared.
> Is it because externs are in contrib/externs/... as opposed to externs/browser/... where they were added initially in a PR #2534? All the other APIs in contrib/externs/... seem to be for 3rd-party libs, whereas WebAssembly is a stable API available in all browsers, seems weird that it's not colocated with other standard APIs.

Aransentin:
Fixes #3552


Aransentin:
Fixed by a323968e315b2da556f7d89bb93a5f51ce6061df instead.

PR #3573, opened 2020-03-25
Add optimization to replace `Array.of` with an array literal

### LINKED ISSUE #3566: Array.of can be optimized
The compiler should always be able to optimize `Array.of` to `[]` when `language_out` is ES6. Is this something that can be added?

Example:

```js
function arrayOf(...values) {
    return Array.of(...values);
}
console.log(arrayOf(1, 2, 3));
```

```
$ closure-compiler --language_out ES6 --compilation_level ADVANCED array_of.js
```

Output (v20200224):

```js
'use strict';console.log(Array.of(...[1,2,3]));
```

Expected output:

```js
'use strict';console.log([1,2,3]);
```

bugreportuser:
Closes #3566


test/com/google/javascript/jscomp/PeepholeReplaceArrayOfTest.java
```diff
@@ -0,0 +1,76 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/**
+ * Unit tests for {#link {@link PeepholeReplaceArrayOf}
+ *
+ */
+@RunWith(JUnit4.class)
+public final class PeepholeReplaceArrayOfTest extends CompilerTestCase {
```
niloc132: Consider perhaps a few tests of things that shouldn't change? Array.of.apply, etc? Not sure how realistic those are though.

src/com/google/javascript/jscomp/PeepholeReplaceArrayOf.java
```diff
@@ -0,0 +1,68 @@
...
+            break;
+          default: // fall out
+        }
+      }
+    }
+
+    return subtree;
+  }
+
+  private strictfp Node replaceArrayOf(Node subtree) {
```
niloc132: I don't think `strictfp` makes sense here.

bugreportuser: Thanks. That was accidentally left from PeepholeReplaceKnownMethods.java.

test/com/google/javascript/jscomp/PeepholeReplaceArrayOfTest.java
bugreportuser: Good idea.

bugreportuser:
@blickly Can you review these changes? It doesn't look like PRs have been reviewed recently.

lauraharker:
Hi, sorry about the lack of support for recent PRs.

We have a triage process to review all incoming bugs/feature requests/customer issues. Unfortunately at some point, we changed how GitHub PRs are mirrored internally at Google, which caused them to stop showing up in our triage queue.

I'll ask Ben to look at this PR and then work on fixing our triage process.

bugreportuser:
Thanks for the explanation @lauraharker. It's great to see that the review process will be operational again.

blickly:
Hi, sorry this took so long to work on. I think I've got the import process working now.

src/com/google/javascript/jscomp/PeepholeReplaceArrayOf.java
```diff
@@ -0,0 +1,68 @@
...
+  private Node tryReplaceArrayOf(Node subtree) {
+    checkArgument(subtree.isCall(), subtree);
+    Node callTarget = checkNotNull(subtree.getFirstChild());
+
+    if (NodeUtil.isGet(callTarget)) {
+      Node clazz = callTarget.getFirstChild();
+      if (clazz.isQualifiedName()) {
+        switch (clazz.getQualifiedName()) {
+          case "Array":
+            if (clazz.getNext().getString().equals("of")) {
```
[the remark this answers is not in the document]
bugreportuser: That's sensible and now fixed.

src/com/google/javascript/jscomp/PeepholeReplaceKnownMethods.java
```diff
@@ -73,10 +74,15 @@ private Node tryFoldKnownMethods(Node subtree) {
       Node callTarget = checkNotNull(subtree.getFirstChild());
 
       if (NodeUtil.isGet(callTarget)) {
-        if (isASTNormalized() && callTarget.getFirstChild().isQualifiedName()) {
```
blickly, line 76: I'm not sure I follow why this `isASTNormalized()` check is incompatible with the new optimization.

I would have thought that waiting for the AST to be normalized would work just as well with "Array" as it did with "Math". Doesn't it?

bugreportuser: I don't know either. I couldn't figure out why that method is used. Is this optimization called multiple times but only once when the AST is normalized?

The strange part is when I use that condition, there's never any change for any `Array.of` call and no errors. Since the optimization only cares about the function name, I thought it would be fine to remove to fix the problem.

blickly: These passes are always run after normalization, but there is also a "late peephole passes" run that runs after denormalization when the names may no longer be unique.

I think that by that point, any calls to Array.of should have already been replaced, so it should be reasonable to backoff here (just in the very uncommon case that denormalization named another namespace to "Array").

bugreportuser: I see. Is there a problem with my tests then? I only changed it because [this commit](https://github.com/google/closure-compiler/pull/3573/commits/e6d3e055e6ee93e7a9d4b61a47900fcc1088b142) failed.

blickly: Oh interesting.  Looking at the similar `Math.abs` case, for example, it looks like the unit tests need an explicit call to `enableNormalize();`

Does that work?

bugreportuser: That did it! Will it normalize automatically for [this test](https://github.com/google/closure-compiler/pull/3573/files#diff-a41f00968b9e5a04015c0b2ecaacfcffR8106), since it's an integration test?

blickly: That's right.  In the integration tests, there's no way to avoid normalization.

bugreportuser: Great, so I think this is done if you approve.

blickly: Looks good. Sorry it was slow to figure out the new process, but thanks for going through the whole code review process!

Importing with copybara now.

bugreportuser: Don't worry. This change wasn't urgent.

And thanks for being so helpful through it!

PR #3602, opened 2020-05-26
DocumentRange should be an interface

niloc132:
From the linked spec, this is the idl, indicating that this should not have a constructor, but just be an interface.
```
// Introduced in DOM Level 2:
interface DocumentRange {
  Range              createRange();
};
```
Would it be more appropriate to also/instead declare `Document.prototype.createRange=function(){};` in this same file?


blickly:
Looks good, thanks! This should be merged in the next push.

PR #3603, opened 2020-05-26
Add missing optional arguments to `Path2D` constructor in extern definition

### LINKED ISSUE #3598: Path2D externs definition is missing optional constructor arguments
The `Path2D` constructor has an optional argument that is either a string or another `Path2D` object: https://developer.mozilla.org/en-US/docs/Web/API/Path2D/Path2D

The externs are currently missing this https://github.com/google/closure-compiler/blob/master/externs/browser/html5.js#L264

Version: any version of Closure.

timdawborn:
Fixes #3598

### DIFF
--- a/externs/browser/html5.js
+++ b/externs/browser/html5.js
@@ -258,10 +258,11 @@ CanvasPathMethods.prototype.arc = function(
 
 /**
  * @constructor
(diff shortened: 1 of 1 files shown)


blickly:
Looks good, thanks!

This should be merged in the next push.

PR #3619, opened 2020-06-17
Test travis build when upgrading to GWT 2.9

jDramaix:
Do not review nor submit. See cl/316743154


lauraharker:
@jDramaix can we revert this after https://github.com/google/closure-compiler/commit/d43678e994af52668ee56ed01c3343100d8340f9?

PR #3620, opened 2020-06-22
Added enum flag to allow for code instrumentation and refactored code to better support CoverageOption Enum

patrick-sko:


src/com/google/javascript/jscomp/CommandLineRunner.java
```diff
@@ -851,6 +852,17 @@
     )
     private boolean helpMarkdown = false;
 
+    @Option(
+            name = "--instrument_code",
+            usage = "Enable code instrumentation to perform code coverage analysis. Options are: "
+            + " 1. NONE (deault)"
```
nbeloglazov, line 858: Can you run compiler with --help and see how this is printed? This will be a single line, maybe you can you add "\n" at the end?

src/com/google/javascript/jscomp/CommandLineRunner.java
```diff
@@ -971,6 +991,7 @@ private void parse(List<String> args) throws CmdLineException {
                     "help",
                     "third_party",
                     "use_types_for_optimization",
+                        "instrument_code",
```
nbeloglazov, line 994: Odd indentation.

src/com/google/javascript/jscomp/Compiler.java
```diff
@@ -94,6 +95,7 @@
 import java.util.regex.Matcher;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
+import javax.sound.midi.Instrument;
```
nbeloglazov, line 98: Accidental import?

src/com/google/javascript/jscomp/Compiler.java
```diff
@@ -892,18 +894,20 @@ public void instrumentForCoverage() {
         () -> {
           checkState(options.getInstrumentForCoverageOnly());
           checkState(!hasErrors());
-          instrumentForCoverageInternal(options.instrumentBranchCoverage);
+          if(options.instrumentForCoverageOption != InstrumentOption.NONE) {
+            instrumentForCoverageInternal(options.instrumentForCoverageOption);
+          }else if(options.instrumentForCoverage){
```
nbeloglazov, line 899: add spaces between `}` and `else`.

src/com/google/javascript/jscomp/CompilerOptions.java
```diff
@@ -1189,12 +1190,16 @@ public boolean getAssumeStaticInheritanceRequired() {
    */
   transient ErrorHandler errorHandler;
 
+  public InstrumentOption instrumentForCoverageOption;
+
   /**
    * Instrument code for the purpose of collecting coverage data.
    */
+  @Deprecated
```
nbeloglazov, line 1198: Update comment to mention that `instrumentForCoverageOption` should be used insted.

src/com/google/javascript/jscomp/CompilerOptions.java
```diff
@@ -1189,12 +1190,16 @@ public boolean getAssumeStaticInheritanceRequired() {
    */
   transient ErrorHandler errorHandler;
 
+  public InstrumentOption instrumentForCoverageOption;
```
nbeloglazov, line 1193: Add comment explaining what it does.

src/com/google/javascript/jscomp/DefaultPassConfig.java
```diff
@@ -570,8 +570,13 @@ private void addModuleRewritingPasses(List<PassFactory> checks, CompilerOptions
       passes.add(j2clUtilGetDefineRewriterPass);
     }
 
+    /*
```
nbeloglazov, line 573: Remove commented out code.

src/com/google/javascript/jscomp/DefaultPassConfig.java
```diff
@@ -570,8 +570,13 @@ private void addModuleRewritingPasses(List<PassFactory> checks, CompilerOptions
...
     if (options.instrumentForCoverage) {
       passes.add(instrumentForCodeCoverage);
+    }*/
+
+    if(options.instrumentForCoverageOption != null && options.instrumentForCoverageOption != InstrumentOption.NONE){
```
nbeloglazov, line 578: Can instrumentForCoverageOption be null? It would be good if we could guarantee it's always set to some value.

src/com/google/javascript/jscomp/bundle/CoverageInstrumenter.java
```diff
@@ -49,6 +49,7 @@ protected void setOptions(CompilerOptions options) {
       options.setSourceMapOutputPath("/dev/null");
       options.setVariableRenaming(VariableRenamingPolicy.OFF);
       options.instrumentForCoverage = true;
+      //options.setInstrumentForCoverageOption(2);
```
nbeloglazov, line 52: Remove?

src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
```diff
@@ -1231,6 +1231,8 @@ protected int doRun() throws IOException {
     }
 
 
+
```
lauraharker, line 1234: revert?

src/com/google/javascript/jscomp/CompilerOptions.java
lauraharker: Also please make this private or package-private

public members are harder to refactor

src/com/google/javascript/jscomp/CompilerOptions.java
lauraharker: Hmm I didn't realize that these were public booleans instead of getter/setter methods.

I was hoping these would be methods we could just make call your new enum setter, so we could preserve the current behavior.

Given that, I'd prefer either:
 - making these options still enable coverage in DefaultPassConfig
 - or if you think that is too unwieldly, just deleting them so that people will get a compiler error if using them instead of a silent behavior change

test/com/google/javascript/jscomp/CommandLineRunnerTest.java
```diff
@@ -2532,6 +2532,34 @@ public void testWebpackModuleIds() throws IOException {
                 + "\"}]");
   }
 
+  @Test
+  public void testInstrumentCodeByLine() {
+    args.add("--instrument_code=LINE");
+
+    test(
+            "function foo(){ const answerToAll = 42; }",
+            "self.window||(self.window=self,self.window.top=self);var __jscov=window.top.__jscov||"
```
lauraharker, line 2541: For newer unit tests we've been using the `lines` helper function instead of string concatenation to create long strings.

Could you replace this test case and the one below with 

```
lines(
    "self.window||(self.window=self,self.window.top=self);var __jscov=window.top.__jscov||",
    "(window.top.__jscov={fileNames:[],instrumentedLines:[],executedLines:[]}),",
[ etc. ]
)

File: src/com/google/javascript/jscomp/CommandLineRunner.java [hunk 12]
```diff
@@ -851,6 +852,17 @@
     )
     private boolean helpMarkdown = false;
 
+    @Option(
+        name = "--instrument_code",
+        usage = "Enable code instrumentation to perform code coverage analysis. Options are:\n"
+            + " 1. NONE (deault)\n"
```
L3P3, line 858: deault?

Issue #2459, opened 2017-04-27
INTERNAL COMPILER ERROR with branch coverage enabled


ribrdb:
I compiled with
       options.setInstrumentForCoverage(true);
      options.setInstrumentBranchCoverage(true);

java.lang.RuntimeException: java.lang.RuntimeException: INTERNAL COMPILER ERROR.
Please report this problem.

null
  Node(FOR): external/com_google_javascript_closure_library/closure/goog/base.js:134:2
  for (var part; parts.length && (part = parts.shift());) {
  Parent(BLOCK): external/com_google_javascript_closure_library/closure/goog/base.js:117:68
goog.exportPath_ = function(name, opt_object, opt_objectToExportTo) {

	at com.google.javascript.jscomp.CompilerExecutor.runInCompilerThread(CompilerExecutor.java:126)
	at com.google.javascript.jscomp.Compiler.runInCompilerThread(Compiler.java:740)
	at com.google.javascript.jscomp.Compiler.compile(Compiler.java:710)
	at com.google.javascript.jscomp.Compiler.compile(Compiler.java:680)
	at com.google.javascript.jscomp.AbstractCommandLineRunner.doRun(AbstractCommandLineRunner.java:1080)
	at com.google.javascript.jscomp.JsCompilerRunner.go(JsCompilerRunner.java:47)
	at com.google.javascript.jscomp.JsCompiler.run(JsCompiler.java:198)
	at com.google.javascript.jscomp.JsCompiler.apply(JsCompiler.java:82)
	at io.bazel.rules.closure.worker.LegacyAspect.run(LegacyAspect.java:38)
	at io.bazel.rules.closure.ClosureWorker.run(ClosureWorker.java:72)
	at io.bazel.rules.closure.worker.PersistentWorker.runProgram(PersistentWorker.java:109)
	at io.bazel.rules.closure.worker.PersistentWorker.runAsPersistentWorker(PersistentWorker.java:144)
	at io.bazel.rules.closure.worker.PersistentWorker.run(PersistentWorker.java:84)
	at io.bazel.rules.closure.ClosureWorker.main(ClosureWorker.java:111)
Caused by: java.lang.RuntimeException: INTERNAL COMPILER ERROR.
Please report this problem.

null
  Node(FOR): external/com_google_javascript_closure_library/closure/goog/base.js:134:2
  for (var part; parts.length && (part = parts.shift());) {
[...]

PR #3626, opened 2020-06-29
Fixed bugs to improve branch instrumentation

patrick-sko:
Solves three different bugs:

1) NPE that was raised when a for loop occurred at the end of a function
2) Illegal State exception that was raised when a for loop was nested within a for loop
3) NPE that was raised when a source file had a branching condition on line number and a Polyfill js file which is injected into the source file also has a branching condition at the same line number. The unit test which catches this `ForLoopBranchConflictWithPolyfill.jsdata` has a branching condition on line 21 which matches to a branch condition in `find.js` within the project.

Added three unit tests to help prevent future occurrences of these bugs.

This fix should also solve the following issue: https://github.com/google/closure-compiler/issues/2459


test/com/google/javascript/jscomp/GoldenFileComparer.java
```diff
@@ -104,6 +105,33 @@ private static void compileAndCompare(
     compare(compiledSource, referenceSource);
   }
 
+  /**
+   * Compile sourceFileName with given CompilerOptions and check that all lines of code from
+   * goldenFileName are present in compiled sourceFile.
+   */
```
nbeloglazov, line 111: Would be good to mention use case of this function: when output contains polyfills and polyfills should be ignored.

src/com/google/javascript/jscomp/BranchCoverageInstrumentationCallback.java
```diff
@@ -89,11 +94,16 @@ public void visit(NodeTraversal traversal, Node node, Node parent) {
...
-          if (destination.isBlock()) {
+          if (destination != null && destination.isBlock()) {
             blocks.add(destination);
           } else {
             Node exitBlock = IR.block();
-            destination.getParent().addChildBefore(exitBlock, destination);
+            if (destination != null && destination.getParent().isBlock()) {
+              destination.getParent().addChildBefore(exitBlock, destination);
+            } else {
+              outEdge.getSource().getValue().getParent()
```
nbeloglazov, line 104: Personally when I'm reading compiler code it's useful to describe each case in comments as what exactly it handles. Otherwise it's hard to visualize AST tree here by just looking at checks in `if` statements.

test/com/google/javascript/jscomp/CoverageInstrumentationPassTest.java
```diff
@@ -165,4 +165,31 @@ public void testEsModule() throws Exception {
...
+    GoldenFileComparer.compileAndCompare(
+        "CoverageInstrumentationPassTest/ForLoopBranchNoEndingBlockGolden.jsdata",
+        branchOptions(LanguageMode.STABLE),
+        "CoverageInstrumentationPassTest/ForLoopBranchNoEndingBlock.jsdata");
+  }
+
+  @Test
+  public void testForLoopConflictWithPolyfill() throws Exception {
+    CompilerOptions options = branchOptions(LanguageMode.STABLE);
+    options.setRewritePolyfills(true);
```
lauraharker, line 188: Consider using `options.forceLibraryInjection` instead of `options.RewritePolyfills` to make this less brittle.

Also I'd suggest adding a comment on what exactly this case covers and explaining that it will become less useful if the relevant polyfill file changes.

test/com/google/javascript/jscomp/GoldenFileComparer.java
```diff
@@ -104,6 +105,33 @@ private static void compileAndCompare(
...
+
+    int identicalLinesBetweenFiles = 0;
+
+    for (int i = 0; i < compiledLines.length; i++) {
+      if (goldenLines[identicalLinesBetweenFiles].equals(compiledLines[i])) {
+        identicalLinesBetweenFiles++;
+      }
+    }
+
+    assertThat(identicalLinesBetweenFiles).isEqualTo(expectedNumberOfIdenticalLines);
```
lauraharker, line 132: If this assertion fails the error message won't be very helpful.

Are the identical lines always going to be the last `expectedNumberOfIdenticalLines` of `compiledLines`? If so we could instead just truncate `compiledLines` then expect exact string equality.

test/com/google/javascript/jscomp/GoldenFileComparer.java
```diff
@@ -104,6 +105,33 @@ private static void compileAndCompare(
     compare(compiledSource, referenceSource);
   }
 
+  /**
+   * Compile sourceFileName with given CompilerOptions and check that all lines of code from
+   * goldenFileName are present in compiled sourceFile.
```
lauraharker, line 110: Also mention that the golden file lines must appear in the same order as in the golden file, not just in any arbitrary order.

src/com/google/javascript/jscomp/BranchCoverageInstrumentationCallback.java
```diff
@@ -49,6 +49,11 @@ public BranchCoverageInstrumentationCallback(
   public void visit(NodeTraversal traversal, Node node, Node parent) {
     String fileName = traversal.getSourceName();
 
+    // If origin of node is not from sourceFile, do not instrument
```
lauraharker, line 52: Add more context here - why don't we instrument? When does this happen?

test/com/google/javascript/jscomp/GoldenFileComparer.java
patrick-sko: Updated.

src/com/google/javascript/jscomp/BranchCoverageInstrumentationCallback.java
patrick-sko: Updated.

test/com/google/javascript/jscomp/CoverageInstrumentationPassTest.java
patrick-sko: Updated.

test/com/google/javascript/jscomp/GoldenFileComparer.java
patrick-sko: The beginning and end lines of the compiledLines are what is expected to match with expectedNumberOfIdenticalLines. Changed to add a message of what is the first line that is not found.

test/com/google/javascript/jscomp/GoldenFileComparer.java
patrick-sko: Updated.

src/com/google/javascript/jscomp/BranchCoverageInstrumentationCallback.java
patrick-sko: Updated.

PR #3634, opened 2020-07-10
Update typing for Performance.prototype.mark

samouri:
The spec includes a second optional param called `PerformanceMarkOptions`.
See: https://w3c.github.io/user-timing/#extensions-performance-interface


samouri:
Learned that this is still a draft spec. closing

PR #3643, opened 2020-07-17
Implementation of function instrumentation and new callback

patrick-sko:
This PR represents the first step in a series of PRs for implementing compiled code instrumentation as explained here: https://docs.google.com/document/d/1S84jaH_57-vxhqVMoMQb8pfMZ-griTtU09V3YbHkAeI/edit?usp=sharing

This PR contains some of the basic functionality that will be expanded on when adding additional features. This includes encoding parameter values as defined in the above doc as well as defines constants for the Instrument Code function name. 
It also implements function instrumentation.


src/com/google/javascript/jscomp/AdvancedCoverageInstrumentationCallback.java
```diff
@@ -0,0 +1,205 @@
+package com.google.javascript.jscomp;
```
lauraharker, line 1: All Java files need a license included. You can copy one from any other file in the repo like CoverageInstrumentationCallback.java.

src/com/google/javascript/jscomp/AdvancedCoverageInstrumentationCallback.java
```diff
@@ -0,0 +1,205 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+
+
+/**
+ * Instrument advanced coverage for javascript.
+ */
+@GwtIncompatible("AdvancedFileInstrumentationData")
+public class AdvancedCoverageInstrumentationCallback extends
```
lauraharker, line 21: Can this be package-private instead of public?

src/com/google/javascript/jscomp/AdvancedCoverageInstrumentationCallback.java
```diff
@@ -0,0 +1,205 @@
...
+    // avoid instrumentation as this callback does not distinguish between sourceFile code and
+    // injected code and can result in an error.
+    if (!Objects.equals(fileName, sourceFileName)) {
+      return;
+    }
+
+    // If Source File is base.js or INSTRUMENT_CODE_FILE_NAME, do not instrument as the instrument
+    // function has not yet been defined when base.js will be executing and the implementation file
+    // can not call the Code instrumentation function on itself
+    if (sourceFileName.endsWith("base.js") || sourceFileName.endsWith(INSTRUMENT_CODE_FILE_NAME)) {
```
lauraharker, line 54: Let's not hardcode "base.js" here. Could we instead have a check like "if this is the first source file in the program, do not instrument"?

src/com/google/javascript/jscomp/CompilerOptions.java
```diff
@@ -3142,7 +3142,8 @@ public String toString() {
   public enum InstrumentOption {
     NONE, // No coverage instrumentation is performed
     LINE_ONLY, // Collect coverage for every executable statement.
-    BRANCH_ONLY; // Collect coverage for control-flow branches.
+    BRANCH_ONLY, // Collect coverage for control-flow branches.
+    ADVANCED;
```
lauraharker, line 3146: Please add a short comment describing what ADVANCED means

src/com/google/javascript/jscomp/CoverageInstrumentationPass.java
```diff
@@ -78,6 +79,13 @@ public void process(Node externsNode, Node rootNode) {
             compiler,
             rootNode,
             new BranchCoverageInstrumentationCallback(compiler, instrumentationData));
+      } else if (instrumentOption == InstrumentOption.ADVANCED) {
+        NodeTraversal.traverse(
+            compiler,
+            rootNode,
+            new AdvancedCoverageInstrumentationCallback(compiler, new HashMap<>()));
```
lauraharker, line 86: Is passing in the Map as a parameter necessary? Looks like it's never used, and could just be initialized in the callback itself.

If not, prefer to use LinkedHashMap here, as it has deterministic iteration order

test/com/google/javascript/jscomp/integration/AdvancedCoverageInstrumentationPassIntegrationTest.java
```diff
@@ -0,0 +1,138 @@
+package com.google.javascript.jscomp.integration;
```
lauraharker, line 1: This file also needs a license

test/com/google/javascript/jscomp/integration/AdvancedCoverageInstrumentationPassIntegrationTest.java
```diff
@@ -0,0 +1,138 @@
...
+import java.util.List;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public final class AdvancedCoverageInstrumentationPassIntegrationTest extends IntegrationTestCase {
+
+
+  private final String instrumentCodeSource = lines("goog.module('instrument.code'); ",
```
lauraharker, line 25: formatting looks off here

test/com/google/javascript/jscomp/integration/AdvancedCoverageInstrumentationPassIntegrationTest.java
```diff
@@ -0,0 +1,138 @@
...
+        options);
+
+    return compiler;
+  }
+
+  @Override
+  public CompilerOptions createCompilerOptions() {
+    CompilerOptions options = new CompilerOptions();
+
+    options.setLanguageOut(LanguageMode.ECMASCRIPT5_STRICT);
```
lauraharker, line 128: Could you also add tests verifying this works without transpilation?

test/com/google/javascript/jscomp/integration/AdvancedCoverageInstrumentationPassIntegrationTest.java
```diff
@@ -0,0 +1,138 @@
...
+    String[] sourceArr = {instrumentCodeSource, source};
+
+    String expected =
+        lines("var global = 23; console.log(global);");
+
+    String[] expectedArr = {instrumentCodeExpected, expected};
+    test(options, sourceArr, expectedArr);
+  }
+
+  protected void test(CompilerOptions options,
```
lauraharker, line 77: Is it necessary to override this method?

src/com/google/javascript/jscomp/AdvancedCoverageInstrumentationCallback.java
```diff
@@ -0,0 +1,205 @@
...
+   * nodes by the function name and what file they are enclosed within.
+   */
+  private static final class AdvancedFileInstrumentationData {
+
+    private final String fileName;
+    private final Map<String, List<Node>> functionScopedNodes;
+
+    AdvancedFileInstrumentationData(String fileName) {
+      this.fileName = fileName;
+      this.functionScopedNodes = new HashMap<>();
```
lauraharker, line 188: Prefer to use LinkedHashMap here, as it has deterministic iteration order

Even if this particular map is never actually iterated over it's still safest to avoid nondeterminism if this code is ever changed

src/com/google/javascript/jscomp/AdvancedCoverageInstrumentationCallback.java
```diff
@@ -0,0 +1,205 @@
...
+      throw new ArithmeticException(
+          "Unique Identifier exceeds value of Integer.MAX_VALUE, could not encode with Base 64 VLQ");
+    }
+
+    StringBuilder sb = new StringBuilder();
+
+    try {
+      Base64VLQ.encode(sb, uniqueIdentifier.intValue());
+    } catch (IOException e) {
+      // If encoding in Base64 VLQ fails, we will use the original identifier as it will still
```
lauraharker, line 124: In what cases does this fail?

This might be a good candidate for a unit test.

src/com/google/javascript/jscomp/AdvancedCoverageInstrumentationCallback.java
```diff
@@ -0,0 +1,205 @@
...
+   *
+   * @param traversal The context of the current traversal.
+   * @param node      The block node to be instrumented.
+   * @param fnName    The function name that the node exists within.
+   * @return The newly constructed function call node.
+   */
+  private Node newInstrumentationNode(NodeTraversal traversal, Node node, String fnName) {
+
+    String type = "Type.FUNCTION";
+    String combinedParam = traversal.getSourceName() + " " + fnName + " " + type;
```
lauraharker, line 109: suggestion:
  import static com.google.common.base.Strings.lenientFormat;

  Strings.lenientFormat("%s %s %s", traversal.getSourceName(), fnName, type);

src/com/google/javascript/jscomp/AdvancedCoverageInstrumentationCallback.java
```diff
@@ -0,0 +1,205 @@
...
+    } catch (IOException e) {
+      // If encoding in Base64 VLQ fails, we will use the original identifier as it will still
+      // maintain obfuscation and partial optimization
+      sb.append(uniqueIdentifier.intValue());
+    }
+
+    parameterMapping.addParamMapping(sb.toString(), combinedParam);
+
+    Node inner_prop = IR
+        .getprop(IR.name("module$exports$instrument$code"), IR.string("instrumentCodeInstance"));
```
lauraharker, line 132: Could these also be static constants? I think it would be useful to comment on where this function name comes from, too.

src/com/google/javascript/jscomp/CompilerOptions.java
nbeloglazov: Also I feel like `ADVANCED` isn't descriptive enough. Internally we don't use "advance" optimization. While I can't think of a good name that describes *what* this type of coverage does - I can suggest `PRODUCTION` that describes *where* it should be used: it's intended to be used in production builds that run in end users browsers. 

Any other suggestions/ideas?

src/com/google/javascript/jscomp/AdvancedCoverageInstrumentationCallback.java
patrick-sko: Changed as discussed.

src/com/google/javascript/jscomp/CompilerOptions.java
patrick-sko: For now I changed it to `PRODUCTION` to make it less ambiguous but will change it later if a better name comes up.

test/com/google/javascript/jscomp/integration/AdvancedCoverageInstrumentationPassIntegrationTest.java
patrick-sko: This one is not so I reverted it but the below Compile method does need to be overridden.

src/com/google/javascript/jscomp/AdvancedCoverageInstrumentationCallback.java
patrick-sko: Base64VLQ can potentially raise an exception because the Java Appendable `append` function can throw an `IOException`. I considered writing a unit test case, but I couldn't figure out in what way does the Java append function raise an IO exception.

src/com/google/javascript/jscomp/CompilerOptions.java
```diff
@@ -3142,7 +3142,8 @@ public String toString() {
   public enum InstrumentOption {
     NONE, // No coverage instrumentation is performed
     LINE_ONLY, // Collect coverage for every executable statement.
-    BRANCH_ONLY; // Collect coverage for control-flow branches.
+    BRANCH_ONLY, // Collect coverage for control-flow branches.
+    PRODUCTION;
```
lauraharker, line 3146: Please add a comment here too

src/com/google/javascript/jscomp/ProductionCoverageInstrumentationCallback.java
```diff
@@ -0,0 +1,190 @@
+/*
+ * Copyright 2009 The Closure Compiler Authors.
```
lauraharker, line 2: nit: s/2009/2020

src/com/google/javascript/jscomp/ProductionCoverageInstrumentationCallback.java
```diff
@@ -0,0 +1,190 @@
...
+import java.math.BigInteger;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+
+
+/**
+ * Instrument advanced coverage for javascript.
+ */
+class ProductionCoverageInstrumentationCallback extends
```
lauraharker, line 34: Can this be final?

Google style is to make classes final unless they are intended to be extended

test/com/google/javascript/jscomp/integration/ProductionCoverageInstrumentationPassIntegrationTest.java
```diff
@@ -0,0 +1,190 @@
+/*
+ * Copyright 2009 The Closure Compiler Authors.
```
lauraharker, line 2: nit: s/2009/2020

test/com/google/javascript/jscomp/integration/ProductionCoverageInstrumentationPassIntegrationTest.java
```diff
@@ -0,0 +1,190 @@
...
+@RunWith(JUnit4.class)
+public final class ProductionCoverageInstrumentationPassIntegrationTest extends
+    IntegrationTestCase {
+
+
+  private final String instrumentCodeSource =
+      lines(
+          "goog.module('instrument.code');",
+          "class InstrumentCode {",
+          "instrumentCode(a,b) {}",
```
lauraharker, line 41: minor issue: I'd find this code + a few other places in this file more readable if you indented the source code. e.g.
```
          "goog.module('instrument.code');",
          "class InstrumentCode {",
          "  instrumentCode(a,b) {}",          
          "}",
```

test/com/google/javascript/jscomp/integration/ProductionCoverageInstrumentationPassIntegrationTest.java
```diff
@@ -0,0 +1,190 @@
...
+            "function foo() { ",
+            "module$exports$instrument$code.instrumentCodeInstance.instrumentCode(\"C\", 1);",
+            "console.log('Hello'); ",
+            "}");
+
+    String[] expectedArr = {noTranspilationInstrumentCodeExpected, expected};
+    test(options, sourceArr, expectedArr);
+  }
+
+  /*
```
lauraharker, line 127: Please remove this

src/com/google/javascript/jscomp/AdvancedCoverageInstrumentationCallback.java
lauraharker: Looks like Base64VLQ shouldn't throw an exception here, because StringBuilder doesn't report that it throws an `IOException`. See also https://stackoverflow.com/questions/15607060/java-interface-throws-an-exception-but-interface-implementation-does-not-throw-a: StringBuilder doesn't have to throw an IOException even though the Appendable interface does.

My suggestion would be to catch the IOException and then throw an AssertionError in the catch block. This is the suggested "impossible condition check" from https://github.com/google/guava/wiki/ConditionalFailuresExplained.

patrick-sko: Interesting. I updated to throw AssertionError with message saying that an unexpected error was encounter.

src/com/google/javascript/jscomp/CoverageInstrumentationPass.java
```diff
@@ -78,6 +78,13 @@ public void process(Node externsNode, Node rootNode) {
             compiler,
             rootNode,
             new BranchCoverageInstrumentationCallback(compiler, instrumentationData));
+      } else if (instrumentOption == InstrumentOption.PRODUCTION) {
+        NodeTraversal.traverse(
+            compiler,
+            rootNode,
+            new ProductionCoverageInstrumentationCallback(compiler));
+
+        return;
```
nbeloglazov, line 87: Add comment explaining that we don't need to add header code for production coverage because it relies on instrument function being part of source code.

src/com/google/javascript/jscomp/ProductionCoverageInstrumentationCallback.java
```diff
@@ -0,0 +1,188 @@
...
+import java.util.Objects;
+
+
+/**
+ * Instrument advanced coverage for javascript.
+ */
+final class ProductionCoverageInstrumentationCallback extends
+    NodeTraversal.AbstractPostOrderCallback {
+
+  private static final String INSTRUMENT_CODE_FUNCTION_NAME = "instrumentCode";
```
nbeloglazov, line 37: Same comment as in old version: let's add TODO to make these part dynamic so that we don't rely on hardcoded namespaces/files.

src/com/google/javascript/jscomp/ProductionCoverageInstrumentationCallback.java
```diff
@@ -0,0 +1,188 @@
...
+  /**
+   * INSTRUMENT_CODE_FILE_NAME will contain an instance of the instrumentCode class and that
+   * instance name is stored in this constant.
+   */
+  private static final String INSTRUMENT_CODE_INSTANCE = "instrumentCodeInstance";
+
+  private final AbstractCompiler compiler;
+  private final ParameterMapping parameterMapping;
+  boolean visitedInstrumentCodeFile = false;
+  private String functionName = "Anonymous";
```
nbeloglazov, line 56: Don't need this class member. It can be method-local.

src/com/google/javascript/jscomp/ProductionCoverageInstrumentationCallback.java
```diff
@@ -0,0 +1,188 @@
...
+   * @param fnName    The function name that the node exists within.
+   * @return The newly constructed function call node.
+   */
+  private Node newInstrumentationNode(NodeTraversal traversal, Node node, String fnName) {
+
+    String type = "Type.FUNCTION";
+
+    String encodedParam = parameterMapping.getEncodedParam(traversal.getSourceName(), fnName, type);
+
+    Node inner_prop = IR
```
nbeloglazov, line 122: innerProp

src/com/google/javascript/jscomp/ProductionCoverageInstrumentationCallback.java
```diff
@@ -0,0 +1,188 @@
...
+   */
+  private void instrumentCode(NodeTraversal traversal, Node block, String fnName) {
+    block.addChildToFront(
+        newInstrumentationNode(traversal, block, fnName));
+    compiler.reportChangeToEnclosingScope(block);
+  }
+
+  /**
+   * Create a function call to the Instrument Code function with properly encoded parameters.
+   *
```
nbeloglazov, line 110: Can you add example of how that function call looks like so that it's easier to understand structure of the node you are building.

src/com/google/javascript/jscomp/ProductionCoverageInstrumentationCallback.java
```diff
@@ -0,0 +1,188 @@
...
+  /**
+   * A class the maintains a mapping of unique identifiers to parameter values. It also generates
+   * unique identifiers by creating a counter starting form 0 and increments this value when
+   * assigning a new unique identifier.
+   */
+  private static final class ParameterMapping {
+
+    private final List<String> uniqueIdentifier;
+    private final List<String> paramValue;
+    private BigInteger nextUniqueIdentifier;
```
nbeloglazov, line 140: let's use long instead of BigInteger. long is 2^64 which should be plenty.

src/com/google/javascript/jscomp/ProductionCoverageInstrumentationCallback.java
```diff
@@ -0,0 +1,188 @@
...
+      paramValue = new ArrayList<>();
+    }
+
+    public String getEncodedParam(String fileName, String functionName, String type) {
+      String combinedParam = lenientFormat("%s %s %s", fileName, functionName, type);
+
+      BigInteger uniqueIdentifier = getUniqueIdentifier(combinedParam);
+      BigInteger maxInteger = new BigInteger(Integer.toString(Integer.MAX_VALUE));
+
+      if (uniqueIdentifier.compareTo(maxInteger) > 0) {
```
nbeloglazov, line 154: With longs it will be just

```java
long id = ++nextId;
if (id > Integer.MAX_VALUE) {
  throw ...
}
```

src/com/google/javascript/jscomp/ProductionCoverageInstrumentationCallback.java
patrick-sko: In this PR, the method can be made local I agree but it is class scope in anticipation of additional PRs that implement branch coverage. In that case, this field will be reused to avoid making redundant calls to `getBestLValueName`.

src/com/google/javascript/jscomp/ProductionCoverageInstrumentationCallback.java
patrick-sko: Added two examples, one with the constants and another with what the current constants are to. We can remove the example of what the current constants evaluate to when we make them dynamic.

src/com/google/javascript/jscomp/AdvancedCoverageInstrumentationCallback.java
```diff
@@ -0,0 +1,205 @@
...
+  private final AbstractCompiler compiler;
+  private final ParameterMapping parameterMapping;
+  private final Map<String, AdvancedFileInstrumentationData> instrumentationData;
+  private String functionName = "Anonymous";
+
+  public AdvancedCoverageInstrumentationCallback(
+      AbstractCompiler compiler, Map<String, AdvancedFileInstrumentationData> instrumentationData) {
+    this.compiler = compiler;
+    this.instrumentationData = instrumentationData;
+    this.parameterMapping = new ParameterMapping();
```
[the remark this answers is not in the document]
patrick-sko: Not sure what you mean? ParameterMapping is private and, at least with the current implementation wont interact outside this class except possibly with the compiler param passed into this class.

PR #3645, opened 2020-07-19
Added support for UnicodeCombiningMark, fixes #3639.

### LINKED ISSUE #3639: Parser error on "combining character" (U+0307)
### There seems to be an issue with parsing composite character sequences.

Composite character `İ` consist of the base character and combining character sequence (U+0049)(U+0307)  
Using this character causing the compiler to fail with the parser error.

### Example code

```
var bar = {
İ: "foo"
};
```

### Compiler error

```
input0:2: ERROR - [JSC_PARSE_ERROR] Parse error. Character '̇' (U+0307) is not a valid identifier start char
  2| İ: "foo"
      ^

1 error(s), 0 warning(s)
```

### Demo
[Example](https://closure-compiler-debugger.appspot.com/#input0%3Dvar%2520bar%2520%253D%2520%257B%250AI%25CC%2587%253A%2520%2522foo%2522%250A%257D%253B%26input1%26conformanceConfig%26externs%26refasterjs-template%26includeDefaultExterns%3Dtrue%26ENABLE_ALL_DIAGNOSTIC_GROUPS%3Dtrue%26CHECK_TYPES%3Dtrue%26CHECK_SYMBOLS%3Dtrue%26MISSING_PROPERTIES%3Dtrue%26TRANSPILE%3Dtrue%26CLOSURE_PASS%3Dtrue%26PRESERVE_TYPE_ANNOTATIONS%3Dtrue%26PRETTY_PRINT%3Dtrue)

christiantjl:
Some notes:

1. Annotating functions with spec grammar is likely wise (and will help with issues like this in the future), so I did not remove that JSDOC block. 
2. A `utils/` directory was added, which contains a script for building the source (without GWT) and a script for calling that built version with the given args. This directory will be a good place to put bash scripts in the future, including `runtime_tests.sh` to execute the jsComp runtime tests (a separate issue I'm working on).
3. There is more grammar to implement if we want to be consistent with the spec (`isConnectorPunctuation`, `isZeroWidthJoiner`, etc.), which will prevent issues like this in the future. I intend to do this over the coming days, I just wanted to get this PR in first.

This PR fixes #3639, and you can see that the source:
```javascript
var bar = {
  İ: "foo"
};
```
now compiles as expected to:
```javascript
var bar={"I\u0307":"foo"};
```


christiantjl:
Build tests are failing because:
```none
/src/com/google/javascript/jscomp/parsing/parser/Scanner.java:909: error: cannot find symbol
    return Character.getType(ch) == Character.NON_SPACING_MARK;
                    ^
  symbol:   method getType(char)
  location: class Character
/src/com/google/javascript/jscomp/parsing/parser/Scanner.java:909: error: cannot find symbol
    return Character.getType(ch) == Character.NON_SPACING_MARK;
```

Anybody have recommendations?  Not sure why `Character.getType` would not be defined, I even added a manual import for `java.lang.Character`, and `getType` method has been defined since JDK 1.1. 

I noticed the rest of the character checks did not use builtins like `Character.NO_SPACING_MARK`, but this syntax is much more legible on top of being future-proofed and likely just as performant.

christiantjl:
Opened an issue at google/j2cl#103 regarding future-proofed Unicode category checks, just need J2CL to support `Character.getType()` and the category constants.

.gitignore
```diff
@@ -12,3 +12,9 @@ target/
 /yarn.lock
 /node_modules
 /.metadata/
+
+# Bazel stuff
+/bazel-bin
+/bazel-closure-compiler
+/bazel-out
+/bazel-testlogs
```
christiantjl, line 20: Running `./build_test.sh` created bazel output directories that were added to the git tree by default. These lines prevent that.

src/com/google/javascript/jscomp/parsing/parser/Scanner.java
```diff
@@ -901,8 +901,65 @@ private static boolean isIdentifierStart(char ch) {
         | (ch >= 0x03B1 & ch <= 0x03C9); // Greek lowercase letters
   }
 
+  // Check if char is Unicode Category "Combining spacing mark (Mc)"
```
christiantjl, line 904: Not compliant (see #3647), per spec `UnicodeCombiningMark` also includes "Non-spacing mark (Mn)" category, but this addresses #3639 and a decent scope of similar possibilities with as minimal change as possible.

utils/run.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+# Run the local build of CC in target/. Make sure to run build.sh before running
+# this script.
+java -jar target/closure-compiler-1.0-SNAPSHOT.jar $@
```
christiantjl, line 4: Handy for debugging local build.

utils/build.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+# Build the compiler without GWT.
+mvn -DskipTests -pl externs/pom.xml,pom-main.xml,pom-main-shaded.xml
```
christiantjl, line 3: Handy for debugging local build.

rishipal: Can you move the utils/ scripts to be inside src/com/google/javascript/jscomp/?

christiantjl: Sure - there's just an existing `/travis_utils` dir, and I am going to be adding a `runtime_tests.sh` script for jscomp runtime tests as part of a separate PR, and I figured this would be a good time to raise the idea of just moving those all to a root-level `utils/` dir for organization.

Will move them, just something to think about.

PR #3647, opened 2020-07-21
Adding full Unicode support for Javascript identifiers.

### LINKED ISSUE #3639: Parser error on "combining character" (U+0307)

christiantjl:
[Build Status](https://travis-ci.org/christiantjl/closure-compiler)

Further resolves #3639, aims to prevent 99.999% of future issues that might occur from using non-English characters (except the few Greek and Latin chars manually included) as an identifier. In the current build, the following errors will occur (as well as a known issue with `U+FF3F`:
```
ERROR - [JSC_PARSE_ERROR] Parse error. Character '〩' (U+3029) is not a valid identifier start char
  1| var a〩 = 10;
          ^
```
```
ERROR - [JSC_PARSE_ERROR] Parse error. Character '〺' (U+303A) is not a valid identifier start char
  1| var 〺b = 10;
         ^
```
```
ERROR - [JSC_PARSE_ERROR] Parse error. Character '＿' (U+FF3F) is not a valid identifier start char
  1| var ａｅｓｔｈｅｔｉｃ＿;
                  ^
```

After:
```
var a〩 = 10;
var a\u3029=10;
```
```
var 〺b = 10;
var \u303ab=10;
```
```
var ａｅｓｔｈｅｔｉｃ＿;
var \uff41\uff45\uff53\uff54\uff48\uff45\uff54\uff49\uff43\uff3f;
```

The RegExpU approach is not very performant for compilation, but all previous optimizations were migrated into the `UnicodeMatch` class, so the only impact is if you use a previously unsupported character (which would have thrown an error and exited altogether). I will likely transform the matches into large `(a <= ch & ch <= b)` blocks to further optimize, but I wanted to put this up for review and feedback (**especially regarding short-circuiting**, which I am not super familiar with).

Apologies if this is an unwanted PR, I just figured I'd take a crack at it.


src/com/google/javascript/jscomp/parsing/parser/util/UnicodeMatch.java
```diff
@@ -0,0 +1,106 @@
...
+ * regexpu: https://github.com/mathiasbynens/regexpu
+ * 
+ * IdentifierStart and IdentifierPart regex patterns based on ECMA's
+ * documentation:
+ * https://github.com/tc39/proposal-regexp-unicode-property-escapes#other-examples
+ * 
+ * const regexIdentifierStart = /[$_\p{ID_Start}]/u;
+ * const regexIdentifierPart = /[$_\u200C\u200D\p{ID_Continue}]/u;
+ */
+public class UnicodeMatch {
```
lauraharker, line 37: Some visibility comments:

This class should be final. I think it shouldn't be public, unless we later find a compelling use case outside this package.

The Google Java best practices also recommends adding an empty private constructor for utility classes to make it clear they should not be instantiated, i.e.

``` java
  private UnicodeMatch() {}
```

src/com/google/javascript/jscomp/parsing/parser/Scanner.java
```diff
@@ -869,52 +870,18 @@ private static String processUnicodeEscapes(String value) {
...
-        | (ch >= 0x00F8 & ch <= 0x00FF) // Latin letters
-        | ch == 0x0275 // Latin Barred O
-        | (ch >= 0x0391 & ch <= 0x03A1) // Greek uppercase letters
-        // 0x03A2 = unassigned
-        | (ch >= 0x03A3 & ch <= 0x03A9) // Remaining Greek uppercase letters
-        | (ch >= 0x03B1 & ch <= 0x03C9); // Greek lowercase letters
-  }
-
-  @SuppressWarnings("ShortCircuitBoolean") // Intentional to minimize branches in this code
+    return UnicodeMatch.isJavascriptIdentifierStart(ch);
```
lauraharker, line 877: Can we inline calls to this method and `isIdentifierPart`?

src/com/google/javascript/jscomp/parsing/parser/util/UnicodeMatch.java
```diff
@@ -0,0 +1,106 @@
...
+  // Intentional to minimize branches in this code
+  @SuppressWarnings("ShortCircuitBoolean")
+  private static boolean fastPathLetters(char ch) {
```
lauraharker, line 40: Could you also copy over the comment about minimizing branch prediction misses? It's not really clear to me otherwise why we are trying to minimize branches.

About adding the Latin and Greek letters to this fast path: I'm concerned that by adding those letters to this function, we're making the more common case of ASCII characters slower due to the lack of short-circuiting.

It seems potentially useful to have a fast path for these character, if they're common in Angular, but not at the expense of ASCII characters. @concavelenz was the one who initially added this fast path - John, do you have any thoughts?

src/com/google/javascript/jscomp/parsing/parser/Scanner.java
```diff
@@ -869,52 +870,18 @@ private static String processUnicodeEscapes(String value) {
...
-          | (ch >= 'a' & ch <= 'z')
-          | (ch >= '0' & ch <= '9')
-          | (ch == '_' | ch == '$'));
-    }
-
-    // Handle non-ASCII characters.
-    // TODO(tjgq): This should include all characters with the ID_Continue property, plus
-    // Zero Width Non-Joiner and Zero Width Joiner.
-    return isIdentifierStart(ch) || Character.isDigit(ch);
+    return UnicodeMatch.isJavascriptIdentifierPart(ch);
```
concavelenz, line 884: We still want to maintain the "ch <= 127" check.  This is a fast path for performance reasons.

src/com/google/javascript/jscomp/parsing/parser/Scanner.java
concavelenz, line 877: We still want to maintain the "ch <= 127" check. This is a fast path for performance reasons.

src/com/google/javascript/jscomp/parsing/parser/util/UnicodeMatch.java
concavelenz: We don't need to include non-ascii (>127) in the fast path.

src/com/google/javascript/jscomp/parsing/parser/util/UnicodeMatch.java
```diff
@@ -0,0 +1,106 @@
...
+
+  // Intentional to minimize branches in this code
+  @SuppressWarnings("ShortCircuitBoolean")
+  private static boolean fastPathDigits(char ch) {
+    return '0' <= ch & ch <= '9';
+  }
+
+  public static boolean isJavascriptIdentifierStart(char ch) {
+    return fastPathLetters(ch)
+        || JavascriptIdentifierStart
```
concavelenz, line 71: because you aren't checking for <127 you go through the slow path even for ASCII characters if they aren't part of an identifier.

src/com/google/javascript/jscomp/parsing/parser/util/UnicodeMatch.java
```diff
@@ -0,0 +1,106 @@
...
...uDC70-\\uDC7C\\uDC80-\\uDC88\\uDC90-\\uDC99]|\\uD835[\\uDC00-\\uDC54\\uDC56-\\uDC9C\\uDC9E\\uDC9F\\uDCA2\\uDCA5\\uDCA6\\uDCA9-\\uDCAC\\uDCAE-\\uDCB9\\uDCBB\\uDCBD-\\uDCC3\\uDCC5-\\uDD05\\uDD07-\\uDD0A\\uDD0D-\\uDD14\\uDD16-\\uDD1C\\uDD1E-\\uDD39\\uDD3B-\\uDD3E\\uDD40-\\uDD44\\uDD46\\uDD4A-\\uDD50\\uDD52-\\uDEA5\\uDEA8-\\uDEC0\\uDEC2-\\uDEDA\\uDEDC-\\uDEFA\\uDEFC-\\uDF14\\uDF16-\\uDF34\\uDF36-\\uDF4E\\uDF50-\\uDF6E\\uDF70-\\uDF88\\uDF8A-\\uDFA8\\uDFAA-\\uDFC2\\uDFC4-\\uDFCB]|\\uD838[\\uDD00-\\uDD2C\\uDD37-\\uDD3D\\uDD4E\\uDEC0-\\uDEEB]|\\uD83A[\\uDC00-\\uDCC4\\uDD00-\\uDD43\\uDD4B]|\\uD83B[\\uDE00-\\uDE03\\uDE05-\\uDE1F\\uDE21\\uDE22\\uDE24\\uDE27\\uDE29-\\uDE32\\uDE34-\\uDE37\\uDE39\\uDE3B\\uDE42\\uDE47\\uDE49\\uDE4B\\uDE4D-\\uDE4F\\uDE51\\uDE52\\uDE54\\uDE57\\uDE59\\uDE5B\\uDE5D\\uDE5F\\uDE61\\uDE62\\uDE64\\uDE67-\\uDE6A\\uDE6C-\\uDE72\\uDE74-\\uDE77\\uDE79-\\uDE7C\\uDE7E\\uDE80-\\uDE89\\uDE8B-\\uDE9B\\uDEA1-\\uDEA3\\uDEA5-\\uDEA9\\uDEAB-\\uDEBB]|\\uD869[\\uDC00-\\uDEDD\\uDF00-\\uDFFF]|\\uD86D[\\uDC00-\\uDF34\\uDF40-\\uDFFF]|\\uD86E[\\uDC00-\\uDC1D\\uDC20-\\uDFFF]|\\uD873[\\uDC00-\\uDEA1\\uDEB0-\\uDFFF]|\\uD87A[\\uDC00-\\uDFE0]|\\uD87E[\\uDC00-\\uDE1D]|\\uD884[\\uDC00-\\uDF4A])");
```
concavelenz, line 92: How did you generate these values?

src/com/google/javascript/jscomp/parsing/parser/Scanner.java
```diff
@@ -869,52 +870,18 @@ private static String processUnicodeEscapes(String value) {
...
-    // Most code is written in pure ASCII, so create a fast path here.
-    if (ch <= 127) {
-      // Intentionally avoiding short circuiting behavior of "||" and "&&".
-      // This minimizes branches in this code which minimizes branch prediction misses.
-      return ((ch >= 'A' & ch <= 'Z') | (ch >= 'a' & ch <= 'z') | (ch == '_' | ch == '$'));
-    }
-
-    // Handle non-ASCII characters.
-    // TODO(tjgq): This should include all characters with the ID_Start property.
-    if (Character.isLetter(ch)) {
```
concavelenz, line 883: Do you feel this is insufficient for the native Java builds?   Where do you believe you need the JavaScript version?

src/com/google/javascript/jscomp/parsing/parser/util/UnicodeMatch.java
christiantjl: These are the [regexpu](https://github.com/mathiasbynens/regexpu)-transpiled patterns provided by the TC39: https://github.com/tc39/proposal-regexp-unicode-property-escapes#other-examples

Online regexpu portal at: https://mothereff.in/regexpu

See the **Updates** section of the OP.

src/com/google/javascript/jscomp/parsing/parser/Scanner.java
christiantjl: @lauraharker Absolutely, I just wanted to make as few code changes as possible. Will revise.

src/com/google/javascript/jscomp/parsing/parser/util/UnicodeMatch.java
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  public static boolean isJavascriptIdentifierStart(char ch) {
+    return fastPathLetters(ch)
+        || JavascriptIdentifierStart
+              .matcher(Character.toString(ch))
+              .matches();
+  }
+
+  public static boolean isJavascriptIdentifierPart(char ch) {
```
tjgq, line 76: I would add a comment here noting that this won't correctly identify supplementary characters (codepoint > 0xFFFF), because they don't fit in a `char`. Fixing this would require additional changes to the parser, which currently uses single-char lookahead.

src/com/google/javascript/jscomp/parsing/parser/Scanner.java
christiantjl: @concavelenz I retained the fastpath in `UnicodeMatch.java` (line 69):
```java
@SuppressWarnings("ShortCircuitBoolean")
  private static boolean fastPathLetters(char ch) {
    return (
      // ASCII
      ('A' <= ch & ch <= 'Z') |
      ('a' <= ch & ch <= 'z') |
      (ch == '_' | ch == '$') |
      // Latin letters
      (0x00C0 <= ch & ch <= 0x00D6) |
      // Latin letters
      (0x00D8 <= ch & ch <= 0x00F6) |
      // Latin letters
      (0x00F8 <= ch & ch <= 0x00FF) |
      // Latin Barred O
      (ch == 0x0275) |
      // Greek uppercase letters
      (0x0391 <= ch & ch <= 0x03A1) |
      // Remaining Greek uppercase letters
      (0x03A3 <= ch & ch <= 0x03A9) |
      // Greek lowercase letters
      (0x03B1 <= ch & ch <= 0x03C9)
    );
  }

  // Intentional to minimize branches in this code
  @SuppressWarnings("ShortCircuitBoolean")
  private static boolean fastPathDigits(char ch) {
    return '0' <= ch & ch <= '9';
  }

  public static boolean isJavascriptIdentifierStart(char ch) {
    return fastPathLetters(ch)
        || JavascriptIdentifierStart
              .matcher(Character.toString(ch))
              .matches();
  }

  public static boolean isJavascriptIdentifierPart(char ch) {
    return (fastPathLetters(ch) | fastPathDigits(ch))
        || JavascriptIdentifierPart
              .matcher(Character.toString(ch))
              .matches();
  }
```

I tried to retain all existing optimizations.

src/com/google/javascript/jscomp/parsing/parser/util/UnicodeMatch.java
christiantjl: Thank you very much Laura, I will make the relevant changes.

christiantjl: If the class is made `private`, I can't import it inside `Scanner.java` - extremely sorry if this is basic syntax, I tried to research it, but is there any way to import *just* `public isJavascriptIdentifierStart` and `public isJavascriptIdentifierPart` from `private UnicodeMatch` without importing the whole class? I tried:
```java
import com.google.javascript.jscomp.parsing.parser.util.UnicodeMatch.isJavascriptIdentifierStart;
import com.google.javascript.jscomp.parsing.parser.util.UnicodeMatch.isJavascriptIdentifierPart;
```

But it will not build without importing the whole class (while `public`) via
```java
import com.google.javascript.jscomp.parsing.parser.util.UnicodeMatch
```

And referencing with `UnicodeMatch.isJavascriptIdentifierStart` and so on.

src/com/google/javascript/jscomp/parsing/parser/util/UnicodeMatch.java
christiantjl: So, to be clear, I should remove the non-ASCII part of the fast path (fast path effectively becomes just `0 < ch < 127`)? 

The Latin and Greek optimizations were there before I touched the code - will adjust, but just looking for acknowledgement that I'm not misunderstanding the direction, which may impact Angular build times. I did not insert those optimizations.

src/com/google/javascript/jscomp/parsing/parser/Scanner.java
christiantjl: @concavelenz The fast path was moved directly into the `UnicodeMatch` class, and is evaluated for every call to `UnicodeMatch.isJavascriptIdentifierStart` and `UnicodeMatch.isJavascriptIdentifierPart`.

src/com/google/javascript/jscomp/parsing/parser/util/UnicodeMatch.java
christiantjl: @tjgq You're right - I will update to `int`.

src/com/google/javascript/jscomp/parsing/parser/Scanner.java
christiantjl: @concavelenz I believe the reason I removed `Character.isLetter` was because the GWT build would fail, as J2CL / Guava implementations did not exist or would not compile to JS. See: google/j2cl#103

Apologies for the lack of specificity, this issue is about a month old and is not super fresh in my mind. Thank you for your reviews, John.

src/com/google/javascript/jscomp/parsing/parser/Scanner.java
christiantjl: Also @concavelenz, doesn't this violate the spec? Won't `ch <= 127` include a bunch of characters that aren't valid for `isIdentifierPart` or `isIdentifierStart`?

src/com/google/javascript/jscomp/parsing/parser/util/UnicodeMatch.java
christiantjl: I'm admittedly confused about the note here, my apologies - it seems to me like there are a lot of ASCII chars which are not valid JS identifier characters (start or part), and the Greek/Latin fast paths are there to catch those before falling back to more expensive checks (before, I believe it was `Character.isLetter()`, now it would be regex).

Please advise, and thank you both for your patience.

src/com/google/javascript/jscomp/parsing/parser/util/UnicodeMatch.java
```diff
@@ -0,0 +1,174 @@
...
+     * Intentionally avoiding short circuiting behavior of "||" and "&&". This
+     * minimizes branches in this code which minimizes branch prediction misses.
+     */
+    if (ch <= 127) {
+      return (
+        (ch >= 'A' & ch <= 'Z') |
+        (ch >= 'a' & ch <= 'z') |
+        (ch == '$' | ch == '_')
+      );
+    }
```
christiantjl, line 96: ASCII fast path will work in every call to `fastPathIdentifierStart` and `fastPathIdentifierPart` (because the former is evaluated in the latter).

rishipal:
This isn't merged yet. I'm testing this internally and will update soon.

rishipal:
I'm seeing multiple internal targets failing to build with this error message - 'The import java.util.regex cannot be resolved'.

rishipal:
I'll take another stab at this today and repro+debug the failing tests.

ctjlewis:
Hey @rishipal, I saw your comment about errors, I did not see any when I submitted the PR. Apologies for the delay, I have not had much free time to review this.

rishipal:
The 'The import java.util.regex cannot be resolved' failures get fixed after switching to re2j regex library. However, I see multiple Scanner tests failing with that. I will update shortly.

concavelenz:
We have dropped support for the JS (GWT version) of the Closure Compiler in favor of the Graal native image versions.  Do you feel that this is still useful?

ctjlewis:
@concavelenz As I understand it, this was an issue regardless of build - it was just an issue of extended Unicode characters not accurately being counted as valid identifier parts, IIRC (this work is not super fresh in my mind, apologies).

rishipal:
Update: I tested this again over the weekend. The internal tests are passing now. I will commit this internally.

ctjlewis:
@rishipal Thank you for your help on this PR. Was this merged internally?

Sorry to bother!

rishipal:
Hi, this is not merged yet. A comment from John on the pending CL suggests that this may not be needed after we stopped support for GWT/J2CL version of the compiler. I will confirm the priority of this and respond by EOD tomorrow.

rishipal:
The Scanner test failures did not get fully resolved contrary to my https://github.com/google/closure-compiler/pull/3647#issuecomment-779989990. This PR is decided to be deprioritized and shelved.

.gitignore
```diff
@@ -12,3 +12,12 @@ target/
...
 /.metadata/
+
+# Bazel stuff
+/bazel-bin
+/bazel-closure-compiler
+/bazel-out
+/bazel-testlogs
+
+# IDE
+.vscode
```
Githugged, line 23: ctjlewis:unicode

PR #3651, opened 2020-07-28
Implemented Production Instrumentation Parameter Mapping

patrick-sko:
Implemented the mapping that is created when the parameter passed to the InstrumentCode function is encoded. The details of the implementation can be found in this [Implementation Doc](https://docs.google.com/document/d/1cdseVYW0x8LWccxYbT6GavrhjL00AA9SoafxAkqyu-8/edit?usp=sharing) in the _Parameter Obfustication and Optimization_ section. 

This implementation will add a new flag that takes in the file name of the generated parameter mapping report.

### DIFF
--- a/src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
+++ b/src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
@@ -1439,6 +1439,9 @@ public abstract class AbstractCommandLineRunner<A extends Compiler,
       outputManifest();
       outputBundle();
 
+      // Output the production instrumentation param mapping if requested.
+      outputInstrumentationMapping();
+
       if (isOutputInJson()) {
         outputJsonStream();
       }
@@ -2013,6 +2016,13 @@ public abstract class AbstractCommandLineRunner<A extends Compiler,
     outputManifestOrBundle(config.outputBundles, false);
   }
 
+  @GwtIncompatible("Unnecessary")
+  private void outputInstrumentationMapping() throws IOException {
+    if (!Strings.isNullOrEmpty(config.instrumentationMappingFile)) {
+      compiler.getInstrumentationMapping().save(config.instrumentationMappingFile);
+    }
+  }
+
   /**
    * Writes the manifest or bundle of all compiler input files that were included as controlled by
    * --dependency_mode, if requested.
@@ -2412,6 +2422,14 @@ public abstract class AbstractCommandLineRunner<A extends Compiler,
      */
     public CommandLineConfig setStringMapOutputFile(String stringMapOutputPath) {
       this.stringMapOutputPath = stringMapOutputPath;
+      return this;
+    }
+
+    private String instrumentationMappingFile = "";
+
+    public CommandLineConfig setInstrumentationMappingFile(
+        String instrumentationMappingFile) {
+      this.instrumentationMappingFile = instrumentationMappingFile;
       return this;
     }
 
--- a/src/com/google/javascript/jscomp/CommandLineRunner.java
+++ b/src/com/google/javascript/jscomp/CommandLineRunner.java
@@ -22,6 +22,7 @@ import com.google.common.annotations.GwtIncompatible;
 import com.google.common.base.Ascii;
 import com.google.common.base.Joiner;
 import com.google.common.base.Splitter;
+import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableMultimap;
@@ -270,6 +271,12 @@ public class CommandLineRunner extends
         usage = "File where the serialized version of the variable "
         + "renaming map produced should be saved")
     private String variableMapOutputFile = "";
+
+    @Option(name = "--instrument_mapping_report",
+        usage = "File where the encoded parameters created by Production "
+            + "Instrumentation are mapped to their pre-encoded values. Must "
+            + "be used in tandem with --instrument_code=PRODUCTION")
+    private String instrumentationMappingOutputFile = "";
 
     @Option(name = "--create_renaming_reports",
         hidden = true,
@@ -1708,6 +1715,7 @@ public class CommandLineRunner extends
           .setVariableMapOutputFile(flags.variableMapOutputFile)
           .setCreateNameMapFiles(flags.createNameMapFiles)
           .setPropertyMapOutputFile(flags.propertyMapOutputFile)
+          .setInstrumentationMappingFile(flags.instrumentationMappingOutputFile)
           .setCodingConvention(conv)
           .setSummaryDetailLevel(flags.summaryDetailLevel)
           .setOutputWrapper(flags.outputWrapper)
@@ -1931,6 +1939,18 @@ public class CommandLineRunner extends
       options.setPropertyRenaming(PropertyRenamingPolicy.OFF);
     }
 
+    if (flags.instrumentCodeParsed == InstrumentOption.PRODUCTION &&
+            (Strings.isNullOrEmpty(flags.instrumentationMappingOutputFile))) {
+      throw new FlagUsageException("Expected --instrument_mapping_report to be set when "
+          + "--instrument_code is set to Production");
+    }
+
+    if (!Strings.isNullOrEmpty(flags.instrumentationMappingOutputFile)
+        && flags.instrumentCodeParsed != InstrumentOption.PRODUCTION) {
+      throw new FlagUsageException("Expected --instrument_code to be passed with PRODUCTION " +
+          "when --instrument_mapping_report is set");
+
+    }
     options.setInstrumentForCoverageOption(flags.instrumentCodeParsed);
 
     return options;
--- a/src/com/google/javascript/jscomp/ProductionCoverageInstrumentationCallback.java
+++ b/src/com/google/javascript/jscomp/ProductionCoverageInstrumentationCallback.java
@@ -16,15 +16,15 @@
 
 package com.google.javascript.jscomp;
 
-import static com.google.common.base.Strings.lenientFormat;
-
 import com.google.common.annotations.GwtIncompatible;
+import com.google.common.collect.ImmutableMap;
 import com.google.debugging.sourcemap.Base64VLQ;
 import com.google.javascript.rhino.IR;
 import com.google.javascript.rhino.Node;
 import java.io.IOException;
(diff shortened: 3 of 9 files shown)


src/com/google/javascript/jscomp/ProductionCoverageInstrumentationCallback.java
```diff
@@ -137,56 +137,123 @@ private Node newInstrumentationNode(NodeTraversal traversal, Node node, String f
...
+    // encoded uniqueIdentifier. This is so we can check if an encoded param has already been
+    // defined so that we do not create a duplicate. (Ex. Key: ACA (Base64 VLQ encoding of [0,1,0]),
+    // Value: C (Base64 VLQ encoding of the uniqueIdentifier). This map will later be inversed so
+    // that it is printed in the following form: C:ACA.
+    private final Map<String, String> paramValueEncodings;
+
+    // Values are stored as a mapping of String to Integers so that we can lookup the index of the
+    // encoded (file|function|type) name and also check if it is present in constant time. These
+    // mappings are added to the VariableMap once instrumentation is complete.
+    private final Map<String, Integer> fileNamesMap;
```
lauraharker, line 163: I think names like `fileNameToIndex` would be clearer. It makes it easier to understand what the values are in this map without having to go back o this comment.

src/com/google/javascript/jscomp/ProductionCoverageInstrumentationCallback.java
```diff
@@ -137,56 +137,123 @@ private Node newInstrumentationNode(NodeTraversal traversal, Node node, String f
...
+    private final Map<String, Integer> functionNamesMap;
+    private final Map<String, Integer> typeMap;
+    private int fileNamesCounter;
```
lauraharker, line 166: Could you replace the counters with looking up `fileNamesMap.size()`?

src/com/google/javascript/jscomp/ProductionCoverageInstrumentationCallback.java
```diff
@@ -137,56 +137,123 @@ private Node newInstrumentationNode(NodeTraversal traversal, Node node, String f
...
-    public String getEncodedParam(String fileName, String functionName, String type) {
-      String combinedParam = lenientFormat("%s %s %s", fileName, functionName, type);
-      long uniqueIdentifier = getUniqueIdentifier();
+    private String getEncodedParam(String fileName, String functionName, String type) {
 
-      if (uniqueIdentifier > Integer.MAX_VALUE) {
-        throw new ArithmeticException(
-            "Unique Identifier exceeds value of Integer.MAX_VALUE, could not encode with Base 64"
-                + " VLQ");
+      if (!fileNamesMap.containsKey(fileName)) {
```
lauraharker, line 190: optional suggestion: make this more concise by replacing the ifs with `fileNamesMap.putIfAbsent(fileName, fileNamesMap.size());`

src/com/google/javascript/jscomp/ProductionCoverageInstrumentationCallback.java
```diff
@@ -137,56 +137,123 @@ private Node newInstrumentationNode(NodeTraversal traversal, Node node, String f
...
+      if (!functionNamesMap.containsKey(functionName)) {
+        functionNamesMap.put(functionName, functionNamesCounter);
+        functionNamesCounter++;
       }
+      if (!typeMap.containsKey(type)) {
+        typeMap.put(type, typeCounter);
+        typeCounter++;
+      }
+
+      Integer fnNameIndex = functionNamesMap.get(functionName);
```
lauraharker, line 203: How about declaring all of these within the try/catch?

General good practice is to limit the scope of variables as much as possible to the places where they are used

test/com/google/javascript/jscomp/integration/ProductionCoverageInstrumentationPassIntegrationTest.java
```diff
@@ -116,6 +121,32 @@ public void testNoTranspilationFunctionInstrumentation() {
     test(options, sourceArr, expectedArr);
   }
 
+  @Test
+  public void testParamMappingIsCreated() {
+    CompilerOptions options = createCompilerOptions();
+
+    String source = lines("function foo() { ", "   console.log('Hello'); ", "}");
```
lauraharker, line 128: nit: add `//` to preserve line breaks here

lines(
    "function foo() { ", //
    "   console.log('Hello'); ",
    "}")

src/com/google/javascript/jscomp/CommandLineRunner.java
```diff
@@ -271,6 +272,12 @@
         + "renaming map produced should be saved")
     private String variableMapOutputFile = "";
 
+    @Option(name = "--instrument_param_mapping_report",
```
nbeloglazov, line 275: Nit: just `--instrumentation_mapping_report` is cleaner. `param` is pretty generic and doesn't really add value.

src/com/google/javascript/jscomp/CommandLineRunner.java
```diff
@@ -1931,6 +1939,11 @@ protected CompilerOptions createOptions() {
       options.setPropertyRenaming(PropertyRenamingPolicy.OFF);
     }
 
+    if (flags.instrumentCodeParsed == InstrumentOption.PRODUCTION &&
```
nbeloglazov, line 1942: Also check the other way around? If report is paramMappingOutputFile is provided while `PRODUCTION` is not enabled?

src/com/google/javascript/jscomp/CommandLineRunner.java
```diff
@@ -271,6 +272,12 @@
         + "renaming map produced should be saved")
     private String variableMapOutputFile = "";
 
+    @Option(name = "--instrument_param_mapping_report",
+        usage = "File where the encoded parameters created by Production "
+            + "Instrumentation are mapped to their pre-encoded values. Must "
+            + "be used in tandem with --instrument_code=PRODUCTION")
+    private String paramMappingOutputFile = "";
```
nbeloglazov, line 279: Let's name this variable similar to flag name. It should at least mention `instrumentation` in it otherwise it's too generic.

src/com/google/javascript/jscomp/ProductionCoverageInstrumentationCallback.java
nbeloglazov, line 203: Use `int`, not `Integer`.

src/com/google/javascript/jscomp/ProductionCoverageInstrumentationCallback.java
```diff
@@ -137,56 +137,123 @@ private Node newInstrumentationNode(NodeTraversal traversal, Node node, String f
...
+        }
+
+        paramValueEncodings.put(encodedParam, sb.toString());
+      }
 
-      return result;
+      return paramValueEncodings.get(encodedParam);
     }
 
     private long getUniqueIdentifier() {
```
nbeloglazov, line 241: Let's rename this to `generateUniqueIdentifier`. `get` operations don't change state. Especially in java where getters are expected to be dumb accessors.

src/com/google/javascript/jscomp/ProductionCoverageInstrumentationCallback.java
```diff
@@ -137,56 +137,123 @@ private Node newInstrumentationNode(NodeTraversal traversal, Node node, String f
...
+        Base64VLQ.encode(sb, fnNameIndex);
+        Base64VLQ.encode(sb, typeIndex);
       } catch (IOException e) {
         throw new AssertionError(e);
       }
 
-      String result = sb.toString();
+      String encodedParam = sb.toString();
+
+      if (!paramValueEncodings.containsKey(encodedParam)) {
```
nbeloglazov, line 219: Curious, do we expect situations where `encodedParam` repeats?

src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
```diff
@@ -2013,6 +2017,19 @@ private void outputBundle() throws IOException {
     outputManifestOrBundle(config.outputBundles, false);
   }
 
+  @GwtIncompatible("Unnecessary")
+  private void outputInstrumentationParamMapping() throws IOException {
+
+    if (!Strings.isNullOrEmpty(config.instrumentationParamMappingFile)) {
+      if (compiler.getOptions().getInstrumentForCoverageOption() != InstrumentOption.PRODUCTION) {
+        throw new FlagUsageException("Expected --instrument_code to be passed with PRODUCTION " +
```
nbeloglazov, line 2025: That check is more appropriate in next to the place where we parse flags.

test/com/google/javascript/jscomp/integration/ProductionCoverageInstrumentationPassIntegrationTest.java
patrick-sko: Not sure what you mean by "add `//` to preserve line breaks here"? I did break it down for readability however
```
lines( 
   "function foo() {",
   ...
);
```

src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
patrick-sko: Moved.

src/com/google/javascript/jscomp/CommandLineRunner.java
patrick-sko: Changed to `instrumentationMappingOutputFile`

src/com/google/javascript/jscomp/ProductionCoverageInstrumentationCallback.java
patrick-sko: I inlined the variables instead.

src/com/google/javascript/jscomp/ProductionCoverageInstrumentationCallback.java
patrick-sko: Yes, a common case would be an anonymous (ex, lambda) function. It would have the same File Name, same type and _technically_ same function name. Also, it would occur when branching is implemented.

src/com/google/javascript/jscomp/ProductionCoverageInstrumentationCallback.java
patrick-sko: Neat function, makes a lot more sense with the `fileNameToIndex.size()` so I updated it.

PR #3652, opened 2020-07-29
Fixed issue with switchValidator and added unit test

patrick-sko:
Fixed issue that `validateSwitch(Node n)` in `AstValidator.java` where it was not checking each child of the switch statement.

### DIFF
--- a/test/com/google/javascript/jscomp/AstValidatorTest.java
+++ b/test/com/google/javascript/jscomp/AstValidatorTest.java
@@ -1275,8 +1275,6 @@ public final class AstValidatorTest extends CompilerTestCase {
         "  switch(x) {",
         "    case 1: ",
         "      var y = 5;",
-        "    default:",
-        "      var y = 5;",
         "    case 2: ",
         "      var y = 5;",
         "  }",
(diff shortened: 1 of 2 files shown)


test/com/google/javascript/jscomp/AstValidatorTest.java
```diff
@@ -1265,6 +1265,38 @@ public void testUnresolvedTypesAreBanned() {
...
+  public void testSwitchStatement() {
+    // Since we're building the AST by hand, there won't be any types on it.
+    enableTypeInfoValidation = false;
+
+    String switchStatement = lines(
+        "function foo(x) {",
+        "  switch(x) {",
+        "    case 1: ",
+        "      var y = 5;",
+        "    default:",
```
nbeloglazov, line 1278: Is `default` here to to make this expression invalid? Or you add `defaultBlock` node later to make the whole tree invalid? Not sure what is the point of `defaultBlock` below.

patrick-sko: No, there is no reason for this. I deleted it to make it more clear. I also changed the `defaultBlock` to just `blockNode`. This node is added to give an extra child where it should not be allowed and that makes it invalid.

PR #3653, opened 2020-08-02
Remove number from CSSProperties properties

bugreportuser:
AFAICT, these properties are [always strings](https://www.w3.org/TR/DOM-Level-2-Style/css.html#CSS-CSS2Properties).

### DIFF
--- a/externs/browser/w3c_css.js
+++ b/externs/browser/w3c_css.js
@@ -1129,61 +1129,61 @@ CSSProperties.prototype.borderBottomStyle;
 CSSProperties.prototype.borderLeftStyle;
 
 /**
- * @type {string|number}
+ * @type {string}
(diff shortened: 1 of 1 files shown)


blickly:
Looks like a good change, and also brings our externs more in line with the TypeScript definitions.  @mprobst, would you mind trying out importing this PR?

mprobst:
Googlers: see http://test/OCL:324978548:BASE:324978753:1596618226888:fffe8c0f

brad4d:
Sorry this fell through the cracks in our PR reviewing process.

However, it turns out that making this change causes a large number of build failures within Google, so it really doesn't seem to be worth the cost of landing it to us.

bugreportuser:
Thanks for the explanation @brad4d. Does one property cause most of the failures or are they mostly even? I changed them all since they were all in the same file but I only need `height` and `width`

PR #3654, opened 2020-08-03
Make storage properties non-null

bugreportuser:
I don't think these can be null.
https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage
https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage


blickly:
Looks good, I'll take a look at importing this to see if there're any violations in google.

bugreportuser:
Thanks!

PR #3655, opened 2020-08-03
Implementation of branch instrumentation

patrick-sko:
Implemented branch instrumented. This PR is part of a series of PRs to implemented production type code instrumentation. The details of the entire proposal can be found in this [Implementation Doc](https://docs.google.com/document/d/1cdseVYW0x8LWccxYbT6GavrhjL00AA9SoafxAkqyu-8/edit?usp=sharing). 

This PR adds instrumentation to `If` statements, loop structures, `switch` statements,  `||`, `&&` , `??` operations, and ternary operations. It also adds unit tests for all these cases.

Another change was the addition of passing the column number to the `InstrumentCode` function to help distinguish between nested operators. In addition to this, I also created a new type, `Type.BRANCH_DEFAULT` which will help to distinguish if a statement is an `else, default` case. This should make it easier to later determine if code is always executed
and an if statement is not needed. The new type is also used for the line after a loop structure to make it possible to distinguish between the instrumented line in the loop vs outside the loop.

A final modification was to the unit test structure, by adding a function called `getInstrumentCodeLine(...)` which returns the InstrumentCode function call as a string. This helps with maintainability if the function is ever changed, and it also makes it easier to write unit tests and is less prone to human error.

### DIFF
--- a/src/com/google/javascript/jscomp/ProductionCoverageInstrumentationCallback.java
+++ b/src/com/google/javascript/jscomp/ProductionCoverageInstrumentationCallback.java
@@ -64,9 +64,17 @@ final class ProductionCoverageInstrumentationCallback
   private final ParameterMapping parameterMapping;
   boolean visitedInstrumentCodeFile = false;
 
-  private final String FUNCTION_TYPE = "Type.FUNCTION";
-  private final String BRANCH_TYPE = "Type.BRANCH";
-  private final String BRANCH_DEFAULT_TYPE = "Type.BRANCH_DEFAULT";
+  private enum Type {
+    FUNCTION("Type.FUNCTION"),
+    BRANCH("Type.BRANCH"),
+    BRANCH_DEFAULT("Type.BRANCH_DEFAULT");
+
+    private String value;
+
+    Type(String s){
+      this.value = s;
+    }
+  }
 
   /**
    * Stores a stack of function names that encapsulates the children nodes being instrumented. The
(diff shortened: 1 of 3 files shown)


test/com/google/javascript/jscomp/integration/ProductionCoverageInstrumentationPassIntegrationTest.java
```diff
@@ -101,26 +110,195 @@ public void testNoTranspilationFunctionInstrumentation() {
...
+
+    String[] sourceArr = {instrumentCodeSource, source};
+
+    String expected =
+        lines(
+            "for(var i = 0 ; i < 10 ; ++i) {",
+                getInstrumentCodeLine("C", 1, 0), ";",
+            "   console.log('*');",
+            "}",
+            getInstrumentCodeLine("E", 1, 0), ";");
```
nbeloglazov, line 242: Tbh I'm not sure what benefit this case adds. In `if` or `switch` statements you use default statements to detect when main condition never runs. While here there is no default case for loop. So if this line runs - it doesn't give any useful info about whether loop ran or not.

src/com/google/javascript/jscomp/ProductionCoverageInstrumentationCallback.java
```diff
@@ -91,23 +115,99 @@ public void visit(NodeTraversal traversal, Node node, Node parent) {
       return;
     }
 
+    String functionName = functionNameStack.peek();
+
     if (node.isFunction()) {
```
lauraharker, line 120: I'd find this chain of its more readable as a switch (node.getToken()), since the token names (FUNCTION, IF) pop out more.

You could put `NodeUtil.isLoopStructure(node)` in the default case since I still think using that function is a good idea.

test/com/google/javascript/jscomp/integration/ProductionCoverageInstrumentationPassIntegrationTest.java
```diff
@@ -150,6 +328,11 @@ public void testInstrumentationMappingIsCreated() {
         .isEqualTo("AAA");
   }
 
+  private String getInstrumentCodeLine(String encodedParam, int lineNo, int colNo){
+    return "module$exports$instrument$code.instrumentCodeInstance.instrumentCode(\"" +
```
lauraharker, line 332: Could you use Strings.lenientFormat here?

src/com/google/javascript/jscomp/ProductionCoverageInstrumentationCallback.java
```diff
@@ -91,23 +115,99 @@ public void visit(NodeTraversal traversal, Node node, Node parent) {
...
-      cachedFunctionName = NodeUtil.getBestLValueName(NodeUtil.getBestLValue(node));
-      instrumentCode(traversal, node.getLastChild(), cachedFunctionName);
+      // If the function node has been visited by visit() then we can be assured that all its
+      // children nodes have been visited and properly instrumented.
+      functionNameStack.pop();
+      instrumentBlockNode(node.getLastChild(), fileName, functionName, FUNCTION_TYPE);
+    } else if (node.isIf()) {
+      if (node.getChildCount() == 2) {
+        addDefaultBlock(node);
```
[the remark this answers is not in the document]
patrick-sko: Yes, changed name to be more clear.

src/com/google/javascript/jscomp/ProductionCoverageInstrumentationCallback.java
```diff
@@ -91,23 +115,99 @@ public void visit(NodeTraversal traversal, Node node, Node parent) {
...
+    parentNode.removeChild(originalNode);
+    Node newInstrumentationNode =
+        newInstrumentationNode(originalNode, fileName, functionName, type);
+
+    // newInstrumentationNode returns an EXPR_RESULT which cannot be a child of a COMMA node.
+    // Instead we use the child of of the newInstrumentatioNode which is a CALL node.
+    Node childOfInstrumentationNode = newInstrumentationNode.getFirstChild().detach();
+    Node infusedExp =
+        StatementFusion.fuseExpressionIntoExpression(childOfInstrumentationNode, originalNode);
+    parentNode.addChildToBack(infusedExp);
```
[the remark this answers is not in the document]
patrick-sko: Given the way I implemented it, yes this will always be the case. Although this is dependent on some behavior that isn't to obvious. I did some small changes by creating a clone of the original node and replacing the original node with the infusedNode to make it a but more clear.

test/com/google/javascript/jscomp/integration/ProductionCoverageInstrumentationPassIntegrationTest.java
patrick-sko: I don't really see that much benefit either. We can also add it back after if a use case presents itself.

src/com/google/javascript/jscomp/ProductionCoverageInstrumentationCallback.java
```diff
@@ -91,23 +123,109 @@ public void visit(NodeTraversal traversal, Node node, Node parent) {
...
+        instrumentBlockNode(node.getLastChild(), fileName, functionName, Type.FUNCTION.value);
+        break;
+      case IF:
+        Node ifTrueNode = node.getSecondChild();
+        instrumentBlockNode(ifTrueNode, sourceFileName, functionName, Type.BRANCH.value);
+        if (node.getChildCount() == 2) {
+          addElseBlock(node);
+        }
+        Node ifFalseNode = node.getLastChild();
+        if (NodeUtil.isEmptyBlock(ifFalseNode)
```
nbeloglazov, line 142: Add comment explaining this `else if` case.

src/com/google/javascript/jscomp/ProductionCoverageInstrumentationCallback.java
```diff
@@ -91,23 +123,109 @@ public void visit(NodeTraversal traversal, Node node, Node parent) {
...
   }
 
   /**
-   * Iterate over all collected block nodes within a Script node and add a new child to the front of
-   * each block node which is the instrumentation Node
+   * Given a node, this function will create a new instrumentationNode and combine it with the
+   * original node using a COMMA node.
+   */
+  private void addInstrumentationNodeWithComma(
+      Node originalNode, String fileName, String functionName, String type) {
```
nbeloglazov, line 204: Pass `Type` as enum, not as a string and then you can get its value only when you need it.

src/com/google/javascript/jscomp/ProductionCoverageInstrumentationCallback.java
```diff
@@ -60,14 +64,42 @@
   private final ParameterMapping parameterMapping;
   boolean visitedInstrumentCodeFile = false;
 
-  /** Stores the name of the current function that encapsulates the node being instrumented */
-  private String cachedFunctionName = "Anonymous";
+  private enum Type {
+    FUNCTION("Type.FUNCTION"),
```
nbeloglazov, line 68: You don't need value given that it's just repeats name. Instead you get enum name: `Type.FUNCTION.name()` which returns `FUNCTION`.

patrick-sko: In this particular case, the value I was passing is not a repeat of the enum name `"Type.FUNCTION"` vs `FUNCTION` but thinking about it more, it could be and probably should be since it doesn't make it any easier to convert back to an enum later so I changed it.

Issue #3563, opened 2020-02-27
Closure compiler doesn't recognize `\u{XXXX}` in regular expression


upsuper:
The following code:
```javascript
let x = /\u{ffff}/u;
```
would yield:
```
WARNING - [JSC_MALFORMED_REGEXP] Malformed Regular Expression: {ffff}
```
from closure compiler, but it's a valid regular expression, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Character_Classes

blickly:
Thanks for the report!
I think this might be a good starter issue for someone learning about the compiler.

muachilin:
I am interested in handling this issue! Can I work on this? Thank you!

blickly:
Sure, the `parsing` package is the right place to look.  I'd start with a unit test in:
https://github.com/google/closure-compiler/blob/master/test/com/google/javascript/jscomp/parsing/ParserTest.java

VoronD:
Hi! I would like to start working on it.

blickly:
Sounds good. Thanks!

I stand by my advice in https://github.com/google/closure-compiler/issues/3563#issuecomment-593548756 on where I'd recommend starting.

lauraharker:
The original repro was completely fixed by https://github.com/google/closure-compiler/pull/3656 (thanks @VoronD!) so I'm closing this issue.

Please file a new issue if you find any remaining bugs.

PR #3656, opened 2020-08-08
Improve unicode escape in regex

HenryRLee:
This PR supports regex Unicode escape in curly brackets format, provided that the 'u' flag presents.

For example:

```javascript
let x = /\u{01ac}/u;
```

Also, it extends Unicode up to 0x10FFFF. For example:

```javascript
let x = /\u{10FFFF}/u;
```

I believe this PR also fixes issue #3563.


christiantjl:
Nice work, Henry! Code LGTM.

I actually submitted a couple of Unicode logic PRs myself to try to learn more about the standard - I'd love if you could review one of them for me. See #3645, #3647.

HenryRLee:
Thanks @christiantjl.

I saw your PRs. #3645 looks good to me. When I have time, I will verify the correctness of the Unicode categories in #3647.

PR #3661, opened 2020-08-15
Add missing NodeList externs

bugreportuser:
Some of these methods don't have externs:
https://developer.mozilla.org/en-US/docs/Web/API/NodeList#Methods

### DIFF
--- a/externs/browser/w3c_dom1.js
+++ b/externs/browser/w3c_dom1.js
@@ -474,6 +474,24 @@ NodeList.prototype.item = function(index) {};
 NodeList.prototype.forEach = function(callback, opt_thisobj) {};
 
 /**
+ * @return {!Iterator<!Array<number|T>>}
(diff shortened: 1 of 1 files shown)


brad4d:

externs/browser/w3c_dom1.js
```diff
@@ -473,6 +473,24 @@ NodeList.prototype.item = function(index) {};
  */
 NodeList.prototype.forEach = function(callback, opt_thisobj) {};
 
+/**
+ * @return {!Iterator<!Array<number|T>>}
```
brad4d, line 477: BTW, these should be `IteratorIterable` instead of just `Iterator`.
I've fixed that in the version I've imported, so no need to fix it yourself.

bugreportuser: Oh, good catch

brad4d:
There were quite a few breakages from this PR.
It's going to take awhile to sort them out.

PR #3665, opened 2020-08-26
Adding null and extern checks to production instrumentation

patrick-sko:
PR to add several checks to improve usability. 

If global code that is instrumented is declared, set function name to `<Anonymous>`
Add check to ensure that array name passed to `--production_instrumentation_array_name` is also declared as an extern

Added unit test for these cases as well.


src/com/google/javascript/jscomp/CoverageInstrumentationPass.java
```diff
@@ -154,4 +164,9 @@ private Node createConditionalObjectDecl(String name, Node srcref) {
                     jscovData)));
     return var.useSourceInfoIfMissingFromForTree(srcref);
   }
+
+  public enum CoverageReach {
```
nbeloglazov, line 168: What is this enum for? I don't see any usages of it in among changed lines.

src/com/google/javascript/jscomp/ProductionCoverageInstrumentationCallback.java
```diff
@@ -52,6 +53,8 @@
   private final AbstractCompiler compiler;
   private final ParameterMapping parameterMapping;
 
+  private static final String GLOBAL_FUNCTION_NAME = "<Anonymous>";
```
nbeloglazov, line 56: ANONYMOUS_FUNCTION_NAME? not sure why it's called global

src/com/google/javascript/jscomp/CoverageInstrumentationPass.java
patrick-sko: The auto-formatter for whatever reason moved it down here. I'm not sure why but I undid it. I never worked with this enum.

src/com/google/javascript/jscomp/ProductionCoverageInstrumentationCallback.java
patrick-sko: I just global since this is also used when there is code in global scope that isn't a function. But changed to use Anonymous instead.

nbeloglazov (approved):

PR #3666, opened 2020-08-27
Organizing all instrumentation files

patrick-sko:
This PR takes care of some small clean-up.

It takes all files related to instrumentation and places them in their own directory.


src/com/google/javascript/jscomp/AbstractCompiler.java
```diff
@@ -477,7 +477,7 @@ public boolean isNormalizedObfuscated() {
    */
   public abstract Node getRoot();
 
-  abstract CompilerOptions getOptions();
+  public abstract CompilerOptions getOptions();
```
lauraharker, line 480: I'd rather this stay package-private. I think it's a better design to refactor the coverage passes to take in the specific option values they need.

But I'm fine with making these other methods public.

patrick-sko: Updated.

PR #3672, opened 2020-09-03
Try reducing bazel memory to fix travis build

rishipal:
Adds three flags to the bazel build:
--discard_analysis_cache
--nokeep_state_after_build
--notrack_incremental_state

They're defined here - https://docs.bazel.build/versions/master/memory-saving-mode.html


rishipal:
Closing as obsolete since we decided to move to using GitHub actions.

PR #3677, opened 2020-09-07
Replicate Travis setup using GitHub Actions v0

nreid260:


nreid260:
This was only meant to test GitHub Actions behaviour but that behaviour wasn't triggered.

PR #3681, opened 2020-09-09
Add support for Screen Wake Lock API

vobruba-martin:
This API is now enabled by default in Chrome.

See https://www.chromestatus.com/feature/4636879949398016
See https://w3c.github.io/screen-wake-lock/


brad4d:
I've imported, tested, reviewed, and submitted this internally.

PR #3682, opened 2020-09-09
Remove incorrect WebWorker extern type

niloc132:
The link in the jsdoc ` http://dev.w3.org/html5/workers/` describes the `Worker` type, and has no `WebWorker` defined at all. As far as I can tell (going back in the history of the web worker spec) it was never defined, perhaps just an early typo in this file that was never corrected?

### DIFF
--- a/externs/browser/html5.js
+++ b/externs/browser/html5.js
@@ -1400,48 +1400,6 @@ function importScripts(var_args) {}
 
 /**
  * @see http://dev.w3.org/html5/workers/
- * @constructor
- * @implements {EventTarget}
- */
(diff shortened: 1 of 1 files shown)


brad4d:
By Jove, I think you're right!

A search within Google doesn't *seem* to be turning up any code that actually needs this `WebWorker` externs definition.
I'll have to import and test the PR to be sure, though.

brad4d:

brad4d:
I've imported this and sent it for internal testing and review.
Looks like there's one file in closure library that uses `WebWorker`, which I have to fix before submitting this.

PR #3683, opened 2020-09-09
Reduce the number of targets to build using bazel

rishipal:


nreid260:
This change need to be made internally.

rishipal:
Closing as obsolete since we decided to move to using GitHub actions.

PR #3700, opened 2020-10-05
Added window.queueMicrotask() to externs

pdaehne:
The declaration for window.queueMicrotask() is missing in the externs declarations.


pdaehne:
> @googlebot I signed it!

PR #3706, opened 2020-10-19
Initial support for dynamic imports

ChadKillingsworth:
Allows dynamic import statements to be recognized by the compiler and utilized. Dynamic import expressions can be used as a script or module loader even without full module rewriting. Contains only basic support for type checking in that the type of the import is a Promise resolving to the unknown type.

Left for future PRs:
 - Module rewriting
 - Transpilation (aliases)

See the design document for the full plan: https://docs.google.com/document/d/1thSbGV5i96jltVwn76KYrPkWoNc-yj_KU3HtBGUCgMc/edit#heading=h.c037si364tjn


src/com/google/javascript/jscomp/GatherModuleMetadata.java
```diff
@@ -233,6 +235,13 @@ private void visitImportOrExport(NodeTraversal t, Node importOrExport) {
       }
     }
 
+    private void visitDynamicImport(NodeTraversal t, Node dynamicImport) {
+      currentModule
+          .metadataBuilder
+          .es6ImportSpecifiersBuilder()
+          .add("*");
```
brad4d, line 242: Shouldn't this be looking at the children of the dynamic import node?

src/com/google/javascript/jscomp/parsing/parser/FeatureSet.java
```diff
@@ -199,10 +199,8 @@
     // https://github.com/tc39/proposal-optional-catch-binding
     OPTIONAL_CATCH_BINDING("Optional catch binding", LangVersion.ES2019),
 
-    // Stage 3 proposals likely to be part of ES2020
-    DYNAMIC_IMPORT("Dynamic module import", LangVersion.ES_UNSUPPORTED),
-
     // ES 2020 Stage 4
+    DYNAMIC_IMPORT("Dynamic module import", LangVersion.ES2020),
```
brad4d, line 203: Yes, it's officially in ES_2020, but from our point of view, we don't really support it, so we can't put it in that language version yet. If you move this into our definition of ES_2020, you'll also need to modify the definition of the ES_NEXT FeatureSet to subtract this feature, because we cannot safely output it. (It will most likely cause miscellaneous exceptions in optimization passes that don't expect to see this node type.)

The usual procedure for a feature is something like:

1. It moves to ES_NEXT_IN when we can safely accept it in the input and transpile down to at least ES_NEXT level output (or just fail with an error because we cannot transpile it, since that is "safe" behavior).
2. It moves to ES_NEXT when we can safely carry the feature all the way through compilation and output it.

brad4d (changes requested):

src/com/google/javascript/jscomp/GatherModuleMetadata.java
ChadKillingsworth: I don't understand the module metadata completely, but these both refer to the same thing:

```js
import * as foo from './foo.js';
import('./foo.js').then((foo2) => {
   console.log(foo === foo2);
});
```

My logic was that a dynamic import was a "*" import so this was simply indicating that the currentModule referred to the module export namespace.

src/com/google/javascript/jscomp/parsing/parser/FeatureSet.java
ChadKillingsworth: @brad4d Based off the call Thursday, I left dynamic import in the ES_2020 category, but added a warning when any usage is encountered.

ChadKillingsworth:
@brad4d This should be ready for review. I ran my local project with this pass (combined with #3367 and #3724). Dynamic imports were successfully passed all the way through the compiler. I did find a couple of passes that needed small changes to support dynamic imports and those are included in this PR.

brad4d (approved):

brad4d:
I'll import and test this...

brad4d:
As this is currently written:

1. Dynamic will be rejected unless the language in is specified as `ES_2020` or newer.
2. The compiler will generate a warning (not an error), if the output language is older than `ES_2020`.
3. Regardless of the output language level, the `import(modulePath)` syntax will be left unmodified in the output.

Number 2 above is rather unusual for the compiler. I'd expect either transpilation to occur or an error to be reported.
I believe we're making an exception in this case for these reasons.

1. Many browsers had support for dynamic import long before the other features of `ES_2020`.
2. @ChadKillingsworth currently has a way to post-process the code to replace the `import()` calls with a method that will be available at runtime.
3. @ChadKillingsworth  plans to have a follow-up change that allow the user to specify a function name for the compiler to use to replace the `import`.

brad4d:
@ChadKillingsworth 

It's not going to work to diverge from the usual practice as I described in my last comment.
There's too much logic in the compiler that depends on it having a correct understanding of what features are currently present in the AST. And, that's a good thing, really.

The "transpilation" pass needs to report an error and stop compilation, not a warning.

I think making the dynamic import "work" for `--language_out < ES_2020` should wait for the follow up change that allows specifying a replacement function to call, so the transpilation can actually get rid of the dynamic import syntax.

OR, you could make the transpilation pass change the `DYNAMIC_IMPORT` node into a `CALL` node that just calls a function named `import`. When the code is output, it will look the same.

To make that work, the transpilation will also need to update the externs AST to include a declaration of a global `import` function, so it won't appear to be undeclared. Note that we cannot add such a declaration to our normal externs, because it is a syntax error to declare a function named `import`. Inserting it directly into the externs AST avoids the parsing, so we don't have this probllem.

I've also noticed a few other problems when trying to get this code to work inside Google.
I'll create a new code review for those.

src/com/google/javascript/jscomp/AstAnalyzer.java
```diff
@@ -520,6 +524,7 @@ boolean nodeTypeMayHaveSideEffects(Node n) {
       case FOR_IN: // assigns to a loop LHS
       case FOR_OF: // assigns to a loop LHS, runs an iterator
       case FOR_AWAIT_OF: // assigns to a loop LHS, runs an iterator, async operations.
+      case DYNAMIC_IMPORT:
         return true;
```
brad4d, line 528: missing `break;` here

src/com/google/javascript/jscomp/DefaultPassConfig.java
```diff
@@ -261,6 +262,11 @@ private void addModuleRewritingPasses(List<PassFactory> checks, CompilerOptions
       checks.add(rewriteCommonJsModules);
     }
 
+    if (options.getLanguageIn().toFeatureSet().contains(Feature.DYNAMIC_IMPORT)) {
+      // Temporary location until this pass does more than warn about usage
+      checks.add(es2020RewriteDynamicImport);
```
brad4d, line 267: Instead of being added here, the pass should be added in `addModuleRewritingPasses()`.

Also, it should be added by calling a new method `TranspilationPasses.addDynamicImportPass(checks)`.
We want to keep the transpilation pass creation in the `TranspilationPasses` class.

src/com/google/javascript/jscomp/Es2020RewriteDynamicImport.java
```diff
@@ -0,0 +1,41 @@
+package com.google.javascript.jscomp;
+
+import com.google.javascript.rhino.Node;
+
+public class Es2020RewriteDynamicImport implements CompilerPass, NodeTraversal.Callback {
```
brad4d, line 5: need a JavaDoc comment for this class

src/com/google/javascript/jscomp/DefaultPassConfig.java
brad4d: See `TranspliationPasses.addEs6ModulePass()` as an example.

src/com/google/javascript/jscomp/Es2020RewriteDynamicImport.java
```diff
@@ -0,0 +1,41 @@
...
+          "JSC_DYNAMIC_IMPORT_EXPERIMENTAL",
+          "Dynamic import support is experimental. No module rewriting is performed.");
+
+  public Es2020RewriteDynamicImport(AbstractCompiler compiler) {
+    this.compiler = compiler;
+  }
+
+  @Override
+  public void process(Node externs, Node root) {
+    NodeTraversal.traverse(compiler, root, this);
```
brad4d, line 20: This should follow the pattern usual to transplation passes.

1. Use `TranspilationPasses.processTranspile()` instead of `NodeTraversal.traverse()`
2. Use `TranspilationPasses.maybeMarkFeaturesAsTranspiledAway()`

See `Es6ForOfConverter#process()` as an example. (Note `Es6RewriteModules` isn't a good pattern to copy, though it may seem tempting since it's related to modules.)

src/com/google/javascript/jscomp/DefaultPassConfig.java
brad4d: Sorry, scratch that.

It's important that the dynamic import still be in the AST when type checking runs so we have a chance to do the type checking we want there.

The pass should get added at the start of `TranspilationPasses.addPostCheckTranspilationPasses()`.

brad4d (changes requested):

src/com/google/javascript/jscomp/AstAnalyzer.java
ChadKillingsworth: you want a `break;` even though there is a `return`?

src/com/google/javascript/jscomp/DefaultPassConfig.java
ChadKillingsworth: Moved

src/com/google/javascript/jscomp/Es2020RewriteDynamicImport.java
ChadKillingsworth: Done

ChadKillingsworth:
The function node conversion is a great idea - but I'll save that until we can actually make this useful.

src/com/google/javascript/jscomp/GatherModuleMetadata.java
```diff
@@ -213,6 +213,8 @@ public boolean shouldTraverse(NodeTraversal t, Node n, Node parent) {
         case MODULE_BODY:
           currentModule.hasModuleBody = true;
           break;
+        case DYNAMIC_IMPORT:
+          visitDynamicImport(t, n);
```
brad4d, line 217: Actually, this is where the break is missing.

src/com/google/javascript/jscomp/AstAnalyzer.java
brad4d: Sorry, I flagged the wrong spot.

brad4d (changes requested):

src/com/google/javascript/jscomp/GatherModuleMetadata.java
ChadKillingsworth: That makes a lot more sense - fixed!

brad4d:
I'm doing another trial import...

src/com/google/javascript/jscomp/parsing/parser/Parser.java
```diff
@@ -1112,6 +1112,8 @@ private boolean peekStatementStandard() {
       case TILDE:
       case BANG:
         return true;
+      case IMPORT:
+        return peekType(1) == TokenType.OPEN_PAREN;
```
brad4d, line 1116: Why is this needed?

I presume we forgot some case that should be added to ParserTest in order to test this line.
Could you add a test case there?

ChadKillingsworth: I was being overly cautious. I can indeed add a test case and ensure this isn't needed.

src/com/google/javascript/jscomp/DefaultPassConfig.java
```diff
@@ -64,6 +64,7 @@
 import com.google.javascript.jscomp.modules.ModuleMapCreator;
```
brad4d, line 64: please drop the now-no-op changes to this file

src/com/google/javascript/jscomp/TypeCheck.java
```diff
@@ -944,6 +944,7 @@ public void visit(NodeTraversal t, Node n, Node parent) {
       case IMPORT_SPEC:
       case IMPORT_SPECS:
       case IMPORT_STAR:
+      case DYNAMIC_IMPORT:
```
brad4d, line 947: This isn't right. `import(something)` is typeable.
Its type should have been already set by TypeInference.

DYNAMIC_IMPORT needs its own case that calls a dedicated `visitDynamicImport()` method.
That method should assert that the child node is typed as string and that the `DYNAMIC_IMPORT` node itself has already been given a type by TypeInference.

If you look around at the other `visitBlah()` methods, I'm sure you can find examples of how to do this.

brad4d: Also, need to add test cases to TypeCheckNoTranspileTest and TypeInferenceTest.

src/com/google/javascript/jscomp/GatherModuleMetadata.java
```diff
@@ -213,6 +213,9 @@ public boolean shouldTraverse(NodeTraversal t, Node n, Node parent) {
         case MODULE_BODY:
           currentModule.hasModuleBody = true;
           break;
+        case DYNAMIC_IMPORT:
```
brad4d, line 216: Could you add a test case to cover this to `GatherModuleMetadataTest`?

ChadKillingsworth: Done

src/com/google/javascript/jscomp/TypeCheck.java
ChadKillingsworth: That helped - I refactored this significantly. I added a check so that the import specifier cannot be undefined or null.

src/com/google/javascript/jscomp/DefaultPassConfig.java
ChadKillingsworth: done

src/com/google/javascript/jscomp/parsing/parser/Parser.java
ChadKillingsworth: It wasn't needed. I removed it and added a test case to prove it wasn't needed.

ChadKillingsworth:
@brad4d Spent a lot of time this weekend tracing through code and was able to finally figure out how to get the specific type of a dynamic import expression.

brad4d:
Importing latest for further review/testiing...

brad4d (changes requested):
I think this is almost ready to land.

My new comments are mostly requests for tests.

General rule we try to follow: If you modify Foo, there should be some corresponding update to FooTest.

Very nice that you worked out where to get the type information for the dynamic import nodes.
That code looks solid to me.

I do need to ask for one other thing, though.
Since this behavior is definitely on the experimental side, I need to make sure people don't trip over it.

Could you add a CommandLineRunner flag and corresponding CompilerOptions field that must be enabled in order for the pass-through behavior to occur? When not enabled, the compiler should report an error as it does now.

Thanks,
Bradford

src/com/google/javascript/jscomp/GatherModuleMetadata.java
```diff
@@ -233,6 +236,15 @@ private void visitImportOrExport(NodeTraversal t, Node importOrExport) {
       }
     }
 
+    private void visitDynamicImport(NodeTraversal t, Node dynamicImport) {
```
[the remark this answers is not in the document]
ChadKillingsworth: done

src/com/google/javascript/jscomp/AstValidator.java
```diff
@@ -433,6 +433,11 @@ public void validateExpression(Node n) {
         validateAwait(n);
         return;
 
+      case DYNAMIC_IMPORT:
```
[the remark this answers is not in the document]
ChadKillingsworth: done

src/com/google/javascript/jscomp/PureFunctionIdentifier.java
```diff
@@ -879,6 +879,12 @@ private void updateSideEffectsForNode(
           }
           break;
 
+        case DYNAMIC_IMPORT:
```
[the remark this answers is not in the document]
ChadKillingsworth: done

test/com/google/javascript/jscomp/TypeInferenceTest.java
```diff
@@ -3751,6 +3765,31 @@ public void testShneTightensUnknownOperandOnRhs() {
...
+            registry.getNativeObjectType(JSTypeNative.PROMISE_TYPE),
+            registry.getNativeType(JSTypeNative.UNKNOWN_TYPE));
+    assertType(getType("foo")).isSubtypeOf(promiseOfUnknownType);
+  }
+
+  @Test
+  public void testDynamicImport2() {
+    withModules(ImmutableList.<String>builder()
+        .add("export default 1; export /** @return {string} */ function Bar() { return 'bar'; };")
+        .add("const foo = import('./testcode0');")
```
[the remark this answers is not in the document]
ChadKillingsworth: done

test/com/google/javascript/jscomp/TypeInferenceTest.java
```diff
@@ -3751,6 +3765,31 @@ public void testShneTightensUnknownOperandOnRhs() {
...
+    JSType promiseOfUnknownType =
+        registry.createTemplatizedType(
+            registry.getNativeObjectType(JSTypeNative.PROMISE_TYPE),
+            registry.getNativeType(JSTypeNative.UNKNOWN_TYPE));
+    assertType(getType("foo")).isSubtypeOf(promiseOfUnknownType);
+  }
+
+  @Test
+  public void testDynamicImport2() {
+    withModules(ImmutableList.<String>builder()
```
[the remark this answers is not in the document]
ChadKillingsworth: done

test/com/google/javascript/jscomp/RewriteDynamicImportTest.java
```diff
@@ -0,0 +1,51 @@
+/*
+ * Copyright 2014 The Closure Compiler Authors.
```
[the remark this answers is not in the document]
ChadKillingsworth: done

src/com/google/javascript/jscomp/TypeValidator.java
```diff
@@ -345,6 +350,21 @@ void expectAsyncGeneratorSupertype(Node n, JSType type, String msg) {
     }
   }
 
+  /**
+   * Expect the type to be a supertype of `Promise`.
+   *
+   * <p>`Promise` is the <em>lower</em> bound of the declared return type, since that's what async
+   * functions always return; the user can't return an instance of a more specific type.
+   */
+  void expectValidDynamicImportType(Node n, JSType type) {
```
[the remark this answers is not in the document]
ChadKillingsworth: This entire function ended up being unnecessary

src/com/google/javascript/jscomp/TypeValidator.java
```diff
@@ -345,6 +350,21 @@ void expectAsyncGeneratorSupertype(Node n, JSType type, String msg) {
...
+    if (promiseOfUnknownType.isSubtypeOf(type)) {
```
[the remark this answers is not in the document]
ChadKillingsworth: Trying to write a test for this made it very clear this was impossible to trigger. So I just removed the the entire function.

src/com/google/javascript/jscomp/AstAnalyzer.java
```diff
@@ -261,6 +261,10 @@ private boolean checkForStateChangeHelper(Node n, boolean checkForNewObjects) {
         // Variable declarations are side-effects.
         return true;
 
+      // import() expressions have side effects
```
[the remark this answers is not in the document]
ChadKillingsworth: done

src/com/google/javascript/jscomp/NodeUtil.java
```diff
@@ -4754,6 +4754,7 @@ static boolean evaluatesToLocalValue(Node value) {
             && evaluatesToLocalValue(value.getLastChild());
       case THIS:
       case SUPER:
+      case DYNAMIC_IMPORT:
```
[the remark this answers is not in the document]
ChadKillingsworth: done

src/com/google/javascript/jscomp/TypeValidator.java
```diff
@@ -345,6 +350,21 @@ void expectAsyncGeneratorSupertype(Node n, JSType type, String msg) {
...
+      return;
+    }
+
+    JSError err = JSError.make(n, INVALID_DYNAMIC_IMPORT_TYPE, type.toString());
+    registerMismatchAndReport(type, promiseOfUnknownType, err);
```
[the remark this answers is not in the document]
ChadKillingsworth: I added additional tests to TypeCheckTest and removed this entire function. I found it impossible to trigger this case. One note: this did outline a parse exception when trying to type-case a dynamic import. I fixed that too.

ChadKillingsworth:
I've now added a `--allow_dynamic_import` flag. Without this flag a special check pass will error if encountering any DYNAMIC_IMPORT node.

brad4d:
I've imported the latest version and started standard testing on it.

brad4d:
I've sent this to @concavelenz for internal review

brad4d:
Internal review is nearly finished. Addressed more review comments from @concavelenz today.

PR #3718, opened 2020-11-03
Fix NavigatorUAData definition

vobruba-martin:
See https://wicg.github.io/ua-client-hints/#navigatoruadata


mprobst:
Internal issue created at b/172411122/

PR #3724, opened 2020-11-13
Refactor ExtractPrototypeMemberDeclarations to use a variable per chunk

ChadKillingsworth:
Switches from a single global variable to a per chunk variable. The prevents RescopeGlobalSymbolNames from adding a global namespace lookup to every variable reference thus undoing some of the benefit of the pass.

Also required for #3367 to prevent attempting to write to a variable in a separate chunk.

### DIFF
--- a/test/com/google/javascript/jscomp/CompilerTestCase.java
+++ b/test/com/google/javascript/jscomp/CompilerTestCase.java
@@ -1125,6 +1125,16 @@ public abstract class CompilerTestCase {
   }
 
   /**
+   * Verifies that the compiler pass's JS output matches the expected output.
+   *
+   * @param js Input
+   * @param expected Expected JS output
+   */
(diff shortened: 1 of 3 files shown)


test/com/google/javascript/jscomp/CompilerTestCase.java
```diff
@@ -1124,6 +1124,16 @@ protected void test(String js, String expected) {
     test(srcs(js), expected(expected));
   }
 
+  /**
+   * Verifies that the compiler pass's JS output matches the expected output.
+   *
+   * @param js Input
```
rishipal, line 1130: Incorrect param name and comment.

ChadKillingsworth: Fixed

brad4d:
imported and sent for internal review and testing

PR #3734, opened 2020-11-30
Add support for File System Access API

vobruba-martin:
This API is enabled in Chrome 86 by default.

See https://www.chromestatus.com/feature/6284708426022912
See https://wicg.github.io/file-system-access/

### DIFF
--- a/externs/browser/wicg_file_system_access.js
+++ b/externs/browser/wicg_file_system_access.js
@@ -122,8 +122,9 @@ FileSystemRemoveOptions.prototype.recursive;
 /**
  * @interface
  * @extends {FileSystemHandle}
- * @extends {AsyncIterable<string, !FileSystemHandle>}
+ * @extends {AsyncIterable<!Array<string|!FileSystemHandle>>}
  * @see https://wicg.github.io/file-system-access/#api-filesystemdirectoryhandle
+ * @see https://wicg.github.io/file-system-access/#api-filesystemdirectoryhandle-asynciterable
  */
 var FileSystemDirectoryHandle = function() {};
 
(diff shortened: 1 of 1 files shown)


blickly:
Thanks for the PR!  I'll work on merging this in.

vobruba-martin:
Hi @blickly , I found out that `AsyncIterable<string, !FileSystemHandle>` is wrong and the directory iterator should be typed [differently](https://wicg.github.io/file-system-access/#api-filesystemdirectoryhandle-asynciterable). I will update this PR tomorrow.

externs/browser/wicg_file_system_access.js
```diff
@@ -0,0 +1,296 @@
...
+var FileSystemHandleKind;
+
+
+/**
+ * @interface
+ * @see https://wicg.github.io/file-system-access/#api-filesystemhandle
+ */
+var FileSystemHandle = function() {};
+
+/** @type {!FileSystemHandleKind} @const */
```
blickly, line 34: Nit, we often combine `@type {t} @const` into just `@const {t}`

externs/browser/wicg_file_system_access.js
```diff
@@ -0,0 +1,296 @@
...
+FileSystemHandle.prototype.kind;
+
+/** @type {string} @const */
```
blickly, line 37: Same here

blickly (commented):
OK, thanks for the heads up. I'll hold off for now, then.

Just a few minor style comments.

vobruba-martin:
@blickly Done. If you need to squash commits let me know.

blickly:
We've improved our process so that we no longer require squashing before merging.

I'll kick of the tests right now for starting the import.

blickly:
Ok, looks good, thanks! I think I've found all the other folks in google who've written conflicting definitions and added updates for them.

The one last thing is I think we'll need to include a copyright in this file like the ones at the top of the other externs files, like:
https://github.com/google/closure-compiler/blob/master/externs/es3.js#L1-L15

blickly:
OK, everything looks good. Looks like I missed getting this in today, but I think everything has been sorted out to submit on Monday.

blickly:
Not sure why https://github.com/google/closure-compiler/commit/b8b0dd5bfde8428cf2a61d6724130cf0766e20b8 isn't listed as a merge of this PR, but I think it should be.

PR #3737, opened 2020-12-05
Preserve parentheses on spread operator's AssignmentExpression

### LINKED ISSUE #3736: Incorrect removal of parentheses
Closure Compiler sometimes incorrectly removes semantically important parentheses.

`example.js`:
```javascript
function spreadSet() {
    var _a;
    return [...(_a = new Set([1]), (_a !== null && _a !== void 0 ? _a : []))];
}
console.log(spreadSet());
```
```
$ java -jar closure-compiler-v20201102.jar --compilation_level ADVANCED_OPTIMIZATIONS --formatting PRETTY_PRINT --language_out=ECMASCRIPT_2015 --js_output_file=bug.js example.js && cat bug.js && node bug.js
'use strict';
console.log(function() {
  var a;
  return [...a = new Set([1]), null !== a && void 0 !== a ? a : []];
}());

[ 1, Set(1) { 1 } ]
```
Notice the difference to just running the original script directly:
```
$ node example.js
[ 1 ]
```

While the input syntax may seem odd, it is the result of TypeScript+tsickle running on the following ES2020->ES2017 code:
```javascript
function spreadSet() {
    return [...(new Set([1]) ?? [])];
}

console.log(spreadSet());
```

The incorrect behavior can also be observed in this simpler-to-read code, though it does throw a suspicious code warning:
```javascript
console.log(...([1], [2])); // 2
```

Incorrect emit:
```javascript
console.log(...[1], [2]); // 1 [ 2 ]
```

When outputting to ES5 the bug is not observed.

DavidANeil:
Closes #3736.

### DIFF
--- a/src/com/google/javascript/jscomp/CodeGenerator.java
+++ b/src/com/google/javascript/jscomp/CodeGenerator.java
@@ -436,7 +436,7 @@ public class CodeGenerator {
       case ITER_SPREAD:
       case OBJECT_SPREAD:
         add("...");
-        add(node.getFirstChild());
+        addExpr(first, NodeUtil.precedence(type), Context.OTHER);
         break;
 
       case EXPORT:
--- a/test/com/google/javascript/jscomp/CodePrinterTest.java
+++ b/test/com/google/javascript/jscomp/CodePrinterTest.java
@@ -3034,6 +3034,7 @@ public final class CodePrinterTest extends CodePrinterTestBase {
     assertPrintSame("f(...args)");
     assertPrintSame("f(...arrayOfArrays[0])");
     assertPrintSame("f(...[1,2,3])");
+    assertPrintSame("f(...([1],[2]))");
   }
 
   @Test


brad4d:
I've imported this PR and am reviewing it now.
Thanks!

src/com/google/javascript/jscomp/CodeGenerator.java
```diff
@@ -436,7 +436,15 @@ protected void add(Node node, Context context) {
       case ITER_SPREAD:
       case OBJECT_SPREAD:
         add("...");
-        add(node.getFirstChild());
+        if (first.getIsParenthesized()) {
```
brad4d, line 439: I believe the right way to do this is actually to use

`addExpr(first, NodeUtil.precedence(type), Context.OTHER);`

instead of this if-statement.

brad4d (changes requested):

src/com/google/javascript/jscomp/CodeGenerator.java
DavidANeil: I figured as much, since this seemed like a common problem for a code generator, I just couldn't find the correct way.
I've force pushed the branch with the fix.

brad4d:
Thanks for addressing my review comment.

I'm sending this for internal review / test / submit now.

PR #3746, opened 2020-12-30
Update package.json, runtime_tests to use Bazel build

### LINKED ISSUE #3743: Open source build instructions unclear
I intended to add details to the README regarding the runtime tests and building/testing with NPM/yarn (i.e., via `yarn build` and `yarn test`) when I realized I cannot get the Maven build to work anymore, now that `pom.xml` is no longer in the root directory. Despite my best efforts to repair this (moving POM to `pom.xml`, adding `project.source` version to silence "source version must be > 8" errors, etc.) so that `mvn -DskipTests` and so on execute as expected, I cannot the project to compile successfully.

Any help from someone more experienced with Maven would be greatly appreciated. I will send my README PR as soon as this is working again.

ctjlewis:
Please let me know if any changes are needed. This PR mainly just updates the old package.json scripts to use Bazel instead of Maven and closes #3743, but some other small package.json updates are included:

- Upgraded NPM deps
- Added `yarn compiler` script so you can easily run `yarn compiler -O ADVANCED file.js` to call the locally built version of the compiler in `bazel-bin/`

And a few runtime test tweaks:
- Missing runtime test was added
- `runtime_tests:build` can now accept a test pattern as an argument (e.g. `yarn runtime_tests:build module_runtime` to build `module_runtime_test.js` only)

---

I will submit README updates about building the compiler with `yarn build` and building/running the runtime tests as part of a PR dedicated strictly to refreshing the README.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "closure-compiler",
   "config": {
-    "runtime_tests_dir": "./test/com/google/javascript/jscomp/runtime_tests"
+    "runtime_tests": "./test/com/google/javascript/jscomp/runtime_tests"
   },
   "version": "1.0.0",
   "repository": "git@github.com:google/closure-compiler.git",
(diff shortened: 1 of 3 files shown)


test/com/google/javascript/jscomp/runtime_tests/module_tests/module_test_resources/+subdir/module_es6.js
```diff
@@ -0,0 +1 @@
+export const a = 1;
```
ctjlewis, line 1: This was a missing runtime test and is intentionally included.

package.json
```diff
@@ -14,18 +14,20 @@
...
-    "test": "mvn test"
+    "build": "bazel build //:compiler_unshaded_deploy.jar",
+    "build:all": "bazel build :all",
+    "compile": "java -jar bazel-bin/compiler_unshaded_deploy.jar",
+    "clean": "yarn runtime_tests:clean && bazel clean",
+    "runtime_tests:clean": "rm -rf $npm_package_config_runtime_tests/**/build/*",
+    "runtime_tests:build": "$npm_package_config_runtime_tests/utils/build.sh",
+    "runtime_tests:run": "jest --testRegex $npm_package_config_runtime_tests/utils/test/* --verbose=true",
+    "runtime_tests": "yarn --silent runtime_tests:clean && yarn --silent runtime_tests:build && yarn run runtime_tests:run",
+    "test": "bazel test //:compiler_unshaded_deploy.jar",
```
nreid260, line 30: There's nothing to test here. Bazel makes a distinction between targets that are tests and targets that are not.

ctjlewis: Can you leave a more helpful suggestion, like how it should be changed?

nreid260: Oh, can we just delete it? It won't do anything. I think "test:all" covers it, if it's even worth making a distinction.

ctjlewis: I can absolutely delete it, I just have the script here so people can interface completely with `yarn`/`npm` without needing to deal with Bazel-y stuff: `yarn build`, `yarn test`, etc.

I will actually probably add the `runtime_tests` test to `yarn test`, such that `yarn test` is: ```bazel test:all && runtime_tests```

Thoughts? It's just a way to do all the Bazel logic for OSS people ahead of time so they can just easily interface with the package.json scripts.

nreid260: I would probably make `test` = `bazel test //:all` and `test:all` = `bazel test //:all && runtime_tests`, but whatever you decide is fine. The important thing is that `bazel test //:compiler_unshaded_deploy.jar` is a no-op.

nreid260 (approved):

package.json
ctjlewis: Awesome, I will make those changes right now.

ctjlewis:
I made that change as you recommended, and renamed `yarn` calls to `npm` in package.json so people can still call them with just NPM.

PR #3747, opened 2020-12-30
Revamp README [WIP]

ctjlewis:
This is a draft of my suggested changes to the README, with the ultimate goal of making it much easier for users to install, build, and interact/integrate with the compiler.

I will be updating this over the coming days, and it is quite a mess atm, but I wanted to leave it here in case anyone had thoughts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,7 +176,7 @@ command-line executable (compatible with the one in linux).
     If you want to integrate the compiler into a larger Java program, depend on
     `//:compiler_shaded_deploy.jar` instead.
 
-### Using a terminal
+### Building from a terminal
 
 You can trigger the build process easily with package.json scripts or by calling
 Bazel manually. 
@@ -189,7 +189,7 @@ yarn build
 yarn build:all
 ```
 
-### Using an IDE
+### Building from an IDE
 
 See [Bazel IDE Integrations](https://docs.bazel.build/versions/master/ide.html).
 
@@ -204,7 +204,7 @@ package.json script:
 yarn compile [...args]
 ```
 
-#### Run using Eclipse
+#### Running using Eclipse
 
 1.  Open the class `src/com/google/javascript/jscomp/CommandLineRunner.java` or
     create your own extended version of the class.
@@ -214,7 +214,7 @@ yarn compile [...args]
     [bug](https://stackoverflow.com/questions/4711098/passing-end-of-transmission-ctrl-d-character-in-eclipse-cdt-console)
     regarding passing "End of Transmission" in the Eclipse console.
 
-## How to Contribute
+## Contributing
 
 ### Contributor code of conduct
 
--- a/README.md
+++ b/README.md
@@ -137,9 +137,9 @@ the inputs automatically.
 
 1.  Post in the [Closure Compiler Discuss
     Group](https://groups.google.com/forum/#!forum/closure-compiler-discuss).
-1.  Ask a question on [Stack
+2.  Ask a question on [Stack
     Overflow](https://stackoverflow.com/questions/tagged/google-closure-compiler).
-1.  Consult the [FAQ](https://github.com/google/closure-compiler/wiki/FAQ).
+3.  Consult the [FAQ](https://github.com/google/closure-compiler/wiki/FAQ).
 
 ## Building the Compiler
 
--- a/README.md
+++ b/README.md
@@ -193,7 +193,7 @@ yarn build:all
 
 See [Bazel IDE Integrations](https://docs.bazel.build/versions/master/ide.html).
 
-## Running
+### Running
 
 Once the compiler has been built, the compiled JAR will be in the `bazel-bin/`
 directory. You can access it with a call to `java -jar ...` or by using the
@@ -204,7 +204,7 @@ package.json script:
 yarn compile [...args]
 ```
 
-### Run using Eclipse
+#### Run using Eclipse
 
 1.  Open the class `src/com/google/javascript/jscomp/CommandLineRunner.java` or
     create your own extended version of the class.
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ the inputs automatically.
     Overflow](https://stackoverflow.com/questions/tagged/google-closure-compiler).
 1.  Consult the [FAQ](https://github.com/google/closure-compiler/wiki/FAQ).
 
-## Building
+## Building the Compiler
 
 To build the compiler yourself, you will need the following:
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -15,34 +18,144 @@ pitfalls.
...
-    command line.
-
-## Options for Getting Help
-
-1.  Post in the
-    [Closure Compiler Discuss Group](https://groups.google.com/forum/#!forum/closure-compiler-discuss).
-1.  Ask a question on
-    [Stack Overflow](https://stackoverflow.com/questions/tagged/google-closure-compiler).
+The easiest way to get the compiler is to install the precompiled binary
+globally using [NPM](https://npmjs.com) or [Yarn](https://yarnpkg.com):
```
ChadKillingsworth, line 22: Do you want to mention using `npx` to use the compiler without even installing?

ctjlewis: My reasoning here was to help people get a global binary in their `PATH` as quickly as possible, cross-platform. I believe `npm i -g` and `yarn global add` provide the best way to do this, and this is how I update and interact with the compiler on my own machine.

README.md
```diff
@@ -15,34 +18,144 @@ pitfalls.
...
+## Getting Help
+
+1.  Post in the [Closure Compiler Discuss
+    Group](https://groups.google.com/forum/#!forum/closure-compiler-discuss).
+1.  Ask a question on [Stack
+    Overflow](https://stackoverflow.com/questions/tagged/google-closure-compiler).
 1.  Consult the [FAQ](https://github.com/google/closure-compiler/wiki/FAQ).
 
-## Building it Yourself
+## Building
```
nreid260, line 142: Because this is a compiler, we should be clear that we're building the compiler itself, not building JS with the compiler.

README.md
```diff
@@ -15,34 +18,144 @@ pitfalls.
...
 
-### Using [Bazel](https://bazel.build/)
+| Prerequisite | Description |
+| --- | --- |
+| [Java 8 or later](https://java.com) | Used to compile the compiler's source code. |
+| [Git](https://git-scm.com/) | Used by Bazel to download dependencies. |
+| [Bazel](https://docs.bazel.build/versions/master/install.html) | Used to build the various compiler targets. |
 
-Building using bazel requires `git` (to download dependencies).
+### Installing Bazel
```
nreid260, line 152: In general users should prefer bazelisk. Otherwise they'll get errors as the default bazel version changes but the BUILD files are not.

README.md
```diff
@@ -61,41 +174,33 @@ command-line executable (compatible with the one in linux).
...
 
-The Closure Compiler has many options for reading input from a file, writing
-output to a file, checking your code, and running optimizations. To learn more,
-type
+Once the compiler has been built, the compiled JAR will be in the `bazel-bin/`
+directory. You can access it with a call to `java -jar ...` or by using the
+package.json script:
 
+```bash
+# java -jar bazel-bin/compiler_unshaded_deploy.jar [...args]
```
nreid260, line 201: Am I mistaken that the deploy JAR is self-executable? Something about ELF...

ctjlewis: IIRC it *purported* to be executable, but did not seem to be:
```none
christian@workspaces:~/PersonalProjects/closure-compiler$ ./bazel-bin/compiler_unshaded.jar
./bazel-bin/compiler_unshaded.jar: line 1: $'PK\003\004': command not found
./bazel-bin/compiler_unshaded.jar: line 2: META-INF/��PK: No such file or directory
./bazel-bin/compiler_unshaded.jar: line 3: zD�PPMETA-INF/MANIFEST.MFManifest-Version:: No such file or directory
./bazel-bin/compiler_unshaded.jar: line 4: Created-By:: command not found
./bazel-bin/compiler_unshaded.jar: line 5: Target-Label:: command not found
./bazel-bin/compiler_unshaded.jar: line 6: $'\r': command not found
./bazel-bin/compiler_unshaded.jar: line 7: $'PK\001\002': command not found
./bazel-bin/compiler_unshaded.jar: line 9: META-INF/��PK: No such file or directory
./bazel-bin/compiler_unshaded.jar: line 11: zD�PP+META-INF/MANIFEST.MFPK}�: No such file or directory
```

README.md
ctjlewis: I didn't see any references to Bazelisk in the original README. I wonder if @lauraharker has thoughts on this. Happy to make this change, just looking for input.

avindra (changes requested):
Suggested to keep 1

README.md
nreid260: For clarity, Bazelisk is a small wrapper around all the versions of Bazel. It reads the `.bazelversion` file and then delegates to the specified version.

My concern is that any commands we mention should call bazelisk, not bazel. This probably also includes commands in the package JSON.

If we don't, users will get difficult to debug errors. The BUILD files will parse with many versions of Bazel, but the behaviour will be slightly different and in many cases won't build. The diagnostics when this happens point mention missing env variables or build targets; nothing to do with the real problem.

I don't think we need to belabor the distinction between Bazel and Bazelisk. We just want to point the most people to the best chance of success. And possibly convey that they are approximately the same thing.

ctjlewis: But if *I* ran `bazelisk` right now, it would fail. I am morally opposed to catering to Windows users specifically for command-line examples - would adding a note ("on Windows, replace `bazel` with `bazelisk`") or something similar work?

nreid260: I think there was a mistake last time I edited this file. bazel nor bazelisk is installed by default on any OS that I'm aware of. Additionally, bazelisk should be used regardless of OS.

ctjlewis: Oh, very interesting, so this is a substantive change to the installation instructions. I can make that change for you if you summarize what should be swapped out, or you can do it directly.

README.md
```diff
@@ -14,35 +17,147 @@ checks syntax, variable references, and types, and warns about common JavaScript
...
+
+### Installing Bazel
+
+#### Linux / MacOS
+All other required tools should be available in a [standard
+distribution](https://docs.bazel.build/versions/master/install.html).
+
+#### Windows
 On Windows, building closure-compiler additionally requires MSYS2, and a zip
 command-line executable (compatible with the one in linux).
```
nreid260, line 162: I think this is also stale https://docs.bazel.build/versions/master/windows.html#running-bazel-msys2-shell-vs-command-prompt-vs-powershell

I'm not even sure it's worthwhile to differentiate OSes here, but it's tough to say because I don't think anyone on the main team ever builds on Windows.

nreid260 (commented):
I'm about to merge this (with some tweaks to the Bazelisk stuff). Please let me know if there are more edits you want to make.

Generally, our philosophy with documentation (vs code) is to update first and refine later.

ctjlewis:
It should be good to go for now @nreid260, my working tree is clean.

ctjlewis:
@nreid260 Table formatting is all messed up. :cry: I will send a fix in.

nreid260:
Don't worry about formatting. I'm running the Google MD formatter.

ctjlewis:
*Oh*, I was very confused at how that happened. I see now.

nreid260:
I see now. That table did get butchered. Apparently we use a different MD parser than GitHub.

Issue #2316, opened 2017-02-17
Feature request: Cross module code motion should move pure calls


nicks:
Create two modules:

```
// module 1
/** @constructor */ function Point(x, y) { this.x = x; this.y = y; }
var ZERO = new Point(0, 0);
```

```
// module 2
alert(ZERO)
```

It would be nice if cross-module code motion could figure out that Point is "pure" and so everything can be moved into the second module.

http://closure-compiler-debugger.appspot.com/#input0%3D%252F**%2520%2540constructor%2520*%252F%2520function%2520Point(x%252C%2520y)%2520%257B%2520this.x%2520%253D%2520x%253B%2520this.y%2520%253D%2520y%253B%2520%257D%250Avar%2520ZERO%2520%253D%2520new%2520Point(0%252C%25200)%253B%26input1%3Dalert(ZERO)%26conformanceConfig%26externs%26refasterjs-template%26includeDefaultExterns%3D1%26CHECK_SYMBOLS%3D1%26MISSING_PROPERTIES%3D1%26TRANSPILE%3D1%26CHECK_TYPES%3D1%26COMPUTE_FUNCTION_SIDE_EFFECTS%3D1%26MARK_NO_SIDE_EFFECT_CALLS%3D1%26CROSS_MODULE_CODE_MOTION%3D1%26CROSS_MODULE_METHOD_MOTION%3D1%26CLOSURE_PASS%3D1%26PRESERVE_TYPE_ANNOTATIONS%3D1%26PRETTY_PRINT%3D1

WearyMonkey:
I've raised a PR for this feature here:
https://github.com/google/closure-compiler/pull/3751

PR #3751, opened 2021-01-12
Feature: Enable cross chunk code motion for pure function values

WearyMonkey:
This PR enables calls to pure function to be moved to deeper chunks.

For example,

if the root chunk contains:

```
function pure(a) {
   return a + 1;
}

var b = pure(1);
```

and a child chunk with a dependency on the root contains:

```
var c = b;
```

Then the `pure` and `b` globals are moved to the child chunk.

This has a significant impact on large code bases that make use of factories to create globals. One of our builds is 10mb in size, with a root chunk of 1.8mb, this change reduces the root chunk to 1.6mb, or a 11% reduction.

Related feature request from a few years ago:
https://github.com/google/closure-compiler/issues/2316

### DIFF
--- a/src/com/google/javascript/jscomp/CrossChunkReferenceCollector.java
+++ b/src/com/google/javascript/jscomp/CrossChunkReferenceCollector.java
@@ -350,6 +350,10 @@ public final class CrossChunkReferenceCollector implements ScopedCallback, Compi
         return true;
 
       case CALL:
+        if (valueNode.getSideEffectFlags() == 0) {
+          return true;
+        }
+
         // In general it is not safe to move function calls, but we carve out an exception
         // for the special stub method calls used for CrossChunkMethodMotion.
         // Case: `JSCompiler_stubMethod(x)`
--- a/src/com/google/javascript/jscomp/CrossChunkReferenceCollector.java
+++ b/src/com/google/javascript/jscomp/CrossChunkReferenceCollector.java
@@ -351,7 +351,7 @@ public final class CrossChunkReferenceCollector implements ScopedCallback, Compi
 
       case CALL:
         // If the function has been marked as having no side effects then it is safe to move to another module.
-        if (valueNode.getSideEffectFlags() == 0) {
+        if (valueNode.isNoSideEffectsCall()) {
           return true;
         }
 
--- a/src/com/google/javascript/jscomp/CrossChunkReferenceCollector.java
+++ b/src/com/google/javascript/jscomp/CrossChunkReferenceCollector.java
@@ -16,16 +16,21 @@
 
 package com.google.javascript.jscomp;
 
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkState;
+
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Iterables;
 import com.google.javascript.jscomp.NodeTraversal.ScopedCallback;
 import com.google.javascript.rhino.Node;
-
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
 import javax.annotation.Nullable;
-import java.util.*;
-
-import static com.google.common.base.Preconditions.checkNotNull;
-import static com.google.common.base.Preconditions.checkState;
 
 /** Collects global variable references for use by {@link CrossChunkCodeMotion}. */
 public final class CrossChunkReferenceCollector implements ScopedCallback, CompilerPass {
--- a/src/com/google/javascript/jscomp/CrossChunkReferenceCollector.java
+++ b/src/com/google/javascript/jscomp/CrossChunkReferenceCollector.java
@@ -16,21 +16,16 @@
 
 package com.google.javascript.jscomp;
 
-import static com.google.common.base.Preconditions.checkNotNull;
-import static com.google.common.base.Preconditions.checkState;
-
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Iterables;
 import com.google.javascript.jscomp.NodeTraversal.ScopedCallback;
 import com.google.javascript.rhino.Node;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
+
 import javax.annotation.Nullable;
+import java.util.*;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkState;
 
 /** Collects global variable references for use by {@link CrossChunkCodeMotion}. */
 public final class CrossChunkReferenceCollector implements ScopedCallback, CompilerPass {
@@ -350,11 +345,12 @@ public final class CrossChunkReferenceCollector implements ScopedCallback, Compi
         return true;
 
       case CALL:
+        // If the function has been marked as having no side effects then it is safe to move to another module.
         if (valueNode.getSideEffectFlags() == 0) {
           return true;
         }
 
-        // In general it is not safe to move function calls, but we carve out an exception
+        // In general it is not safe to move function calls with side effects, but we carve out an exception
         // for the special stub method calls used for CrossChunkMethodMotion.
         // Case: `JSCompiler_stubMethod(x)`
         Node functionName = checkNotNull(valueNode.getFirstChild());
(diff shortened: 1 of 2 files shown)


nreid260:
Unfortunately there's an issue here in the difference between side-effect**ing** and side-effected**ed**. `PureFunctionIdentifier` only records what the side-effects of calling a function are, not whether the result of the call might be influenced by some other calls side-effects.

Here's an example of where it would matter under your proposed PR:

```java
  @Test
  public void testSideEffectedFunctionCall_isNotMoved() {
    // Test a function call with side effects is not moved.
    JSModule[] modules =
        JSChunkGraphBuilder.forChain()
            // m1
            .addChunk(
                lines(
                    "var a = 0;", //
                    "function f1(a) { return a + 1 }",
                    "var b = f1(1);",
                    "a += 1;"))
            // m2
            .addChunk("var c = b")
            .build();

    test(
        modules,
        new String[] {
          // m1
          lines(
              "var a = 0;", //
              "a += 1;"),
          // m2
          lines(
              "function f1(a) { return a + 1 }", //
              "var b = f1(1);",
              "var c = b"),
        });
  }
```
Notice how `a += 1` now happens before `var b = f1(1)`, which changes the value of `b`.

It's our fault in that the name `PureFunctionIdentifier` is inaccurate. There's a lot of disagreement over the correct terminology in these matters.

I'm not aware of anything in the compiler that tracks the information we'd need to make this safe. One could expand `PureFunctionIdentifier` to try, but there are a lot of ways a function can be side-effected, so you'd have to be pretty careful, if it's even possible. `ExpressionDecomposer` has some code that attempts to do this in limited cases.

WearyMonkey:
Thanks for your detailed reply @nreid260, I hadn't considered that case.

Do you think it would be possible to add 'pure function' tagging? Similar to the `@nosideeffects` annotation, e.g. `@notsideeffected`, or just `@pure` (although I can imagine people may get `@pure` confused with `@nosideeffects`).

```
/** @pure */
function fn(a) {
  return a + 1;
}
```

nreid260:
@shicks Is there an existing feature request for `@pure`/`@nosideeffected`? I know we've talked about this before to interoperate with other optimizers.

A new annotation it's not something we want to jump into head-first. At the very least, there are terminology decisions we'd want to hammer out because "pure" is not something programmers agree on a deffiniton of. Beyond that, there are concerns that the proposed API is error-prone and therefore more dangerous that it is valuable.

We also recognize that the existing system has sharp-edges. It's that recognition that makes us more cautious about changing it without a thoroughly designed replacement.

WearyMonkey:
Thank @nreid260 , agree with your concerns and the need to move carefully.

To give you some more context on our particular use case. Around one quarter (400kb) of our render blocking entry chunk is made up of generated code for a JS protocol buffer runtime. E.g. there is a lot of code of the form:

```
const FooProto = createMessage(() => ({ 
  field1: { type: 'number' },
 ... more fields...
}))
```

We recently switched from generating `class`s to generating the `createMessage` calls which generate the classes at runtime instead to reduce bundle footprint.

As the code is generated, we can guarantee it follows side-effect free rules, and would be a good candidate for marking with a `order-doesnt-matter` tag. 

But as we have a large engineering team, we probably wouldn't want to expose the same tag to all developers.

nreid260:
Within Google incredible care is spent making sure that generated protocol buffer code can be deleted or moved by the compiler. Such code often represents a huge percentage of the input code size and because generated code tends to be very homogeneous, can cause catastrophic reductions in optimization efficacy.

The sample you linked doesn't look like any generator I'm familiar with. If it's custom (or just not one our team has added support for) I suspect you're going to encounter a lot of problems we already solved, but the solutions won't trigger on your code.

However, as quick fix for this specific issue, you can exploit the `valueOf` trick. Basically, because `valueOf` might be called implicitly all over a JS program, we unconditionally assume that all implementations are pure functions. Considering them impure would make a lot of statements unremovable. To use the trick, do something like this:

```js
const createMessage = {
   /** @nocollapse @suppress {checkTypes} */
   valueOf(x) { ... },
};

const FooProto = createMessage.valueOf(() => ({ 
  field1: { type: 'number' },
 ... more fields...
}))
```

WearyMonkey:
Thanks @nreid260, you're right we're using custom generators. I also carefully crafted them so CC would detect the `createMessage` call as side-effect free and remove it if the return value is not used. So dead code elimination is working well. 

But I wasn't aware of the code motion rules until recently and that they aren't being applied, which means if a message is only used in the a child chunk, it's still included in the entry chunk.

I just had a go at using the `valueOf` trick but couldn't get it to work. Looking at the code in [canMoveValue](https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/CrossChunkReferenceCollector.java#L355-L361) method, an exception is only made for the `'JSCompiler_stubMethod'`  method name, generated from the method motion phase.

Here is the code I used to test:

```javascript
/** a.js */

import { apple } from './b.js';

// prevent inlining
export const a1 = apple.map(b => b + 1);
export const a2 = apple.map(b => b + 1);

console.log(a1, a2);


/** b.js */

const Foo = {
    /** @nocollapse @suppress {checkTypes} */
    valueOf(v) {
        return v;
    }
}

export const apple = Foo.valueOf([5,6,7,2,4,2,3]);
export const cherry = Foo.valueOf([1,2,3,4,5,6,6]);

/** c.js */

import { cherry } from './b.js';
import { a1, a2 } from './a.js';

// prevent inlining
const c1 = cherry.map(b => b + 1);
const c2 = cherry.map(b => b + 1);

console.log(a1, a2, c1, c2);
```

And the command: 

```bash
java -jar compiler.jar \
  --compilation_level=ADVANCED \
  --warning_level=VERBOSE \
  --language_in=ECMASCRIPT_2019 \
  --language_out=ECMASCRIPT_2017 \
  --chunk entry:2 \
  --js ./a.js \
  --js ./b.js \
  --chunk child:1:entry \
  --js ./c.js
```

Then both the `apple` and `cherry` values are include in `entry.js`:


```javascript
/** entry.js */

'use strict';const b={valueOf(a){return a}},c=b.valueOf([5,6,7,2,4,2,3]),d=b.valueOf([1,2,3,4,5,6,6]);const e=c.map(a=>a+1),f=c.map(a=>a+1);console.log(e,f);

/** child.js */

'use strict';const h=d.map(a=>a+1),k=d.map(a=>a+1);console.log(e,f,h,k);
```

If I remove `Foo.valueOf` and any other function call so `b.js` is:

```javascript
export const apple = [5,6,7,2,4,2,3];
export const cherry = [1,2,3,4,5,6,6]
```

Then code motion works well and the output is:

```javascript
/** entry.js */

'use strict';const b=[5,6,7,2,4,2,3];const c=b.map(a=>a+1),d=b.map(a=>a+1);console.log(c,d);

/** child.js */

'use strict';const e=[1,2,3,4,5,6,6],f=e.map(a=>a+1),g=e.map(a=>a+1);console.log(c,d,f,g);
```

If I use `JSCompiler_stubMethod` as the function name, like so:

```javascript
/** b.js */

export const apple = JSCompiler_stubMethod([5,6,7,2,4,2,3]);
export const cherry = JSCompiler_stubMethod([1,2,3,4,5,6,6]);

/** externs.js */

function JSCompiler_stubMethod(x) {}
```

Then code motion does work:

```javascript
/** entry.js */
'use strict';const b=JSCompiler_stubMethod([5,6,7,2,4,2,3]);const c=b.map(a=>a+1),d=b.map(a=>a+1);console.log(c,d);

/** child.js */
'use strict';const e=JSCompiler_stubMethod([1,2,3,4,5,6,6]),f=e.map(a=>a+1),g=e.map(a=>a+1);console.log(c,d,f,g);
```

nreid260:
Ok, yes sorry. I forgot `valueOf` is only useful if you're trying to lie about function purity. In this case we've determined the function has no side-effects but cross-chunk code motion still can't do anything.

If code motion is your concern, I think the solution is to generate the code as one proto per file. The idea is to make the generated code more granular so that it can be allocated to the appropriate chunk during graph assembly rather than via cross-chunk motion.

I'm assuming from our own experience that the problem you're facing is that
  - One proto file contains multiple messages
  - Your generator creates on JS file per proto file
  - Your core chunk depends on only some of the messages in a file
  - All the rest of the messages end up in the core chunk
 
This happens because files are the unit of chunk assembly. Any dep on anything from a file means the entire file ends of in that chunk. If you can make more files, and have finer grained requires, this will stop happening.

Please don't generate calls to `JSCompiler_stubMethod`. That's not intended to be an API and we can't guarantee stability or correctness.

WearyMonkey:
> If code motion is your concern, I think the solution is to generate the code as one proto per file

That's what I've ended up doing :)

We're using webpack, so I wrote a loaded that splits the proto modules into lots of per-module exports. The original proto module re-exports the generated modules:

e.g.
```
/** Foo.js */

export { Thing } from './foo_generated_thing.js';
```

Webpack is pretty good at detecting dependencies on re-exports and replacing them with direct dependencies. We feed that into our chunking algorithm and then into the closure compilation and we get the proto code motion across chunks.

So I've found a solution for our use case. But I'd still like to through in my support for some form of side effect safeness tagging in CC, as it would simplify our tooling a lot.

WearyMonkey:
Closing this PR. As per discussion moving side-effect free calls is not safe, as there is no guarantee that the function is not affected by outside side-effects.

PR #3753, opened 2021-01-15
Add support for rewriting dynamic import expressions

ChadKillingsworth:
Completes support for rewriting dynamic imports. See https://docs.google.com/document/d/1thSbGV5i96jltVwn76KYrPkWoNc-yj_KU3HtBGUCgMc/edit#heading=h.c037si364tjn for the full design document.

The new pass rewrites dynamic imports:
 - Optionally allows aliasing the dynamic import as a function call as a form of transpilation. Setting the alias is exposed as a cli flag.
 - Retargets import specifier paths to the correct output chunk.
 - Ensures the return value is a promise resolving to the rewritten module namespace.

### DIFF
--- a/src/com/google/javascript/jscomp/RewriteDynamicImports.java
+++ b/src/com/google/javascript/jscomp/RewriteDynamicImports.java
@@ -31,6 +31,7 @@ import com.google.javascript.rhino.JSDocInfo;
 import com.google.javascript.rhino.JSTypeExpression;
 import com.google.javascript.rhino.Node;
 import com.google.javascript.rhino.Token;
+import com.google.javascript.rhino.jstype.FunctionType;
 import com.google.javascript.rhino.jstype.JSType;
 import com.google.javascript.rhino.jstype.JSTypeNative;
 import com.google.javascript.rhino.jstype.JSTypeRegistry;
@@ -38,19 +39,21 @@ import com.google.javascript.rhino.jstype.TemplatizedType;
 import javax.annotation.Nullable;
 
 /**
- * Rewrite dynamic import expressions to account for bundling and module rewriting.
- * Since dynamic imports cannot be fully polyfilled, optionally support replacing the expression
- * with a function call which indicates an external polyfill is utilized.
+ * Rewrite dynamic import expressions to account for bundling and module rewriting. Since dynamic
+ * imports cannot be fully polyfilled, optionally support replacing the expression with a function
+ * call which indicates an external polyfill is utilized.
  *
- * If the import specifier is a string literal and the module resolver recognizes the target,
- * the pass retargets the specifier to the correct output chunk.
+ * <p>If the import specifier is a string literal and the module resolver recognizes the target, the
+ * pass retargets the specifier to the correct output chunk.
  *
- * Example from:
+ * <p>Example from:
+ *
  * <pre>
  *   import('./input-module1.js');
  * </pre>
  *
  * To:
+ *
  * <pre>
  *   imprt_('./output-chunk0.js').then(function() { return module$output$chunk0; });
  * </pre>
@@ -58,27 +61,27 @@ import javax.annotation.Nullable;
 public class RewriteDynamicImports extends NodeTraversal.AbstractPostOrderCallback
     implements CompilerPass {
 
-  static final DiagnosticType UNABLE_TO_COMPUTE_RELATIVE_PATH = DiagnosticType.error(
-      "JSC_UNABLE_TO_COMPUTE_RELATIVE_PATH",
-      "Unable to compute relative import path from \"{0}\" to \"{1}\"");
+  static final DiagnosticType UNABLE_TO_COMPUTE_RELATIVE_PATH =
+      DiagnosticType.error(
+          "JSC_UNABLE_TO_COMPUTE_RELATIVE_PATH",
+          "Unable to compute relative import path from \"{0}\" to \"{1}\"");
 
-  static final DiagnosticType DYNAMIC_IMPORT_ALIAS_MISSING = DiagnosticType.warning(
-      "JSC_DYNAMIC_IMPORT_ALIAS_MISSING",
-      "A dynamic import alias is specified as a qualified name but the definition " +
-          "is missing: \"{0}\"");
+  static final DiagnosticType DYNAMIC_IMPORT_ALIAS_MISSING =
+      DiagnosticType.warning(
+          "JSC_DYNAMIC_IMPORT_ALIAS_MISSING",
+          "A dynamic import alias is specified as a qualified name but the definition "
+              + "is missing: \"{0}\"");
 
   private final AbstractCompiler compiler;
+  private final AstFactory astFactory;
   private final String alias;
   private boolean dynamicImportsRemoved = false;
   private boolean shouldInjectAliasExtern = false;
 
-  /**
-   *
-   *
-   * @param compiler The compiler
-   */
-  public RewriteDynamicImports(AbstractCompiler compiler, @Nullable  String alias) {
+  /** @param compiler The compiler */
+  public RewriteDynamicImports(AbstractCompiler compiler, @Nullable String alias) {
     this.compiler = compiler;
+    this.astFactory = compiler.createAstFactory();
     this.alias = alias;
   }
 
@@ -88,11 +91,11 @@ public class RewriteDynamicImports extends NodeTraversal.AbstractPostOrderCallba
     shouldInjectAliasExtern = false;
     checkArgument(externs.isRoot(), externs);
     checkArgument(root.isRoot(), root);
-    ScopeCreator scopeCreator = new TypedScopeCreator(compiler);
-    NodeTraversal nodeTraversal = new NodeTraversal(compiler, this, scopeCreator);
-    nodeTraversal.traverse(root.getParent());
+    NodeTraversal.traverse(compiler, root, this);
     if (dynamicImportsRemoved) {
-      compiler.setFeatureSet(compiler.getFeatureSet().without(Feature.DYNAMIC_IMPORT));
+      // This pass removes dynamic import, but adds arrow functions.
+      compiler.setFeatureSet(
+          compiler.getFeatureSet().without(Feature.DYNAMIC_IMPORT).with(Feature.ARROW_FUNCTIONS));
     }
     if (shouldInjectAliasExtern) {
       injectAliasAsExtern(externs);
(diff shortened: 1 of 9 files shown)


rishipal:
The build is failing in this PR because `JSDocInfoBuilder` is replaced with `JSDocInfo.Builder` after 35e25ea82d08502017c76b16022ac9e28a450acf.

ChadKillingsworth:
Thanks - that would have taken a bit to figure out. I've force pushed the fix.

brad4d:
I've just spent several hours working on the imported PR.
I'm having to massage it quite a lot, I'm afraid, and I don't have the tests passing yet.

A fairly large chunk of the work is rewriting the pass to use AstFactory, which is the preferred way to create new AST nodes and handles a good bit of the heavy lifting to make sure nodes have the right types.
Unfortunately, I'm also having to add new methods to that class, because there are some use cases here that we haven't previously needed to support.

brad4d:
I've pushed back several changes to the PR branch.
See the commit message for more details.
Currently tests are failing, but I think @ChadKillingsworth needs to evaluate the failures.

brad4d:
I've just imported the latest version, performed necessary massaging, and sent it for internal review and testing.

src/com/google/javascript/jscomp/RewriteDynamicImports.java
```diff
@@ -0,0 +1,312 @@
...
+    checkNotNull(this.alias);
+    final Node aliasNode = astFactory.createQNameWithUnknownType(this.alias);
+    aliasNode.setOriginalName("import");
+    final Node moduleSpecifier = dynamicImport.getFirstChild().detach();
+    Node importAliasCall =
+        astFactory
+            .createCall(aliasNode, moduleSpecifier)
+            .srcrefTreeIfMissing(dynamicImport);
+    if (dynamicImport.getJSType() != null) {
+      importAliasCall.setJSType(dynamicImport.getJSType());
```
brad4d, line 298: @blickly tells me that the preferred way to do this is now `importAliasCall.copyTypeFrom(dynamicImport)`, because this works correctly with the new, simplified "colors" type system we're introducing for use in optimizations.

There are several places where this change needs to be made.

src/com/google/javascript/jscomp/RewriteDynamicImports.java
```diff
@@ -0,0 +1,312 @@
...
+   * <p>Before
+   *
+   * <pre>
+   *   import('./foo.js');
+   * </pre>
+   *
+   * After
+   *
+   * <pre>
+   *   import('./foo.js').then(function() { return module$foo; });
```
brad4d, line 244: This code actually adds an arrow function.
Could you update this comment to match?

src/com/google/javascript/jscomp/RewriteDynamicImports.java
```diff
@@ -0,0 +1,312 @@
...
+    final JSTypeRegistry registry = compiler.getTypeRegistry();
+    final Node promiseDotResolve =
+        astFactory.createQName(scope.getGlobalScope(), "Promise.resolve");
+    final Node promiseResolveCall = astFactory.createCall(promiseDotResolve);
+    final boolean isExpressionUsed = NodeUtil.isExpressionResultUsed(dynamicImport);
+    if (isExpressionUsed) {
+      JSType moduleNamespaceType;
+      if (dynamicImport.getJSType() != null && dynamicImport.getJSType().isTemplatizedType()) {
+        TemplatizedType templatizedType = (TemplatizedType) dynamicImport.getJSType();
+        moduleNamespaceType = templatizedType.getTemplateTypes().asList().get(0);
```
brad4d, line 159: Could you drop the ".asList()" here? It shouldn't be necessary.

src/com/google/javascript/jscomp/RewriteDynamicImports.java
```diff
@@ -0,0 +1,312 @@
...
+
+  static final DiagnosticType UNABLE_TO_COMPUTE_RELATIVE_PATH =
+      DiagnosticType.error(
+          "JSC_UNABLE_TO_COMPUTE_RELATIVE_PATH",
+          "Unable to compute relative import path from \"{0}\" to \"{1}\"");
+
+  private final AbstractCompiler compiler;
+  private final AstFactory astFactory;
+  private final String alias;
+  private boolean dynamicImportsRemoved = false;
```
brad4d, line 67: It looks like this field isn't really needed anymore.

Our internal automated checks report several warnings at its usages, including saying that the place where its value is used it could be a local variable.

src/com/google/javascript/jscomp/deps/ModuleLoader.java
```diff
@@ -298,6 +299,28 @@ static String normalize(String path, Iterable<String> moduleRootPaths) {
...
+  public static String relativePathFrom(String fromUriPath, String toUriPath) {
+    Path fromPath = Paths.get(fromUriPath);
+    Path toPath = Paths.get(toUriPath);
+    Path fromFolder = fromPath.getParent();
+
+    // if the from URIs are simple names without paths, they are in the same folder
+    // example: m0.js
+    if (fromFolder == null && toPath.getParent() == null) {
+      return "./" + toUriPath;
+    } else if (fromFolder == null && toUriPath.startsWith(".") || toPath.startsWith("/")) {
```
brad4d, line 311: our internal checks want grouping parentheses to make operator precedence explicit here.

test/com/google/javascript/jscomp/RewriteDynamicImportsTest.java
```diff
@@ -0,0 +1,303 @@
...
+    super.setUp();
+    enableCreateModuleMap();
+    enableTypeInfoValidation();
+    disableScriptFeatureValidation();
+  }
+
+  @Override
+  protected CompilerOptions getOptions() {
+    CompilerOptions options = super.getOptions();
+    options.setLanguage(LanguageMode.ECMASCRIPT_2020);
```
brad4d, line 48: Since we're rewriting the dynamic imports, should at least some of these test cases use a lower language out than this?

brad4d (changes requested):
Alas! we haven't reached the point where I can just make a few tweaks and submit.

We're definitely getting closer, though.

src/com/google/javascript/jscomp/RewriteDynamicImports.java
blickly: I think this one may be a bug in our internal automated checks. I'll file a bug internally

test/com/google/javascript/jscomp/RewriteDynamicImportsTest.java
blickly: In general, I think that unit tests should be able to remove the language mode completely, since they (recently changed) now default to the most recent language mode with no transpilation.

src/com/google/javascript/jscomp/RewriteDynamicImports.java
ChadKillingsworth: My editor is telling me `.copyTypeFrom` is marked as `@deprecated`

src/com/google/javascript/jscomp/RewriteDynamicImports.java
ChadKillingsworth: It's definitely used.

test/com/google/javascript/jscomp/RewriteDynamicImportsTest.java
ChadKillingsworth: I can test that an error is thrown with a lower language level without an alias. But that's the only impact a lower language level has.

src/com/google/javascript/jscomp/RewriteDynamicImports.java
ChadKillingsworth: Done

src/com/google/javascript/jscomp/RewriteDynamicImports.java
ChadKillingsworth: Done

src/com/google/javascript/jscomp/deps/ModuleLoader.java
ChadKillingsworth: Rightly so - added.

src/com/google/javascript/jscomp/RewriteDynamicImports.java
ChadKillingsworth: Changed.

blickly: Thanks for catching that.  I think it was unintentionally deprecated in https://github.com/google/closure-compiler/tree/f272e4d991f51976f77949d92372178280c11bf4.  I'll send out a change to undeprecate.

test/com/google/javascript/jscomp/RewriteDynamicImportsTest.java
blickly: I think that Bradford was just trying to forward my comment here, which is that I've been trying to remove the language mode settings from the unit tests completely.  Do the unit tests in this file still work with this line removed?

ChadKillingsworth: They do - but I added a warning (and test) for when the output language level doesn't support dynamic imports and an alias wasn't specified.

brad4d:
I imported the latest version and tried to submit it.
Unfortunately, I'm getting failures because apparent j2cl's version of Path doesn't have a `startsWith()` method, which is being used in `ModuleLoader` now.

I'm leanding toward cheating and converting the `Path` to a string and using the `startsWith()` from there.

brad4d:
I'm checking to see if all the tests pass when I replace `path.startsWith("/")` with `path.toString().startsWith("/")`.

BTW, I checked and `path.isAbsolute()` isn't available either.

brad4d:
I got one test failure that looks to be a flake.

@ChadKillingsworth 

I realize doing `path.toString().startsWith("/")`  is a crappy fix and may well mean the check is broken on Windows,
but I figure it's better to get this all submitted now. You could do a second PR to address this problem, which will be much smaller and easier to land.

ChadKillingsworth:
Yes that would be an easy follow up. I'll roll it into the next PR where I update the `ConvertChunksToESModules` pass.

brad4d:
@ChadKillingsworth 

It may only be necessary to add a method to this interface and then put the call to `Path::startsWith()` back as it was in `ModuleLoader`.
I had forgotten that we were providing the `java.nio.file.Path` implementation for j2cl builds.

PR #3757, opened 2021-01-16
Prevent arrow functions from being improperly inlined

### LINKED ISSUE #3756: InlineSimpleMethods makes a semantic change in ES2015+
Given the following input code:
```javascript
class Holder {
    constructor() {
        this.val = {internal: true};
        this.context =  {getVal: () => this.val}
    }
}

console.log(new Holder().context.getVal().internal);
```

Outputs:
```javascript
class a {
  constructor() {
    this.g = {h:!0};
    this.i = {j:() => this.g};
  }
}
console.log((new a).i.g.h);
```

The `InlineSimpleMethods` replaces the call of `getVal()` with `val`, which is incorrect because the `this` in the definition of `getVal` is not `context`, but rather `Holder`.
This is not a problem in ES5, because the arrow function is already rewritten before running this pass.

DavidANeil:
Closes #3756.

### DIFF
--- a/src/com/google/javascript/jscomp/InlineSimpleMethods.java
+++ b/src/com/google/javascript/jscomp/InlineSimpleMethods.java
@@ -99,7 +99,7 @@ class InlineSimpleMethods extends MethodCompilerPass {
           // Verify this is a trivial return
           Node returned = returnedExpression(firstDefinition);
           if (returned != null) {
-            if (isPropertyTree(returned)) {
+            if (isPropertyTree(returned) && !firstDefinition.isArrowFunction()) {
               if (logger.isLoggable(Level.FINE)) {
                 logger.fine("Inlining property accessor: " + callName);
               }
--- a/test/com/google/javascript/jscomp/InlineSimpleMethodsTest.java
+++ b/test/com/google/javascript/jscomp/InlineSimpleMethodsTest.java
@@ -387,6 +387,23 @@ public final class InlineSimpleMethodsTest extends CompilerTestCase {
   }
 
   @Test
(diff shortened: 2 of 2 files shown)


blickly:
And thanks for the fix as well!

This looks good to me, so I'll work on getting it imported right away

blickly:
This PR is now merged. Thanks!

One suggested improvement I heard during review was: only backing off for arrow functions that reference `this`/`super`/`arguments` rather than all arrow functions.  This would still avoid the bug of #3756, while at the same time being able to inline arrow functions that don't reference those references from outer scopes (the more common case).

DavidANeil:
My understanding was that `isPropertyTree` is a recursive function that only returns true if the lhs of the returned property access is `this`, so I feel like what you've suggested is already the case.

blickly:
I think this might be best demonstrated with another unit test, similar to the one you created already, but checking that arrow functions can be inlined, something like:

```
    test(
        lines(
            "class Holder {",
            "    constructor() {",
            "        this.val = {internal: true};",
            "        this.context = {getVal: () => 5}",
            "    }",
            "}",
            "",
            "console.log(new Holder().context.getVal().internal);"),
        lines(
            "class Holder {",
            "    constructor() {",
            "        this.val = {internal: true};",
            "        this.context = {getVal: () => 5}",
            "    }",
            "}",
            "",
            "console.log(new Holder().context.val.internal);"));
```

If that unit test passes, that would prove that arrow functions can still be inlined by the pass.

blickly:
I filed https://github.com/google/closure-compiler/issues/3764 to track this work.

PR #3778, opened 2021-03-01
Rollforward of #3724: Refactor ExtractPrototypeMemberDeclarations

ChadKillingsworth:
Re-adds the changes from #3724 but only when chunk output mode is ES Modules. Prevents the code size regressions that triggered the previous rollback.


brad4d:
I've imported this, applied necessary internal changes, and sent it for testing and review.

Issue #3781, opened 2021-03-09
SourceMapConsumerV3 feature request: allow disabling of approximated mappings


lauraharker:
Filing off of https://groups.google.com/g/closure-compiler-discuss/c/rhAsFdwsnmQ so that this shows up in our bug scrub meeting. (and so that someone more familiar with SourceMapConsumerV3 than me will look at it)

Here's Kyle's original request:

> I am a consumer of SourceMapConsumerV3 :)
>
> I am using the getMappingForLine API to retrieve a source line from a transpiled line and column.
>
> There are some cases where calling this API will not find a direct mapping and approximate it, either by retrieving previous transpiled line mappings or by grabbing a mapping with a different column than was requested.
>
> It would be nice if one could opt out of this behavior, so that the caller could have the visibility that a mapping was not found vs it being approximated.
>
> Are there any objections to this feature request? If not I would be willing to contribute the change.

and a followup message:

> rough example here: https://github.com/kmantzel/closure-compiler/pull/1/files
>
> Also I know I originally mentioned that the approximation can "grab a mapping with a different column than was requested", but I think that was an error on my part for not understanding source mapping. As a variable name can span multiple columns. So I just ended up making the approximation around the existing `getPreviousLine` method for now.
>
> Just out of curiosity, is changing the behavior without this feature flag an option?

lauraharker:
Bradford offered to look more into this & if it would be reasonable to always disable "approximate mappings"

brad4d:
I believe the "approximate mappings" mentioned here represent the two places in the method where `getPreviousMapping()` gets called.

https://github.com/google/closure-compiler/blob/master/src/com/google/debugging/sourcemap/SourceMapConsumerV3.java#L163

As an experiment I modified these 2 places to return `null` instead and then ran our usual tests.
I got one failure in `CompilerTest` and several in an internal-only test.

My interpretation of this result is that we do actually depend on this approximation behavior.
However, I think it would be reasonable to expand the `OriginalMapping` data structure defined in `mapping.proto` to include an enum with values `UNKNOWN`, `EXACT`, and `APPROXIMATE`. (default of `UNKNOWN`), then modify the API to populate this field.

Kyle would you be interested in doing that?

kmantzel:
Hi Bradford, thanks for looking into this. I'd definitely be willing to make any changes.

Your understanding is correct that the behavior I'd like to change is specifically around the `getPreviousMapping` method. 

[...]

PR #3785, opened 2021-03-22
Add retrieval type enum to OriginalMappings

kmantzelSNow:
This will allow callers to determine whether or not a mapping was approximated for the `SourceMapConsumerV3#getMappingForLine` API as discussed here https://github.com/google/closure-compiler/issues/3781

### DIFF
--- a/src/com/google/debugging/sourcemap/proto/mapping.proto
+++ b/src/com/google/debugging/sourcemap/proto/mapping.proto
@@ -22,4 +22,13 @@ message OriginalMapping {
 
   // The original name of the identifier.
   optional string identifier = 4;
+  
+  enum RetrievalType {
+    UNKNOWN = 0;
+    EXACT = 1;
+    APPROXIMATED_LINE = 2;
+  }
+  
+  // The type of retrieval performed to get this mapping 
+  optional RetrievalType retrieval_type = 5 [default = UNKNOWN];
 }
--- a/src/com/google/debugging/sourcemap/proto/mapping.proto
+++ b/src/com/google/debugging/sourcemap/proto/mapping.proto
@@ -23,12 +23,12 @@ message OriginalMapping {
   // The original name of the identifier.
   optional string identifier = 4;
   
-  enum RetrievalType {
-    UNKNOWN = 0;
+  enum Precision {
+    UNKNOWN_PRECISION = 0;
     EXACT = 1;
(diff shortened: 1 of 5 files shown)


src/com/google/debugging/sourcemap/proto/mapping.proto
```diff
@@ -22,4 +22,13 @@ message OriginalMapping {
 
   // The original name of the identifier.
   optional string identifier = 4;
+  
+  enum RetrievalType {
```
nreid260, line 26:
```
  enum Precision {
    UNKNOWN_PRECISION = 0,
    EXACT = 1,
    APPROXIMATE_LINE = 2,
  }
```

Proto bindings in C++ and Python dump all the enum elements in the surrounding namespace, so it's a good idea to qualify UNKNOWN in some way.

nreid260:
Hey @kmantzelSNow, we need you to sign a CLA release before we can merge this.

kmantzelSNow:
Hi @nreid260, thanks for the response.

I'm still waiting on approval from legal at my company before I sign the CLA. I'll update the PR via googlebot reply once I get the go-ahead

src/com/google/debugging/sourcemap/proto/mapping.proto
kmantzelSNow: I'm somewhat inexperienced with protobuffs, but I believe I understand your point. 

Does it make sense to qualify `EXACT` and `APPROXIMATE_LINE` with `_PRECISION` as well, or would the example you provided suffice?

nreid260: Those should be fine, but it's a judgement call. `UNKNOWN` is a particularly bad example because almost every enum has an `UNKNOWN = 0` entry, so conflicts are very likely for that name.

kmantzelSNow: Ah I see, that makes sense.

kmantzelSNow:
@googlebot I signed it!

kmantzelSNow:
@googlebot I signed it!

kmantzelSNow:
@googlebot I signed it!

brad4d:
I've imported this for internal code review and testing.

brad4d:
I expect this will be submitted in the next couple of days.

PR #3790, opened 2021-03-26
Add ability to use renaming map produced by a previous compilation

yennien:
--variable_map_input_file, --variable_map_output_file, --property_map_input_file and --property_map_output_file flags were removed from the master branch on April 8, 2014.
Later, --variable_map_output_file and --property_map_output_file  were added back as --variable_renaming_report and --property_renaming_report flags on  June 4, 2014 
Here is to add back --variable_map_input_file and --property_map_input_file flags as --variable_renaming_input and --property_renaming_input flags.

### DIFF
--- a/src/com/google/javascript/rhino/Node.java
+++ b/src/com/google/javascript/rhino/Node.java
@@ -2194,10 +2194,9 @@ public class Node implements Serializable {
         return null;
       }
 
-      Node stringNode = this;
-      String right = stringNode.getOriginalName();
+      String right = this.getOriginalName();
       if (right == null) {
-        right = stringNode.getString();
+        right = this.getString();
       }
 
       return left + "." + right;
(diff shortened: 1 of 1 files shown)


yennien:
@googlebot I signed it!

concavelenz:
Can you explain your use case for these?  Without a strong use case I wouldn't support adding them back. Generally, folks misuse these in attempt to facilitate cross-compilation unit communication  rather than their intended use case of reducing patch size between build.  The former is not reliable because the compiler must obey other restrictions for renaming first (externs, blacklists, etc)

yennien:
Hi John,

Office Online used to have an old patch based on 2016.07.13 that added
those flags. Now, we are trying to use the latest package directly, not the
local patch. The use case for it is we want to separate our js files into
more smaller js files. Some of them are core sections that don't get
changed a lot. We want to re-use the same symbol in order to not change
those js and can be cached.

Thanks,
Sam

concavelenz:
Yeah, that is generally the use case we are trying not to support.   I wrote up the explanation in more detail here:

https://docs.google.com/document/d/1ie-sbEhaMF0oVeOcdpvl9cBh6TRCeRycwWSVaBXSTeY/edit?resourcekey=0-AWmt5JhUEyQirZSrYWD9xA#

Please let me know if you want to continue to discuss this after you have read it.

yennien:
Thanks, it makes sense. I will abandon it

PR #3791, opened 2021-03-31
Add chunk dependencies as explicit import statements

ChadKillingsworth:
Dependent chunks may create side effects which are required. Ensure every dependent chunk is explicitly imported.

### DIFF
--- a/src/com/google/javascript/jscomp/ConvertChunksToESModules.java
+++ b/src/com/google/javascript/jscomp/ConvertChunksToESModules.java
@@ -27,6 +27,7 @@ import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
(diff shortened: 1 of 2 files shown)


brad4d:
I have imported this and sent it for internal testing and review.

src/com/google/javascript/jscomp/ConvertChunksToESModules.java
```diff
@@ -225,7 +233,22 @@ private void addImportStatements() {
...
     @Override
     public void visit(NodeTraversal t, Node n, Node parent) {
-      if (n.isName()) {
+      if (n.isScript()) {
+        JSModule chunk = t.getModule();
+        List<JSModule> chunkDependencies = chunk.getDependencies();
+
+        // Ensure every chunk dependency is explicitly listed with an import
+        // Dependent chunks may have side effects even if there isn't an explicit name reference
+        if (chunkDependencies.size() > 0) {
```
brad4d, line 242: The "Error Prone" tool complains that this would be better as `!chunkDependencies.isEmpty()`.
Any reason not to change to that?

src/com/google/javascript/jscomp/ConvertChunksToESModules.java
```diff
@@ -209,8 +210,15 @@ private void addImportStatements() {
                   getChunkName(exportingChunk)));
```
brad4d, line 210: Error Prone is complaining about line 203's reference to static method `relativePath` via `this`.
Is there some reason not to drop the `this.`?

brad4d (commented):
I'll make the 2 changes I've noted here myself, as long as you're cool with them.

src/com/google/javascript/jscomp/ConvertChunksToESModules.java
ChadKillingsworth: That would be equivalent and better

src/com/google/javascript/jscomp/ConvertChunksToESModules.java
ChadKillingsworth: Dropping the `this.` reference should be just fine.

PR #3796, opened 2021-04-10
Infer the correct type for dynamic imports after module rewriting 

ChadKillingsworth:
When rewriting common js modules, all module rewriting occurs before type inference. Adjust the type inference pass to correctly find the type of these rewritten module namespaces.

Also switch to using the module loader to resolve relative paths between modules. Some modules were not found in the previous methods and no warning was issued either.

### DIFF
--- a/src/com/google/javascript/jscomp/TypeInference.java
+++ b/src/com/google/javascript/jscomp/TypeInference.java
@@ -97,7 +97,6 @@ class TypeInference extends DataFlowAnalysis.BranchedForwardDataFlowAnalysis<Nod
   private final TypedScopeCreator scopeCreator;
   private final AssertionFunctionLookup assertionFunctionLookup;
   private final ModuleImportResolver moduleImportResolver;
-  private final boolean moduleTranspilationBeforeTypeInference;
   // A record is pushed onto this stack during the traversal of each optional chain.
   // Inference of the type at the end of the optional chain requires scope information
   // from traversing the start of the chain. This stack serves as a communication channel
@@ -136,7 +135,6 @@ class TypeInference extends DataFlowAnalysis.BranchedForwardDataFlowAnalysis<Nod
     this.bottomScope =
         LinkedFlowScope.createEntryLattice(
             compiler, TypedScope.createLatticeBottom(syntacticScope.getRootNode()));
-    this.moduleTranspilationBeforeTypeInference = compiler.getOptions().processCommonJSModules;
   }
 
   @CheckReturnValue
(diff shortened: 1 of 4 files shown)


src/com/google/javascript/jscomp/TypeInference.java
```diff
@@ -134,6 +136,7 @@
     this.bottomScope =
         LinkedFlowScope.createEntryLattice(
             compiler, TypedScope.createLatticeBottom(syntacticScope.getRootNode()));
+    this.moduleTranspilationBeforeTypeInference = compiler.getOptions().processCommonJSModules;
```
brad4d, line 139: We try to avoid having classes other than Compiler base their behavior directly on Compiler options.
Instead, we prefer to have a boolean constructor parameter.
In this case, you'll need to also add one to `TypeInferencePass`, since that's where this constructor gets called.

test/com/google/javascript/jscomp/TypeInferenceTest.java
```diff
@@ -3798,6 +3806,25 @@ public void testDynamicImport2() {
...
+                "/** @const */ module$testcode0.default = 1;",
+                "/** @return {string} */ function Bar() { return 'bar'; };",
+                "/** @const */ module$testcode0.Bar = Bar;"),
+            // modules are named of the format `testcode#` based off their index.
+            // testcode0 refers the first module.
+            "const foo = import('./testcode0');"));
+
+    assertType(getType("foo"))
+        .toStringIsEqualTo(
+            "Promise<{\n" + "  Bar: function(): string,\n" + "  default: number\n" + "}>");
```
brad4d, line 3825: could you use `lines()` and a trailing `//` comment on the first argument to make these more readable?

brad4d (changes requested):

test/com/google/javascript/jscomp/TypeInferenceTest.java
ChadKillingsworth: fixed

src/com/google/javascript/jscomp/TypeInference.java
ChadKillingsworth: I wondered about that as I wrote it.

I've made the changes, but it affected more code than it probably should have. The only purpose of this is to prevent having to traverse the AST if a MODULE_BODY isn't found. I'm wondering now if this is even needed. What do you think?

ChadKillingsworth: After reviewing it again, this just seemed like a premature optimization. I removed the explicit test and left this as the fallback case.

brad4d:
I've imported the latest version of this and sent it for internal testing and review.

blickly:
I guess it's a little late for it now, but how feasible would it be to move commonjs rewriting after typechecking.  This matches the direction that we are moving for `goog.module`/ESM module rewriting, and is likely to be more future proof.

That said, I understand that such a request might be a bigger project than what's going on in this PR.

ChadKillingsworth:
Moving common js rewriting after type checking is definitely something I want to do, but until very recently I didn't understand enough about how this was accomplished. Moving the pass also means splitting it in two: Pre-type checking common js modules need normalized (things like removing universal module definition wrappers). Post type checking the actual rewriting should occur.

I've actually started this work several times, but it's quite challenging. Needless to say its more than we would want in this PR and something I may actually need some assistance on.

blickly:
OK, that makes sense to me. Thanks for the explanation, Chad!

PR #3800, opened 2021-04-20
Provide externs for FinalizationRegistry

cpcallen:
Provide externs for `FinalizationRegistry` and its`register` and `unregister` methods.

I believe the templating is correct (it has caught type errors in my code when I updated from the previous `FinalizationGroup` API) but of course a second pair of eyes is always a good idea…

### DIFF
--- a/externs/es6.js
+++ b/externs/es6.js
@@ -2036,6 +2036,31 @@ function WeakRef(value) {}
 WeakRef.prototype.deref = function() {};
 
 /**
+ * @constructor
+ * @struct
+ * @param {function(HOLDINGS)} cleanupCallback
+ * @template TARGET, HOLDINGS, TOKEN
+ * @nosideeffects
+ * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry
+ */
+function FinalizationRegistry(cleanupCallback) {}
+
+/**
+ * @param {TARGET} target
+ * @param {HOLDINGS} holdings
+ * @param {TOKEN=} unregisterToken
+ * @return {void}
+ */
+FinalizationRegistry.prototype.register =
+    function(target, holdings, unregisterToken) {};
+
+/**
+ * @param {TOKEN} unregisterToken
+ * @return {void}
+ */
+FinalizationRegistry.prototype.unregister = function(unregisterToken) {};
+
+/**
  * @type {!Global}
  */
 var globalThis;
--- a/externs/es6.js
+++ b/externs/es6.js
@@ -2038,8 +2038,8 @@ WeakRef.prototype.deref = function() {};
 /**
  * @constructor
  * @struct
- * @param {function(HOLDINGS)} cleanupCallback
- * @template TARGET, HOLDINGS, TOKEN
+ * @param {function(HELDVALUE)} cleanupCallback
+ * @template TARGET, HELDVALUE, TOKEN
  * @nosideeffects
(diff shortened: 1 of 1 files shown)


externs/es6.js
```diff
@@ -2035,6 +2035,31 @@ function WeakRef(value) {}
  */
 WeakRef.prototype.deref = function() {};
 
+/**
+ * @constructor
+ * @struct
+ * @param {function(HOLDINGS)} cleanupCallback
+ * @template TARGET, HOLDINGS, TOKEN
```
rishipal, line 2042: From the spec it seems like the `cleanupCallback` accepts a single `heldValue` at any time. So maybe change `HOLDINGS` to `HELDVALUE` to avoid confusion.

https://tc39.es/ecma262/#sec-finalization-registry-constructor

externs/es6.js
```diff
@@ -2035,6 +2035,31 @@ function WeakRef(value) {}
...
+function FinalizationRegistry(cleanupCallback) {}
+
+/**
+ * @param {TARGET} target
+ * @param {HOLDINGS} holdings
+ * @param {TOKEN=} unregisterToken
+ * @return {void}
+ */
+FinalizationRegistry.prototype.register =
+    function(target, holdings, unregisterToken) {};
```
rishipal, line 2055: nit: rename parameter to `heldValue` to stay consistent with the spec.

externs/es6.js
```diff
@@ -2035,6 +2035,31 @@ function WeakRef(value) {}
  */
 WeakRef.prototype.deref = function() {};
 
+/**
+ * @constructor
+ * @struct
+ * @param {function(HOLDINGS)} cleanupCallback
+ * @template TARGET, HOLDINGS, TOKEN
+ * @nosideeffects
```
rishipal, line 2043: I could not find any information on whether these could be side-effectful. Just out of curiosity, did you?

externs/es6.js
cpcallen: Fixed.

externs/es6.js
cpcallen: My externs _were_ consistent with the spec before [they decided to change the spec](https://github.com/tc39/proposal-weakrefs/issues/113)!  🙄  Fixed.

externs/es6.js
cpcallen: As I understand it (please correct me if I am wrong), the `@nosideffects` declaration here just means that the call the the constructor can be omitted if the returned value is not used.  That appears to be the case: [the `FinalizationRegistry` constructor](https://tc39.es/ecma262/#sec-finalization-registry-cleanup-callback)  only really does three things:

* It calls [OrdinaryCreateFromConstructor](https://tc39.es/ecma262/#sec-ordinarycreatefromconstructor), which is called by basically every constructor in the spec and so I presume is regarded as side-effect free.
* It calls [HostMakeJobCallback](https://tc39.es/ecma262/#sec-hostmakejobcallback)(_cleanupCallback_), and
* It initialises a few slots on the new instance, including one with the result of the HostMakeJobCallback call.

HostMakeJobCallback appears to be somewhat loosely specified.  The default implementation is clearly side-effect free, but it's permissible that a web browser could implement it differently. There does not appear to be any guarantee that such an alternative implementation is strictly side-effect free, but this possibility seems sufficiently remote that it would seem unreasonable for compiler users to depend on a side effect of calling `void new FinalizationRegistry()`.

cpcallen:
I believe this is now ready to go.

rishipal:
Thanks for addressing the review comments. Started internal merge process.

PR #3801, opened 2021-04-21
Remove long deprecated CLI flags

ChadKillingsworth:
- Remove long deprecated flags
- Allow dynamic import expressions by default
- Document the --json_streams flag
- Enable optimal flags automatically when --chunk_output_type=ES_MODULES is specified

### DIFF
--- a/src/com/google/javascript/jscomp/CommandLineRunner.java
+++ b/src/com/google/javascript/jscomp/CommandLineRunner.java
@@ -255,8 +255,7 @@ public class CommandLineRunner extends AbstractCommandLineRunner<Compiler, Compi
                 + "<num-js-files> may be set to 'auto' for the first chunk if it "
                 + "has no dependencies. "
                 + "Provide the value 'auto' to trigger chunk creation from CommonJS"
-                + "modules.",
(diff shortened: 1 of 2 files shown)


rishipal:
Hi Chad, I'm not aware of a good way to gauge usage of these flags externally, but it seems that this could break external users? Should we have a forward-looking announcement / seek approval?

ChadKillingsworth:
Once it's merged, I can make a post in the group. But most of these flags were renamed years ago. I don't see this being a problem.

concavelenz:
This has been merged.

jespertheend:
What is the correct way to provide source maps using the `json_streams` flags? The docs say it's `srcmap` but the returned json contains `source_map`.

ChadKillingsworth:
I think unfortunately it's asymmetrical. If you file an issue for that, I will try to fix it. Here's code that uses this as an example: https://github.com/google/closure-compiler-npm/blob/master/packages/google-closure-compiler/lib/gulp/concat-to-json.js

jespertheend:
Ah yes `sourceMap` seems to work properly. I'll file an isue 👍

PR #3814, opened 2021-05-15
Convert the first argument to optional to match the underlying API.

realityforge:
See https://tc39.es/ecma262/#sec-sharedarraybuffer.prototype.slice and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer/slice


realityforge:
Anything that needs to done to get this looked at?

PR #3817, opened 2021-05-15
Add missing extern for Symbol.split

realityforge:


realityforge:
Any chance that this can be looked at?

brad4d:
I've just reviewed and approved this internally.

rishipal:
Adding this extern causes code size regression in internal tests. 

It's possible that disambiguation is conflating the Symbol's split property with another split property. Adding a misspelt `Symbol.splitt` extern instead does not cause the size regression.

I will need to investigate this more. Is this PR blocking?

realityforge:
Not blocking for us as we have added a local extern but it would great if it eventually landed ;)

realityforge:
Closing this to get it off my plate as it seems unlikely to be merged

PR #3819, opened 2021-05-15
Month is an optional argument to the Date.UTC function

realityforge:
For confirmation see
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/UTC
- https://tc39.es/ecma262/#sec-date.utc


realityforge:
It would be great if this could be looked at ;)

PR #3826, opened 2021-05-29
Add GamepadEvent to w3c_gamepad.js

treblereel:
It would be nice to have GamepadEvent in gamepas externs

ps: there are some additional useful Gamepad APIs ( like GamepadHapticActuator ) that are experimental and because of that, they can't be added, right ?


brad4d:

PR #3828, opened 2021-06-05
Avoid creating dirs when the output is JSON

ChadKillingsworth:
When the output is a JSON array of source files to stdout, avoid creating unnecessary filesystem directories.


brad4d:
@ChadKillingsworth has a PR to fix this, which I intend to get merged this week.

brad4d:
I've sent this for internal testing and review.

Also, TIL that I could fix a merge conflict with a manual edit directly in GitHub.
Sweet!

PR #3833, opened 2021-06-16
added extern for resolveLocalFileSystemURL()

phatpaul:
resolveLocalFileSystemURL() is missing from built-in extern nonstandard_fileapi.js.  (resolveLocalFileSystemURI is there, but it's not what cordova-plugin-file uses)


phatpaul:
@googlebot I signed it!

rishipal (approved):

PR #3845, opened 2021-07-28
Fold declaration identity. `var x = x;` -> `var x;`

caojoshua:
`var x = x;` -> `var x;` is always correct.

* var allows re-declaration of variables
* If `x` is undefined, then `var x = x` is same as `var x = undefined`, or just `var x;`
* If `x` is already defined, then `var x = x` does not do anything

Note that the same does not apply to `const` or `let`, which does not allow re-declaration of variables. Therefore, I first try to optimize `var` name declarations, then pass the resulting subtree to the generic name declaration optimization.

### DIFF
--- a/src/com/google/javascript/jscomp/PeepholeRemoveDeadCode.java
+++ b/src/com/google/javascript/jscomp/PeepholeRemoveDeadCode.java
@@ -83,6 +83,7 @@ class PeepholeRemoveDeadCode extends AbstractPeepholeOptimization {
       case OBJECT_PATTERN:
         return tryOptimizeObjectPattern(subtree);
       case VAR:
+        return tryOptimizeVarNameDeclaration(subtree);
       case CONST:
       case LET:
         return tryOptimizeNameDeclaration(subtree);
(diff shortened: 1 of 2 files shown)


caojoshua:
@googlebot I signed it!

mprobst:
@caojoshua can you help us share the use case here a bit? The optimization seems legit, but I have a hard time coming up with reasonable code where this pattern would occur, even after inlining or other previous code transformations.

What kind of code would this help with? Where did you encounter that?

caojoshua:
I did not encounter it, just writing some code for fun.

My thinking was that if we optimize `x=x`, we could also optimize `var x=x`. I did not consider that other optimizations could result in `x=x`, but probably not `var x=x`.

I think that the role of a compiler or code optimizer is to make code as optimal as possible, no matter how the source code is written. So while I agree that `var x=x` is an unusual use case, we should still optimize it since its within our power.

mprobst:
@caojoshua there are two issues with this. \

* One is that `var x = x;` is not equivalent to `var x;`. `var x` doesn't shadow its own initializer in JS, so if you e.g. run `var name = name;` in a browser, it'll pick up `window.name` for the initializer value.
* The other is that compiler optimizations aren't free: they cost time to develop, maintain, test, and also to execute. So each optimization should actually benefit code user code, scaled to how expensive it'll be to run and maintain. As I said above, I cannot see what code this optimization might apply to in the wild.

With that, I'll close this issue. Please speak up if that doesn't make sense.

Issue #3559, opened 2020-02-25
need flag to enable `CompilerOptions.setAssumeStaticInheritanceRequired(true)`


brad4d:
To avoid compilation breaking uses of `this` and `super` in static methods of ES6 classes, you have to get `CommandLineRunner` to do  `CompilerOptions.setAssumeStaticInheritanceRequired(true)` when creating the `Compiler` object.

Unfortunately, there is no flag for this yet, and we can't make it the default behavior because of increase in output code size it causes for Google projects.

This would be a good community-volunteer thing to do.

cshung:
@brad4d
I am interested in this one.

brad4d:
@cshung thanks for your interest. We'll look out for a PR from you.

PR #3864, opened 2021-09-08
Adding command line flag for setAssumeStaticInheritanceRequired

rorye:
Fix for #3559


brad4d:
I've imported this PR and sent it for internal review and testing.

brad4d:
FYI, I've had to reverse the name of the flag and the logic around it to be consistent with our flag conventions.
`--assume_something=true` is expected to always enable more optimizations, not disable optimizations.

I expect to land the modified PR in the next day or two.

Issue #3834, opened 2021-06-16
Cannot compile import.meta


AshleyScirra:
[This comment](https://github.com/google/closure-compiler/issues/3707#issuecomment-834744655) appears to indicate pass-through for `import.meta` is supported, but it still doesn't seem to work in v20210601.

Input test.js: `console.log(import.meta.url);`

Command line: `java -jar ./closure-compiler.jar --compilation_level SIMPLE --language_in ECMASCRIPT_2020 --language_out ECMASCRIPT_2020 --allow_dynamic_import --js test.js --js_output_file out.js`

Output:
```
test.js:1:12: ERROR - [JSC_CANNOT_CONVERT] This code cannot be converted from ES6. import.meta
  1| console.log(import.meta.url);
                 ^^^^^^^^^^^

1 error(s), 0 warning(s)
```

We previously used `document.currentScript`, which Closure accepts fine (and our code knows about how Closure can transform and combine files, and handles it accordingly). However the lack of Closure support for `import.meta` blocks us switching from classic scripts to modules. It should have the same kind of pass-through support as was added for dynamic import.

rishipal:
I can repro this with v20210601. 

@ChadKillingsworth , looks like `import.meta` is not supported in code here - https://github.com/google/closure-compiler/blob/bba5661a295b607d0f935cb06a00f2ea6eb3824a/src/com/google/javascript/jscomp/Es6RewriteModules.java#L401 ?

Issue #3707, opened 2020-10-19
Support passthrough dynamic import, or basic code doesn't work


AshleyScirra:
Please at least support a basic passthrough for dynamic import for Closure Compiler. In the current version, any presence of dynamic import at all causes a compilation failure. This causes even basic use cases of dynamic import to cause the entire compilation to fail.

Example in.js:

```js
(async () =>
{
	await import("https://cdn.example.com/library.js");
	externalLibrary.someFunc();
})();
```

Command line:

`java -jar ./closure-compiler.jar --js in.js --js_output_file out.js --compilation_level SIMPLE --formatting PRETTY_PRINT --language_in ECMASCRIPT_2020 --language_out ECMASCRIPT_2020`

Observed result:
```
in.js:4:7: WARNING - [JSC_PARSE_ERROR] Parse error. This language feature is not currently supported by the compiler: Dynamic module import
  4|    await import("https://cdn.example.com/library.js");
              ^

ERROR - [JSC_FEATURES_NOT_SUPPORTED_BY_PASS] Attempted to run pass "markUntranspilableFeaturesAsRemoved" on input with features it does not support. Running pass anyway.
Unsupported features: [Dynamic module import]

1 error(s), 1 warning(s)
```
No output file is generated.

Expected result:

Just output the file with all transforms, but passthrough `import(XYZ)` -> `import(XYZ)`

This is blocking us using JavaScript modules at all in our code. We want to insert a dynamic import, but then Closure Compiler will refuse to process any of our code at all.

AshleyScirra:
The same goes for `import.meta` too - we also need to use `import.meta.url`, but the fact all compilation fails if it's present anywhere is a major limitation. That should be able to passthrough too. We know the url or other metadata might change after compilation; that's fine, it either changes nothing in our case or is something we deal with.

concavelenz:
We need some way to distinguish between imports that should be rewritten and those that folks want to preserve.

AshleyScirra:
Just want to link to [this comment](https://github.com/google/closure-compiler/issues/3708#issuecomment-713488361) here as well - for the "just minify" case I think it would be fine to have a flag that just says "pass through all imports/modules features without changing them".

ctjlewis:
Yeah, it would be great if I could tell the compiler not to worry about `import.meta`. `@supress` or `@external import.meta` would be good approaches - tried both, compiler won't take the source if it contains an `import.meta` declaration.

brad4d:
[...]

PR #3904, opened 2022-01-06
Allow import.meta references when the chunk output type is ES_MODULES

ChadKillingsworth:
import.meta references are only valid inside an ES module. While we cannot transpile import.meta, we should allow it to pass through the compiler when the chunk output type is ES_MODULES.

### DIFF
--- a/src/com/google/javascript/jscomp/Es6RewriteModules.java
+++ b/src/com/google/javascript/jscomp/Es6RewriteModules.java
@@ -34,6 +34,7 @@ import static com.google.javascript.jscomp.parsing.parser.FeatureSet.Feature.MOD
 import com.google.common.base.Splitter;
 import com.google.common.collect.HashBasedTable;
 import com.google.common.collect.Table;
+import com.google.javascript.jscomp.CompilerOptions.ChunkOutputType;
 import com.google.javascript.jscomp.ModuleRenaming.GlobalizedModuleName;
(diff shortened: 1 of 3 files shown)


brad4d:
> import.meta references are only valid inside an ES module. While we cannot transpile import.meta, we should allow it to pass through the compiler when the chunk output type is ES_MODULES.

I'm not convinced of this yet. I'm concerned that allowing `import.meta` will just increase user confusion without providing meaningful functionality.

When writing the code the user likely is thinking of the file containing the `import.meta` is in as the "module" it should report about, but at runtime the real ES module is a bunch of code assembled by the compiler from contents of multiple files.

Is there a specific use-case you have in mind that requires access to `import.meta`?

ChadKillingsworth:
You mean beyond #3707 and #3834?

It's the module equivalent of Document.currentScript so there are all kinds of uses. It's perfectly valid with ES Module output and right now we are erroring out without anyway to suppress or accept the consequences - no matter what output language or type is used.

ChadKillingsworth:
It does look like bundlers do different things:

- [Webpack](https://github.com/webpack/webpack/issues/6719#issuecomment-710680859)
- [Rollup](https://rollupjs.org/guide/en/#resolveimportmeta)

There are also a couple different plugins for Babel.

brad4d:
I've imported this and sent it for internal testing and review.

PR #3906, opened 2022-01-11
Adjust types for crypto.getRandomValues

### LINKED ISSUE #3905: Wrong types in Crypto.getRandomValues
### The "real life" problem:
```js
const randomBinaryString8 = String.fromCharCode(...crypto.getRandomValues(new Uint16Array(4)));
console.log(randomBinaryString8);
```
```bash
$ java -jar closure-compiler-v20220104.jar -O ADVANCED -W VERBOSE --formatting PRETTY_PRINT --language_out ECMASCRIPT_NEXT good.js
good.js:1:51: WARNING - [JSC_TYPE_MISMATCH] Spread operator only applies to Iterable types
found   : ArrayBufferView
required: Iterable
  1| const randomBinaryString8 = String.fromCharCode(...crypto.getRandomValues(new Uint16Array(4)));
                                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

const a = String.fromCharCode(...crypto.getRandomValues(new Uint16Array(4)));
console.log(a);
```
Also a problem:
```js
const throwme = crypto.getRandomValues(new DataView(new ArrayBuffer(8))); // obvious TypeMismatchError
console.log(throwme);
```
```bash
$ java -jar closure-compiler-v20220104.jar -O ADVANCED -W VERBOSE --formatting PRETTY_PRINT --language_out ECMASCRIPT_NEXT bad.js
const a = crypto.getRandomValues(new DataView(new ArrayBuffer(8)));
console.log(a);
```

### The reason:
closure-compiler thinks that `Crypto.getRandomValues` takes and returns generic `ArrayBufferView`, which:
- doesn't implement everything `TypedArray` has, e.g. array methods
- can be `Float##Array` or even `DataView`

According to [W3C docs](https://www.w3.org/TR/WebCryptoAPI/#Crypto-method-getRandomValues), it must throw `TypeMismatchError` if argument isn't an integer-based (well, or bigint-based) array.
According to [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues), it takes and returns integer-based `TypedArray` as well.

### Proposed solution:
```diff
diff --git a/externs/browser/w3c_webcrypto.js b/externs/browser/w3c_webcrypto.js
--- a/externs/browser/w3c_webcrypto.js
+++ b/externs/browser/w3c_webcrypto.js
@@ -796,8 +796,8 @@ webCrypto.Crypto = function() {};
 
 /**
  * @see https://developer.mozilla.org/en/DOM/window.crypto.getRandomValues
- * @param {!ArrayBufferView} typedArray
- * @return {!ArrayBufferView}
+ * @param {!TypedArray} typedArray
+ * @return {!TypedArray}
  * @throws {Error}
  */
 webCrypto.Crypto.prototype.getRandomValues = function(typedArray) {};
```
Or in ugly way to raise warning for `Float32Array` and `Float64Array` (totes uncommon mistake):
```diff
- * @param {!ArrayBufferView} typedArray
- * @return {!ArrayBufferView}
+ * @param {!Int8Array|!Uint8Array|!Uint8ClampedArray|!Int16Array|!Uint16Array|!Int32Array|!Uint32Array|!BigInt64Array|!BigUint64Array} typedArray
+ * @return {!Int8Array|!Uint8Array|!Uint8ClampedArray|!Int16Array|!Uint16Array|!Int32Array|!Uint32Array|!BigInt64Array|!BigUint64Array}
```
PR for any of these is ready to be opened.

LiviaMedeiros:
Fixes #3905 

The union type is "any `TypedArray` but not `Float32Array` nor `Float64Array`" according to specifications: [W3C](https://www.w3.org/TR/WebCryptoAPI/#Crypto-method-getRandomValues), [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues).

With this adjustment, correct `JSC_TYPE_MISMATCH` warning is raised for `Float##Array` or `DataView`.
Also compiler always correctly assumes the returned value to be a subset of `TypedArray`.

This "integer-based TypedArray" is needed only for this particular method (to explicitly prevent a footgun, i.e. generating "random floats"), so it doesn't deserve its own typedef.

### DIFF
--- a/externs/browser/w3c_webcrypto.js
+++ b/externs/browser/w3c_webcrypto.js
@@ -796,8 +796,8 @@ webCrypto.Crypto = function() {};
 
 /**
  * @see https://developer.mozilla.org/en/DOM/window.crypto.getRandomValues
- * @param {!ArrayBufferView} typedArray
- * @return {!ArrayBufferView}
+ * @param {!Int8Array|!Uint8Array|!Uint8ClampedArray|!Int16Array|!Uint16Array|!Int32Array|!Uint32Array|!BigInt64Array|!BigUint64Array} typedArray
+ * @return {!Int8Array|!Uint8Array|!Uint8ClampedArray|!Int16Array|!Uint16Array|!Int32Array|!Uint32Array|!BigInt64Array|!BigUint64Array}
  * @throws {Error}
  */
 webCrypto.Crypto.prototype.getRandomValues = function(typedArray) {};


brad4d:
I've imported this change and sent it for internal testing.

brad4d:
Unfortunately this change causes some breakages within Google's codebase.

I'm investigating to see how hard they are to fix.

LiviaMedeiros:
Thanks! Huh, it's hard to imagine what could break from that.

I've also checked most of well-known implementations of Web Crypto API, all modern browsers have the same typecheck as suggested there.

Node.js has inverse check, prohibiting `Float##Array` and allowing `DataView`; and one of their tests will definitely trigger a new warning. Patches are submitted.
They also have `Buffer` type which is a subclass of `Uint8Array`. I see that type partially defined in closure-compiler's externs, but no sign of its inheritance (trying `.every()` raises `JSC_INEXISTENT_PROPERTY`, as well as `.poolSize`).

And a bit strange definition of `DataView` here:
https://github.com/google/closure-compiler/blob/2bf27051dcc96213a976fa8fc263071efcdbf777/externs/es6.js#L1073-L1077
It seems that draft of https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView was at `/Typed_arrays/`.
Snapshots: [2013-08-24](https://web.archive.org/web/20130824075637/https://developer.mozilla.org/en-US/docs/Web/JavaScript/Typed_arrays/DataView) (draft), [2014-02-08](https://web.archive.org/web/20140208023347/https://developer.mozilla.org/en-US/docs/Web/JavaScript/Typed_arrays/DataView) (HTTP 301)
Maybe there was some ambiguity back then.
Hope it would help.

brad4d:
So far the errors I've found were all tests that created a mock for `crypto.getRandomValues()`.
The assignment `crypto.getRandomValues = mockGetRandomValues;` wouldn't compile because the types didn't match anymore.

I'm fixing those by doing this.

```javascript
/**
 * @type {typeof crypto.getRandomValues}
 */
function mockGetRandomValues(typedArray) {
  // implementation here
}
```

brad4d:
I got those fixed, and it looks like there were no further errors.
I've sent this for final review.

PR #3918, opened 2022-02-10
Add externs for ParentNode.replaceChildren

LiviaMedeiros:
This convenient DOM manipulation method is well-supported for a long enough time to be acknowledged.

Refs: [WHATWG](https://dom.spec.whatwg.org/#dom-parentnode-replacechildren), [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Element/replaceChildren), [Can I use](https://caniuse.com/mdn-api_element_replacechildren)


mprobst:
Internal change http://cl/427972679, now running tests.

Issue #3919, opened 2022-02-13
url params are not accepting URL objects


LiviaMedeiros:
A lot of networking methods in JS are accepting any stringable objects as url-string arguments.
```js
const amStringable = {toString: () => 'https://example.com'};
fetch(amStringable).then(response => console.log(response.url)); // 'https://example.com/' if fetched

/* closure-compiler warns:
WARNING - [JSC_TYPE_MISMATCH] actual parameter 1 of fetch does not match formal parameter
found   : {}
required: (Request|string)
*/
```
Most implementations are de facto omnivorous: browsers are casting everything to strings and in case of error happily use `undefined`, `null` and `[object%20Object]`. Usually this is an undesirable outcome, and closure-compiler saves the day with strict typechecking for strings only.

However, [URL](https://url.spec.whatwg.org/#url-class) objects should be acceptable.
Aside from other advantages over raw strings, these objects come with validation and proper encodings. Using them directly as parameters would reinforce safety.

Therefore, I propose replacing `string` with `!URL|string` in externs, whenever method expects an url-string and allows such objects.

LiviaMedeiros:
List (perhaps non-exhaustive) of methods where it is applicable:
- [History.prototype.pushState](https://developer.mozilla.org/en-US/docs/Web/API/History/pushState), [History.prototype.replaceState](https://developer.mozilla.org/en-US/docs/Web/API/History/replaceState)
- [Location.prototype.assign](https://developer.mozilla.org/en-US/docs/Web/API/Location/assign), [Location.prototype.replace](https://developer.mozilla.org/en-US/docs/Web/API/Location/replace)
- [Navigator.prototype.sendBeacon](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon) ※ closure-compiler thinks it's gecko-specific
- [Navigator.prototype.registerProtocolHandler](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/registerProtocolHandler), [Navigator.prototype.unregisterProtocolHandler](https://html.spec.whatwg.org/multipage/system-state.html#dom-navigator-unregisterprotocolhandler) ※ have additional rules
- [Response.redirect](https://developer.mozilla.org/en-US/docs/Web/API/Response/redirect)
- [URL.revokeObjectURL](https://developer.mozilla.org/en-US/docs/Web/API/URL/revokeObjectURL) ※ `URL.createObjectURL` returns raw string
[...]

PR #3920, opened 2022-02-21
Allow URL objects in methods expecting a url-string

LiviaMedeiros:
Fixes https://github.com/google/closure-compiler/issues/3919

This might encourage users to prefer transparent usage of well-encoded and well-validated absolute URLs over raw strings where applicable; and prevent potential errors where URL objects are explicitly stringified just to avoid falsy warning.

Trivia/clarification on `URL` [constructor](https://url.spec.whatwg.org/#dom-url-url):
Apparently, it does allow first param to be an absolute URL even if `base` is given. E.g.:
```js
new URL('ftp://url.com/url', 'https://base.com/base') == 'ftp://url.com/url'; // true
```
This means, it is safe to declare first param as `URL|string`, which is essential for constucting from another `URL` object:
```js
const a = new URL('https://a.com');
const b = new URL(a); // clone of a
```
...which currently raises `JSC_TYPE_MISMATCH` warning.

### DIFF
--- a/externs/browser/fetchapi.js
+++ b/externs/browser/fetchapi.js
@@ -143,7 +143,7 @@ Body.prototype.text = function() {};
 
 
 /**
- * @typedef {!Request|string}
+ * @typedef {!Request|!URL|string}
  * @see https://fetch.spec.whatwg.org/#requestinfo
  */
 var RequestInfo;
@@ -151,12 +151,12 @@ var RequestInfo;
 
 /**
  * @param {!RequestInfo} input
- * @param {!RequestInit=} opt_init
+ * @param {!RequestInit=} init
  * @constructor
  * @implements {Body}
  * @see https://fetch.spec.whatwg.org/#request
  */
-function Request(input, opt_init) {}
+function Request(input, init) {}
 
 /** @override */
 Request.prototype.bodyUsed;
@@ -329,7 +329,7 @@ function Response(opt_body, opt_init) {}
 Response.error = function() {};
 
 /**
- * @param {string} url
+ * @param {!URL|string} url
  * @param {number=} opt_status
  * @return {!Response}
  */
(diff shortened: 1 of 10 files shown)


LiviaMedeiros:
Adjusted more methods for better compatibility with TypeScript.
Refs: https://github.com/LiviaMedeiros/TypeScript-DOM-lib-generator/commit/ead3804f54d5b9ff070a87fef23706d0ffcaaa24
https://github.com/LiviaMedeiros/TypeScript-DOM-lib-generator/compare/acceptsUrl...URL for all changes (includes to-be-adjusted `RequestInfo`)

Trivia:

`RequestInfo`
While changing typedef from `!Request|string` to `!Request|!URL|string` seems to be the most correct, concise and failproof solution, on TypeScript side it might break compatibility with third-party libraries. So in their definitions signatures will have `RequestInfo | URL`, `(RequestInfo | URL)[]`, etc.
I'd prefer to keep typedef change here, but if there are reasons to change signatures instead, it is doable as well.

`Navigator.prototype.registerContentHandler`, `Navigator.prototype.unregisterContentHandler`
Completely obsolete, left untouched.

`Window.prototype.open`
First argument was declared as `*=`. Does hardening it break something important?

`importScripts`
[MDN](https://developer.mozilla.org/en-US/docs/Web/API/WorkerGlobalScope/importScripts) says "These paths are relative to html document base URL" atm. Sounds ambiguous, but [specs](https://html.spec.whatwg.org/multipage/workers.html#dom-workerglobalscope-importscripts-dev) and implementations are happy with absolute URLs.

h-joo:
@mprobst ping

mprobst (approved):

mprobst:
Sorry for the long wait. I'm picking this back up. Change looks good to me, I'll work on landing this in our code base.

Issue #3836, opened 2021-06-20
MediaQueryListEvent is an unknown type


GlennKnight:
I'm getting the following error when trying to use the MediaQueryListEvent type:

**WARNING - [JSC_UNRECOGNIZED_TYPE_ERROR] Bad type annotation. Unknown type MediaQueryListEvent**

Code:
```
    /**
     * Handle MediaQueryList change event
     * @param {MediaQueryListEvent} ev
     */
    handleChange(ev) {
        ...
    }
```

The code passes type checking in VSCode and 'Go to Definition' works.

I'm running "google-closure-compiler": "^20210601.0.0"

It looks like MediaQueryListEvent has been around for a while (https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryListEvent) so maybe it's just a missing extern?

I've tried suppressing the warning for now by adding `@suppress {checkTypes}`, but that doesn't seem to work.

rishipal:
I think that the `MediaQueryListEvent` should've been defined inside `https://github.com/google/closure-compiler/blob/master/externs/browser/w3c_css.js`, but I see that it isn't. Were you able to check that adding it to `w3c_css.js` fixes it?

GlennKnight:
> I think that the `MediaQueryListEvent` should've been defined inside `https://github.com/google/closure-compiler/blob/master/externs/browser/w3c_css.js`, but I see that it isn't. Were you able to check that adding it to `w3c_css.js` fixes it?
>  
> 
> 

I'm using the npm package isn't of building it myself so I haven't been able to try adding a fix. Instead, I created my own local externs file to remove the error

PR #3939, opened 2022-05-09
Add extern for MediaQueryListEvent

treblereel:
missed https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryListEvent

https://github.com/google/closure-compiler/issues/3836


mprobst:
@treblereel thanks for the contribution!

treblereel:
@mprobst is it rollbacked ? what was wrong ?

PR #3960, opened 2022-06-16
little optimize to array iteration

Yuddomack:
Hi! While studying the closure compiler, found something that would be good to trying.

It may not have much meaning in a some javascript engine, but I think optimizing the loop would still be meaningful. (for [efficiency](https://developers.google.com/closure/compiler)!)

thank you for read :)

### DIFF
--- a/src/com/google/javascript/jscomp/js/es6/array/flat.js
+++ b/src/com/google/javascript/jscomp/js/es6/array/flat.js
@@ -35,7 +35,8 @@ $jscomp.polyfill('Array.prototype.flat', function(orig) {
     // TODO(sdh): Consider respecting Symbol.species (b/121061255).
     depth = depth === undefined ? 1 : depth;
     var flattened = [];
-    for (var i = 0; i < this.length; i++) {
+    var length = this.length;
+    for (var i = 0; i < length; i++) {
       var element = this[i];
       if (Array.isArray(element) && depth > 0) {
         var inner = Array.prototype.flat.call(element, depth - 1);
--- a/src/com/google/javascript/jscomp/js/es6/array/flatmap.js
+++ b/src/com/google/javascript/jscomp/js/es6/array/flatmap.js
@@ -34,7 +34,8 @@ $jscomp.polyfill('Array.prototype.flatMap', function(orig) {
    */
   var flatMap = function(callback, thisArg) {
     var mapped = [];
-    for (var i = 0; i < this.length; i++) {
+    var length = this.length;
+    for (var i = 0; i < length; i++) {
       var result = callback.call(thisArg, this[i], i, this);
       if (Array.isArray(result)) {
         mapped.push.apply(mapped, result);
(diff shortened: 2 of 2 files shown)


niloc132:
`flat()` won't need to worry about that, since there is no callback, and `flatMap()` seems to be the same:

My read of the spec says that the size cannot change - https://tc39.es/ecma262/multipage/indexed-collections.html#sec-array.prototype.flatmap takes the array size at start and delegates to https://tc39.es/ecma262/multipage/indexed-collections.html#sec-flattenintoarray, and step 4 explicitly compares against that initial `sourceLen`.

Testing in firefox/chrome:
```
var arr = [1,2] 
arr.flatMap((val, index, array) => {if (index == 0) { array[2] = 3} return [0, val]}) 
```
In both cases, the result of the flatmap is always [0,1,0,2] as expected, the new value isn't included. Take care when repeating this to replace `arr`'s contents, since it correctly grows during this operation, so just running the flatMap again will mean you see the new value the second time.

Yuddomack:
@frost-cy 

> My read of the spec says that the size cannot change - https://tc39.es/ecma262/multipage/indexed-collections.html#sec-array.prototype.flatmap takes the array size at start and delegates to https://tc39.es/ecma262/multipage/indexed-collections.html#sec-flattenintoarray, and step 4 explicitly compares against that initial sourceLen.

I agree with this opinion. (thank you for description @niloc132)
Could it have been helpful?

### but my point is

If you refer to `.length` inside the `for` statement, it is executed every time.
this is the behavior of javascript.

```javascript
var arr = [1,2,3,4,5];
for(var i = 0; i < arr.length; i++){
  // first loop - reference executed in array 'length: 5'
  // second loop - reference executed in array 'length: 5'
  // ...
  // last loop - reference executed in array 'length: 5'
}
```

This is an internal implementation of the JavaScript engine, but it can reproduce.

see this code https://stackblitz.com/edit/node-qmymnz?file=index.js and try to run.
then you can see logs

image

If the user declares a very large array and calls it, many references will occur.

There are many different JavaScript engines(V8 of google, Rhino, SpiderMonkey, Chakra, Webkit), and some engines optimize these problems themselves.
but it depends on the user's environment anyway.

decisively, this solve is already using in this repository's code

https://github.com/google/closure-compiler/blob/00fa46abfb7bb98366f23685ce0a728a0f5e4cea/src/com/google/javascript/jscomp/js/es6/array/from.js#L55-L59

https://github.com/google/closure-compiler/blob/00fa46abfb7bb98366f23685ce0a728a0f5e4cea/src/com/google/javascript/jscomp/js/es6/string/raw.js#L41-L48

https://github.com/google/closure-compiler/blob/00fa46abfb7bb98366f23685ce0a728a0f5e4cea/src/com/google/javascript/jscomp/js/util/findinternal.js#L39-L43

...

I'm not good at English, so I hope it's communicated well.

PR #3974, opened 2022-08-02
Fix useless tests for LiveVariablesAnalysis

### LINKED ISSUE #3973: Potentially incorrect/useless test case for LiveVariablesAnalysis
TL;DR: a test doesn't appear to verify anything useful. No impact on users.

Hi, I was recently looking at the tests for `LiveVariablesAnalysis` and came across some test cases that appear to be incorrect/useless. 
https://github.com/google/closure-compiler/blob/1b7b32d78640297d8814ddabaf971da530c98659/test/com/google/javascript/jscomp/LiveVariablesAnalysisTest.java#L360-L368
The assertions appear to test that various destructuring parameters aren't escaped by references to the `arguments` array. However, the params in the input code will _never_ be considered escaped because they, and the `arguments` references, are in nested functions; The inputs contain a function e.g. `function f([a]) { arguments; }` but are then wrapped in another function by `computeLiveness`:
https://github.com/google/closure-compiler/blob/1b7b32d78640297d8814ddabaf971da530c98659/test/com/google/javascript/jscomp/LiveVariablesAnalysisTest.java#L677-L678
which then configures `ControlFlowAnalysis` to not visit nested functions (note the lack of `builder.setTraverseFunctions(true)`):
https://github.com/google/closure-compiler/blob/1b7b32d78640297d8814ddabaf971da530c98659/test/com/google/javascript/jscomp/LiveVariablesAnalysisTest.java#L691-L696
This means that the params will never be escaped because:

1. The `arguments` references are never visited during data flow analysis, so [`markAllParametersEscaped`](https://github.com/google/closure-compiler/blob/1b7b32d78640297d8814ddabaf971da530c98659/src/com/google/javascript/jscomp/LiveVariablesAnalysis.java#L433) is never called.
2. Even if `markAllParametersEscaped` were called, the _outer_ params (`param1`, `param2`, `param3`) would be escaped, not the inner params (`a`).

The issue can be shown by adding the following test cases to LiveVariablesAnalysisTest.java:
```java
assertNotEscaped("function f(a) { arguments; }", "a");
   assertEscaped("function f(a) { arguments; }", "a");
```
The param `a` should be escaped because it's a `NAME` param, and therefore `assertNotEscaped` should fail and `assertEscaped` should pass, but the _reverse_ happens.

One way to resolve this would be to add an alternate `assertNotEscaped` helper function that doesn't wrap the input in a function. I believe this would correct the tests so they verify the intended behaviour.

If you agree with the above, I'm happy to open a PR to explore the aforementioned fix.

CamWass:
This PR fixes #3973 by allowing callers to configure whether `assertNotEscaped` (and by extension `computeLiveness`) wrap the input in a function. The previously useless tests in `testArgumentsArray_doesNotEscape_destructuredParams` now leverage this flag so they test the correct behaviour. These were the only affected tests, so all others remain unchanged. This was accomplished by overloading the relevant helpers, attempting to match the style of other helpers and minimise changes to unrelated code.

Not sure about the new param name and java doc comments - I tried to make them descriptive to prevent the same mistake being made in the future. Feedback and suggestions are most welcome.

The code has been formatted by [google-java-format](https://github.com/google/google-java-format).

### DIFF
--- a/test/com/google/javascript/jscomp/LiveVariablesAnalysisTest.java
+++ b/test/com/google/javascript/jscomp/LiveVariablesAnalysisTest.java
@@ -360,11 +360,11 @@ public final class LiveVariablesAnalysisTest {
   @Test
   public void testArgumentsArray_doesNotEscape_destructuredParams() {
     // These cases also cover a crash related to assuming all RESTs have a NAME child.
-    assertNotEscaped("function f([a]) { arguments; }", "a");
-    assertNotEscaped("function f([a] = []) { arguments; }", "a");
-    assertNotEscaped("function f(...[a]) { arguments; }", "a");
-    assertNotEscaped("function f({a}) { arguments; }", "a");
-    assertNotEscaped("function f({a} = {}) { arguments; }", "a");
+    assertNotEscaped("function f([a]) { arguments; }", "a", false);
+    assertNotEscaped("function f([a] = []) { arguments; }", "a", false);
+    assertNotEscaped("function f(...[a]) { arguments; }", "a", false);
+    assertNotEscaped("function f({a}) { arguments; }", "a", false);
+    assertNotEscaped("function f({a} = {}) { arguments; }", "a", false);
   }
 
   @Test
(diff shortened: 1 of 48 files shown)


test/com/google/javascript/jscomp/LiveVariablesAnalysisTest.java
```diff
@@ -659,12 +659,29 @@ private static void assertEscaped(String src, String name) {
...
+    return computeLiveness(src, async, true);
+  }
+
+  /**
+   * @param wrapInFunction Whether the input should be wrapped in a function e.g. <code>var a = 1;
+   *     </code> -> <code>function f() { var a = 1; }</code>
+   * @param async Whether the wrapper function should be {@code async}.
+   */
+  private static LiveVariablesAnalysis computeLiveness(
+      String src, boolean async, boolean wrapInFunction) {
```
lauraharker, line 684: Could we replace the two boolean parameters with an enum? something like

``` java
enum Wrapper {
  NONE,
  FUNCTION,
  ASYNC_FUNCTION
}
```

lauraharker (changes requested):
Thanks for looking into this! Just one minor suggestion on my end

test/com/google/javascript/jscomp/LiveVariablesAnalysisTest.java
CamWass: Good idea, the enum makes more sense. Done

test/com/google/javascript/jscomp/LiveVariablesAnalysisTest.java
```diff
@@ -360,11 +360,11 @@ public void testArgumentsArray() {
   @Test
   public void testArgumentsArray_doesNotEscape_destructuredParams() {
     // These cases also cover a crash related to assuming all RESTs have a NAME child.
-    assertNotEscaped("function f([a]) { arguments; }", "a");
-    assertNotEscaped("function f([a] = []) { arguments; }", "a");
-    assertNotEscaped("function f(...[a]) { arguments; }", "a");
-    assertNotEscaped("function f({a}) { arguments; }", "a");
-    assertNotEscaped("function f({a} = {}) { arguments; }", "a");
+    assertNotEscaped("function f([a]) { arguments; }", "a", Wrapper.FUNCTION);
```
lauraharker, line 363: Should these be `Wrapper.NONE`?

CamWass: Whoops! I managed to re-break the tests. Nice catch; fixed

PR #3975, opened 2022-08-04
#3840 - allow user to pass any path separator as --chunk_output…

### LINKED ISSUE #3840: --create_source_map throws exception on windows when using input source maps and chunks
Closure throws an exception when I try to build source maps for chunks:
java.io.FileNotFoundException: .\dist\closure\common.js.map (The system cannot find the path specified)
It seems like there's a race condition issue if the source map of a depending module is built before the dependency. It seems to work properly if I run the build without `--create_source_map`, and then run it again _with_ `--create_source_map`.

Reproducer is available here: https://github.com/DerGernTod/ts-esbuild-closure-reproducer/blob/main/index.js

The used flags:
--chunk common:2
--chunk_wrapper common:(function(){%s})()
--source_map_input dist/esbuild/chunk-7THFJKJQ.js|dist/esbuild/chunk-7THFJKJQ.js.map
--js dist/esbuild/chunk-7THFJKJQ.js
--source_map_input dist/esbuild/filecommon.js|dist/esbuild/filecommon.js.map
--js dist/esbuild/filecommon.js
--chunk filea:1:common
--chunk_wrapper filea:(function(){%s})()
--source_map_input dist/esbuild/filea.js|dist/esbuild/filea.js.map
--js dist/esbuild/filea.js
--chunk fileb:1:common
--chunk_wrapper fileb:(function(){%s})()
--source_map_input dist/esbuild/fileb.js|dist/esbuild/fileb.js.map
--js dist/esbuild/fileb.js
--language_in ECMASCRIPT_2017
--language_out ECMASCRIPT_2017
--chunk_output_path_prefix ./dist/closure/
--rename_prefix_namespace myNamespace
--compilation_level ADVANCED_OPTIMIZATIONS
--create_source_map %outname%.map

I also tried `--source_map_include_content`, but that doesn't seem to work either (even if you build it once without source maps to prevent the error mentioned above, and then build it with source maps). The source is not embedded at all.

DerGernTod:
fixes #3840 - allow user to pass any path separator as --chunk_output_path_prefix

as a user, you might want to use unix-style path separators always, since they're also compatible with windows. doing so however causes issues when writing source maps, since `maybeCreateDirsForPath` only creates output directories if path separators are in os-style.

### DIFF
--- a/src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
+++ b/src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
@@ -1139,10 +1139,14 @@ public abstract class AbstractCommandLineRunner<A extends Compiler, B extends Co
   @GwtIncompatible("Unnecessary")
   private static void maybeCreateDirsForPath(String pathPrefix) {
     if (!Strings.isNullOrEmpty(pathPrefix)) {
-      String dirName =
-          pathPrefix.charAt(pathPrefix.length() - 1) == File.separatorChar
-              ? pathPrefix.substring(0, pathPrefix.length() - 1)
-              : new File(pathPrefix).getParent();
+      char sep = pathPrefix.charAt(pathPrefix.length() - 1);
+      String dirName;
+      // allow the user to provide any path separator
+      if (sep == '/' || sep == '\\') {
+        dirName = pathPrefix.substring(0, pathPrefix.length() - 1);
+      } else {
+        dirName = new File(pathPrefix).getParent();
+      }
       if (dirName != null) {
         new File(dirName).mkdirs();
       }
--- a/src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
+++ b/src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
@@ -1139,17 +1139,7 @@ public abstract class AbstractCommandLineRunner<A extends Compiler, B extends Co
   @GwtIncompatible("Unnecessary")
   private static void maybeCreateDirsForPath(String pathPrefix) {
     if (!Strings.isNullOrEmpty(pathPrefix)) {
-      char sep = pathPrefix.charAt(pathPrefix.length() - 1);
-      String dirName;
-      // allow the user to provide any path separator
-      if (sep == '/' || sep == '\\') {
-        dirName = pathPrefix.substring(0, pathPrefix.length() - 1);
-      } else {
-        dirName = new File(pathPrefix).getParent();
-      }
-      if (dirName != null) {
-        new File(dirName).mkdirs();
-      }
+      new File(filePath).getParent().mkdirs();
     }
   }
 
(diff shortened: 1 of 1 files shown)


src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
```diff
@@ -1139,10 +1139,14 @@ void writeOutput(
...
   private static void maybeCreateDirsForPath(String pathPrefix) {
     if (!Strings.isNullOrEmpty(pathPrefix)) {
-      String dirName =
-          pathPrefix.charAt(pathPrefix.length() - 1) == File.separatorChar
-              ? pathPrefix.substring(0, pathPrefix.length() - 1)
-              : new File(pathPrefix).getParent();
+      char sep = pathPrefix.charAt(pathPrefix.length() - 1);
+      String dirName;
+      // allow the user to provide any path separator
+      if (sep == '/' || sep == '\\') {
```
brad4d, line 1145: This was a really good catch.

This method is just badly designed.
IIUC, it's trying to interpret a trailing file separator to mean "create this full path as a directory", but if there isn't a trailing separator it just creates directories up to the last path element.

how confusing.

I've just taken a look at all callers of this function.
All of them pass paths that will be used to create a file except for the one on line 1778

Presumably the idea there is a user may specify either `path/to/my/output/chunks/` or `path/to/my/output/chunks/chunk-` as a prefix

How about lets drop the check for a trailing file separator from this method and instead fix that one call on 1778 to pass a string with a dummy file name appended to eliminate the need for the check?

The parameter to this file then should be named "filePath" instead of "pathPrefix" and the body is much simpler

```
if (!Strings.isNullOrEmpty(filePath)) {
  new File(filePath).getParent().mkdirs();
}
```

brad4d (changes requested):

src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
theseanl: > How about lets drop the check for a trailing file separator from this method and instead fix that one call on 1778 to pass a string with a dummy file name appended to eliminate the need for the check?

Would this change the behavior of the command line flag `chunk_output_path_prefix`? 
https://github.com/google/closure-compiler/blob/736de5253399c743b439ae03696e9eca3fb86aff/src/com/google/javascript/jscomp/CommandLineRunner.java#L404-L410

https://github.com/theseanl/tscc depends on the behavior of the flag `chunk_output_path_prefix` of treating trailing path separator as a signal for creating a new directory.

DerGernTod: if i understand @brad4d correctly, it would behave exactly the same (which is the confusing part). if i pass `out/dir/` and pass a dummy file like `out/dir/dummy`, or use `out/dir/mypath` and still use the dummy like `out/dir/mypathdummy`, the result would be exactly the same, the paths would be created up to the last found `/`, so `out/dir`

DerGernTod:
looks like the cla bot doesn't recognize my account, i'll try to resubmit a new pr

Issue #3957, opened 2022-06-09
gson version is not upgraded in thelatest version of the closure-compiler


GaneshKaparapu:
We needed to use 2.8.9 or later vesrion of gson as the older versions are having a security vulnerability. We are seeing 2.7 version in almost last one year released versions of closure compiler.

rahul-kamat:
Thanks! We've imported the latest version of `gson` into google now, so the next release of closure compiler should include this update

cmuchinsky:
I'm still seeing the older `2.7` version referenced in [WORKSPACE.bazel](https://github.com/google/closure-compiler/blob/master/WORKSPACE.bazel#L58), is this not where the dependency version is defined?

cmuchinsky:
I confirmed that the `master` branch still seems to build using gson `2.7`. With the following changes I was able to build with `2.9.0` and run the tests successfully, should this issue be re-opened @rahul-kamat @GaneshKaparapu @brad4d ?
```diff --git a/README.md b/README.md
index 39ea3e9e3..6c859a867 100644
--- a/README.md
+++ b/README.md
@@ -642,7 +642,7 @@ without make's wrinkles and with the full portability of pure java code.</td>
 
   <tr>
     <td>Version</td>
-    <td>2.7</td>
+    <td>2.9.0</td>
   </tr>
 
   <tr>
diff --git a/WORKSPACE.bazel b/WORKSPACE.bazel
index e6dfcc5fe..3cf8c43d0 100644
--- a/WORKSPACE.bazel
+++ b/WORKSPACE.bazel
@@ -55,8 +55,8 @@ maven_import(
     # https://github.com/google/gson
     group_id = "com.google.code.gson",
     artifact_id = "gson",
-    version = "2.7",
-    sha256 = "2d43eb5ea9e133d2ee2405cc14f5ee08951b8361302fdd93494a3a997b508d32",
+    version = "2.9.0",
+    sha256 = "c96d60551331a196dac54b745aa642cd078ef89b6f267146b705f2c2cbef052d",
     licenses = ["notice"],
 )

PR #3982, opened 2022-08-10
Bump gson from 2.7 to 2.9.1 to address CVE-2022-25647

cmuchinsky:
This should resolve https://github.com/google/closure-compiler/issues/3957, which I believe was closed prematurely without a fix. This is needed because the closure-compiler jar is being flagged with CVE-2022-25647 as a result of the shaded version of gson `2.7` inside it.


cmuchinsky:
I updated the PR to pick up the latest nodejs test fixes from the `master` branch

PR #3997, opened 2022-09-27
Prevent replacing "super" references when inlining

ChadKillingsworth:
When static inheritance is utilized, references to super must not be replaced while inlining as it changes the refence of "this" when calling methods and getters.

While the `GlobalNamespace` class correctly utilized the option, there still existed an unconditional call to `StaticSuperPropReplacer`. Even though the final reference was not inlined, the `super` reference was replaced with the classname.

### DIFF
--- a/src/com/google/javascript/jscomp/InlineAndCollapseProperties.java
+++ b/src/com/google/javascript/jscomp/InlineAndCollapseProperties.java
@@ -95,6 +95,7 @@ class InlineAndCollapseProperties implements CompilerPass {
   private final ChunkOutputType chunkOutputType;
   private final boolean haveModulesBeenRewritten;
   private final ResolutionMode moduleResolutionMode;
+  private final boolean staticInheritanceUsed;
 
   /**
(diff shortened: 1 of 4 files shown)


src/com/google/javascript/jscomp/InlineAndCollapseProperties.java
```diff
@@ -176,6 +179,12 @@ public Builder testAggressiveInliningOnly(Consumer<GlobalNamespace> globalNamesp
       return this;
     }
 
+    @CanIgnoreReturnValue
+    public Builder setAssumeStaticInheritanceIsNotUsed(boolean assumeStaticInheritanceIsNotUsed) {
+      this.staticInheritanceUsed = assumeStaticInheritanceIsNotUsed == false;
```
rishipal, line 184: Suggested replacement:
      this.staticInheritanceUsed = !assumeStaticInheritanceIsNotUsed;

rishipal (approved):

Issue #3996, opened 2022-09-25
Make character set for renaming customizable


LiviaMedeiros:
Character set seems to be hardcoded:
https://github.com/google/closure-compiler/blob/440d1f188e63079fc451a5010c9dbf3e744f5f4d/src/com/google/javascript/jscomp/DefaultNameGenerator.java#L89-L95

- The set is not exhaustive: variable/property name can start with `_`.
- The set contains `$`, which sometimes might be undesired. For example, we can do `Object.assign(someDiv.dataset, dataObj)` and be okay with properties of `dataObj` being mangled; but at some point users will get `DOMException: Failed to set a named property on 'DOMStringMap': 'data-$' is not a valid attribute name.` in runtime.

How about adding something like `--rename_variable_characters` option to adjust naming generation?

gunan:
Thank you for bringing this issue to our attention.

Can we get some more information on why you need renaming on these attributes?

If you like to contribute a means of excluding `$` from the renaming, we can consider that PR.

blickly:
Chad pointed out on the PR that setting `--polymer_version=2` effectively gives this desired behavior already.

So it seems like using that flag even on non-Polymer projects might be a workaround to get the behavior you want without landing a compiler change.

blickly:
@LiviaMedeiros, can you try out adding that flag to your project and let us know if that works for your use case?

LiviaMedeiros:
Yes, adding `--polymer_version=2` did the trick for me. As far as I can tell, it restricts all characters except lowercase letters, which is a bit overkill but is technically desired because it also helps to avoid camelCase conversion.
Also https://github.com/google/closure-compiler/pull/4001#issuecomment-1269869127 mentioned first characters but it seems to be applied to all characters, which is also desired.

If this flag looks like an adequate workaround in terms of reliability (i.e. not just convenient unintended side effect, and safe to use by default), feel free to close the issue.

Thanks!

PR #4001, opened 2022-10-05
Exclude `$` from appearing in renamed variable and property names

LiviaMedeiros:
Mangled property names can be accessed via `Object.keys()`, `Object.assign()`, etc.
In most scenarios, assuming renaming consistency (i.e. all related code is processed into one file by compiler), it is completely safe to do.

However, using property names with `$` might be undesirable in some scenarios, for example in `data-*` html attritubes or regular expressions.

Refs: https://github.com/google/closure-compiler/issues/3996

### DIFF
--- a/src/com/google/javascript/jscomp/DefaultNameGenerator.java
+++ b/src/com/google/javascript/jscomp/DefaultNameGenerator.java
@@ -87,11 +87,11 @@ public final class DefaultNameGenerator implements NameGenerator {
   // end up balancing the huffman tree and result is bad compression.
   /** Generate short name with this first character */
   static final char[] FIRST_CHAR =
-    "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ$".toCharArray();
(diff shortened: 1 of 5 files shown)


blickly:
I can see how this might be helpful in some scenarios, but always excluding any names that include a `$` character seems very likely to cause code-size increases, and likely not acceptable for all users.  If we do add such functionality, it probably belongs behind some sort of flag rather than hard-coded.

ChadKillingsworth:
A side effect of using the `--polymer_version=3` flag is that the characters variable names can start with are restricted for exactly this reason.

blickly:
Good point Chad, that's a good workaround.  Would using the `--polymer_version=2` (I think that's what you meant since the polymer version flag can only be set to 1 or 2 AFAIU) be an adequate workaround?  I'm pretty sure that it's safe to set on non-polymer projects...

blickly:
Since we have what seems like a reasonable workaround without needing to change the naming for everyone, I'll close this PR for now, and we can continue the discussion on #3996

PR #4007, opened 2022-11-03
Avoid parsing JS files when building a bundle

### LINKED ISSUE #4006: Compilation level BUNDLE always fully parses file instead of only using regex parser
The idea of BUNDLE is that files are reordered and slightly wrapped, transpiling only when necessary to allow them all to be in the same file instead of either requiring that the browser load multiple files, or fully transpiling contents (in another optimization level). In theory, this means that the `JsFileRegexParser` should be sufficient to pick up dependencies between the files to correctly order them for output, but at this time `JsAst.getAstRoot` is called for every file passed to the command line runner.

This can be easily confirmed by trying to bundle a small set of files with a breakpoint in `JsAst.parse`. There are roughly two paths to this call - `Compiler.orderInputs` runs before the bundle work gets started (only if the --dependencyMode is set to something that requires dependencies to be directly managed), but `AbstractCommandLineRunner.printBundleTo` also hits this in potentially two ways, if numParallelThreads is greater than 1 (ironically asking the compiler to parallelize will cause it to do substantially more work), or if preventLibraryInjection is false. 

There doesn't seem to be a way to "fix" PrebuildAst directly, but perhaps we could instead invoke CompilerInput.getDependencyInfo() on other threads, which in turn delegates to JsFileRegexParser. With that said, `parseInputs`, which calls `orderInputs` above, also does PrebuildAst work. It might be better to just encourage BUNDLE users to not set numParallelThreads.

But for the other two, the root issue is that while JsFileRegexParser does the parsing work as expected, it is then wrapped by a `LazyParsedDependencyInfo` instead before being returned. Then, several callers invoke `getLoadFlags().get("module")` to see if the file is an "es6" module (uses `import`/`export`), or a "goog" module (uses goog.provide/require/module/etc). However, while LazyParsedDependencyInfo.getLoadFlags() calls to the underlying `AutoValue_SimpleDependencyInfo` delegate to see the initial set of flags, it then calls `ast.getFeatures(compiler)`, which will actually run the parser. Some validation will then occur - if the parsed features happen to suggest that the file is an es6 module but the JsFileRegexParser didn't detect this, there will be an error.

That validation is the only value I see at a glance, and to my untrained eye it would be a compiler bug for the two parsers to disagree (the user shouldn't need to explicitly enable BUNDLE to use es6 modules, but perhaps other compilation levels would need that?), and Feature.MODULES only appears to be set when the parser encounters `import`, `import.meta`, or `export`. At a brief glance, `import.meta` might not be checked for in JsFileRegexParser, which could contribute to this...?

With this in mind, it seems reasonable to introduce convinience methods that call `getLoadFlags().get("module")` and compare to the two known values, "goog" and "es6", returning a boolean for each one. See the linked PR for a proposed draft of this change.

Finally, this doesn't completely remove all parsing - any es6 module needs to be parsed and rewritten to correctly map its imports and exports. Perhaps the PrebuildAst could be limited to only files with module=es6 to let that happen up front.

Here's a quick perf run, new then old. This run is on closure-library's source, some 1k JS files, showing a substantial speedup
```
[colin@runes closure-compiler (master *)]$ time java -jar compiler_new.jar --js ../closure-library/closure/goog/**/*.js --js ../closure-library/closure/goog/*.js --dependency_mode SORT_ONLY --compilation_level BUNDLE --js_output_file `pwd`/out.js

real    0m1.640s
user    0m6.588s
sys     0m0.418s
[colin@runes closure-compiler (master *)]$ time java -jar compiler_new.jar --js ../closure-library/closure/goog/**/*.js --js ../closure-library/closure/goog/*.js --dependency_mode SORT_ONLY --compilation_level BUNDLE --js_output_file `pwd`/out.js

real    0m1.597s
user    0m6.644s
sys     0m0.432s
[colin@runes closure-compiler (master *)]$ time java -jar compiler_new.jar --js ../closure-library/closure/goog/**/*.js --js ../closure-library/closure/goog/*.js --dependency_mode SORT_ONLY --compilation_level BUNDLE --js_output_file `pwd`/out.js

real    0m1.714s
user    0m6.759s
sys     0m0.420s
```
```
[colin@runes closure-compiler (master *)]$ time java -jar compiler_old.jar --js ../closure-library/closure/goog/**/*.js --js ../closure-library/closure/goog/*.js --dependency_mode SORT_ONLY --compilation_level BUNDLE --js_output_file `pwd`/out-old.js

real    0m4.232s
user    0m31.733s
sys     0m1.321s
[colin@runes closure-compiler (master *)]$ time java -jar compiler_old.jar --js ../closure-library/closure/goog/**/*.js --js ../closure-library/closure/goog/*.js --dependency_mode SORT_ONLY --compilation_level BUNDLE --js_output_file `pwd`/out-old.js

real    0m4.187s
user    0m30.468s
sys     0m1.193s
[colin@runes closure-compiler (master *)]$ time java -jar compiler_old.jar --js ../closure-library/closure/goog/**/*.js --js ../closure-library/closure/goog/*.js --dependency_mode SORT_ONLY --compilation_level BUNDLE --js_output_file `pwd`/out-old.js

real    0m4.010s
user    0m32.340s
sys     0m1.040s
```

The contents of the output files are identical, before/after.

niloc132:
This avoids the case where LazyParsedDependencyInfo.getLoadFlags() might parse the entire file to check if a JS file can be bundled as a Closure or ES6 module.

Fixes #4006

### DIFF
--- a/src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
+++ b/src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
@@ -2403,7 +2403,7 @@ public abstract class AbstractCommandLineRunner<A extends Compiler, B extends Co
       // ES6 modules will need a runtime in a bundle. Skip appending this runtime if there are no
       // ES6 modules to cut down on size.
       for (CompilerInput input : inputs) {
-        if ("es6".equals(input.getLoadFlags().get("module"))) {
+        if (input.isEs6Module()) {
           appendRuntimeTo(out);
           break;
         }
--- a/src/com/google/javascript/jscomp/deps/DependencyInfo.java
+++ b/src/com/google/javascript/jscomp/deps/DependencyInfo.java
@@ -129,7 +129,9 @@ public interface DependencyInfo {
   /** Gets the symbols required by this file. */
   ImmutableList<Require> getRequires();
 
-  ImmutableList<String> getRequiredSymbols();
+  default ImmutableList<String> getRequiredSymbols() {
+    return Require.asSymbolList(getRequires());
+  }
 
   /** Gets the symbols type-required by this file (i.e. for typechecking only). */
   ImmutableList<String> getTypeRequires();
@@ -137,12 +139,11 @@ public interface DependencyInfo {
   /** Gets the loading information for this file. */
   ImmutableMap<String, String> getLoadFlags();
 
-  /** Whether the symbol is provided by a module */
-  boolean isModule();
-
+  /** Whether the symbol is provided by an ES6 module */
   default boolean isEs6Module() {
     return "es6".equals(getLoadFlags().get("module"));
   }
+  /** Whether the symbol is provided by a goog module */
   default boolean isGoogModule() {
     return "goog".equals(getLoadFlags().get("module"));
   }
@@ -152,21 +153,6 @@ public interface DependencyInfo {
 
   /** Whether the file has the '@nocompile' annotation. */
   boolean getHasNoCompileAnnotation();
-
-  /**
-   * Abstract base implementation that defines derived accessors such
-   * as {@link #isModule}.
-   */
-  abstract class Base implements DependencyInfo {
-    @Override public boolean isModule() {
-      return isGoogModule();
-    }
-
-    @Override
-    public ImmutableList<String> getRequiredSymbols() {
-      return Require.asSymbolList(getRequires());
-    }
-  }
 
   /** Utility methods. */
   class Util {
--- a/src/com/google/javascript/jscomp/deps/DependencyInfo.java
+++ b/src/com/google/javascript/jscomp/deps/DependencyInfo.java
@@ -143,6 +143,7 @@ public interface DependencyInfo {
   default boolean isEs6Module() {
     return "es6".equals(getLoadFlags().get("module"));
   }
+
   /** Whether the symbol is provided by a goog module */
   default boolean isGoogModule() {
     return "goog".equals(getLoadFlags().get("module"));
--- a/src/com/google/javascript/jscomp/deps/DependencyInfo.java
+++ b/src/com/google/javascript/jscomp/deps/DependencyInfo.java
@@ -140,6 +140,13 @@ public interface DependencyInfo {
   /** Whether the symbol is provided by a module */
   boolean isModule();
 
+  default boolean isEs6Module() {
+    return "es6".equals(getLoadFlags().get("module"));
+  }
+  default boolean isGoogModule() {
+    return "goog".equals(getLoadFlags().get("module"));
+  }
+
   /** Whether the file '@externs' annotation. */
   boolean getHasExternsAnnotation();
 
@@ -152,7 +159,7 @@ public interface DependencyInfo {
    */
   abstract class Base implements DependencyInfo {
     @Override public boolean isModule() {
-      return "goog".equals(getLoadFlags().get("module"));
+      return isGoogModule();
     }
 
     @Override
--- a/src/com/google/javascript/jscomp/Compiler.java
+++ b/src/com/google/javascript/jscomp/Compiler.java
@@ -2232,7 +2232,7 @@ public class Compiler extends AbstractCompiler implements ErrorHandler, SourceFi
       // Only process files that are detected as ES6 modules
       if (!options.getDependencyOptions().shouldPrune()
           || !JsFileRegexParser.isSupported()
-          || "es6".equals(input.getLoadFlags().get("module"))) {
+          || input.isEs6Module()) {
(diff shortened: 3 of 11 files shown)


src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
```diff
@@ -2403,7 +2403,7 @@ void printBundleTo(JSChunk module, Appendable out) throws IOException {
       // ES6 modules will need a runtime in a bundle. Skip appending this runtime if there are no
```
niloc132, line 2403: Just above this is where the PrebuiltAst check runs - we could simplify that as well in this patch to only prebuild the inputs that return true for isEs6Module()?

lauraharker:
I'm trying to figure out why @googlebot hasn't run for this PR - usually it automatically runs to check if you've signed a Contributor's Licensing Agreement. Example: https://github.com/google/closure-compiler/pull/3833#issuecomment-861923992.

I'll try to retrigger @googlebot. Meanwhile, if you haven't already signed at https://cla.developers.google.com/, please go ahead and do so.

lauraharker:
Ok, I reran the CLA check and looks like you've already signed, thanks.

I've imported this for some internal testing.

src/com/google/javascript/jscomp/deps/DependencyInfo.java
```diff
@@ -140,6 +140,13 @@ abstract static class Builder {
   /** Whether the symbol is provided by a module */
```
lauraharker, line 140: Should `isModule()` and `isGoogModule()` always return the same thing?

If so let's be more explicit about that - could we make DependencyInfo.Base.isModule `final`? The only overrides I see are in JSChunk.java and DependencyMap.java.

lauraharker (changes requested):

src/com/google/javascript/jscomp/deps/DependencyInfo.java
niloc132: Probably, but I'm not totally sure that it is always the same thing. From JSChunk.isModule():
```
  @Override
  public boolean isModule() {
    // NOTE: The meaning of "module" has changed over time.  A "JsChunk" is
    // a collection of inputs that are loaded together. A "module" file,
    // is a CommonJs module, ES6 module, goog.module or other file whose
    // top level symbols are not in global scope.
    throw new UnsupportedOperationException();
  }
```
I was concerned that, despite the implementation in DependencyInfo.Base.isModule() explicitly checking for goog modules, that this JsChunk implementation required that isModule continue to throw. The comment _seems_ to be generally inaccurate, or at least in conflict with other implementations, so I only changed DependencyInfo.Base.isModule() to take the fast path (and skip parsing), but stil permit JSChunk to throw in this specific case.

I didn't notice at the time (but am seeing now) that JSChunk also throws for getLoadFlags(), so perhaps I'm being overly cautious. Will be guided by your knowledge/preference here - if I'm being too careful about code that I can't see (that assumes isModule() might in fact be referring to commonjs/es6/goog as the comment describes), yes, we can probably delete JSChunk.isModule().

--

I can see no source file with the name DependencyMap.java, nor type with that name - typo, or internal class I can't see?

src/com/google/javascript/jscomp/AbstractCommandLineRunner.java
lauraharker: Could you leave this as-is, perhaps with a comment? I agree this is potentially useful, but I'd like to have someone profile some of our internal Closure Compiler users before making any changes.

lauraharker (changes requested):

lauraharker:
Update: I'm looking at a few test failures in our internal test suite. Currently investigating whether they're caused by some internal-only changes I made/failed to make, or something in this PR.

lauraharker:
Another update:

I'm currently working on trying to avoid using the regex parser for compilations that need to fully parse each input file anyway. So the goals of these two changes don't conflict, but I still need to verify my implementation won't mess up your PR or vice versa.

niloc132:
Thanks for the update, I appreciate it. Let me know if it makes more sense to merge your changes and have me try to rework this instead of you managing both. 

We're trying to pick up the typed_ast feature for ADVANCED compiles without re-parsing everything on small changes, but have a sort of simplified version for BUNDLE that serializes the results of the regex parser to avoid even redoing that parsing work on the bulk of a given library. While this is working, we're still figuring out typed_ast, so redoing some of this work wouldn't be a big impediment if it helps generally improve performance.

lauraharker:
Thanks for all your patience - this change looks good to go, no conflicts with what I'm working on. I've gotten it approved in internal code review and plan to submit it on Monday.

Issue #4009, opened 2022-11-14
Protobuf java version 3.19.3 affected by CVE-2022-3171


cmuchinsky:
The `protobuf-java` version declared in `WORKSPACE.bazel` (`"com.google.protobuf:protobuf-java:3.19.3"`) is affected by CVE-2022-3171. Version `3.19.6` should be used instead as it contains the fix.

mprobst:
This has been fixed in https://github.com/google/closure-compiler/commit/6da46fcc7fe64b74eeb87dc4a2493611cdf30f6a.

PR #4010, opened 2022-11-15
Bump protobuf-java from 3.19.3 to 3.19.6 to remediate CVE-2022-3171

cmuchinsky:
Fixes https://github.com/google/closure-compiler/issues/4009


mprobst:
Thanks for the PR!

I believe we've already fixed this in https://github.com/google/closure-compiler/commit/6da46fcc7fe64b74eeb87dc4a2493611cdf30f6a though, don't we?

PR #4012, opened 2022-11-21
Update AbortSignal and AbortController externs to match spec, mdn.

### LINKED ISSUE #4011: Is it a bug that the extern for AbortController doesn't allow a reason argument?
I tried [asking this on SO](https://stackoverflow.com/questions/74445883/can-i-update-the-extern-for-abortcontroller-in-the-google-closure-compiler) but didn't get any replies. It appears to me that a "reason" argument is allowed in the browser when calling `abort()` on a Fetch `AbortController`, but the externs built into Closure don't allow it.

Any fixes or workarounds?

niloc132:
Fixes #4011

### DIFF
--- a/externs/browser/w3c_abort.js
+++ b/externs/browser/w3c_abort.js
@@ -29,12 +29,29 @@
  */
 function AbortSignal() {}
 
-/** @type {boolean} */
+/** @const {boolean} */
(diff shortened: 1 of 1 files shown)


lvelden (approved):
lgtm

ndw:
Thank you!

PR #4019, opened 2022-12-17
webgpu externs

treblereel:
Final WebGPU specs are almost there, so, in case Google team doesn't have their internal webgpu externs, maybe we can use this as a starting point. Externs are based on specs from December 9th.


treblereel:
good news: https://bugs.chromium.org/p/chromium/issues/detail?id=1420483

PR #4052, opened 2023-01-25
Update File System and File System Access APIs to the latest spec

vobruba-martin:
Originally there was only one specification - File System Access API. Later it
was divided into 2 separate APIs and that's why I divided it into 2 separate
files.

On top of that I've updated both APIs to the latest specifications.

### DIFF
--- a/externs/browser/w3c_permissions.js
+++ b/externs/browser/w3c_permissions.js
@@ -23,10 +23,14 @@
 
 
 /**
- * @typedef {{name: PermissionName}}
+ * @record
+ * @struct
  * @see https://w3c.github.io/permissions/#permission-descriptor
  */
-var PermissionDescriptor;
+var PermissionDescriptor = function() {};
+
+/** @type {!PermissionName} */
+PermissionDescriptor.prototype.name;
 
 
 /**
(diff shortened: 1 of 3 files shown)


brad4d:
I'm afraid this is a very risky change, but I'll import it and test it to see what happens.

brad4d:
My first attempt at testing failed miserably, because I forgot to add the new file into our internal build setup.
I'm trying again now.

There's also a list of externs somewhere hard-coded into the java code that is used to enforce the standard externs get read in the right order. I suspect it may need updating, but it seems the 2 files that already existed aren't listed in it (i.e. searching across the compiler code for their names didn't find that java file), so maybe the new one doesn't need to be listed either. We'll see.

vobruba-martin:
Thank you for your time @brad4d

externs/browser/w3c_permissions.js
```diff
@@ -23,10 +23,14 @@
 
 
 /**
- * @typedef {{name: PermissionName}}
+ * @record
```
brad4d, line 26: Just FYI,

changing this to no longer be a typedef exposed the fact that `Permissions.prototype.query` is declared incorrectly below. It's argument should be of type `!PermissionDescriptor`.

This broke some code within Google.
I've already applied the fix and am testing again to see what other problems there are.

vobruba-martin: Thanks, I will apply the fix too.

brad4d:
I think I've dealt with all the breakages.
I'm sending this for review and will hopefully submit it this week.

brad4d:
This was merged by 17a20f5 but it looks like the commit comment didn't contain what it was supposed to in order to close this issue.

PR #4058, opened 2023-02-11
Don't create ON_FALSE control flow edges for infinite DO loops

CamWass:
This PR changes the handling of infinite do-while loops in the control flow analysis to match other infinite loops.

Currently, for `FOR` and `WHILE` loops, no `ON_FALSE` edges are created from the loop condition if it is a `true` boolean literal (i.e. infinite loops). This is done by first retrieveing the loop's condition and then checking that it is not a true literal.
For `WHILE` loops, `handleWhile` does:
```java
Node cond = node.getFirstChild();
if (!cond.isTrue()) {
    // Create ON_FALSE branch
}
```
`handleDo` also contains this exact code, but because the first child of a `DO` loop is its body, not its condition, the value of `cond` is actually a statement (the body) and so `cond.isTrue()` is always false and therefore an `ON_FALSE` branch is always created, even if the condition is a `true` literal.

This PR corrects `handleDo`'s behaviour, bringing it in line with the other methods.

### DIFF
--- a/src/com/google/javascript/jscomp/ControlFlowAnalysis.java
+++ b/src/com/google/javascript/jscomp/ControlFlowAnalysis.java
@@ -476,10 +476,11 @@ public final class ControlFlowAnalysis implements NodeTraversal.Callback {
   }
 
   private void handleDo(Node node) {
-    Node cond = node.getFirstChild();
+    Node body = node.getFirstChild();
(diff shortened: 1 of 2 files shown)


brad4d:
I've reviewed the change and convinced myself it is correct.
I'll import it and run internal tests to see if I can land it.

Thanks for finding and fixing this!

PR #4061, opened 2023-03-02
Update to latest tagged rules_jvm_external to avoid / in unshaded jar

### LINKED ISSUE #4060: Unshaded jar contains bogus "/" entry
The unshaded jar contains a bogus "/" directory entry. While this isn't a problem for using it, it does cause trouble with [DependencyCheck's](https://github.com/jeremylong/DependencyCheck) archive analyzer which fails with the following error when it runs into it:

    Archive contains a file (/) that would be extracted outside of the target directory.

niloc132:
Fixes #4060

### DIFF
--- a/WORKSPACE.bazel
+++ b/WORKSPACE.bazel
@@ -2,8 +2,8 @@ workspace(name = "com_google_javascript_jscomp")
 
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
-RULES_JVM_EXTERNAL_TAG = "4.2"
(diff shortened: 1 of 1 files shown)


blickly:
LGTM

blickly:
Sorry, I merged this one the wrong way which caused it to get reverted.  Do you mind re-opening this PR again so that I can submit it the right way and give you attribution for the change?

niloc132:
Done, thanks. Don't worry about the attribution on my account, it just looked like a straightforward fix.

PR #4069, opened 2023-03-14
Externs inner namespaces fix

0c3d7r4:
**Inner namespace consolidation.** 

📝 The write up: [The Bitter Taste Of Bazel In My Morning Java, Or How I Consolidated Google Namespaces ](https://www.typal.dev/blog-bitter-bazel.html). 

I understand you might not go with the `isFromEmptyObjLitExtern` as in the PR and choose to properly fix `if (isObjectLiteralThatCanBeSkipped(type)` instead, i.e., by maintaining `droppedPropertiesOfUnions` correctly. It's all in the blog post.

Please leave a comment to acknowledge the presence of the bug. I work hard for this.

### DIFF
--- a/src/com/google/javascript/rhino/jstype/JSType.java
+++ b/src/com/google/javascript/rhino/jstype/JSType.java
@@ -210,6 +210,10 @@ public abstract class JSType {
     return false;
   }
 
+  public boolean isFromEmptyObjLitExtern() {
+    return false;
+  }
+
   /** Whether this is the prototype of a function. */
   // TODO(sdh): consider renaming this to isPrototypeObject.
   public boolean isFunctionPrototypeType() {
--- a/src/com/google/javascript/rhino/jstype/JSTypeRegistry.java
+++ b/src/com/google/javascript/rhino/jstype/JSTypeRegistry.java
@@ -1110,7 +1110,7 @@ public final class JSTypeRegistry {
   // we don't need to store these properties in the propertyIndex separately.
   private static boolean isObjectLiteralThatCanBeSkipped(JSType t) {
     t = t.restrictByNotNullOrUndefined();
-    return t.isRecordType() || t.isLiteralObject();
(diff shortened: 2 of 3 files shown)


niloc132:
Maybe add tests for this? 

We've personally run into this especially when trying to patch closure's own externs from a downstream project - there are lots of cases where merging not only namespaces, but also fully unioning externs, which we've had trouble with (but maybe it has gotten better).

Finally, in lieu of a comments section in the blog, bazel itself advertises [a plugin for IntelliJ](https://ij.bazel.build/docs/bazel-plugin.html), which is not terrible, though does have some pretty substantial warts (for example, it effectively never works with a recent IJ build, and you sometimes need to exclude some targets from being indexed). Once installed and configured, as long as you're comfortable with the bazel command line, it is pretty decent though.

lauraharker:
Thanks doing all this investigation. I'll take a more thorough look after the weekend, but a few early comments:

1. as niloc132 suggested, could you add unit tests? That would help me review this change (and prevent future regressions of the sort you found in r20220104 😀 )
2. please attempt to sign the Google CLA and let me know if you have any problems. import this PR to google's internal repo for testing is blocked on a valid CLA.
3. I'll look closer at what's going on with `droppedPropertiesOfUnions` since as you suggest in the PR summary, I find having this very specific PrototypeObjectType method & also the fact that it's only used in the missing property checks to be suspicious.

lauraharker:
Ok, I definitely believe the right fix is to unify the behavior of `isObjectLiteralThatCanBeSkipped` and the code adding to droppedPropertiesOfUnions, so that we either stop skipping literal objects or start tracking their dropped properties. I'd expect that just tracking them in droppedPropertiesInUnions would be better for performance and bloat the size of the type registry.

Are you interested in making that change in this PR? Sounds great if you are, otherwise please hand this off via filing an issue.

dandov:
No response from the PR author.

PR #4075, opened 2023-03-25
Fix wicg_scheduler.js externs

vobruba-martin:
Due to commit 876761b33479c1cade1e37e648d6afc4693d9ba6 I'm no longer able to compile my project using `--language_in ECMASCRIPT5_STRICT` because compilation fails with `externs.zip//wicg_scheduler.js:56:50: ERROR - [JSC_LANGUAGE_FEATURE] This language feature is only supported for ECMASCRIPT_2015 mode or better: default parameter.` error.


frost-cy:
bug scrub: The change that caused this problem is now rolledback

PR #4093, opened 2023-05-25
Annotate Node.js readdirSync's last optional parameter

kripken:
Node.js docs for that API:

https://nodejs.org/api/fs.html#fsreaddirsyncpath-options


brad4d:
I've sent this for internal review and test. Sorry for the delay.

Issue #4104, opened 2023-07-31
License and origin details may be missing for Jarjar'red JARs and other bundled 3rd-party?


pombredanne:
The distributed JARs contain jarjar'red code in /com/google/javascript/jscomp/jarjar/ and I cannot find the corresponding copyright/notices/licenses for these.
The same applies to com/google/javascript/rhino

This seems to be caused by a combo of things:
- the https://github.com/google/closure-compiler/blob/master/rules.jarjar rules and jarjar seem to be ignoring any form of license, attribution or third-party origin documentation
- the definitions in the https://github.com/google/closure-compiler/blob/master/WORKSPACE.bazel do not seem to get correct details in the built JARs

As result, there is a double whammy:
- some license information is missing  from the deployed JAR resulting in possible license compliance issues.
- third-party origin information is mostly missing from the deployed JAR resulting in difficulties to track which packages are included and which if any is subject to known vulnerabilities, resulting in security risks

blickly:
Thanks for the report!

Do you know if the unshaded version (https://mvnrepository.com/artifact/com.google.javascript/closure-compiler-unshaded) fairs better, or if it also suffers from the same issues?

niloc132:
While the unshaded jar shouldn't have those problems for its _dependencies_ (since by definition, it isn't shaded, so each dependency has to be fetched separately, and should have been packaged properly), the closure-compiler-unshaded jar doesn't contain a LICENSE/NOTICE/COPYING file, nor does its pom have a `<license>` block (see https://search.maven.org/artifact/com.google.javascript/closure-compiler-unshaded/v20230502/jar). It does at least have a license for the pom.xml file itself, but putting it in a comment seems to imply that the license covers the pom file, rather than being declared for the associated jar (in a `<license>` block within the `<project>` itself).

I've proposed https://github.com/google/closure-compiler/pull/4105 as a draft to at least rectify the missing `<license>` tag, but adding a file to the jar would probably also be good to cover all bases, though I'll need a little time to remember how to dance with bazel to make that happen. 

If someone can point me in the right direction for jarjar, I can see about fixing that side as well?

niloc132:
[...]

PR #4105, opened 2023-08-02
Include a license tag for the unshaded jar's pom

niloc132:


gunan (approved):

pombredanne:
@niloc132 As explained in https://github.com/google/closure-compiler/issues/4104#issuecomment-1761940716 this license notice is a good start, but this is incomplete.

niloc132:
Definitely, these changes are not complete for the shaded distribution, but I take pains to avoid the shaded distribution - the task of making the unshaded distribution correct seems complete at least?

Issue #3451, opened 2019-08-13
Option to fail on warnings


l1bbcsg:
I would like to propose an option that could make the compiler fail if at least one warning is encountered, regardless of errors and exit with a non-zero exit code.

This could be a simple flags such as `--fail-on-warnings` or a more sophisticated setting like `--error-level` that could introduce more granular control, i.e. still exit with 0 if compilation failed with errors.

Use case for this is either a purist approach when no warnings are tolerated or code quality control CI job that verifies new code with `--checks-only`.

brad4d:
I believe this option already exists. The magic words are `--jscomp_error=*` to say "treat all diagnostic groups as errors".

l1bbcsg:
I don't think this is quite the right option. It seems to change behaviour much more elevating a lot of other warnings that are not normally reported to errors.

This example:

```js
/**
 * @type {number}
 */
let a = 10;

a = 'foo'
```

Produces a warning:

```sh
google-closure-compiler a.js --checks_only --warning_level verbose
```

```
a.js:6: WARNING - [JSC_TYPE_MISMATCH] assignment
found   : string
required: number
a = 'foo'
^^^^^^^^^

0 error(s), 1 warning(s), 100.0% typed
```

But with `--jscomp_error=*` there are now two errors:

```sh
google-closure-compiler a.js --checks_only --warning_level verbose --jscomp_error=*
```

```
a.js:6: ERROR - [JSC_MISSING_SEMICOLON] Missing semicolon
a = 'foo'
^^^^^^^^^

a.js:6: ERROR - [JSC_TYPE_MISMATCH] assignment
found   : string
required: number
a = 'foo'
^^^^^^^^^

2 error(s), 0 warning(s), 100.0% typed
```

On a large code base I encountered many more new errors and compiler actually crashed with `java.lang.RuntimeException: INTERNAL COMPILER ERROR`.

brad4d:
Created internal issue
http://b/139437890

brad4d:
This is looking harder to do than I'd wish, but I think the gist is that the proposed option would need to trigger doing `options.addWarningsGuard(new StrictWarningsGuard())` when the compiler options are created in `CommandLineRunner#createOptions()`.

kaovilai:
Bump. Really wanted to use it with GitHub Actions for CI. Currently all exits regardless of error/warning levels are zero exits. If anyone have a bash workaround that detects a non-zero length output and fail

PR #4124, opened 2023-09-14
Add '--fail_on_warnings', issue #3451

doublep:
This PR adds option `--fail_on_warnings` as requested in #3451 (and that I missed myself too). It is obviously not finished, but before adding documentations, tests, etc. I'd like to see some feedback if this option is fine in general.

Instead of meddling with existing warning guards as suggested by brad4d in the issue, I decided to keep a simple boolean flag and at the last possible point check it: if we still have a warning level by then, simply turn it into an error. I feel it is cleaner this way (and also easier, yeah), because configuration of warning guards (which is quite extensive, with all the different named categories) is completely independ from "fail on warnings" setting and you don't need to coordinate changes in them anywhere. In effect, the "fail on warnings" setting is simple "whatever I want as warnings usually, should become an error in this context" and is not dependent on how you set that "whatever" — be that the defaults or extensive custom configuration.

A possible improvement would be to pass to error reporter a boolean "error elevated from warning" so that it could report that if it so chooses.

I verified that it does produce the desired change:

    $ java -jar bazel-bin/compiler_uberjar_deploy.jar ~/test/test.js --checks_only --warning_level verbose || echo DOES NOT COMPILE
    /home/paul/test/test.js:3:0: WARNING - [JSC_TYPE_MISMATCH] assignment
    found   : string
    required: number
      3| a = 'foo'
         ^^^^^^^^^

    $ java -jar bazel-bin/compiler_uberjar_deploy.jar ~/test/test.js --checks_only --warning_level verbose --fail_on_warnings || echo DOES NOT COMPILE
    /home/paul/test/test.js:3:0: ERROR - [JSC_TYPE_MISMATCH] assignment
    found   : string
    required: number
      3| a = 'foo'
         ^^^^^^^^^

    1 error(s), 0 warning(s), 100.0% typed
    DOES NOT COMPILE

Where `~/test/test.js` is:

    /** @type {number} */
    let a = 10;
    a = 'foo'

### DIFF
--- a/src/com/google/javascript/jscomp/CommandLineRunner.java
+++ b/src/com/google/javascript/jscomp/CommandLineRunner.java
@@ -462,6 +462,12 @@ public class CommandLineRunner extends AbstractCommandLineRunner<Compiler, Compi
                 + " --parse_inline_source_maps.")
     private boolean applyInputSourceMaps = true;
 
+    @Option(
+        name = "--fail_on_warnings",
+        handler = BooleanOptionHandler.class,
+        usage = "Elevate all compiler warnings to errors")
+    private boolean failOnWarnings = false;
+
     // Used to define the flag, values are stored by the handler.
     @SuppressWarnings("unused")
     @Option(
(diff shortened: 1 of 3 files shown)


brad4d:
I've imported this change and have looked at it.
I see the reason for not doing what I suggested on the related issue.
Maybe this is OK, but I feel like this change should probably be in the error manager implementation code.

I'm going to ask @concavelenz to take a look also.

brad4d:
I'm afraid that this approach isn't going to fly.
It is simpler, true, but that's because it's trying to completely ignore the established mechanisms for controlling the reporting of errors and warnings.

Doing that seems likely to have unforeseen consequences.

I quite agree that the current mechanism is complicated, but it was developed to meet particular needs and shouldn't just be worked around.

I honestly do not have the time and energy to put into figuring out the details necessary to make the option desired here work. What I can do is tell you that the inside-google-only version of the compiler does have such an option and it accomplishes it by using `StrictWarningsGuard` as I suggested on #3451

Issue #4129, opened 2023-11-07
Assigning a variable while passing to a function fails AC


alexmalcoci:
Passing a variable as an arg to a function, while at the same time assigning the variable makes the advanced compilation think the two args are supposed to be equal, when they are not.

Below is a distilled example of a [real use case](https://github.com/thysultan/stylis/blob/7b0fac1005b74b2131e7e35f0ea67fe0eaf49451/src/Parser.js#L166):

<details>
  <summary>Advanced compilation example</summary>

```js
function substr (value, begin, end) {
  return value.slice(begin, end)
}

// assigning to window so it doesn't get optimized away
window.bug = function (s, i) {
  return substr(s, i, i = 5);
}
```
`java -jar closure-compiler-v20230802.jar --js in.js --formatting PRETTY_PRINT --compilation_level ADVANCED`
=>
```js
window.g = function(b, a) {
  var c = a = 5;
  return b.slice(a, c);
};
```
</details>

Notice how the `c` and `a` end up with 5, which makes the `slice` return an empty string.

Simple compilation works fine because it keeps the 2 variables separate:
<details>
  <summary>Simple compilation example</summary>

```js
function substr (value, begin, end) {
  return value.slice(begin, end)
}

// assigning to window so it doesn't get optimized away
window.bug = function (s, i) {
  return substr(s, i, i = 5);
}
```
`java -jar closure-compiler-v20230802.jar --js in.js --formatting PRETTY_PRINT --compilation_level SIMPLE`
=>
```js
function substr(a, b, c) {
  return a.slice(b, c);
}
window.bug = function(a, b) {
  return substr(a, b, 5);
};
```
</details>

Interestingly, advanced compilation without the extra function works fine too
<details>
  <summary>Advanced compilation example without extra function</summary>

```js
// assigning to window so it doesn't get optimized away
window.bug = function (s, i) {
  return s.slice(i, i = 5);
}
```
`java -jar closure-compiler-v20230802.jar --js in.js --formatting PRETTY_PRINT --compilation_level ADVANCED`
=>
```js
window.g = function(a, b) {
  return a.slice(b, 5);
};
```
</details>

lauraharker:
Drive-by comment: this might be the same as https://github.com/google/closure-compiler/issues/4115 but I haven't investigated further.

alexmalcoci:
[...]

Issue #4115, opened 2023-09-02
ADVANCED compilation mode produces code in wrong execution order


codedhead:
I found this bug when compiling a third-party library, below is a simplified version to reproduce the bug:

input.js
```
function substr(value, begin, end) { return value.slice(begin, end) }
window.testFunc = function (a, b, c) {
  return substr(a, b+1, b=c);
}
```

command
```
google-closure-compiler -O ADVANCED input.js --js_output_file output.js
```

output.js
```
window.testFunc=function(c,b,a){a=b=a;return c.slice(b+1,a)};
```

The compiler somehow decided to evaluate the third argument `b=c` before the second argument `b+1`, causing the generated code to output incorrect value
```
testFunc('abc',-1,3) // output: '', expected: 'abc'
```

rishipal:
We choose option 2 from the above comment as we don't do good enough side effect tracking to figure out if side-effectable state is actually updated by good side effects, so special casing specific code patterns seems complicated.

rishipal:
Option 2 requires adding / updating unit tests in the following files. I'm currently updating the expected code sections of unit tests to match the results with this change. 

//third_party/java_src/jscomp/javatests/com/google/javascript/jscomp:FunctionArgumentInjectorTest
//third_party/java_src/jscomp/javatests/com/google/javascript/jscomp:FunctionInjectorTest
//third_party/java_src/jscomp/javatests/com/google/javascript/jscomp:InlineFunctionsTest
//third_party/java_src/jscomp/javatests/com/google/javascript/jscomp:ProcessCommonJSModulesTest

Once the tests are updated, I will send out the change for internal code review and update again.

rishipal:
Updated the 4 JSCompiler tests. JSCompiler tests are all passing with the option 2 fix now. 
Started google3-wide presubmit testing for the fix to see what breaks.

rishipal:
Subset of tests (Sample tests) passed. Sent the fix (option 2) for internal code review. google3 presubmit tests still running.

rishipal:
I've submitted the fix internally and it should get auto-pushed to GitHub by copybara in a few ~hours.

rishipal:
This is fixed with bebf834e32d61a14baa608bc49b4081ce0b1ecbc.

PR #4131, opened 2023-11-14
Fix modifying vars passed as function params

alexmalcoci:
Fixes https://github.com/google/closure-compiler/issues/4129 where passing a variable as one param, then changing it as second param would equate the two refs.


niloc132:
I don't see an explicit test for it, but does this also resolve https://github.com/google/closure-compiler/issues/4115?

alexmalcoci:
That issue revealed that looking for assignments wasn't enough, we have to track all usages of vars, and if it happens to be assigned later on, inline all usages. I have added a few more tests and fixed one test related to that issue by adding an extra temp. 3edba2e should cover all use cases. Thanks @niloc132

alexmalcoci:
@rishipal Any updates on this? It's a bit of a blocker for us.

alexmalcoci:
I've uncovered another edge case where one of the args is a variable, and a subsequent arg is a function call that eventually modifies the value of the first variable. What happens is that the call arg is temped, but the first one is not, so the first one ends up with the updated value instead of the original. Below is a contrived example illustrating this (most of the complexity is to prevent the compiler from optimizing it into a form that no longer has the bug).

<details>
  <summary>Example of first param getting a wrong value</summary>

```js
let text = 'abc'
let index = 1

// This function is necessary. Inlining the slice call will result in both args being temped
function slice(begin, end) {
  return text.slice(begin, end);
}

function getEndIndex() {
  if (index)
    // This line is the only relevant line in this function -- the index is being modified.
    // Everything else is artificial complexity to prevent the function from being inlined
    return ++index;

  return getEndIndex();
}

window.bug = slice(index - 1, getEndIndex())
```
`java -jar closure-compiler-v20230802.jar --js in.js --formatting PRETTY_PRINT --compilation_level ADVANCED`
=>
```js
let a = 1;
function b() {
  return a ? ++a : b();
}
var c = window, d = b();
c.g = "abc".slice(a - 1, d);
```
</details>

Notice how `b()` gets invoked first which modifies he value of `a`, and only then it's used in the `.slice()` call. A simpler version of this example would be below, but then the function call is inlined and breaks the example.
<details>
  <summary>A simpler (but broken) version of the example above</summary>

```js
let index = 1

function getEndIndex() {
  return ++index;
}

window.bug = 'abc'.slice(index - 1, getEndIndex())
```
</details>

I propose a fix in 99548d5 where we keep track of all args that contain references to vars, then if subsequently we encounter a function call, unconditionally temp all tracked args before. Ideally only the affected args would be temped, but since there's no way to see what the function call does, we have to temp all previous args. As always the tests are green, and added a few additional tests.

rishipal:
PTAL at my comments on https://github.com/google/closure-compiler/issues/4115. 
I believe all 3 issues (this PR https://github.com/google/closure-compiler/pull/4131 and issues https://github.com/google/closure-compiler/issues/4129 and https://github.com/google/closure-compiler/issues/4115)  reference the same underlying problem in JSCompiler (being fixed in https://github.com/google/closure-compiler/issues/4115).

alexmalcoci:
Thanks, @rishipal, it does seem like your approach is similar to mine. Can you post the patch file with the fix here, or add my tests to your fix? I just want to make sure that your fix fixes all my use cases as well.

rishipal:
I've submitted the fix internally and it should get auto-pushed to GitHub by copybara in a few ~hours.

rishipal:
Sg, for future readers this got fixed with bebf834e32d61a14baa608bc49b4081ce0b1ecbc.

PR #4134, opened 2023-12-04
Add missing AbortSignal static methods

ecrider:
In browser externs added 3 missing static methods to AbortSignal type - defined in w3c_abort.js file. Missing statics are any(), timeout() and abort().

At this point the last 2 of those methods are living standard, any() still lacks support in Firefox and Safari, but it's practical and can be used care-free in the realm of Chromium-based browser extensions so I imagine it won't just disappear overnight.

Anyway I use those methods, can't drag my own build of the compiler to every project and I hate suppressing warnings, so here it is.

### DIFF
--- a/externs/browser/w3c_abort.js
+++ b/externs/browser/w3c_abort.js
@@ -29,6 +29,27 @@
  */
 function AbortSignal() {}
 
+/**
+ * @param {*=} reason
+ * @return {!AbortSignal}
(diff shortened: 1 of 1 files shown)


h-joo:
Hi, thank you for your contribution. I've tried testing this over google codebase, but there's some bug in the infrastructure that's preventing us from merging this right now, so I'll have to hold this PR until the other issue has been resolved, I doubt that the other issue will be resolved soon.

PR #4153, opened 2024-02-20
Specify type of digits in Number.toFixed

zbynek:
According to [the docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed) this argument has to be a number. Using the more specific type is needed for https://github.com/google/elemental2/issues/129

### DIFF
--- a/externs/es3.js
+++ b/externs/es3.js
@@ -1172,7 +1172,7 @@ Number.prototype.toExponential = function(opt_fractionDigits) {};
 
 /**
  * @this {Number|number}
- * @param {*=} opt_digits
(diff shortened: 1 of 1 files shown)


zbynek:
@h-joo thanks for merging my other PR, will you be able to check/merge this one too?

h-joo:
> @h-joo thanks for merging my other PR, will you be able to check/merge this one too?

First I want thank you for your contribution! I actually am happy with the current state of the PR, but this change broke some internal projects so I'm fixing them up before I can merge this in.

Issue #4097, opened 2023-06-07
Re-add Minification Mapping Input Suggestion


tommywilkinson:
Looks like this ability used to be in the compiler, but was [removed long ago](https://github.com/google/closure-compiler/commit/e17ddd9271874add9608c71dd9d257a5f6e6c085). A piece of the functionality was re-added for #433, but not the piece allowing for an input mapping to use as a starting point. 

It's a very common practice for website to use hashed file names to allow browsers to cache unchanged portions of scripts, especially as a part of a webpack bundle, which allows only certain portions of the bundle to need to get re-downloaded by users in the new version. 

However, if you're using Closure Compiler as a minifier (perhaps via the [webpack plugin](https://github.com/webpack-contrib/closure-webpack-plugin)), the instability of the minification will mean that all chunk contents will likely change every build, even if there were no real changes to the contents of some of the chunks. 

In an ideal world, we'd be able to give Closure the minification map of the previous build, which it could then use as a "suggestion" for the current mapping. If there's a big enough upside, it can choose to re-map any or all of the identifiers. This means that for cases where there's no upside to re-mapping, Closure could choose stability, perhaps allowing more of those chunks to have unchanged contents. 

The main pitch here is to make minification slightly less "chaotic" (in the sense that small changes anywhere within the scope can lead to big output differences everywhere) for cases where there is little-to-no upside.

L3P3:
Also good for any diffing use, including differential updates.

nhelfman:
To expand on the use case described - if one wishes to use Compress Dictionary Transport (https://github.com/WICG/compression-dictionary-transport) this is likely a must. Without it, the typical diff between builds will be too large to be useful.

I found that sometimes even small actual difference results in a size diff of over 40% due to renaming which is inconsistent between builds.

What would it take to get this prioritized?

niloc132:
It looks like the feature might still be present, just not plumbed to the command line?
https://github.com/google/closure-compiler/blob/558e96f386c378ff8ef9cbdc36a3ba266d008517/src/com/google/javascript/jscomp/AbstractCommandLineRunner.java#L2759C20-L2769

Have you tried generating the prop/variable maps and then reading them using this to confirm it doesn't work? I recognize that the commit linked at the top claims that it isn't intended for this purpose, but the tests seem to confirm that it works, and there are compiler passes that check for the presence of these input maps and make use of them to keep names consistent:
[...]

PR #4155, opened 2024-02-25
Add variable and input map command line arguments.

nhelfman:
The change re-enables existing functionality which has been previously disabled as part of change  e17ddd9271874add9608c71dd9d257a5f6e6c085.

The reason to re-enable it at this time is to enable minimal binary compression diff between to build versions.
Without using input maps, 2 compilation builds, even when have small changes between them may result with completely different minified file due to renames which occur at different order and pick different rename value. When using input rename maps the resulting minified code remains very similar to minified code from earlier build version.

This is important for adopting Compression Dictionary Transport which greatly benefit from small compressed dictionary diffs.

The change addresses the ask in #4097


brad4d:
I've imported this PR and sent it for internal review and testing.

PR #4156, opened 2024-02-29
Definitions for W3C's Web Locks API - browser externs

ecrider:
Added browser externs for Web Locks API - based on definitions from latest W3C draft
[](https://www.w3.org/TR/2023/WD-web-locks-20230105/). The coverage is already great and API synergy with web workers cannot be overstated. Feel free to poke holes in it.

### DIFF
--- a/externs/browser/w3c_web_locks.js
+++ b/externs/browser/w3c_web_locks.js
@@ -32,10 +32,10 @@ var LockManager = function() {};
 
 
 /**
- * @param  {string}                           name
+ * @param  {string}                            name
  * @param  {!LockOptions|!LockGrantedCallback} options
- * @param  {!LockGrantedCallback|undefined=}  callback
- * @return {!Promise<void>}
+ * @param  {!LockGrantedCallback|undefined=}   callback
+ * @return {!Promise<*>}
  */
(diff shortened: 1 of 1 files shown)


lauraharker (changes requested):
Thanks for the contribution!

brad4d:
I've just discovered that we actually already have an externs file for these declarations, but for some unknown reason it isn't currently being exported to GitHub.

I think this is a historical accident, so I'm going to try to get it exported to GitHub instead of accepting this PR.

ecrider:
> I've just discovered that we actually already have an externs file for these declarations, but for some unknown reason it isn't currently being exported to GitHub.
> 
> I think this is a historical accident, so I'm going to try to get it exported to GitHub instead of accepting this PR.

Sure, whatever helps. I just want those definitions to end up in my production pipeline without forking official repository - whatever it will be by my contribution or someone else's doesn't matter to me much.

brad4d:
OK, https://github.com/google/closure-compiler/blob/master/externs/browser/weblocksapi.js now exists, so I'll close this PR.

@ecrider, thanks for showing us that this was missing.

Issue #890, opened 2015-03-25
@typedef and @template doen't work together


agrafix:
The following code does not work:

    /**
     * @template S, T
     * @typedef {{
     *             type: string,
     *             humanName: string,
     *             keyboard: TriResourceRef,
     *             serverElems: Array.<S>,
     *             getSrvId: function(S): string,
     *             getTitle: function(S): string,
     *             getDescriptions: function(S): Array.<string>,
     *             getUseForGrouping: function(S): boolean,
     *             getDeleted: function(S): boolean,
     *             getCliId: function(T): string,
     *             convertToSrv: function(T): S,
     *             convertToCli: function(S): T,
     *             makeNewCli: function(string, Date): T,
     *             mergeSrvCli: function(S, T): S,
     *             getUpdateId: function(T): string,
     *             isSent: function(T): boolean,
     *             insertDescriptionIfMissing: function(S): S
     *          }}
     */
    var DrgInterf;

Using the recent compiler version, it results this:

    /var/folders/m0/8vx31n6j775g80_0pqhh47k40000gn/T/tmpbDmD3e:77568: ERROR - Bad type annotation. Unknown type S
     *             serverElems: Array.<S>,
                                       ^

concavelenz:
we should warn about a misplaced annotation with the use of @template described here.  The @record proposal might allow this.

silverlyra:
Is this ever going to be supported? We've got:

```js
/**
 * @param {!function(F, null=, (Object<string,*>|null)=): !goog.soy.data.SanitizedContent} template
 * @template F
 */
```

pasted all over our codebase because it's impossible to have a parameterized `@typedef` 😒

dimvar:
maybe, but it's not high priority so it might be a while

shicks:
I think TypeScript allows adding some sort of "call" property to their interfaces.  That might be one way to handle this, but allowing to define an `@record` (which can, of course, be templated) with a typed `call` method.  In any case, if any approach is taken, it's not going to happen until after NTI is fully rolled out.

mbrukman:
I'm also running into this issue. I'm trying to translate the [TypeScript type annotations in Preact](https://github.com/preactjs/preact/blob/10.4.6/src/index.d.ts) to Closure Compiler-compatible JsDoc (to be able to use Preact from a project which is compiled with Closure Compiler).

However, I cannot translate some of the types completely due to limitations with Closure Compiler support for `@typedef` and `@template`.

[...]

PR #4173, opened 2024-06-22
Typedefs with templates

0c3d7r4:
Allows to add templates to typedefs, like

```js
/**
 * @typedef {Array<T>}
 * @template T
 */
var TypedList
```

Address https://github.com/google/closure-compiler/issues/890

### DIFF
--- a/src/com/google/javascript/jscomp/CheckJSDoc.java
+++ b/src/com/google/javascript/jscomp/CheckJSDoc.java
@@ -254,6 +254,7 @@ final class CheckJSDoc extends AbstractPostOrderCallback implements CompilerPass
         && !info.getTemplateTypeNames().isEmpty()
         && !info.isConstructorOrInterface()
         && !isClassDecl(n)
+        && info.getTypedefType() == null
         && !info.containsFunctionDeclaration()
         && getFunctionDecl(n) == null) {
       reportMisplaced(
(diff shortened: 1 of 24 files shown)


shicks:
In general, we're not really investing in the Closure type system at this point, so this is unlikely to land.  You're certainly welcome to use it in your own fork, but our general recommendation is to defer to TypeScript for sophisticated type system support.

That said, this sort of thing is already supported pretty well with `@record` types, which can indeed be templatized:

```js
/**
 * @record
 * @template T
 */
function Record() {
  /** @type {function(): T} */
  this.run;
}
```

should already behave the same way.

0c3d7r4:
@shicks yeah i've figured since, you could do it via fields on records... but it doesn't cancel the fact that ideally typedefs _could_ have templates also as there are issues like #890 (which you should close as "won't fix" now) which shows that people expect that functionality from a compiler. So if your policy is to not have a product of exceptional quality where _everything_ works and to set example to the industry, but to have a substandard tool that works to the bare minimum and defer everyone to TS for something as basic as templates for typedefs, then I can't really blame you, I just thought Google wasn't not that poor to afford to **invest** in experts who wanted to work on professional software and not some sort of MVP.

Also, for a lot of people, TS is _too_ sophisticated. For many people the best things are the older things @niloc132 and JS was our thing now even that's being implicitly taken away with TS, all according to MS's plan to wipe out ye olde simple plain JS as language.

Screenshot 2024-09-20 at 11 12 35

Why even have typedefs if there are structural interfaces? Why even bother with Closure Compiler when there's TypeScript right? I guess the fate similar to Closure Library is awaiting Closure Compiler. So when the next round of tech-layoffs ensues, don't blame me for not being able to justify why there needs to be a dedicated compiler team at Google 😇 just saying, no hard feelings, i have a fork anyway i just thought i'd help to improve the project i loved.

PR #4180, opened 2024-07-17
Support @closurePrimitive for reflect and object primitives.

ribrdb:


concavelenz:
What is the motivation for this change?

ribrdb:
The main motivation is basically what Steve Hicks said several years ago on
the closure-library-discuss list:
>
> The Closure dependency management system (specifically - goog.provide,
> goog.require, and goog.module), while revolutionary when it was first
> introduced, is now a bit of a dead end, and the best bet is to move away
> from it whenever possible.


Although Google has deprecated the Closure Library, we still need to use it
until we're able to migrate fully away from it. We still want to use
Closure Compiler for our production, advanced mode builds. But we also need
a maintainable story for development.

The closure library debug loader works, but is unmaintained, awkward to
use, and makes debugging difficult because sourcemaps don't work.
The closure compiler readme says:

> Compilation modes other than `ADVANCED` were always an afterthought and we
> have deprecated those modes. We believe that other tools perform comparably
> for non-`ADVANCED` modes and are better integrated into the broader JS
> ecosystem.


We would like to be able to use tools from the broader JS ecosystem for
development, but those tools do not work with the closure dependency
management system. However, with Closure Library converted to ES6 we can
load closure .js files directly into the browser without any bundling. Or,
when necessary, we can use standard tools that understand es modules like
esbuild or jest.

ribrdb:
Oh, I didn't realize this was a pr comment. There was also this discussion on the closure compiler mailing list:
https://groups.google.com/g/closure-compiler-discuss/c/jW5BUa6WV1M

shicks:
I don't think we want to do this.  `@closurePrimitive` was indeed intended to increase the separation betwen compiler and library, but it didn't turn out to work as effectively as we'd hoped.  In particular, it relies way too much on the traditional Closure type system, which isn't always available for all primitives (e.g. the dependency management primitives, among others), and also doesn't work for the newer optimization-only "color" system that we're moving towards now that TypeScript provides much better developer-facing type checking.

In terms of the actual primitives being added here, I don't believe any of them are relevant.  In particular, the `goog.object` methods are long-obsolete in favor of just using object literals with computed properties.  They should probably have been deprecated years ago.  `goog.reflect.cache` is also obsoleted by `@pureOrBreakMyCode`, which can now be used instead of special-purpose caching primitive.  Finally, `goog.reflect.object` and `goog.reflect.objectProperty` are included in the new trimmed-down `base.js` that's been moved into the compiler repo.  This file is trivial and will compile to zero with advanced optimizations enabled, so it should be treated as effectively "the compiler's runtime library".  The benefit of allowing their use from user-written functions is significantly less than the total complexity cost.

ribrdb:
I don't actually intend to use goog.object, or goog.reflect.cache in code I write, but we are using closure library code that depends on those.
Also I still don't see how goog.reflect.objectProperty of goog.reflect.object can be used from an es6 module.

PR #4185, opened 2024-08-11
Replace ! in filenames when converting to JS identifiers

joshuamiller:
When using Closure to emit ES5 code for ClojureScript, the source filename of the input code can be in a jar, such as `file:/Users/myuser/.m2/repository/org/clojure/google-closure-library/0.0-20230227-c7c0a541/google-closure-library-0.0-20230227-c7c0a541.jar!goog/log/log.js`. The `!` is not replaced when converting this filename to a JS identifier, resulting in the following input/output pair:

```
goog.log.Level = class Level { constructor() ... }
```

```
file_$Users$myuser$_m2$repository$org$clojure$google_closure_library$0_0_20201211_3e6c510d$google_closure_library_0_0_20201211_3e6c510d_jar!$goog$log$log$classdecl$var0 = function(name, value) { ...
```

Note that the variable name has `!` in it, which does not parse.

This PR adds `!` to the list of special characters to replace so that jar-sourced input files can produce legal ES5 output.

### DIFF
--- a/src/com/google/javascript/jscomp/deps/ModuleNames.java
+++ b/src/com/google/javascript/jscomp/deps/ModuleNames.java
@@ -62,6 +62,7 @@ public class ModuleNames {
         .replace('\\', '$')
         .replace('@', '$')
         .replace('+', '$')
+        .replace('!', '$')
         .replace('-', '_')
         .replace(':', '_')
         .replace('.', '_')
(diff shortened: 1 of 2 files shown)


mhausner:
Chinthoorie, can you please accept this PR?

PR #4206, opened 2024-12-20
Unset ANDROID_HOME in workflow to fix Android SDK loading error

lauraharker:
Attempt to fix "ERROR: Analysis of target '//:test/com/google/javascript/jscomp/disambiguate/AmbiguatePropertiesTest' failed; build aborted: Android SDK api level 32 was requested but it is not installed in the Android SDK at /usr/local/lib/android/sdk. The api levels found were [35, 34]. Please choose an available api level or install api level 32 from the Android SDK Manager." - https://github.com/google/closure-compiler/actions/runs/12412367428/job/34652003730

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -57,7 +57,7 @@ jobs:
       - uses: bazelbuild/setup-bazelisk@v2
 
       - name: Build and Test
-        run: bazelisk test //:all
+        run: unset ANDROID_HOME && bazelisk test //:all
 
       - name: Unsymlink Bazel Artifacts
         # upload-artifact doesn't support paths with symlinks


lauraharker:
For some additional context -

unsetting `ANDROID_HOME` stops some Bazel code from trying to load the Android SDK.

It seems like something changed in our GitHub CI setup, where previously ANDROID_HOME was unset & now it's set?

There might be a better fix but 1) we already did this in the past for a different build script - https://github.com/google/closure-compiler/commit/f1faba688ddcddacdcfa3df370eb40bb6848afea (for Google's internal CI) and 2) we definitely don't need the Android SDK.

lauraharker:
Note: the Build & Test for this PR did fail, but it failed later on than the current CI runs are doing.

The problem it's hitting now is that we added a Java 21 API usage to the compiler & our Bazel Java rules aren't setup to use Java 21 yet.

But without this PR, we don't even get to the point of seeing the Java 21 error.

brad4d:
I'm going to import this and get it submitted.

PR #4211, opened 2025-01-08
Add direct links to spec definitions for each property

ecrider:
Updated spec links annotations for consistency with extern lib format and some backwards compatibility with IDE plugins fetching JSDoc referenced resources.

Also https://wicg.github.io/web-locks/ is a http-equiv redirect to canonical https://w3c.github.io/web-locks/, returns a body and status code 200 so most crawlers just stop right there and make attempt to parse it as a final destination.


lauraharker:
Thanks! Sending this for internal Google review.

PR #4213, opened 2025-01-20
Add missing extern composition methods for global Set object

ecrider:


dloyda:
An update on this, we'll be adding the polyfills for this before we can add the externs

PR #4221, opened 2025-02-21
Add SUSPICIOUS_BREAKING_OUT_OF_OPTIONAL_CHAIN to SUSPICIOUS_CODE DiagnosticGroup

jnowjack-lucidchart:
Discussion on the Closure Compiler Discuss Group: https://groups.google.com/g/closure-compiler-discuss/c/ZpsidQkqdjk

Reasoning for change: We have set `--jscomp_error suspiciousCode` in our codebase and noticed that we were getting warnings instead of errors for `SUSPICIOUS_BREAKING_OUT_OF_OPTIONAL_CHAIN`. We would like this to be an error instead of a warning in line with the other `DiagnosticTypes` defined in [CheckSuspiciousCode](https://github.com/google/closure-compiler/blob/c2af98e3bd9b1f1c333b7019597eafca3d7018ad/src/com/google/javascript/jscomp/CheckSuspiciousCode.java)

### DIFF
--- a/src/com/google/javascript/jscomp/DiagnosticGroups.java
+++ b/src/com/google/javascript/jscomp/DiagnosticGroups.java
@@ -492,6 +492,7 @@ public class DiagnosticGroups {
           "suspiciousCode",
           CheckDuplicateCase.DUPLICATE_CASE,
           CheckSuspiciousCode.SUSPICIOUS_SEMICOLON,
+          CheckSuspiciousCode.SUSPICIOUS_BREAKING_OUT_OF_OPTIONAL_CHAIN,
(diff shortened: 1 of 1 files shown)


brad4d:
It's worth trying to do this despite my comments on  https://groups.google.com/g/closure-compiler-discuss/c/ZpsidQkqdjk

We do suppress "suspiciousCode" warnings in internal JS code that was generated from TS, so this may actually be OK.

brad4d:
Good news! The test results look good, so I think I'll be able to submit the internal change, which will end up closing this PR.

jnowjack-lucidchart:
Awesome, thanks Bradford!

PR #4222, opened 2025-02-21
Create new DiagnosticGroup for SUSPICIOUS_BREAKING_OUT_OF_OPTIONAL_CHAIN

jnowjack-lucidchart:
Discussion on the Closure Compiler Discuss Group: https://groups.google.com/g/closure-compiler-discuss/c/ZpsidQkqdjk

### Reasoning for change: 
We have set `--jscomp_error suspiciousCode` in our codebase and noticed that we were getting warnings instead of errors for `SUSPICIOUS_BREAKING_OUT_OF_OPTIONAL_CHAIN`. We would like this to be an error instead of a warning in line with the other `DiagnosticType`s defined in [CheckSuspiciousCode](https://github.com/google/closure-compiler/blob/c2af98e3bd9b1f1c333b7019597eafca3d7018ad/src/com/google/javascript/jscomp/CheckSuspiciousCode.java).  

There are concerns about adding this check to the existing `suspiciousCode` group  (see the [discussion group thread](https://groups.google.com/g/closure-compiler-discuss/c/ZpsidQkqdjk/m/rhJuKWTkAQAJ)) so I've created a new `DiagnosticGroup` with just `SUSPICIOUS_BREAKING_OUT_OF_OPTIONAL_CHAIN` so that it can be configured to be an error.

### DIFF
--- a/debian/closure-compiler.1.txt
+++ b/debian/closure-compiler.1.txt
@@ -359,6 +359,7 @@ require as a parameter a `WARNCLASS` warning class name. The following
 names are valid warning class names:
 
 * ``accessControls``
+* ``breakingOptionalChain``
 * ``checkRegExp``
 * ``checkTypes``
 * ``checkVars``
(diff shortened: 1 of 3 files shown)


brad4d:
This is no longer needed, because #4221 was merged.

PR #4227, opened 2025-03-20
use `pkg_zip` to replace `genrule + zip` 

jzhan-canva:
as per title, use `pkg_zip` to bundle externs.zip


lauraharker:
Thanks, @jzhan-canva

PR #4234, opened 2025-05-06
Add KeyboardEvent.prototype.isComposing, correct name for InputEvent.isComposing

zbynek:
MDN:
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/isComposing
https://developer.mozilla.org/en-US/docs/Web/API/InputEvent/isComposing
I believe isComposed was simply a typo and such property never existed (it's unrelated to the existing `composed` property of `Event`)

### DIFF
--- a/externs/browser/w3c_event.js
+++ b/externs/browser/w3c_event.js
@@ -566,7 +566,7 @@ function InputEvent(type, opt_eventInitDict) {}
 InputEvent.prototype.data;
 
 /** @type {boolean} */
-InputEvent.prototype.isComposed;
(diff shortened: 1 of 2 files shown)


brad4d:
After checking the relevant specification and corresponding TS `.d.ts` declaration, I'm inclined to think you are correct.

I will pull in this PR and try to submit it.

brad4d:
Testing is taking a long time for annoying reasons unrelated to the content of this change.

brad4d:
Internal testing looks good. Sent for review.

PR #4247, opened 2025-07-02
Add support for Picture-in-Picture API

vobruba-martin:
See https://w3c.github.io/picture-in-picture/
See https://caniuse.com/picture-in-picture


mhausner:
Note: we have externs for this API but for some reason they were not included in the public release. I added them, they should show up in the next day or so.

mhausner:
externs/browser/wicg_picture_in_picture.js is now included.

PR #4284, opened 2026-01-02
Add missing Uint8Array methods for base64 and hex strings

niloc132:


externs/es6.js
```diff
@@ -844,6 +910,46 @@ Uint8Array.from = function(source, mapFn, thisArg) {};
  */
 Uint8Array.of = function(var_args) {};
 
+/**
+ * Creates a new Uint8Array from a base64 encoded string.
+ * @param {!string} string a base64 string encoded array
+ * @param {Uint8ArrayBase64Options=} options an object specifying how to read the base64 string
+ * @return {!Uint8Array} a newly created array with the specified bytes
```
lauraharker: Please remove the explicit `!` nullability from {!string}, and add explicit nullability with `?` to `Uint8ArrayBase64Options`

externs/es6.js
```diff
@@ -808,6 +808,72 @@ Int8Array.from = function(source, mapFn, thisArg) {};
 Int8Array.of = function(var_args) {};
 
 
+/**
+ * Options to use when decoding base64 into a Uint8Array.
```
lauraharker: Optionally - consider linking to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array/fromBase64#options & the other options descriptions for the other records

lauraharker (changes requested):
Thanks! This looks great overall. I've tried pulling it into the Google repo & all our tests pass.

Just have a few small stylistic/documentation nits.

niloc132:
Thank you for the review, I'll try to turn this around promptly.

lauraharker:
Thanks, I'll pull this in to our internal repo for review

PR #4287, opened 2026-01-14
Modify comment on permissions for private repos

satashi-labs:
Updated comment regarding permissions for private repositories.


12wrigja:
This comment change does not seem to be accurate. Closing.

Issue #4229, opened 2025-03-21
"A terminally deprecated method in sun.misc.Unsafe has been called"


hills:
With the latest commits, and today's download from Oracle, the following error is omitted towards the end of JavaScript compilation:

```
WARNING: A terminally deprecated method in sun.misc.Unsafe has been called
WARNING: sun.misc.Unsafe::arrayBaseOffset has been called by com.google.common.hash.LittleEndianByteArray$UnsafeByteArray (file:/scratch/mark/cache/bazel/_bazel_mark/d465ce32b1ece8e176cd9a6c173f6b09/execroot/_main/bazel-out/k8-fastbuild/bin/compiler_uberjar_deploy.jar)
WARNING: Please consider reporting this to the maintainers of class com.google.common.hash.LittleEndianByteArray$UnsafeByteArray
WARNING: sun.misc.Unsafe::arrayBaseOffset will be removed in a future release
```

```
closure-compiler$ git describe --long
v20250317-0-g455c10971
```

```
$ java --version
java 24 2025-03-18
Java(TM) SE Runtime Environment (build 24+36-3646)
Java HotSpot(TM) 64-Bit Server VM (build 24+36-3646, mixed mode, sharing)
```

rishipal:
What project are you trying to compile? How do I reproduce this?

hills:
To reproduce, just install the latest java from Oracle ([downloads](https://www.oracle.com/java/technologies/downloads/))

It displays the warning with any JavaScript input, for example a blank input:

```
$ echo '' | /path/to/java -jar /path/to/closure-compiler-v20240317.jar
The compiler is waiting for input via stdin.
WARNING: A terminally deprecated method in sun.misc.Unsafe has been called
WARNING: sun.misc.Unsafe::arrayBaseOffset has been called by com.google.javascript.jscomp.jarjar.com.google.protobuf.UnsafeUtil$MemoryAccessor (file:/opt/google/closure-compiler/closure-compiler-v20240317.jar)
WARNING: Please consider reporting this to the maintainers of class com.google.javascript.jscomp.jarjar.com.google.protobuf.UnsafeUtil$MemoryAccessor
WARNING: sun.misc.Unsafe::arrayBaseOffset will be removed in a future release
```

It's also reproducible with the latest master of Closure Compiler.

niloc132:
[...]

PR #4289, opened 2026-01-18
Upgrade to latest Guava

MatrixFrog:
This should fix https://github.com/google/closure-compiler/issues/4229 or at least one instance of it


MatrixFrog:
By the way it doesn't make much difference to me if you want to merge this or not, I was mainly trying to find a relatively straightforward task I could to try out learning [jj](https://www.jj-vcs.dev/)

PR #4290, opened 2026-01-18
Fix PolymerPass crash (#1950)

MatrixFrog:
It's been a while so I don't remember a lot of detail about how Polymer works, but this seems like a pretty straightforward and low-risk fix for #1950


MatrixFrog:
This was another excuse to try out `jj` so again, feel free to skip merging this, but I think it should be pretty safe.

concavelenz:
> This was another excuse to try out `jj` so again, feel free to skip merging this, but I think it should be pretty safe.

Good to hear from you. :-)

PR #4308, opened 2026-03-19
Mark (bigint|number).toString(radix) nosideeffects (fixes #4169)

### LINKED ISSUE #4169: `Number.toString(base)` does not have `@nosideeffects` semantics
It is mentioned in various places that `.toString()` method has a hardcoded `@nosideeffects` property.  (e.g., [Angular nosideeffect annotation)](https://github.com/angular/angular/blob/67b2c336bc0bdce3f7ae054c094990a9831f5b20/packages/core/src/util/closure.ts#L9-L20
)) (On a side note, is this hack still relevant after the @nosideeffects being allowed in non-externs code?)

However the same is not true for `Number.toString(base)` variant, which writes the number in bases other than base 10.

```javascript
// a.js
for (let i = 0; i < 10; ++i)
  i.toString();
for (let j = 0; j < 10; ++j)
  j.toString(16);
```

```shell
google-closure-compiler -O ADVANCED a.js

a.js:2:2: WARNING - [JSC_USELESS_CODE] Suspicious code. This code lacks side-effects. Is there a bug?
  2|   i.toString();
       ^^^^^^^^^^^^

0 error(s), 1 warning(s), 94.7% typed
```

The first one is detected as suspicious code, the second one isn't.

The relevant  [externs definition](https://github.com/google/closure-compiler/blob/573aad7b4f1f7f9320b8d52a0ec2831b2b468ea7/externs/es3.js#L1191-L1200) appears correct, however it looks like there some special handling of `.toString()` in the java code.

This is important because in many fields (crypto, verifiable computation) some constants need to be generated (instead of being written out as literals). If GCC can eliminate the ones that aren't used, it will result in significant reductions. However when the generation code uses `toString(16)` elimination does not happen.

mnsaglam:
- Object.toString() was marked nosideeffects in AstAnalyzer through a special code path.
 - We extend this to (number|bigint).toString(radix) as well.

### DIFF
--- a/src/com/google/javascript/jscomp/AstAnalyzer.java
+++ b/src/com/google/javascript/jscomp/AstAnalyzer.java
@@ -66,8 +66,6 @@ public class AstAnalyzer {
   private static final ImmutableSet<String> BUILTIN_FUNCTIONS_WITHOUT_SIDEEFFECTS =
       ImmutableSet.of(
           "Object", "Array", "String", "Number", "BigInt", "Boolean", "RegExp", "Error");
-  private static final ImmutableSet<String> OBJECT_METHODS_WITHOUT_SIDEEFFECTS =
-      ImmutableSet.of("toString", "valueOf");
   private static final ImmutableSet<String> REGEXP_METHODS = ImmutableSet.of("test", "exec");
   private static final ImmutableSet<String> STRING_REGEXP_METHODS =
       ImmutableSet.of("match", "replace", "search", "split");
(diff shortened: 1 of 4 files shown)


mnsaglam:
@brad4d Does this look about right?

brad4d:
@KimlikDAO-bot I think this is OK, but I'm having trouble importing it due to a completely unrelated problem that it's not worth explaining. Could you please rebase this PR to the current head? That should resolve the problem for me.

mnsaglam:
@brad4d  It's synced to the head now. Let me know if the direction is right, if so, happy to add more tests.

brad4d:
Thanks, that resolved the problem.
I've imported the PR and will start testing and review.

brad4d (changes requested):

brad4d:
I've pulled in the latest version of this PR for testing and review.

FYI, I'll be on vacation April 15 - 27.
It is unlikely that the testing will complete fast enough for me to submit this before I leave.

brad4d:
I'm back from vacation.
There were failures in the testing, but they may well be flakes.
I'm re-running testing now for the failed tests.

brad4d:
So, this change uncovered a bug! Now that `.toString(16)` is considered side-effect-free, closure-compiler can see that a statement is doing nothing and complain about it.

After I fix that I'll be able to submit this change internally, which will be reflected here.

PR #4314, opened 2026-04-23
Only run the release workflow when piper-cut-cl.json changes

lauraharker:
This should remove no-op workflow runs on other
commits, like https://github.com/google/closure-compiler/actions/runs/24802669504, while still running on "Bump version for compiler release" commits like https://github.com/google/closure-compiler/actions/runs/24790507715.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -17,8 +17,7 @@ permissions:
   contents: read
 
 # This workflow is triggered by pushes to the `master` branch that modify '.github/piper-cut-cl.json'.
-# It first checks if `.github/piper-cut-cl.json` has been modified in the latest push.
-# If changes are detected, indicating a new release cut, the workflow proceeds to:
+# When `piper-cut-cl.json` changes, indicating a new release cut, the workflow proceeds to:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -16,16 +16,17 @@ name: Release Project
 permissions:
   contents: read
 
-# This workflow is triggered by pushes to the `master` branch.
+# This workflow is triggered by pushes to the `master` branch that modify '.github/piper-cut-cl.json'.
 # It first checks if `.github/piper-cut-cl.json` has been modified in the latest push.
```
hybrist: I assume we're leaving the check as a ~defense in depth for non-material changes to the file..?

hybrist (approved):

.github/workflows/release.yaml
lauraharker: That was my intent, but now that I look at the check, it's not smart enough to distinguish material vs. non-material changes anyway

It just diffs the entire file contents before and after.

I'd imagine we could try and make it smarter but not sure it's worth it, given that I don't think we've seen any problems historically. I just removed the check altogether.

Thanks for pointing this out.

PR #4330, opened 2026-08-04
Don't rename a function expression's name to match a parameter shadowing it

jzhan-canva:
Older versions of Safari treat a function expression's name as one of its parameters, so they reject a function whose non-simple parameter list binds that same name:

```js
var foo = function bar({bar = baz}) { return bar; };
// output: var b = function a({bar: a = c}) { return a; };
// Safari: SyntaxError: Duplicate parameter 'a' not allowed in function with destructuring parameters
```

See https://bugs.webkit.org/show_bug.cgi?id=220517 and https://bugs.webkit.org/show_bug.cgi?id=247433.

The function name and the parameter are separate bindings, so renaming them apart is free — the parameter shadows the function name throughout the function, making the name unreferenceable.

This gives the function name its own `Assignment` when a parameter shadows it and the parameter list is non-simple. `assignNames` reserves every name it generates, so a distinct key is enough to guarantee a distinct name. Simple parameter lists are left alone, since `function a(a) {}` doesn't trigger the bug and sharing the name is cheaper.

`recordPseudoName` needed the same distinction, otherwise `--debug` output kept the collision.

Tested: `bazel test //...` — 431/431 targets pass.

### DIFF
--- a/src/com/google/javascript/jscomp/RenameVars.java
+++ b/src/com/google/javascript/jscomp/RenameVars.java
@@ -132,6 +132,12 @@ final class RenameVars implements CompilerPass {
   /** A prefix to distinguish temporary local names from global names */
   private static final String LOCAL_VAR_PREFIX = "L ";
 
+  /**
+   * A suffix distinguishing the temporary name of a function expression's name from the temporary
+   * name of the parameter shadowing it. See {@link #isFunctionNameShadowedByParameter}.
+   */
+  private static final String FUNCTION_NAME_SUFFIX = " fn";
+
   // Shared name generator
   private final NameGenerator nameGenerator;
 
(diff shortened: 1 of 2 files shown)


jzhan-canva:
I have signed CLA but claude hasn't?
image

trevorade:
For clarity, does this problem only occur if the input JS already exhibits the shadowing issue or does JS Compiler sometimes introduce the shadowing issue when it wasn't present before?

Also, in what version of Safari was this issue fixed?

jzhan-canva:
@trevorade 

CC won't introduce it — only when the input already has the shadowing.

Worth stressing that the construct is perfectly legal JavaScript; the function name and the parameter are separate bindings and every other engine handles it. It's only WebKit that rejects it. We want to rely on closure compiler to keep it out of our production output.

On the Safari version: the JSC fix landed in trunk as [256478@main](https://github.com/WebKit/WebKit/pull/6257) on 2022-11-08. It can be reproduced in safari <16.3

PR #4331, opened 2026-08-11
Fix: Default resolveSourceMapAnnotations to false to prevent arbitrary file read (#4322)

### LINKED ISSUE #4322: Arbitrary file read via //# sourceMappingURL= path traversal in SourceMapResolver
## Summary
 
`closure-compiler` reads arbitrary files from the local filesystem when compiling JavaScript source that contains a crafted `//# sourceMappingURL=` comment with a relative path-traversal sequence. No bounds check is applied to keep the resolved path within the expected directory.
 
This was reported to the Google OSS VRP (bughunters.google.com). The team closed it as **Infeasible** (does not meet OSS VRP's bar, since it requires closure-compiler to already be compiling untrusted/third-party source) and explicitly authorized public disclosure here. Filing per that guidance so the pattern is documented and users who compile untrusted input can mitigate it.
 
This is the same bug class as [CVE-2026-49356 / GHSA-4x5r-pxfx-6jf8](https://github.com/advisories/GHSA-4x5r-pxfx-6jf8) in `@babel/core` (fixed June 13, 2026) — an inline `sourceMappingURL` comment used to read arbitrary files via unbounded relative path resolution. closure-compiler has an independently vulnerable implementation of the same pattern.
 
## Root cause
 
`src/com/google/javascript/jscomp/SourceMapResolver.java`:
 
```
private static boolean isAbsolute(String url) {
  try {
    return new URI(url).isAbsolute() || url.startsWith("/");
  } ...
}
 
static @Nullable SourceFile getRelativePath(String baseFilePath, String relativePath) {
  return SourceFile.builder()
      .withPath(
          FileSystems.getDefault().getPath(baseFilePath).resolveSibling(relativePath).normalize())
      ...
}
```
 
`isAbsolute()` only rejects URIs that are themselves absolute (`file:///etc/passwd`) or start with `/`. It does **not** reject relative traversal sequences like `../../../../etc/passwd`, so those pass through to `getRelativePath()`. There, `resolveSibling(relativePath).normalize()` resolves the traversal but the result is never checked against the base directory (no `resolved.startsWith(base)` guard anywhere in this function).
 
## Full attack chain
 
1. Attacker supplies a JS file containing:
```
   //# sourceMappingURL=../../../../etc/passwd
```
2. The parser extracts `sourceMapURL = "../../../../etc/passwd"`.
3. `SourceMapResolver.extractSourceMap(jsFile, "../../../../etc/passwd", parseInline)` is invoked.
4. `isAbsolute("../../../../etc/passwd")` → `false` (it's a relative path — not blocked).
5. `getRelativePath(jsFile.getName(), "../../../../etc/passwd")` → e.g. `Path.of("input.js").resolveSibling("../../../../etc/passwd").normalize()` → resolves outside the working directory (e.g. `/etc/passwd` when the base path allows enough `..` segments, or further up the tree relative to CWD otherwise).
6. Returns a `SourceFile` built from that resolved path.
7. `compiler.addInputSourceMap(jsFileName, new SourceMapInput(sourceMapSourceFile))` registers it.
8. Later, `inputSourceMap.getSourceMap(errorManager)` calls `sourceFile.getCode()`, which reads the file's contents from disk.
9. The file contents surface to the attacker via compiler error messages (e.g. `SOURCEMAP_PARSE_FAILED` when the target file isn't valid source-map JSON, which will include a snippet/parse context) and/or via the `sourcesContent` array of the compiler's own output source map, if source map output is enabled.
No special flag is required — `resolveSourceMapAnnotations` (which governs whether `sourceMappingURL` comments are honored) defaults to `true`.
 
## Affected scenarios
 
Impactful wherever closure-compiler processes JavaScript it doesn't fully control:
 
- SaaS/hosted build or minification services accepting user-submitted JS
- CI/CD pipelines compiling third-party or unreviewed dependencies
- Monorepos where one tenant's/team's code is compiled in a shared build environment
- IDE or dev-tool integrations that run closure-compiler over arbitrary opened files
Users who only ever compile their own fully-trusted source are not affected.
 
## Suggested fix
 
Add a bounds check in `getRelativePath()` so the resolved path can't escape the expected base directory:
 
```
static @Nullable SourceFile getRelativePath(String baseFilePath, String relativePath) {
  Path base = FileSystems.getDefault().getPath(baseFilePath).toAbsolutePath().normalize();
  Path resolved = base.resolveSibling(relativePath).toAbsolutePath().normalize();
  if (!resolved.startsWith(base.getParent())) {
    return null; // reject traversal outside the base directory
  }
  return SourceFile.builder()
      .withPath(resolved)
      .withKind(SourceKind.NON_CODE)
      .build();
}
```
 
An alternative, lower-effort mitigation (matching Babel's approach) would be to default `resolveSourceMapAnnotations` to `false`, requiring an explicit opt-in for any environment that needs inline source-map resolution over trusted input.
 
## Comparison with CVE-2026-49356 (@babel/core)
 
| | @babel/core (CVE-2026-49356, patched) | closure-compiler (this issue) |
|---|---|---|
| Trigger | `//# sourceMappingURL=` comment | `//# sourceMappingURL=` comment |
| Enabled by default | Yes | Yes (`resolveSourceMapAnnotations=true`) |
| Bounds check before fix | None | None |
| Disclosure vector | Output source map | Compiler error messages + output `sourcesContent` |
| Status | Fixed June 13, 2026 | Unpatched as of this report |
 
## Verification
 
Dynamically verified against closure-compiler v20260629 (latest at time of testing) on Java 21, by compiling a JS input containing a `sourceMappingURL` comment pointing outside the working directory and observing the target file's contents surface via compiler error output.
 
## Environment
 
- Reported to bughunters.google.com; closed as Infeasible for OSS VRP purposes, with explicit permission granted to disclose publicly here.
- `SourceMapResolver.java` last touched by a refactor-only commit (no security-relevant change) prior to this report; pattern confirmed still present as of the date of this issue.

kunaaaalcodes:
### Summary
Fixes #4322.

Changes `CompilerOptions.resolveSourceMapAnnotations` to default to `false` instead of `true`.

### Rationale
When `resolveSourceMapAnnotations` is `true` by default, compiling untrusted JS containing a `//# sourceMappingURL=` comment causes `SourceMapResolver.extractSourceMap` to resolve local filesystem paths automatically, creating an arbitrary file read vulnerability for untrusted inputs.

Flipping this default ensures `closure-compiler` is secure-by-default when handling untrusted JS inputs, following the mitigation pattern used for CVE-2026-49356 (`@babel/core`). Callers compiling trusted code can explicitly enable resolution via `options.setResolveSourceMapAnnotations(true)`.

### Verification
- Added `testResolveSourceMapAnnotationsDefaultsToFalse` in `CompilerOptionsTest.java`.
- Ran `bazel test //:test/com/google/javascript/jscomp/CompilerOptionsTest` (PASSED).

cc @lauraharker

### DIFF
--- a/src/com/google/javascript/jscomp/CompilerOptions.java
+++ b/src/com/google/javascript/jscomp/CompilerOptions.java
@@ -1165,8 +1165,18 @@ public class CompilerOptions {
   /**
    * Whether to resolve source mapping annotations. Cannot do this in an appengine or js environment
    * since we don't have access to the filesystem.
+   *
+   * <p>Defaults to {@code false}. When enabled, a {@code //# sourceMappingURL=} comment in
+   * compiled JS is followed to a file on the local filesystem (see {@link
+   * SourceMapResolver#extractSourceMap}), including via relative paths such as {@code
+   * ../../../../etc/passwd} that resolve outside the directory of the file being compiled. If
+   * this defaulted to {@code true}, compiling untrusted or third-party JavaScript source (e.g. in
+   * a hosted minification service or CI environment) would let the input JS trigger an arbitrary
+   * local file read with no explicit opt-in required. Callers who need to resolve source mapping
+   * annotations over source they fully trust can opt back in with {@link
+   * #setResolveSourceMapAnnotations}.
    */
-  private boolean resolveSourceMapAnnotations = true;
+  private boolean resolveSourceMapAnnotations = false;
 
   private ImmutableList<? extends SourceMap.LocationMapping> sourceMapLocationMappings =
       ImmutableList.of();
--- a/test/com/google/javascript/jscomp/CompilerOptionsTest.java
+++ b/test/com/google/javascript/jscomp/CompilerOptionsTest.java
@@ -40,6 +40,15 @@ import org.junit.runners.JUnit4;
 public final class CompilerOptionsTest {
 
   @Test
+  public void testResolveSourceMapAnnotationsDefaultsToFalse() {
+    CompilerOptions options = new CompilerOptions();
+    assertThat(options.getResolveSourceMapAnnotations()).isFalse();
+
+    options.setResolveSourceMapAnnotations(true);
+    assertThat(options.getResolveSourceMapAnnotations()).isTrue();
+  }
+
+  @Test
   public void testBrowserFeaturesetYearOptionSetsLanguageOut() {
     CompilerOptions options = new CompilerOptions();
     options.setBrowserFeaturesetYear(2012);
--- a/test/com/google/javascript/jscomp/CompilerTest.java
+++ b/test/com/google/javascript/jscomp/CompilerTest.java
@@ -259,6 +259,7 @@ public final class CompilerTest {
   public void testInputSourceMapInline() {
     Compiler compiler = new Compiler();
     compiler.initCompilerOptionsIfTesting();
+    compiler.getOptions().setResolveSourceMapAnnotations(true);
     String code = SOURCE_MAP_TEST_CODE + "\n//# sourceMappingURL=" + BASE64_ENCODED_SOURCE_MAP;
     CompilerInput input = new CompilerInput(SourceFile.fromCode("tmp", code));
     input.getAstRoot(compiler);
@@ -290,6 +291,7 @@ public final class CompilerTest {
   public void testInputSourceMapInlineContent() {
     Compiler compiler = new Compiler();
     compiler.initCompilerOptionsIfTesting();
+    compiler.getOptions().setResolveSourceMapAnnotations(true);
     String code =
         SOURCE_MAP_TEST_CODE + "\n//# sourceMappingURL=" + BASE64_ENCODED_SOURCE_MAP_WITH_CONTENT;
     CompilerInput input = new CompilerInput(SourceFile.fromCode("tmp", code));
@@ -305,6 +307,7 @@ public final class CompilerTest {
   public void testResolveRelativeSourceMap() throws Exception {
     Compiler compiler = new Compiler();
     compiler.initCompilerOptionsIfTesting();
+    compiler.getOptions().setResolveSourceMapAnnotations(true);
     File tempDir = Files.createTempDir();
     String code = SOURCE_MAP_TEST_CODE + "\n//# sourceMappingURL=foo.js.map";
     File jsFile = new File(tempDir, "foo.js");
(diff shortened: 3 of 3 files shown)


kunaaaalcodes:
Hi @lauraharker,

The Google CLA is signed and verified. 

I've updated the PR with:
- **Default configuration fix**: Set `resolveSourceMapAnnotations` to `false` in `CompilerOptions.java` (matching the secure-by-default pattern used in `@babel/core` for CVE-2026-49356).
- **Unit test coverage**: Added `testResolveSourceMapAnnotationsDefaultsToFalse` in `CompilerOptionsTest.java` and updated test configurations in `CompilerTest.java`.
- **Local verification**: Built and ran `bazel test` cleanly across the test suite.

Please take a look when you have a chance. Thanks!

lauraharker:
Hi Kunal -

I think we're somewhat concerned about flipping the default here breaking existing workflows. What do you think about restricting file reads to some heuristically "safe" set of directories?

kunaaaalcodes:
Hi @lauraharker ,

Appreciate the feedback — agreed, flipping the default isn't really the fix. It just hides the problem, since anyone who turns `resolveSourceMapAnnotations` back on would be fully exposed again. Let's leave the default as-is.

For the "safe directories" idea, I want to keep it simple: a `sourceMappingURL` should only be allowed to resolve to something inside the JS file's own directory (or below it). Anything that tries to escape just gets ignored. No new config needed, existing workflows stay untouched, and the default doesn't change.

One thing worth calling out on scope — `getRelativePath()` is also used when resolving `sources` entries inside sourcemaps that are already loaded, and those can legitimately point outside the map's own directory (`../src/...` is a pretty normal layout). So I don't want to slap the check there. I'll put it in `extractSourceMap` instead, since that's the actual point where we're interpreting an annotation coming from untrusted input. Absolute URLs are already blocked there, so this just closes off the relative traversal case without touching legitimate sourcemap resolution.

**Plan:**

- Revert the `CompilerOptions` default change and its test
- Add the containment check in `extractSourceMap`
- Add a regression test that actually reproduces the `../../../../etc/passwd` case from #4322, with the flag explicitly left on — so we're testing the real traversal, not just checking a default value

Separately, something to think about later: `normalize()` won't catch a symlink sitting inside the source directory that points outward, so that's technically still open. `toRealPath()` would fix it but gets messy with paths that don't exist yet. Happy to fold that into this PR or split it into a follow-up, whichever you'd prefer.

PR #4332, opened 2026-08-14
Fix: Attach a function name identifier to the minified symbol instead of the `function` keyword. 

bdingman-daedalus:
Source maps could lose the original name of a minified identifier for functions that started as anonymous function expressions (e.g. `var f = function() {...}`). `SourceInformationAnnotator` records the original name on the enclosing FUNCTION node but skips the (initially empty) NAME node. When `CollapseAnonymousFunctions` later fills in that NAME node as part of rewriting to `function f() {...}`, it doesn't backfill originalName — so once `RenameVars` minifies the identifier, the source map has no original name to attach to it.

` SourceMap.getOriginalName` now falls back to the enclosing FUNCTION node's originalName for this case, so the minified identifier correctly maps back to "f".

The PR adds a regression test that fails without the implemented fix.

The original purpose of this fix was to integrate with Sentry monitoring. Sourcemaps generated by closure are uploaded to Sentry using their CLI ([docs](https://docs.sentry.io/platforms/javascript/sourcemaps/uploading/cli/)). Sentry's parsing of sourcemaps would incorrectly label or not label certain functions at all due to the issue described here. We implemented this fix on a fork and it has been working for us for several months.

### DIFF
--- a/src/com/google/javascript/jscomp/SourceMap.java
+++ b/src/com/google/javascript/jscomp/SourceMap.java
@@ -232,6 +232,19 @@ public final class SourceMap {
     if (node.getOriginalName() != null) {
       return node.getOriginalName();
     }
+    // If this is the name identifier of a function declaration/expression and the name node itself
+    // has no originalName, fall back to the enclosing FUNCTION node's originalName.
+    // This covers functions that were originally anonymous (class methods compiled to prototype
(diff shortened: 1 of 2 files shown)


concavelenz:
Can you add what your use case is for the symbol that this change matters?

bdingman-daedalus:
Edited the original description to add our use case. This issue was interfering with Sentry's ability to de-minify stack traces from monitoring in our application.

PR #4336, opened 2026-08-25
Optimize compiler, add controls for speed-size tradeoffs, and allow for AST as input

DavidANeil:
This PR is a collection of LLM-driven optimizations to the compiler.

It does include a few changes to the tracer mode, which were used for making these improvements, but if you would like I can revert those changes and only leave the actual optimizations.

For our closure-compiler pipeline we were able to get it down from 290s to 100s for a byte-equivalent version, and even faster when using the exposed speed-size tuning knobs.  Not all of those speed improvements are local to the changes here, as those timings also include improvements to, for example, how we compress the assets after compilation completes.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -2,22 +2,22 @@
 #
 # Copied from https://github.com/google/j2cl/blob/master/.bazelrc since Bazel
 # doesn't support referring to .bazelrc from different workspaces.
-build --watchfs
-build --spawn_strategy=local
-build --strategy=J2cl=worker
-build --strategy=Closure=worker
+common --watchfs
+common --spawn_strategy=local
+common --strategy=J2cl=worker
+common --strategy=Closure=worker
 
 # Avoid the problem where build fails with:
 # //:compiler_unshaded.publish: $(COMPILER_VERSION) not defined.
-build --define=COMPILER_VERSION=1.0-SNAPSHOT
+common --define=COMPILER_VERSION=1.0-SNAPSHOT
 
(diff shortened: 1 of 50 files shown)


blickly:
Thanks for your interest in exploring performance improvements for Closure Compiler and sharing your results!

Unfortunately, we cannot accept this PR. To be transparent about the main blockers:

* Closure Compiler relies on strict architectural invariants that cannot be safely modified through large, automated collections of changes:

* **Scope and Reviewability**: We cannot accept multi-thousand-line PRs that bundle dozens of disparate modifications together. Each change must be small, independently justified, and accompanied by targeted unit tests.
* **Single-Threaded Architecture & Determinism**: The AST and internal compiler state are fundamentally single-threaded by design. Injecting concurrency across compiler passes introduces data races and breaks bit-for-bit reproducible builds.
* **Code Size Prioritization**: Closure strictly prioritizes minimal final JavaScript payload size over compilation speed. Heuristics that terminate optimization loops early regress output size, which is unacceptable for production use cases.
* **Internal Infrastructure Coupling**: Schemas like `typed_ast.proto` and associated flags are tightly coupled to internal Google build systems and cannot easily be modified externally.

If you have specific, targeted improvements that operate within those bounds, we would be open to reviewing a small, standalone PR focused solely on that. But because of the architectural complexity, threading constraints, and strict output size guarantees of this codebase, we cannot accept broad, speculative overhauls or AI-generated passes.

PR #4337, opened 2026-08-27
stop truncating out-of-bmp identifier escapes in the scanner

dxbjavid:
**Out-of-range identifier escapes get silently truncated to a char**

In `processUnicodeEscapes` a braced unicode escape inside an identifier has its parsed code point cast straight to a char, so only the low 16 bits survive. That means an escape like \u{10041} is quietly accepted as the identifier `A`, and even values above the U+10FFFF maximum such as \u{110041} slip through the same way, so two identifiers that ought to be distinct (or rejected outright) can end up collapsing onto one. The scanner already rejects \u{10000} here, and the string-literal path range-checks the code point, so this brings identifiers into line by rejecting anything that does not fit in a char rather than aliasing it to a different character.

### DIFF
--- a/src/com/google/javascript/jscomp/parsing/parser/Scanner.java
+++ b/src/com/google/javascript/jscomp/parsing/parser/Scanner.java
@@ -882,8 +882,18 @@ public class Scanner {
           hexDigits = value.substring(escapeStart + 3, escapeEnd);
           escapeEnd++;
         }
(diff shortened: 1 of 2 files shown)


blickly:
Thanks for the contribution! This looks good to me. I'll work on getting it landed.

dxbjavid:
@blickly Thanks for the merge. Appreciated

